PR #45, opened 2021-03-10
iOS Archiving for MapLibre + MetalANGLE

roblabs:
This PR is the result after testing `Product` > `Archive` from within Xcode.  

* Create an app with MapLibre + MetalANGLE to verify if the embedding of the Frameworks is correct.
* This PR has changes to the Xcode Project, which is difficult to assess in `git diff`, so be sure to see the discussion and screen caps below.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -49,3 +49,6 @@
 [submodule "maplibre-gl-js"]
 	path = maplibre-gl-js
 	url = https://github.com/maplibre/maplibre-gl-js.git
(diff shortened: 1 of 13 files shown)


roblabs:
@petr-pokorny-1 — please hold off on this PR until I can review #50 .  There may be some duplication that would cause us to void this PR.

roblabs:
Closing with out merging.  Cherry picking (manually) into #50.

PR #61, opened 2021-03-31
[ios] Cherry pick changes from from https://github.com/mapbox/mapbox-gl-native-ios/pull/268/files

juhieta:
add allowScrollGesturesDuringRotateOrZoom


juhieta:
This is useful option to make zooming more "stable" by locking it to the original location

Issue #76, opened 2021-04-24
Map opens "zoomed in" when app brought to foreground with `metal-support` -branch


juhieta:
Using iOS SDK built from the `metal-support` -branch, map opens "zoomed in" when bringing front from background, and then updates to correct zoom level. This causes visible flickering.

This doesn't happen with SDK built from master.

# Steps to reproduce
1. Start iOS SDK [demo application](https://github.com/maplibre/maplibre-gl-native/tree/master/platform/ios/platform/ios/app) built from `metal-support` -branch
2. Move app to background (don't close it)
3. Click app icon to bring back to foreground

# Expected behavior
Map opens to correct zoom level without flickering

# Actual behavior
Map opens zoomed in, then quickly shows actual zoom level causing flicker.

Following is a video showing the issue with the iOS SDK Demo App, first in normal speed and then slowed down for better view:

https://user-images.githubusercontent.com/2369584/115957491-7e78c780-a50b-11eb-9589-ade706062091.mov

# Configuration
Maplibre iOS SDK version: `metal-support` -branch commit [bae8d18](https://github.com/maplibre/maplibre-gl-native/commit/bae8d18d5e04651177183c34f869bde7d55c32cb)
iOS/macOS versions: **iOS 14.4**
Device/simulator models: **iPhone 11 Simulator**
Xcode version: **12.4**

juhieta:
This problem can be fixed with PR #83

juhieta:
Was fixed with #83

juhieta:
Reopening until #83 is merged to metal branch

petr-pokorny-1:
> Reopening until #83 is merged to metal branch

@juhieta  I've merged #83 into metal branch, are we good to close this one?

juhieta:
> > Reopening until #83 is merged to metal branch
> 
> @juhieta I've merged #83 into metal branch, are we good to close this one?

Great, thanks and let's close! 👍

PR #83, opened 2021-05-18
[ios] Disable map snapshots by default

juhieta:
MapLibre currently periodically takes snapshots of the map view, and when brought from background to foreground, it will first display the latest snapshot and then change the display to rendered map.

With #68, to prevent iOS crashes, we limited map snapshotting to foreground. Snapshots are now taken every 60 seconds on iOS when map is idle in the foreground. Due to this, map snapshot is not necessarily very recent one. This causes following issues:

* If user has changed map style after last snapshot, wrong map style can flash when app is brought to foreground
* Map can jump quite much when brought to foreground
* Snapshotting also causes this "zoomed in" flash issue with with Metal branch: #76

I have been testing the iOS SDK and I have not noticed any drawbacks yet when disabling the map snapshotting completely. If we don't find any drawbacks, I suggest we first disable the snapshotting with a define (as done in this PR) and later we can remove the snapshotting logic completely if no issues arise.

### DIFF
--- a/platform/ios/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/platform/ios/src/MGLMapView.mm
@@ -6699,7 +6699,9 @@ static void *windowScreenContext = &windowScreenContext;
         return;
     }
 
+#ifdef MBGL_ENABLE_MAP_SNAPSHOTS
     [self queueBackgroundSnapshot];
+#endif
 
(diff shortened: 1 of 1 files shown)


petr-pokorny-1 (approved):
Sounds good to me. Thanks for helping with this issue!

PR #102, opened 2021-07-04
[qt] Update for tile server configuration changes

ntadej:
Update the Qt platform for tile server configuration changes.

This is the first in a set of MRs to add proper Qt support for `maplibre-gl-native`.

Note that the default demo MapLibre style still does not work. I only get blue background. Other styles seem to work fine. Any idea why?

### DIFF
--- a/platform/qt/app/mapwindow.cpp
+++ b/platform/qt/app/mapwindow.cpp
@@ -61,7 +61,7 @@ void MapWindow::changeStyle()
 {
     static uint8_t currentStyleIndex;
 
-    auto& styles = QMapbox::defaultStyles();
+    auto& styles = m_map->defaultStyles();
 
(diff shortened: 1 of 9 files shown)


petr-pokorny-1:
> Note that the default demo MapLibre style still does not work. I only get blue background. Other styles seem to work fine. Any idea why?

I found the issue and fixed it https://github.com/maplibre/maplibre-gl-native/pull/107. Could you please check on your side?

ntadej:
Seems to work now thanks. Interestingly my MapTiler styles still seem to be working, probably because the necessary information is contained inside.

I will also add some more detailed support for these settings in the future, but personally I like to keep MRs small, so I think this one is ready to be reviewed.

petr-pokorny-1 (approved):
Great job!

PR #109, opened 2021-07-06
Address crash when localFontFamilyName is nil

nnhubbard:
Cherry-pick from https://github.com/mapbox/mapbox-gl-native-ios/pull/467


roblabs (approved):
Thanks!

PR #113, opened 2021-07-06
Fix for annotations stuttering and drifting when panning

nnhubbard:
Fixes annotation drifting when panning map. See: https://github.com/maplibre/maplibre-gl-native/issues/112

`mapbox-gl-native-ios` fixes for reference:

- https://github.com/mapbox/mapbox-gl-native-ios/pull/483
- https://github.com/mapbox/mapbox-gl-native-ios/commit/0044577cdd50de5fb2d106b0a4cb5ba4989d3466

### DIFF
--- a/platform/ios/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/platform/ios/src/MGLMapView.mm
@@ -285,6 +285,9 @@ public:
 @property (nonatomic, assign) UIEdgeInsets safeMapViewContentInsets;
 @property (nonatomic, strong) NSNumber *automaticallyAdjustContentInsetHolder;
 
+@property (nonatomic, assign) NSInteger numberOfRenderCalls;
+@property (nonatomic, assign) NSInteger numberOfRenderCallsMoreThanOneSecond;
+
 // Display Link
 @property (nonatomic, weak) UIScreen *displayLinkScreen;
 @property (nonatomic) CADisplayLink *displayLink;
@@ -989,7 +992,7 @@ public:
     }
 }
 
-- (void)updateViewsPostMapRendering {
+- (void)updateViewsWithCurrentUpdateParameters {
     // Update UIKit elements, prior to rendering
     [self updateUserLocationAnnotationView];
     [self updateAnnotationViews];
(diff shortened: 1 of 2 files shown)


roblabs:
I kicked off a CI/CD build.  Please see the _Checks_ tab above.

platform/ios/platform/ios/src/MGLMapView.mm
```diff
@@ -1003,17 +1003,47 @@ - (void)updateViewsPostMapRendering {
...
+      
+        if (_rendererFrontend) {
+            
+            _rendererFrontend->render();
+
+        }
+        
+    }
+
+    if (hasPendingBlocks) {
```
sulewicz, line 1037: I wonder, is this block needed given updateViewsWithCurrentUpdateParameters invokes processPendingBlocks already? Unless you want to invoke processPendingBlocks when the condition in line 1018 evaluates to false, but in this case, can we drop hasPendingBlocks and have an else block in 1035 instead?

sulewicz (commented):
Thanks! Posted one question.

platform/ios/platform/ios/src/MGLMapView.mm
nnhubbard: This code was taken from https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/ios/src/MGLMapView.mm#L1180 in `mapbox-gl-native-ios`. I did not consider removing that as I would assume it was originally placed there for a reason.

What what I see it needs to be called before and after `_rendererFrontend->render();`.

sulewicz: I see that updateViewsWithCurrentUpdateParameters does not contain processPendingBlocks call in the code you posted (mapbox code base: https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/ios/src/MGLMapView.mm#L1180). Removing it would sort out my concern, as it processPendingBlocks would not get called twice in the render method :).

nnhubbard: @sulewicz Thanks, you were right, that shouldn't have been there. It has been removed. 👍

sulewicz (approved):

nnhubbard:
@sulewicz Can this be merged? Thanks!

PR #116, opened 2021-07-07
Scale bar dark mode

nnhubbard:
Allows scale bar to change text outline colors to better show up on a dark map.


roblabs (approved):
I added the patch to test it out, but you mentioned it was for dark maps.  Please add any screen shots that may be useful to others.  A tiny shot of the scale bar would be fine.

I've kicked off a CI/CD build which anyone can read and peruse.

nnhubbard:
@roblabs I actually am closing this one. I think that this could cause issues with users that have dark mode in their apps, but don't provide a dark style for the map. I will update this to allow choosing light or dark style for the scale bar, which will give the user more control.

PR #118, opened 2021-07-07
[qt] Deprecation warnings and Qt compatibility fixes

ntadej:
Cherry-pick some commits from the `qt-staging` branch that were not in master and add additional deprecation warning fixes by @dpaulat and me.

### DIFF
--- a/platform/qt/src/local_glyph_rasterizer.cpp
+++ b/platform/qt/src/local_glyph_rasterizer.cpp
@@ -52,7 +52,11 @@ Glyph LocalGlyphRasterizer::rasterizeGlyph(const FontStack&, GlyphID glyphID) {
         return glyph;
     }
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 11, 0)
+    glyph.metrics.width = impl->metrics->horizontalAdvance(glyphID);
+#else
     glyph.metrics.width = impl->metrics->width(glyphID);
+#endif
     glyph.metrics.height = impl->metrics->height();
     glyph.metrics.left = 3;
     glyph.metrics.top = -8;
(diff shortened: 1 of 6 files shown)


platform/qt/src/local_glyph_rasterizer.cpp
```diff
@@ -68,8 +72,13 @@ Glyph LocalGlyphRasterizer::rasterizeGlyph(const FontStack&, GlyphID glyphID) {
     // Render at constant baseline, to align with glyphs that are rendered by node-fontnik.
     painter.drawText(QPointF(0, 20), QString(QChar(glyphID)));
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+    auto img = std::make_unique<uint8_t[]>(image.sizeInBytes());
+    memcpy(img.get(), image.constBits(), image.sizeInBytes());
+#else
     auto img = std::make_unique<uint8_t[]>(image.byteCount());
     memcpy(img.get(), image.constBits(), image.byteCount());
+#endif
```
petr-pokorny-1, line 81: Maybe just use variable to get sizeInBytes/byteCount in version-specific if-else directive and then use that variable  in make_unique/memcpy so that statements are not repeated?

petr-pokorny-1 (approved):
Just an idea to reduce repeated code, otherwise looks great! Thanks for your work!

platform/qt/src/local_glyph_rasterizer.cpp
ntadej: Sorry, I'm very busy this week. Let me follow-up in one of my next MRs.

petr-pokorny-1: No prob, this was just a "cosmetic" issue. Thanks for your help!

PR #120, opened 2021-07-07
scaleBar enhancements

nnhubbard:
This includes enhancements to the `scaleBar` that allow updating labels colors to better work with dark maps, and also allowing the user to control if the `scaleBar` is using metric.

## Dark Mode Enhancement

Light map with white outlined labels and black text:

mi

Dark map with black outlined labels and white text:

dark

`MGLMapView` includes a new property, `scaleBarShouldShowDarkStyles` to update the `scaleBar`.

## Metric Enhancement

Imperial units:

mi

Metric units:

km

`MGLMapView` includes a new property, `scaleBarUsesMetricSystem` to update the `scaleBar`.

### DIFF
--- a/platform/ios/platform/ios/src/MGLScaleBar.mm
+++ b/platform/ios/platform/ios/src/MGLScaleBar.mm
@@ -169,6 +169,10 @@ static const CGFloat MGLScaleBarMinimumBarWidth = 30.0; // Arbitrary
     self.clipsToBounds = NO;
     self.hidden = YES;
     
+    // Default to current local
+    NSLocale *locale = [NSLocale currentLocale];
+    _usesMetricSystem = [[locale objectForKey:NSLocaleUsesMetricSystem] boolValue];
+    
     _containerView                     = [[UIView alloc] init];
     _containerView.clipsToBounds       = YES;
     _containerView.backgroundColor     = _secondaryColor;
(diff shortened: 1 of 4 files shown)


platform/ios/platform/ios/src/MGLScaleBar.mm
```diff
@@ -282,8 +301,44 @@ - (MGLRow)preferredRow {
...
 #pragma mark - Setters
 
+- (void)setUsesMetricSystem:(BOOL)usesMetricSystem {
+    
+    if (_usesMetricSystem != usesMetricSystem) {
+        
+        _usesMetricSystem = usesMetricSystem;
+        
+        // Set the distance formatter locale using Germany for metric and United States for imperial.
+        self.formatter.numberFormatter.locale = usesMetricSystem ? [NSLocale localeWithLocaleIdentifier:@"de_DE"] : [NSLocale localeWithLocaleIdentifier:@"en_US"];
```
roblabs, line 330: This produces the case of `1,5 km`, but is there locale, say in en-US, where I want metric, but a `.` for a decimal point, such as `1.5 km`?

platform/ios/platform/ios/src/MGLScaleBar.mm
```diff
@@ -111,11 +122,19 @@ - (void)drawTextInRect:(CGRect)rect {
     CGContextSetLineJoin(context, kCGLineJoinRound);
     
     CGContextSetTextDrawingMode(context, kCGTextStroke);
-    self.textColor = [UIColor whiteColor];
+    if (_shouldShowDarkStyles) {
+        self.textColor = [UIColor blackColor];
+    } else {
+        self.textColor = [UIColor whiteColor];
+    }
```
roblabs, line 129: Great idea to use black and white for appropriate contrasts!!

roblabs:
Approved with the caveat that we may have other locales that need testing.  Great contribution, @nnhubbard.  Thanks for the screen shots!

platform/ios/platform/ios/src/MGLScaleBar.mm
nnhubbard: You made a good point about this. I somehow didn't catch this. Will put in a PR to fix this.

PR #121, opened 2021-07-08
scaleBar: Fixes for locales and using decimals or commas as separator

nnhubbard:
This is an additional fix for https://github.com/maplibre/maplibre-gl-native/pull/120 that will take into account the users current local to check if it supports metric and then provide fallbacks.


roblabs (approved):
Thanks for the PR.  Looks like it will test out properly.  Out of scope for this PR, but it would be nice to add Automated UI testing that exhibits this change.  Great work, @nnhubbard

PR #122, opened 2021-07-09
Align remote and local glyph baselines

honjo2:
The baseline of the remote glyph and the local glyph are out of alignment.
glyph1
The black line is the local glyph(Hiragino Sans) and the red line is the remote glyph(Noto Sans).
glyph2
The modified version is below.
glyph3


petr-pokorny-1:
Looks good! Thanks!

petr-pokorny-1 (approved):

PR #129, opened 2021-07-25
[core] fix perf regression in cached tiles of tile pyramid

nickaein:
It seems https://github.com/maplibre/maplibre-gl-native/commit/ac6c5d9ce09b76f0de3872589437b22c1c4984ac introduced a performance regression (apparently during refactoring) in which `setLayers` (and by extent `setObserver`) is called for cached tiles. This continuously triggers parsing for the cached tiles which significantly degrades the performance.

This PR reverts that change. In our benchmarks, this reduced the computation by ~2x in a hand-crafted looping zoom-in/zoom-out scenario.


petr-pokorny-1:
Looks good to me, thanks!

petr-pokorny-1 (approved):

PR #142, opened 2021-08-12
[win] Additional fixes for MSVC and MinGW

ntadej:
This is a collection of some additional fixes for MSVC and MinGW. Not sure why @dpaulat did not need those so keeping it as draft for a bit.

- some additional `ValueBase` and `Value` template specialisations to help resolve them
- MSVC-specific `std::chrono::duration` polyfill
- just use Qt-provided zlib on Windows
- some other minor tweaks


ntadej:
This MR is ready for review.

PR #149, opened 2021-08-28
add methods getTiles and setTiles to mimick maplibre-gl-js interface

xabbu42:
This adds the methods setTiles and getTiles to VectorSource to mimick the maplibre-gl-js interface
introduced in https://github.com/mapbox/mapbox-gl-js/pull/8048. This is useful to change parametrized data sources without having to load a complete new style.

Todo:
- [ ] To completely port the maplibre-gl-js interface, also setUrl should be added.
- [ ] Is the additional call to observer->onSourceLoaded correct/necessary?
- [ ] There is a race and the map sometimes still requests old tiles even after setTile is called.

Any input especially about the race is appreciated.


xabbu42:
We use this patch in production for over a year now. I just finally got around to update to latest maplibre-native opengl-2 branch and it still works as expected. Moreover I can no longer reproduce the race and I doubt it was a problem with this patch anyway.

I no longer thinks we should block this on an implementation of setUrl. That is a separate issue and a bit more complex, so I'd rather just get this in for now.

From my reading of the code the call to `observer->onSourceLoaded` is not necessary for Core funktionality, as `onSourceLoaded` and `onSourceChanged` ist handled exactly the same as `onSourceLoaded` in https://github.com/maplibre/maplibre-native/blob/main/src/mbgl/style/style_impl.cpp#L314. I'm happy to leave it in or take it out whatever the semantics of an external Observer are supposed to be.

So this ready for reviewing from my point of view.

wipfli:
Nice @xabbu42! Please remove the draft status in that case

xabbu42:
I'll have to look in the failed checks tomorrow.

louwers (commented):
Thanks for your contribution.

Could you add:

- A test
- A doxygen docstring (since these methods are part of the public API)

xabbu42:
Sorry for the long silence. I'll be away for the next 3 weeks but I plan to get back to this after that.

xabbu42:
The checks run through before the merge and it looks like the failures now have nothing to do with my changes. @louwers I hope I addressed your requests adequately (I'm not familiar with doxygen or GoogleTest). So this should be ready to merge now.

louwers:
@xabbu42 Seems like it runs through cleanly now. Let me have another look.

include/mbgl/style/sources/vector_source.hpp
```diff
@@ -26,6 +26,14 @@ class VectorSource final : public Source {
 
     void loadDescription(FileSource&) final;
 
+    /// @brief Gets the tile urls for this vector source.
+    /// @return std::vector<std::string> List of tile urls.
+    const std::vector<std::string> getTiles();
```
louwers, line 31: Should be marked `const`.

louwers (changes requested):
Two more nits, the getter can be marked `const`. And you can remove the type from the Doxygen comments. Then it looks good to merge!

include/mbgl/style/sources/vector_source.hpp
xabbu42: I removed the type from the docstring. The return type already is marked `const` or am I missing something?

louwers: Right, but the method itself can also be marked const right? It does not modify the object.

xabbu42: ah yes. I don't use c++ too often and forgot about that. fixed it now.

louwers (approved):
@xabbu42 Yup, that is what I meant.

https://en.cppreference.com/w/cpp/language/member_functions#Member_functions_with_cv-qualifiers

Thanks for your contribution!

PR #157, opened 2021-09-05
MGLNetworkConfiguration.sessionDelegate

### LINKED ISSUE #156: MGLNetworkConfiguration.sessionDelegate
It would be nice to be able to set an optional NSURLSessionDelegate to the NSURLSession used to download tiles to handle HTTP Basic Authentication.

For Mapbox GL, I reported this in https://github.com/mapbox/mapbox-gl-native/issues/11888 and I suggested a code change for it, but Mapbox ended up adding MGLNetworkConfiguration.sessionDelegate for this.

halset:
Add MGLNetworkConfiguration.sessionDelegate to configure a custom NSURLSessionDelegate.

Closes #156

### DIFF
--- a/platform/darwin/include/mbgl/interface/native_apple_interface.h
+++ b/platform/darwin/include/mbgl/interface/native_apple_interface.h
@@ -14,6 +14,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSURLSessionConfiguration *)sessionConfiguration;
 
+- (id<NSURLSessionDelegate>)sessionDelegate;
+
 - (void)startDownloadEvent:(NSString *)event type:(NSString *)type;
 
 - (void)cancelDownloadEventForResponse:(NSURLResponse *)response;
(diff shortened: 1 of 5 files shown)


petr-pokorny-1:
I believe this can be done outside of the SDK using NSURLSessionConfiguration.

rudolfkopriva:
You can do it outside SDK like this (Swift variant):
1. Create class which implements URLProtocol and URLSessionDataDelegate
2. Register it using the URLProtocol.registerClass
3. Create URLSessionConfiguration and insert your URLProtocol using protocolClasses.insert
4. Assign the sessionConfiguration to MGLNetworkConfiguration.sharedManager.sessionConfiguration

A good example of custom URLProtocol can be found in [Alamofire](https://github.com/cpDave/Dairy/blob/004020926318f3dc2ce9c7bf4c3781028fdb4ed1/Carthage/Checkouts/Alamofire/Tests/URLProtocolTests.swift)

halset:
@rudolfkopriva , would that require a custom url scheme or can it be normal http and https?

rudolfkopriva:
@halset normal http or https

halset:
@rudolfkopriva , thanks. That work fine!

nvanfleet:
Does it make sense that this only seems to affect tile loading? I see no calls within canInit for my style downloads.

rudolfkopriva:
@nvanfleet For me it is called for style URL as well

nvanfleet:
> @nvanfleet For me it is called for style URL as well

Thanks for the answer. I did get it working. It just seems like there is some kind of internal caching that makes it behave in a way I didn't expect. Also when I take over the networking and say caching is not allowed I notice it re-downloads the style every few seconds.

PR #159, opened 2021-09-12
[qt] Improve CMake setup

ntadej:
I collected all CMake updates to make `maplibre-gl-native` build with Qt on all platforms that I want to support (macOS, Windows, Linux, Android, iOS).
 - ensure consistent MSVC compile flags
 - proper CMake logging
 - update (seemingly unused) iOS toolchain file
 - add versioning support to the Qt build
 - add static build option
 - add option to build only library (without tests and demo app)
 - add options to choose internal sqlite or ICU libraries (on platforms where supported)
 - macOS framework support
 - add installation target


ntadej:
I don't see how the iOS issue could be related...

PR #172, opened 2021-10-16
add missing include 

nunojpg:
to fix error: ‘exception_ptr’ in namespace ‘std' does not name a type


roblabs:
@nunojpg — What build error does this fix?  Is it for QT?

Looks like the GitHub workflow for iOS-ci on this commit failed for UI Tests.  Android & macOS both passed.

nunojpg:
@roblabs yes, QT with gcc 11.2. GCC probably removed some internal header dependency, so this used header had to be explicitly included.

PR #173, opened 2021-10-16
restore Qt5 compatibility

nunojpg:
OpenGLWidgets only exists in Qt6, in Qt5 is enough to include OpenGL


ntadej:
Should already be fixed (and improved) in #159.

PR #180, opened 2021-11-04
Update Windows build instructions in platform/qt/README.md

srcejon:
The Windows build instructions for Qt appear to be out of date and not to work, so this PR updates them.

I've used Ninja as the cmake generator as it's much faster than "Visual Studio 16 2019", although that does work.


petr-pokorny-1 (approved):
Thanks!

PR #188, opened 2021-11-28
[ios] Fix camera paddings reset on map view gestures performing.

OlexandrStepanov:
This PR fixes the issue with [inconsistent padding reset](https://github.com/maplibre/maplibre-gl-native/issues/171) of the map view camera after performing any gesture on the map other than pan.

The logic to reset camera paddings with pinch/rotation gesture was introduced in [this commit](https://github.com/maplibre/maplibre-gl-native/commit/5c565a5e51be8379ddc2ffb5c859d8cfebe2fcd5) altogether with gesture unit tests. Although I'm not sure what was the original intent, it seems wrong nowadays taking into account the current logic of edge insets handling in `MGLMapView.set...` methods.

Just to be clear, what I mean by "current logic":
* `contentInset` property of `MGLMapView` controls the global padding of the map view. This value persists between camera changes of the map and controls the map elements positioning, like compass, scale bar, logo view etc.
* `edgePadding` value passed in the methods like `setCamera:`, `setVisibleCoordinatesBounds:`, `showAnnotations:` etc. is the transient value for this particular camera update.
* The result map paddings are a combination of the map `contentInset` and passed `edgePadding`. This value is passed to the internal `_mbglMap` as a part of `CameraOptions`.
* The result paddings  persist until the next programmatic camera update or change of `contentInset` value.

There is [another PR](https://github.com/mapbox/mapbox-gl-native-ios/pull/323) which is reported to be fixing the same issue, but there are some new problems introduced there.
* Adding another way to set paddings through the camera object property overcomplicate the things. It seems that semantically camera paddings and the transient `edgePadding` are interchangeable, so having both of them introduces some ambiguity.
* There is an inconsistency in the camera management: if one pass camera object to the map, and then reads it, it's expected that it returns the same object back. Padding value breaks that assumption in that PR.

More discussions on the topic:
- https://github.com/mapbox/mapbox-gl-native-ios/issues/198
- https://github.com/mapbox/mapbox-gl-native/issues/15233


OlexandrStepanov:
[Here](https://github.com/OlexandrStepanov/maplibre-gl-native/tree/fix-asymmetrical-paddings-demo/AsymmetricalPaddingsDemo) one can find the project with the fix demonstration. It's in the separate `fix-asymmetrical-paddings-demo` branch.

https://user-images.githubusercontent.com/971079/143784787-ed11c233-4593-4fa0-9ee8-70000f2390ac.mov

* Red frame demonstrates the current map view `contentInset` padding.
* Blue frame demonstrates the combined `cameraEdgeInsets` padding.
* Black frame is just the test region to be used in `setVisibleCoordinateBounds` method.
* The grey cross demonstrates `centerCoordinate` position, which is efectively the camera center respecting the paddings,
* There are two actions which demonstrates setting symmetrical and asymmetrical paddings.

platform/ios/platform/ios/src/MGLMapView.mm
```diff
@@ -2591,8 +2586,7 @@ - (MGLMapCamera *)cameraByZoomingToZoomLevel:(double)zoom aroundAnchorPoint:(CGP
 
 - (MGLMapCamera *)cameraByRotatingToDirection:(CLLocationDirection)degrees aroundAnchorPoint:(CGPoint)anchorPoint
 {
-    mbgl::EdgeInsets padding = MGLEdgeInsetsFromNSEdgeInsets(self.contentInset);
-    mbgl::CameraOptions currentCameraOptions = self.mbglMap.getCameraOptions(padding);
+    mbgl::CameraOptions currentCameraOptions = self.mbglMap.getCameraOptions();
```
OlexandrStepanov, line 2589: Here and in similar places below the padding value passed in `mbglMap.getCameraOptions()` just replaces the camera edge insets. Although `CameraOptions` is a pure model class which doesn't have any computations and side effects, it doesn't make much sense to replace its padding value.

platform/ios/platform/ios/test/MGLMapViewGestureRecognizerTests.mm
```diff
@@ -67,15 +67,15 @@ - (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
...
+    UIEdgeInsets contentInset = UIEdgeInsetsMake(1, 1, 1, 1);
     self.mapView.contentInset = contentInset;
     mbgl::EdgeInsets padding = MGLEdgeInsetsFromNSEdgeInsets(self.mapView.contentInset);
     auto cameraPadding = self.mapView.mbglMap.getCameraOptions().padding;
     XCTAssertEqual(padding, cameraPadding, @"MGLMapView's contentInset property should match camera's padding.");
     XCTAssertTrue(UIEdgeInsetsEqualToEdgeInsets(self.mapView.contentInset, contentInset));
     
     contentInset = UIEdgeInsetsMake(20, 20, 20, 20);
-    [self.mapView setCamera:self.mapView.camera withDuration:0.1 animationTimingFunction:nil edgePadding:contentInset completionHandler:nil];
+    [self.mapView setCamera:self.mapView.camera withDuration:0.0 animationTimingFunction:nil edgePadding:contentInset completionHandler:nil];
```
OlexandrStepanov, line 78: The change must be immediate to take an effect while the test is running.

petr-pokorny-1:
This looks really good @OlexandrStepanov Thank you for the great work!

petr-pokorny-1 (approved):

PR #189, opened 2021-12-02
mapbox -> maplibre, fix broken links

matkoniecz:
I assume that such rename is desired and should be applied, but link targets still carry Mapbox name

BTW, links were broken and I fixed them


ntadej (approved):

ntadej:
Thanks @matkoniecz!

PR #190, opened 2021-12-02
remove mention of nonexisting labelling system

matkoniecz:
See https://github.com/maplibre/maplibre-gl-native/labels - such elaborate system is not present here

And I have doubts is it useful to spend time on such sorting given current available resources


matkoniecz:
Failed tests are bug on their own, see #191

ntadej:
We should at least revisit labeling so it's probably fine to remove this.

What do you think, @roblabs and @birkskyum?

birkskyum:
yeah, it's not in system now anyway, so better to clean it up

ntadej:
@matkoniecz, can you please try to rebase this PR? Unfortunately the pipeline is too old so I can not retry it anymore.

PR #192, opened 2021-12-05
[qt] allow to build without widgets and other cmake improvements

rinigus:
This PR brings improvements to CMake/Qt which target

- better support for install directories (lib, bin, include all come from defined env vars)
- allows to build Maplibre on platforms missing OpenGL and Qt Widgets
- brings CMake import that can be used for linking against it

Example platforms missing Widgets and OpenGL: Sailfish OS and Ubuntu Touch. 

This PR combines changes in https://github.com/mapbox/mapbox-gl-native/pull/16591 . Large sections are based on @PureTryOut work on CMake config from that PR.

### DIFF
--- a/platform/qt/qt.cmake
+++ b/platform/qt/qt.cmake
@@ -244,6 +244,12 @@ install(FILES
 	DESTINATION ${CMAKECONFIG_INSTALL_DIR}
 	COMPONENT development)
 
+install(
+    DIRECTORY include/mbgl
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+    COMPONENT development
+)
+
 # FIXME: Because of rapidjson conversion
 target_include_directories(
     qmapboxgl
--- a/platform/qt/qt.cmake
+++ b/platform/qt/qt.cmake
@@ -9,17 +9,23 @@ message(STATUS "Version ${MBGL_QT_VERSION}")
 
 option(MBGL_QT_STATIC "Build Mapbox GL Qt bindings staticly" OFF)
 option(MBGL_QT_LIBRARY_ONLY "Build only libraries" OFF)
+option(MBGL_WITH_QT_HEADLESS "Build Mapbox GL Qt with headless support" ON)
 option(MBGL_QT_WITH_INTERNAL_SQLITE "Build Mapbox GL Qt bindings with internal sqlite" $<NOT:$<BOOL:${MBGL_QT_STATIC}>>)
 
 find_package(QT NAMES Qt6 Qt5 COMPONENTS Core REQUIRED)
 find_package(Qt${QT_VERSION_MAJOR}
              COMPONENTS Gui
                         Network
-                        OpenGL
-                        Widgets
              REQUIRED)
 if (Qt6_FOUND)
     find_package(Qt${QT_VERSION_MAJOR}OpenGLWidgets REQUIRED)
+endif()
+
+if(MBGL_WITH_QT_HEADLESS OR NOT MBGL_QT_LIBRARY_ONLY)
+    find_package(Qt${QT_VERSION_MAJOR}
+                 COMPONENTS OpenGL
+                            Widgets
+                 REQUIRED)
 endif()
 
 if(NOT MBGL_QT_WITH_INTERNAL_SQLITE)
@@ -88,7 +94,6 @@ target_sources(
         ${PROJECT_SOURCE_DIR}/platform/qt/src/async_task_impl.hpp
         ${PROJECT_SOURCE_DIR}/platform/qt/src/number_format.cpp
         ${PROJECT_SOURCE_DIR}/platform/qt/src/gl_functions.cpp
-        ${PROJECT_SOURCE_DIR}/platform/qt/src/headless_backend_qt.cpp
         $<$<BOOL:${MBGL_PUBLIC_BUILD}>:${PROJECT_SOURCE_DIR}/platform/qt/src/http_file_source.cpp>
         $<$<BOOL:${MBGL_PUBLIC_BUILD}>:${PROJECT_SOURCE_DIR}/platform/qt/src/http_file_source.hpp>
         $<$<BOOL:${MBGL_PUBLIC_BUILD}>:${PROJECT_SOURCE_DIR}/platform/qt/src/http_request.cpp>
@@ -106,6 +111,19 @@ target_sources(
         ${PROJECT_SOURCE_DIR}/platform/qt/src/utf.cpp
 )
 
+if(MBGL_WITH_QT_HEADLESS OR NOT MBGL_QT_LIBRARY_ONLY)
+    target_sources(
+        mbgl-core
+        PRIVATE
+           ${PROJECT_SOURCE_DIR}/platform/qt/src/headless_backend_qt.cpp
+    )
+    target_link_libraries(
+        mbgl-core
+        PRIVATE
+	   Qt${QT_VERSION_MAJOR}::OpenGL
+    )
+endif()
+
 target_compile_definitions(
     mbgl-core
     PRIVATE QT_IMAGE_DECODERS
@@ -117,16 +135,17 @@ target_include_directories(
     PRIVATE ${PROJECT_SOURCE_DIR}/platform/default/include
 )
 
+include(GNUInstallDirs)
 include(${PROJECT_SOURCE_DIR}/vendor/nunicode.cmake)
 if(MBGL_QT_STATIC)
     install(TARGETS mbgl-vendor-csscolorparser
-            ARCHIVE DESTINATION lib
+            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
     )
     install(TARGETS mbgl-vendor-nunicode
-            ARCHIVE DESTINATION lib
+            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
     )
     install(TARGETS mbgl-vendor-parsedate
-            ARCHIVE DESTINATION lib
+            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
     )
 endif()
 
@@ -139,7 +158,6 @@ target_link_libraries(
         Qt${QT_VERSION_MAJOR}::Core
         Qt${QT_VERSION_MAJOR}::Gui
         Qt${QT_VERSION_MAJOR}::Network
-        Qt${QT_VERSION_MAJOR}::OpenGL
         $<IF:$<BOOL:${MBGL_QT_WITH_INTERNAL_SQLITE}>,mbgl-vendor-sqlite,Qt${QT_VERSION_MAJOR}::Sql>
         $<$<PLATFORM_ID:Linux>:$<IF:$<BOOL:${MBGL_QT_WITH_INTERNAL_ICU}>,mbgl-vendor-icu,ICU::uc>>
         mbgl-vendor-nunicode
@@ -147,7 +165,7 @@ target_link_libraries(
 
 if(MBGL_QT_STATIC)
     install(TARGETS mbgl-core
-            ARCHIVE DESTINATION lib
+            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
     )
 endif()
 
@@ -188,6 +206,7 @@ target_sources(
 set_target_properties(
     qmapboxgl PROPERTIES
     AUTOMOC ON
+    EXPORT_NAME QMapboxGL
     OUTPUT_NAME QMapboxGL
     VERSION ${MBGL_QT_VERSION}
     SOVERSION ${MBGL_QT_VERSION_COMPATIBILITY}
@@ -201,6 +220,30 @@ if (NOT MBGL_QT_STATIC)
     )
 endif()
 
+include(CMakePackageConfigHelpers)
+set(CMAKECONFIG_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/qmapboxgl/)
+
+configure_package_config_file(
+	"platform/qt/QMapboxGLConfig.cmake.in"
+	"${CMAKE_CURRENT_BINARY_DIR}/QMapboxGLConfig.cmake"
+	INSTALL_DESTINATION ${CMAKECONFIG_INSTALL_DIR}
+	PATH_VARS CMAKE_INSTALL_PREFIX CMAKE_INSTALL_INCLUDEDIR
+	CMAKE_INSTALL_LIBDIR NO_CHECK_REQUIRED_COMPONENTS_MACRO)
+
+write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/QMapboxGLConfigVersion.cmake
+	VERSION ${MBGL_QT_VERSION}
+	COMPATIBILITY AnyNewerVersion)
+
+install(EXPORT QMapboxGLTargets
+	DESTINATION ${CMAKECONFIG_INSTALL_DIR}
+	COMPONENT development)
+
+install(FILES
+	"${CMAKE_CURRENT_BINARY_DIR}/QMapboxGLConfig.cmake"
+	"${CMAKE_CURRENT_BINARY_DIR}/QMapboxGLConfigVersion.cmake"
+	DESTINATION ${CMAKECONFIG_INSTALL_DIR}
+	COMPONENT development)
+
 # FIXME: Because of rapidjson conversion
 target_include_directories(
     qmapboxgl
@@ -209,7 +252,7 @@ target_include_directories(
 
 target_include_directories(
     qmapboxgl
-    PUBLIC ${PROJECT_SOURCE_DIR}/platform/qt/include
+    PRIVATE ${PROJECT_SOURCE_DIR}/platform/qt/include
 )
 
 target_compile_definitions(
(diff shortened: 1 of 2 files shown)


rinigus:
Looks like mbgl includes are not installed. Will look into that as well.

kiibimees:
Would be nice if all install commands are wrapped with MBGL_QT_STATIC check. This flag can be used to build and install Qt plugin as one library that has Maplibre linked in - when including Maplibre as a git submodule in the same build tree. Or are there any other ideas how to build Qt plugin in the future?

rinigus:
@kiibimees : I haven't checked with static builds, but I would expect that to work for static as well. Why do you want to limit it to static builds only?

For background: I am using qmapboxgl directly through unofficial QML bindings, without Qt Location plugins. So, I am using dynamic version of the lib myself.

kiibimees:
Yes, I'm sorry - I was a bit hasty. The `MBGL_QT_STATIC` was introduced by @ntadej , to _install_ more libs addition to the main `qmapboxgl` when static linking. It wasn't there in the Mapbox times, when Qt plugin was built as part of QtLocation, and libs were linked in the same build tree, and only the plugin itself was finally _installed_. It would be nice to have a flag that can be used to disable all CMake install commands. Apparently this doesn't exists yet and creating it can easily be a separate task.

ntadej:
Just to add that supporting a working static build is always useful in my opinion as static linking may be easier on some platforms. Probably we could factor out some of those things to not be Qt-specific but in my changes I wanted to touch as little other platform code as possible.

rinigus:
@kiibimees, regarding disabling install flag. Would in that case just not calling `make install` do? If you don't want to install, just run `make`. Or do I miss something?

ntadej:
Personally I'd like to have install also for the static build work correctly.

rinigus:
> Personally I'd like to have install also for the static build work correctly.

yes, that would be reasonable assumption. is it missing something with these changes?

rinigus:
PS: found some errors when configuring at cmake step with `-DMBGL_QT_STATIC=ON`:

```
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "qmapboxgl" which requires target "mbgl-compiler-options" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "qmapboxgl" which requires target "mbgl-core" that is not in any export set.
```

can't make sense out of it, though.

kiibimees:
> @kiibimees, regarding disabling install flag. Would in that case just not calling `make install` do? If you don't want to install, just run `make`. Or do I miss something?

Well, the trouble is that in case we link to the Maplibre in one build tree that builds a library itself (for example Qt plugin, inside one big QtLocation CMake build tree), then we want to install the plugin, but only the plugin, because Maplibre is statically linked to the pluigin lib file.

rinigus:
@kiibimees: understood. then it would be easier to disable it here if requested in config stage

kiibimees:
> PS: found some errors when configuring at cmake step with `-DMBGL_QT_STATIC=ON`:
> 
> ```
> CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "qmapboxgl" which requires target "mbgl-compiler-options" that is not in any export set.
> CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "qmapboxgl" which requires target "mbgl-core" that is not in any export set.
> ```
> 
> can't make sense out of it, though.

It's because you also need to export all dependent targets:

rinigus:
Now, if I set 
```
if(MBGL_QT_STATIC)
    install(TARGETS mbgl-core
            EXPORT QMapboxGLTargets
            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    )
endif()
```

I will get a host of following errors:
```
CMake Error in CMakeLists.txt:
  Target "mbgl-core" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/home/.../maplibre-gl-native/include"

  which is prefixed in the source directory.


CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-extras-kdbush.hpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-supercluster.hpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-shelf-pack-cpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-geojson-vt-cpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-cheap-ruler-cpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-compiler-options" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-boost" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-csscolorparser" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-earcut.hpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-eternal" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-parsedate" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-polylabel" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-protozero" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-unique_resource" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-vector-tile" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-wagyu" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-extras-expected-lite" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-extras-rapidjson" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-geojson.hpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-geometry.hpp" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-optional" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-variant" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mbgl-vendor-nunicode" that is not in any export set.
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "qmapboxgl" which requires target "mbgl-compiler-options" that is not in any export set.
```

Looks like current static lib install is way off already... Tips?

kiibimees:
> Looks like current static lib install is way off already... Tips?
Yes, I might have a good tip. I ran into the same problem when building Qt plugin for iOS. All the libs that you need only on build time, you can link like this:
``target_link_libraries(StaticLibrary PRIVATE "$<BUILD_INTERFACE:PrivateInterface>")``

kiibimees:
Ou, and.... @rinigus , forget my talk about a special flag to disable all install commands. Instead, I can just use EXCLUDE_FROM_ALL when including maplibre subdir and all install targets are ignored.

rinigus:
@kiibimees, thanks! I feel way out of my level with CMake. @PureTryOut - want to take a look?

kiibimees:
But, it seems that it's not enough. It will break the export/install dependencies but it will not combine all dependent libs into one file, and using the lib downstream will give "undefined symbol" linker error.

Described here: https://cmake.org/pipermail/cmake/2018-September/068263.html

kiibimees:
Now, the simplest solution that I found is to [make mbgl-* libs as OBJECT libs, not STATIC libs](https://github.com/kiibimees/maplibre-gl-native/commit/0f6ef7904383474bb77a99ac0f8cd24b01572cdc). In that case it is possible to generate single file **static** lib that has all the other "static" libs bundled in, like my [Qt plugin](https://codereview.qt-project.org/c/qt/qtlocation/+/382087/4/src/plugins/geoservices/maplibregl/CMakeLists.txt) now.

kiibimees:
Now, the simplest solution that I found is to [make mbgl-* libs as OBJECT libs, not STATIC libs](https://github.com/kiibimees/maplibre-gl-native/commit/0f6ef7904383474bb77a99ac0f8cd24b01572cdc). In that case it is possible to generate single file **static** lib that has all the other "static" libs bundled in, like my [Qt plugin](https://codereview.qt-project.org/c/qt/qtlocation/+/382087/4/src/plugins/geoservices/maplibregl/CMakeLists.txt) now.

rinigus:
@kiibimees: disclosure - I am not following what is going on and why STATIC -> OBJECT replacement would help. In general, CMake is relatively new for me and I mainly try to pass by with it. Just hoping that this PR would be more help than harm done in larger picture.

I have tested whether your changes from https://github.com/kiibimees/maplibre-gl-native/commit/0f6ef7904383474bb77a99ac0f8cd24b01572cdc would help. Unfortunately, no, I am still getting long list of 

```
CMake Error: install(EXPORT "QMapboxGLTargets" ...) includes target "mbgl-core" which requires target "mapbox-base-extras-kdbush.hpp" that is not in any export set.
...
```

Some of them can be dropped by adding EXPORT into 
```
if(MBGL_QT_STATIC)
    install(TARGETS mbgl-vendor-csscolorparser
            EXPORT QMapboxGLTargets
            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    )
    install(TARGETS mbgl-vendor-nunicode
            EXPORT QMapboxGLTargets
            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    )
    install(TARGETS mbgl-vendor-parsedate
            EXPORT QMapboxGLTargets
            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    )
endif()
```

but the rest is probably of qt platform somewhere.

platform/qt/qt.cmake
```diff
@@ -9,19 +9,25 @@ message(STATUS "Version ${MBGL_QT_VERSION}")
 
 option(MBGL_QT_STATIC "Build Mapbox GL Qt bindings staticly" OFF)
 option(MBGL_QT_LIBRARY_ONLY "Build only libraries" OFF)
+option(MBGL_WITH_QT_HEADLESS "Build Mapbox GL Qt with headless support" ON)
```
kiibimees, line 12: What's the idea of this headless version? Isn't it basically the same as MBGL_QT_LIBRARY_ONLY?

platform/qt/qt.cmake
```diff
@@ -236,24 +279,25 @@ if(MSVC)
         set_target_properties(mbgl-core PROPERTIES COMPILE_PDB_NAME mbgl-cored)
         set_target_properties(qmapboxgl PROPERTIES COMPILE_PDB_NAME QMapboxGLd)
 
-        install(FILES $<TARGET_FILE_DIR:mbgl-vendor-csscolorparser>/mbgl-vendor-csscolorparserd.pdb CONFIGURATIONS "Debug" DESTINATION lib)
-        install(FILES $<TARGET_FILE_DIR:mbgl-vendor-nunicode>/mbgl-vendor-nunicoded.pdb CONFIGURATIONS "Debug" DESTINATION lib)
-        install(FILES $<TARGET_FILE_DIR:mbgl-vendor-parsedate>/mbgl-vendor-parsedated.pdb CONFIGURATIONS "Debug" DESTINATION lib)
+        install(FILES $<TARGET_FILE_DIR:mbgl-vendor-csscolorparser>/mbgl-vendor-csscolorparserd.pdb CONFIGURATIONS "Debug" DESTINATION "${CMAKE_INSTALL_LIBDIR}")
```
kiibimees, line 288: IMHO these pdb install targets can be remoted totally. They should be automatic in CMake, dependin on the build type and compiler, no?

platform/qt/qt.cmake
```diff
@@ -236,24 +279,25 @@ if(MSVC)
...
 
 install(TARGETS qmapboxgl
-        RUNTIME DESTINATION bin
-        FRAMEWORK DESTINATION lib
-        LIBRARY DESTINATION lib
-        ARCHIVE DESTINATION lib
-        INCLUDES DESTINATION include
-        PUBLIC_HEADER DESTINATION include
+        EXPORT QMapboxGLTargets
+        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
```
kiibimees, line 301: I think that this is not needed to be defined explixitly, if they are set to CMake defaults. The default is set anyway.

platform/qt/qt.cmake
```diff
@@ -201,6 +220,30 @@ if (NOT MBGL_QT_STATIC)
     )
 endif()
 
+include(CMakePackageConfigHelpers)
+set(CMAKECONFIG_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/qmapboxgl/)
+
+configure_package_config_file(
+	"platform/qt/QMapboxGLConfig.cmake.in"
+	"${CMAKE_CURRENT_BINARY_DIR}/QMapboxGLConfig.cmake"
```
kiibimees, line 228: Perhaps putting all CMake package config files somewhere in more general, not under binary dir, is a bit more logical. "PACKAGE_DIR/cmake" or something. Then both libs/bin and include can be referenced from the same general place.

kiibimees:
I'm also a bit new to CMake but as far as I understand, this is how CMake is built. If you want to build and export a single file *static* library like this:
```
add_library(qmapboxgl STATIC)
```

then you have to link all dependency libraries only at build time, by using CMake generator expressions:

```
target_link_libraries(
    qmapboxgl
    PRIVATE
        Qt${QT_VERSION_MAJOR}::Core
        Qt${QT_VERSION_MAJOR}::Gui
        "$<BUILD_INTERFACE:mbgl-compiler-options>"
        "$<BUILD_INTERFACE:mbgl-core>"
)
```

And, it's very important that the libraries are OBJECT libraries, and none of these object libraries have no extra dependent OBJECT library. If they have, you need to link them all directly.

```
target_link_libraries(
    qmapboxgl
    PRIVATE
        Qt${QT_VERSION_MAJOR}::Core
        Qt${QT_VERSION_MAJOR}::Gui
        "$<BUILD_INTERFACE:mbgl-core>"
        "$<BUILD_INTERFACE:mbgl-vendor-parsedate>"
        "$<BUILD_INTERFACE:mbgl-vendor-nunicode>"
        "$<BUILD_INTERFACE:mbgl-vendor-csscolorparser>"
)
```

...otherwise you get "undefined symbol" on linking when consuming that library, because objects of recursive dependency libraries are not copied in to the final library file.

Import/export definition of classes is also important because right now symbols are not exported when buildin static lib:
```
#if !defined(QT_MAPBOXGL_STATIC)
#  if defined(QT_BUILD_MAPBOXGL_LIB)
#    define Q_MAPBOXGL_EXPORT Q_DECL_EXPORT
#  else
#    define Q_MAPBOXGL_EXPORT Q_DECL_IMPORT
#  endif
#else
#  define Q_MAPBOXGL_EXPORT
#endif
```
...this way it cannot be used in a CMake package because downstream linking is not in the same build tree.

I make some some improvements into @ntadej  and @rinigus code and tested it with a separate 'client' app that used CMake package. Have a look here:
https://github.com/kiibimees/maplibre-gl-native/commit/f4748a92190ea0556d8630b1979f6fbda9c3b354

This will now create one single file library file. Both fot static and shared, and no need to install the dependent libs anymore.

platform/qt/qt.cmake
rinigus: Headless requires OpenGL and widgets. Library only allows you to build a version which is running on GLES. This is important for platforms without QtWidgets or OpenGL - such as some mobile Linux distributions.

platform/qt/qt.cmake
rinigus: As far as I remember, this is temporary destination. The install will put it into the correct location on your filesystem.

rinigus:
@kiibimees: thank you very much for your help and review! I will have to look into your suggestions a bit more. Replied to few questions that I knew the answer for from the top of my head, others would need a bit of work on my side. I'll come back to it and will respond to you, hopefully tonight.

platform/qt/qt.cmake
kiibimees: No, the comment was was this line `INSTALL_DESTINATION ${CMAKECONFIG_INSTALL_DIR}`.

`CMAKECONFIG_INSTALL_DIR` is set to `${CMAKE_INSTALL_LIBDIR}/cmake/qmapboxgl/` but CMake cannot find the files when starting from top package dir. It starts to search like this: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

platform/qt/qt.cmake
kiibimees: For some reason I still don't undestand this "headless" business :). Can you provide a few simple examples for all three (none / headless / library) usage? I can think of only three:

1. Running the Maplibre Qt demo app directly from this reposotory.
2. Including Maplibre into some other build tree (as a git submodule, for example, like QtLocation plugin).
3. Building a library, making a package, and making find_package from some other applicationäs build tree.

kiibimees:
No problem. I started testing my changes on different platforms and found one missing dependency (forgot to add mggl-compiler-options to qmapboxgl). Now, @rinigus , I took you changes, added my changes, and a few more smaller improvements, and the [tip of this branch](https://github.com/kiibimees/maplibre-gl-native/commits/qt6_plugin) now compiles in my case on MSVC, Android and iOS. Also as a part of QtLocation. And, it provides one bundled lib file (which is very important for QtLocarion). And it creates a CMake package (in case on install target) as you wanted.

Hane a look!

platform/qt/qt.cmake
rinigus: Headless component gets on a way if you don't have OpenGL, as some Qt platforms do. Corresponding code: https://github.com/maplibre/maplibre-gl-native/blob/master/platform/qt/src/headless_backend_qt.cpp . Hence I want to be able to drop it, as I do on Sailfish. To your examples:

1. demo app - that I haven't checked thoroughly. as it is using QtWidgets and, in Sailfish case we don't have that either, headless is enabled
2. git submodule. that would depend on what you use in submodule, tbh. I don't know what is in QtLocation plugin precisely and don't know whether it is exposing headless functionality. if you don't use it, there is no need for it and you may want to skip OpenGL requirements as well.
3. separate lib: again, as in 2, it depends on platform and your application. in my case, headless is no go and I will disable it before linking with my QML plugin. For some other users (if we have any), it could be different.

As not many know about the flag, I would suggest to keep it enabled by default.

Hopefully, I replied to you.

rinigus:
@kiibimees: looks very promising! Just running `make install` leads to 

```
-- Install configuration: "Release"
CMake Error at cmake_install.cmake:58 (file):
  file cannot create directory: /usr/cmake.  Maybe need administrative
  privileges.
```

instead of trying to install it to `/usr/lib64/cmake/QMapboxGL/`. When looking on my Linux PC, that would be an expected location for it. `/usr/cmake` is empty over here.

kiibimees:
> instead of trying to install it to `/usr/lib64/cmake/QMapboxGL/`. When looking on my Linux PC, that would be an expected location for it. `/usr/cmake` is empty over here.

Perhaps the problem is that you haven't specified [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html)? To make truely relocatable packages, most of the paths must be relative, and they are relative to this variable. And if not set explicitly, it's default is system dependent. And, note that there is a difference in just installing the libs into the system OS and creating CMake package.

platform/qt/qt.cmake
kiibimees: Aha, you mean that you don't have OpenGL Qt module? Because you must have OpenGL in the system in general.

But, about the `headless_backend_qt.cpp` - why do we need it at all? For tests? I think I'm missing something here.

rinigus:
No, that was specified: `-DCMAKE_INSTALL_PREFIX:PATH=/usr`. You tested it on Windows, Android, and iOS, right? 

CMake config path should be set not to cmake as was done in your patch, but the value from this PR. For example, see  https://github.com/google/snappy/blob/master/CMakeLists.txt#L390 for same setting.

In your branch, it is line https://github.com/kiibimees/maplibre-gl-native/blob/qt6_plugin/platform/qt/qt.cmake#L211

platform/qt/qt.cmake
rinigus: Sorry, missed it last night. On some platforms you may not have OpenGL fully. For those, there is no Qt module either :). Same goes for QtWidgets.

I haven't used it, but I think headless could be used if you need to have a map tile rendered using MapboxGL. Could be handy, actually.

platform/qt/qt.cmake
kiibimees: Aha! I'm sorry. Now I understand the point of putting cmake config files under lib folder. This way the package can be installed to a system lib path and it can act both as system libs and CMake package. I reverted my change. Try that one:
https://github.com/kiibimees/maplibre-gl-native/commit/adeaaff713479ca1524c8adc5299e8d242a444f9

kiibimees:
Aha! I'm sorry. Now I understand the point of putting cmake config files under lib folder. This way the package can be installed to a system lib path and it can act both as system libs and CMake package. I reverted my change. Try that one:
https://github.com/kiibimees/maplibre-gl-native/commit/adeaaff713479ca1524c8adc5299e8d242a444f9

rinigus:
@kiibimees, will continue testing tomorrow. So far, dynamic and static lib install worked, excellent work! Will have to check some other combinations as well.

kiibimees:
@rinigus , did you manage to test?

rinigus:
No, not fully yet. Was planning to do it over this weekend. Sorry, had to finish one other project before .

rinigus:
@kiibimees - I can see that you submitted https://github.com/maplibre/maplibre-gl-native/pull/197 - should I test directly that?

rinigus:
Tested and commented at #197

rinigus:
Closing this as development proceeded in #197 by @kiibimees

Issue #32, opened 2021-02-25
Proposal: improvements to local MBTiles FileSource


brendan-ward:
Context: I'm working on Python bindings to build a self-hosted static renderer similar to [`mbgl-renderer`](https://github.com/consbio/mbgl-renderer) that I built on top of the (now unmaintained?) `mapbox-gl-native` NodeJS bindings and was starting to build up a similar FileSource for local mbtiles when I discovered it was already implemented here (thanks!).

I use a somewhat different approach to resolving local mbtiles files to tileJSON and tiles: e.g., `"mbtiles://land/{z}/{x}/{y}.pbf"` => `/<mbtiles root path>/land.mbtiles` tile request.
([style JSON example](https://github.com/consbio/mbgl-renderer/blob/master/tests/fixtures/example-style-mbtiles-tiles-vector.json))

though I am open to other alternatives of resolving this that don't require full filesystem paths in the style JSON files; an approach like is used for assets is more appropriate.  Style JSON strings are sent by client applications to the static renderer; the root path containing the mbtiles files is set as a property for running the static renderer server.

It looks like the current implementation only resolves absolute file paths?  I couldn't find a full request URL or test that would reveal this in the codebase, but from inspection of the code a tile request looks something like `/<absolute mbtiles file path>?file={x}/{y}/{z}.pbf`

Is there more documentation / examples for this FileSource anywhere?

I'd like to make the following changes to the MBTiles FileSource, but wanted to be sure that the overall direction seems reasonable:
* rename `MaptilerFileSource` => `MBTilesFileSource`
* move implementations of tileJSON construction and tile unzipping to separate utility files
* add property to `ResourceOptions` to set root path containing local MBTiles files, similar to asset path
* make resolving tileJSON / tile request URLs more configurable
* add tests for `MBTilesFileSource` and a couple small mbtiles files as test fixtures; I don't see that there are any tests now, so it's a bit hard to know if the above changes would break any downstream usage

petr-pokorny-1:
Good ideas. Unfortunately we currently don't have any documentation for this FileSource, but you can check https://github.com/maplibre/maplibre-gl-native/issues/17

wipfli:
This could be related to relative paths in the style spec. See https://github.com/maplibre/maplibre-gl-js/pull/645

PR #198, opened 2021-12-13
Rename MaptilerFileSource to MBTilesFileSource

brendan-ward:
This achieves the first part of #32, renaming `MaptilerFileSource` to `MBTilesFileSource`, since this file source is about working with local `mbtiles` files (nothing specific to Maptiler).  Does this need to be listed as a breaking change?

Includes minor cleanup to ensure it is declared once and leverage the existing constant `MBTILES_PROTOCOL`.


brendan-ward:
What is needed in order for this PR to proceed?  Do we need to alias `MBTilesFileSource` to `MaptilerFileSource` so as to avoid breaking downstream users that use the previous class?

ntadej:
Sorry for the delay in this one. Do we actually expect many clients of it actually? I'm surprised that none of the platfom needed to be updated. It also does not seem like a public API to me. I think we're fine to rename.

May I ask you to rebase so we run the latest tests? Then I think it would be OK to be merged.

brendan-ward:
Correct: it is not a public API, just an internal class.  I believe this was donated by Maptiler when `maplibre-gl-native` was originally forked from Mapbox and hasn't been more broadly used since then.  For building a static rendering engine, it is quite handy, since it lets us use MBtiles files directly on disk.

ntadej (approved):

PR #205, opened 2022-01-10
Update benchmark to v1.6.0

birkskyum:


roblabs (changes requested):
The check for the iOS & macOS workflow failed for when `benchmark` changed.  I've submitted a re-run of the workflow to see it were an anomaly.  We'll need to monitor and see if it passed.

roblabs:
@birkskyum  — 

The Build Checks via GitHub Actions are only done per _branch_ per _fork_.  There currently is no GitHub Action with all of your PRs for a single GitHub Action.

I recommend that you merge all your dependency related PRs into a single branch (call it `BirksDependencyUpgrades` or whatever), then once they are in a single branch on your fork, then the GitHub Actions will fire for iOS & Android.  Once that is done,  then I can verify that each and every dependency change will indeed work when it hits master.  Right now I can only verify each individually. 

For example,

This build check failed
* https://github.com/maplibre/maplibre-gl-native/actions/runs/1679078607/attempts/2

And the rebuilds I triggered were successful
* https://github.com/maplibre/maplibre-gl-native/actions/runs/1679078607/attempts/3

But the problem is that is only for a single dependency of the many PRs that you submitted.  The way the PRs came in (PR # 205-214), there is no way to build all of the changes until it hits `main`.

*Note Branch name for #205*
image

---

*Note Branch name for #206*
image

birkskyum:
Okay, i will do that. I always try to keep the pr's small, but with the build system queueing up it's easier to have it together. Thanks @roblabs

ntadej:
We could also force the rebase on MRs. It's a bit more tedious to merge many in a row, but one would always test compared to the latest version.

PR #216, opened 2022-01-11
update c and c++ language standards

birkskyum:
Update language standard of C from 1999 to 2011
Update language standard of C++ from 2014 to 2017


ntadej:
Maybe it's due to my job, but can you try to make the descriptions a bit more verbose (it might help with searching later) 😊 

I like the idea though and I think this should not break compatibility with any of the clients.

birkskyum:
> Maybe it's due to my job, but can you try to make the descriptions a bit more verbose (it might help with searching later) 😊 
> 
> I like the idea though and I think this should not break compatibility with any of the clients.

Hi Tadej, is that verbose enough, or did you have something else in mind?

roblabs (approved):
Checks pass for macOS, iOS & Android

PR #217, opened 2022-01-14
Feature - Bring back node support

jutaz:
This PR brings back support for node.js later versions, as well as actually making it work. Maps now render, properly, using node v10, v12, and v16 (these are the ones I've tested, sure others work too).

## Testing

A good way to test is to build this on your platform, and then use sample code from `platform/node/README.md` to render the map. Additionally, here's a simple way to build on macos (x86 at the moment):
```sh
#!/usr/bin/env bash

set -e
set -o pipefail

# Install these dependencies, if they aren't there already.
# brew install cmake ccache glfw ninja pkgconfig qt chargepoint/xcparse/xcparse libuv
# brew cask install google-cloud-sdk
# pip install ansi2html scipy

cmake . -B build -G Ninja -DCMAKE_OSX_DEPLOYMENT_TARGET=12.0 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON

sed -i -e 's/$(ARCHS_STANDARD)/x86_64/g' build/build.ninja
sed -i -e 's/-arch arm64e/-arch x86_64/g' build/build.ninja

cmake --build build -j $(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null)
```

## Questions

What would it take to build and publish `@maplibre/maplibre-gl-native` to NPM, equivalent to what exists [from Mapbox](https://www.npmjs.com/package/@mapbox/mapbox-gl-native) today? 

## Credits

It's been inspired and heavily relied on work in this fork: https://github.com/mnutt/maplibre-gl-native/tree/node-14, specifically the `libuv` integration.


platform/linux/linux.cmake
```diff
@@ -8,6 +8,34 @@ find_package(X11 REQUIRED)
 
 pkg_search_module(LIBUV libuv REQUIRED)
 
+add_library(mbgl-platform-loop STATIC)
```
ntadej, line 11: You add this only to linux but other parts depend on it, so we should be a bit more careful I guess.

We should also make sure this does not interfere with the Qt build.

platform/node/README.md
```diff
@@ -36,9 +36,9 @@ var mbgl = require('@mapbox/mapbox-gl-native');
 var sharp = require('sharp');
 
 var options = {
-  request: function(req, callback) {
+  request: function(req) {
     fs.readFile(path.join(__dirname, 'test', req.url), function(err, data) {
-      callback(err, { data: data });
+      req.respond(err, { data });
```
brendan-ward, line 41: Can you explain a little bit more on the nature of this change?  This is something that will impact downstream users, but the nature of `req` in this context isn't immediately obvious - compared to a simple callback.

brendan-ward (commented):
Thank you for working on this @jutaz !! 👏 

We maintain a [library / server](https://github.com/consbio/mbgl-renderer) based on the Mapbox GL JS bindings, and have been hopeful that someone would be able to made appropriate updates here so we can move to newer NodeJS versions.

I had one question below that impacts the request handlers that users pass in as part of the options, but otherwise this looks like it should be a fairly seamless update for end-users, once the packaging gets worked out.  Is that correct, or are there other interface changes here that users should be aware of?  I'm assuming the `libuv` work here is entirely internal and doesn't impact use as a NodeJS package, correct?

(not qualified to review the NodeJS C++ bindings...)

platform/node/README.md
jutaz: Yeah, so this was actually not necessary - I’ll revert this back 👍

jutaz:
@brendan-ward yeah, this change should mostly be free from any client-side changes. My most important thing is to make sure this works in existing systems (which is why I’m making this change in the first place).

platform/linux/linux.cmake
jutaz: FWIW it really shouldn't - it just introduces a new dep, `libuv`, but that's about it.

ntadej:
Hi @jutaz, I'm still a bit sceptical that you covered all of the CMake changes. Do you mind that I try to setup a test also for the Qt build over the weekend and then we re-run that everything works as expected?

jutaz:
Sure thinkg @ntadej! Please do, I'll appreciate it

roblabs:
I am interested in keeping the TileServer-GL stuff working, so this is great to see progress here.

My notes and feedback based on attempting to test Node bindings using the notes from https://github.com/maplibre/maplibre-gl-native/pull/217#issue-1103730360 (above)

Feedback:

- [ ] I am unable to build. *Node bindings build log* below.  It may be an issue with local missing dependencies?!?
  * The GitHub CI YAML only uses `brew install libuv`
  * Your notes mention others (qt, xcparse, pip):  are they required for a minimal node build?  I'm just looking out for future Node developers.
- [x] Build CI checks seem to pass for iOS & Android.  Likely nothing broken in the C++ code(s):  https://github.com/maplibre/maplibre-gl-native/pull/217/checks
- [ ] Please update `platform/node/DEVELOPING.md` with the your compile/brew/build notes.
- [ ] The test notes from `platform/node/README.md` mention `npm install @mapbox/mapbox-gl-native`, but doesn't that install the Mapbox version from NPM?
- [ ] In your own development, is there an Xcode Project that you launch?  If so, can add documentation on how to launch Xcode to do development.  E.g, `xed platform/node/path/XcodeProj`

---

<details style="border: 1.618px solid gray; border-radius: 2.78px;"><summary>Node bindings build log ❌</summary> 

<pre>
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- The C compiler identification is AppleClang 13.0.0.13000029
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Configuring GL-Native with OpenGL renderer backend
-- Found OpenGL: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/System/Library/Frameworks/OpenGL.framework   
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "0.29.2") 
-- Checking for one of the modules 'glfw3'
-- [Node.js] Downloading Node.js version list...
-- [Node.js] Downloading Nan 2.10.0...
-- [Node.js] Downloading headers for Node v17.4.0...
-- [Node.js] Downloading headers for Node v16.13.2...
-- [Node.js] Downloading headers for Node v15.14.0...
-- [Node.js] Downloading headers for Node v10.24.1...
-- [Node.js] Downloading headers for Node v8.17.0...
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/build
[1/618] Building CXX object test/CMakeFiles/mbgl-test.dir/util/bounding_volumes.test.cpp.o
[2/618] Building CXX object test/CMakeFiles/mbgl-test.dir/util/dtoa.test.cpp.o

: snip :

[321/618] Building C object CMakeFiles/mbgl-core.dir/platform/darwin/src/native_apple_interface.m.o
[322/618] Building CXX object platform/node/CMakeFiles/mbgl-node.abi-102.dir/src/node_logging.cpp.o
FAILED: platform/node/CMakeFiles/mbgl-node.abi-102.dir/src/node_logging.cpp.o 
ccache /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -DBUILDING_NODE_EXTENSION -DGL_SILENCE_DEPRECATION -DMODULE_NAME=mbgl_node -DRAPIDJSON_HAS_STDSTRING=1 -D_DARWIN_USE_64_BIT_INODE=1 -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -Dmbgl_node_abi_102_EXPORTS -I/Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/include -I/Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/platform/default/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/build/headers/node/v17.4.0 -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/build/headers/nan/2.10.0 -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/extras/expected-lite/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/geojson-vt-cpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/geojson.hpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/geometry.hpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/jni.hpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/optional -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/pixelmatch-cpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/shelf-pack-cpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/supercluster.hpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/variant/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/deps/cheap-ruler-cpp/include -isystem /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/vendor/mapbox-base/extras/rapidjson/include -g -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk -mmacosx-version-min=10.11 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -std=c++11 -std=c++14 -MD -MT platform/node/CMakeFiles/mbgl-node.abi-102.dir/src/node_logging.cpp.o -MF platform/node/CMakeFiles/mbgl-node.abi-102.dir/src/node_logging.cpp.o.d -o platform/node/CMakeFiles/mbgl-node.abi-102.dir/src/node_logging.cpp.o -c /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/platform/node/src/node_logging.cpp
In file included from /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/platform/node/src/node_logging.cpp:1:
In file included from /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/platform/node/src/node_logging.hpp:10:
In file included from /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/build/headers/nan/2.10.0/nan.h:202:
In file included from /Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/build/headers/nan/2.10.0/nan_converters.h:67:
/Users/roblabs/Documents/github/mapbox/tmp/robinpowered/maplibre-gl-native/build/headers/nan/2.10.0/nan_converters_43_inl.h:22:1: error: no viable conversion from 'Local<v8::Context>' to 'v8::Isolate *'
X(Boolean)
^~~~~~~~~~

</pre>
</details>
 
---
 
*Build Environment used for Node testing*

`xcodebuild -version`
> Xcode 13.2.1
Build version 13C100

`sw_vers`
> ProductName:	macOS
ProductVersion:	11.6.2
BuildVersion:	20G314

`uname -a`
> Darwin roblabs.local 20.6.0 Darwin Kernel Version 20.6.0: Wed Nov 10 22:23:07 PST 2021; root:xnu-7195.141.14~1/RELEASE_X86_64 x86_64

jutaz:
> Your notes mention others (qt, xcparse, pip): are they required for a minimal node build? I'm just looking out for future Node developers.

As far as I can tell - yes. We might be able to cut that down, but I'd like to make that a follow up rather than a requirement from a get-go. I assume  you've been able to run the following commands to get the deps?
```
brew install cmake ccache glfw ninja pkgconfig qt chargepoint/xcparse/xcparse libuv
brew cask install google-cloud-sdk
pip install ansi2html scipy
```


> In your own development, is there an Xcode Project that you launch? If so, can add documentation on how to launch Xcode to do development. E.g, xed platform/node/path/XcodeProj

Nope no XCode for this stuff, just pure C++ compilation.


I also have a handy `Dockerfile` if you want to build on Linux:

```Dockerfile
FROM debian:bullseye-slim AS builder

RUN apt-get indextargets
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get update     && apt-get dist-upgrade -y     && apt-get autoremove --purge -y && apt-get clean
RUN apt-get install -y       ca-certificates       git       gzip       ssh       tar       colorized-logs       python3-scipy     && apt-get autoremove --purge -y && apt-get clean
RUN apt-get install -y       awscli       ccache       clang-9       clang-format-9       clang-tidy-9       cmake       doxygen       fonts-noto       g++-10       libc++-9-dev       libc++abi-9-dev       mesa-common-dev       ninja-build       nodejs       npm       pkg-config       python3-bs4       python3-pip       python3-requests       python3-git       python3-github       software-properties-common       valgrind       xvfb       zip     && apt-get autoremove --purge -y && apt-get clean
RUN pip3 install cmake-format==0.5.5
RUN apt-get install -y       libcurl4-openssl-dev       libgl1-mesa-dev       libgles2-mesa-dev       libglfw3-dev       libicu-dev       libjpeg62-turbo-dev       libpng-dev       libuv1-dev       zlib1g-dev     && apt-get autoremove --purge -y && apt-get clean
RUN git clone --recursive https://github.com/google/bloaty.git /tmp/bloaty     && cd /tmp/bloaty && git checkout 3cf5c3feca15 && cmake . && make -j $(nproc) bloaty && cp bloaty /usr/local/bin/     && rm -rf /tmp/bloaty && rm -rf ~/.ccache
# RUN apt-get install -y       qdoc-qt5       qt5-default     && apt-get autoremove --purge -y && apt-get clean
# RUN apt-get install -y       coreutils       curl       openjdk-8-jdk-headless       unzip     && apt-get autoremove --purge -y && apt-get clean
RUN apt-get install -y g++-10     && apt-get autoremove --purge -y && apt-get clean
# RUN apt-get install -y       python-yaml     && apt-get autoremove --purge -y && apt-get clean
RUN  /usr/sbin/update-ccache-symlinks

RUN mkdir -p /src
WORKDIR /src

FROM builder AS buildable

COPY . /src/

RUN cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10

RUN ccache --zero-stats --max-size=2G
RUN cmake --build build -j $(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null)
RUN ccache --show-stats
```

.github/workflows/macos-ci.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           brew list ccache || brew install ccache
           brew list pkg-config || brew install pkg-config
           brew list glfw3 || brew install glfw3
+          brew list libuv || brew install libuv
```
ntadej, line 49: Is this a new dependency or just a missing one and loaded implicitly by some other?

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
+name: node-ci
+
+on:
+  workflow_dispatch:
+  push:
+    tags-ignore:
+      - '**'
+    paths-ignore:
+      - "platform/android/**"
+      - "platform/ios/**"
```
ntadej, line 10: You should probably also ignore Qt here.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+    strategy:
+      matrix:
+        os:
+          - macos-11.0
+          - macos-10.15
+          - ubuntu-20.04
+          - ubuntu-18.04
+    runs-on: ${{ matrix.os }}
+    env:
+      BUILDTYPE: Debug
```
ntadej, line 33: Should we build Release builds in the CI? (I have to check other pipelines what they do)

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+        if: runner.os == 'macOS'
+        env:
+          HOMEBREW_NO_AUTO_UPDATE: 1
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+        run: |
+          brew list cmake || brew install cmake
+          brew list ccache || brew install ccache
+          brew list ninja || brew install ninja
+          brew list pkgconfig || brew install pkgconfig
+          brew list qt || brew install qt
```
ntadej, line 70: I know this was already asked, but why exactly is Qt needed?

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+          brew list chargepoint/xcparse/xcparse || brew install chargepoint/xcparse/xcparse
+          brew list pkg-config || brew install pkg-config
+          brew list google-cloud-sdk || brew install --cask google-cloud-sdk
```
ntadej, line 73: I'm also suprised by this one.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Clear ccache statistics
+        run: |
+          ccache --zero-stats
+          ccache --max-size=2G
+          ccache --show-stats
+
+      - name: CMake
+        if: runner.os == 'macOS'
+        run: |
+          cmake . -B build -G Ninja -DCMAKE_OSX_DEPLOYMENT_TARGET=12.0 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON
```
ntadej, line 121: Why set deployment target to such a new OS? This will prevent binaries running on older versions.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+          sed -i -e 's/$(ARCHS_STANDARD)/x86_64/g' build/build.ninja
+          sed -i -e 's/-arch arm64e/-arch x86_64/g' build/build.ninja
+
+      - name: CMake
+        if: runner.os == 'Linux'
+        run: |
+          cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
```
ntadej, line 128: You install `clang`, but you then use `gcc`, please cleanup 😊

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+      # On PRs make sure that the npm package can be packaged.
+      - name: Pack
+        if: github.ref != 'refs/heads/master'
+        run: |
+          npm pack
+
+      - name: Publish package
+        if: github.ref == 'refs/heads/master'
+        run: |
+          npm publish
```
ntadej, line 151: Where is this published?

circle.yml
```diff
@@ -283,7 +283,7 @@ commands:
       - run:
           name: Prepare macOS
           command: |
-            brew install cmake ccache glfw ninja pkgconfig qt chargepoint/xcparse/xcparse
+            brew install cmake ccache glfw ninja pkgconfig qt chargepoint/xcparse/xcparse libuv
```
ntadej, line 286: FYI, we do not use Circle CI anymore, this file should be removed at some point.

platform/macos/macos.cmake
```diff
@@ -9,6 +9,9 @@ set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[vari
 
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES)
 
+find_package(PkgConfig REQUIRED)
+pkg_search_module(LIBUV libuv REQUIRED STATIC)
```
ntadej, line 13: OK, I see it's static here.

platform/node/src/node_request.cpp
```diff
@@ -2,6 +2,7 @@
 #include "node_map.hpp"
 #include <mbgl/storage/response.hpp>
 #include <mbgl/util/chrono.hpp>
+#include <iostream>
```
ntadej, line 5: Debugging leftover?

platform/macos/macos.cmake
```diff
@@ -20,18 +23,46 @@ if(MBGL_WITH_OPENGL)
...
         mbgl-core
         PRIVATE OpenGL::GL
     )
 endif()
 
+add_library(mbgl-platform-loop STATIC)
+
 target_sources(
-    mbgl-core
+    mbgl-platform-loop
```
ntadej, line 38: Why do you need to split this for other platforms than node? Is it because node uses macOS platform when on this OS?

ntadej (changes requested):

.github/workflows/macos-ci.yml
jutaz: It's implicitly loaded due to `node-loop`, which is built as part of the whole package.

.github/workflows/node-ci.yml
jutaz: Good point 👍

.github/workflows/node-ci.yml
jutaz: Others are using `Debug` too. I think we should build debug builds for PRs, for `master` and tag builds build `Release`

.github/workflows/node-ci.yml
jutaz: It's basically because everything is built as a single package. It's not a direct dependency of the `node` build, but it seems to be coupled together.

.github/workflows/node-ci.yml
jutaz: Yeah, it might be that I pulled this from something that Mapbox had in their circleCI pipelines. Will see if I can remove it

.github/workflows/node-ci.yml
jutaz: Yah, good Q. I think this also might have to use the current OS version - will test it out.

.github/workflows/node-ci.yml
jutaz: Ha, good point 👍

.github/workflows/node-ci.yml
ntadej: Then I consider this a bug in the setup. Each platfom should build independently.

circle.yml
jutaz: Yup, for whatever reason I thought otherwise, and took a sec to realize we're using GH pipelines here. Will revert changes to this file 👍

platform/node/src/node_request.cpp
jutaz: Indeed! Will remove

.github/workflows/macos-ci.yml
ntadej: Node dependencies should not be built as part of macOS build.

.github/workflows/node-ci.yml
jutaz: Currently nowhere. There's quite a lot of setup needed to to before we can publish - I really only created this to demonstrate how this would be done. Will remove before merging.

jutaz:
Really appreciate the review @ntadej! Will clean it up, thanks.

.github/workflows/node-ci.yml
ntadej: This is the minimum deployment target so the minimum OS we want to support. Either you should just omit it for now or figure out the minimum version this is still supposed to run on.

ntadej:
Looking closer and closer at this repo, the build is in dire need of cleanup. I wonder if we should split this MR into C++ part and the build part, but this is just my opinion. At least nothing in other platforms should be touched in my opinion.

jutaz:
I agree @ntadej, but maybe we can also get this in and _then_ clean it all up? That way at least we can get this in a working state, and then do the build system overhaul.

FWIW from what I can tell this doesn’t negatively affect already working ports, such as android or ios. I can test Qt too, but I’m pretty sure that’ll be in working order too.

ntadej:
> I agree @ntadej, but maybe we can also get this in and _then_ clean it all up? That way at least we can get this in a working state, and then do the build system overhaul.
> 
> FWIW from what I can tell this doesn’t negatively affect already working ports, such as android or ios. I can test Qt too, but I’m pretty sure that’ll be in working order too.

If you can motivate why you need an additional static library also for platforms other than node, then I guess we could merge this in. I appreciate your changes, but I am a bit confused why `mbgl-platform-loop` is needed.

platform/node/src/node_conversion.hpp
```diff
@@ -79,23 +79,23 @@ class ConversionTraits<v8::Local<v8::Value>> {
...
+        return Nan::To<bool>(value).ToChecked();
     }
 
     static optional<float> toNumber(const v8::Local<v8::Value>& value) {
         Nan::HandleScope scope;
         if (!value->IsNumber()) {
             return {};
         }
-        return value->NumberValue();
+        return (float)Nan::To<double>(value).ToChecked();
```
ntadej, line 90: Please use `static_cast<float>`.

platform/node/src/node_expression.cpp
```diff
@@ -157,8 +160,8 @@ struct ToValue {
     v8::Local<v8::Value> operator()(const std::vector<Value>& array) {
         Nan::EscapableHandleScope scope;
         v8::Local<v8::Array> result = Nan::New<v8::Array>();
-        for (std::size_t i = 0; i < array.size(); i++) {
-            result->Set(i, toJS(array[i]));
+        for (std::uint32_t i = 0; i < array.size(); i++) {
```
ntadej, line 163: Note that static analysers would still find this change offensive as `size_t` can be 64 bit, but it's fine for now.

ntadej (changes requested):

ntadej:
Hi @jutaz, I have just one additional comment (missed it in the first pass, sorry for that).

Thank you very much to avoid touching other platforms but I think I see now why it may have been needed. Still, I think for the current iteration it is better to keep it small.

Note that given that we also have to add Qt platform to the CI, we may want to reduce the matrix a bit and test only on latest macOS and latest LTS Ubuntu, but let's keep it as is for now and see later if too many parallel pipelines cause issues.

jutaz:
> Note that given that we also have to add Qt platform to the CI, we may want to reduce the matrix a bit and test only on latest macOS and latest LTS Ubuntu, but let's keep it as is for now and see later if too many parallel pipelines cause issues.

Yeah; I think it’d be a great idea to optimize the build matrix overall and maybe reduce that further. I’d like to take a look at that separately, however.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+
+      - name: Prepare ccache
+        run: ccache --clear
+
+      - name: Cache ccache
+        uses: actions/cache@v2
+        env:
+          cache-name: ccache-v1
+        with:
+          path: ~/.ccache'
```
ntadej, line 88: There's a typo here

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+          pip3 install cmake-format==0.5.5
+          sudo apt-get install -y libcurl4-openssl-dev libgl1-mesa-dev libgles2-mesa-dev libglfw3-dev libicu-dev libjpeg-turbo8-dev libpng-dev libuv1-dev zlib1g-dev g++-10
+          git clone --recursive https://github.com/google/bloaty.git /tmp/bloaty && cd /tmp/bloaty && git checkout 3cf5c3feca15 && cmake . && make -j $(nproc) bloaty && cp bloaty /usr/local/bin/ && rm -rf /tmp/bloaty && rm -rf ~/.ccache
+          /usr/sbin/update-ccache-symlinks
+
+      - name: npm ci
+        run: npm ci --ignore-scripts
+
+      - name: Prepare ccache
+        run: ccache --clear
```
ntadej, line 81: On mac, something like
```
ccache --set-config cache_dir=~/.cache
```
is also needed (testing in my MR).

platform/node/DEVELOPING.md
```diff
@@ -1,19 +1,71 @@
...
 the [macOS](../macos/INSTALL.md#requirements) or [Linux](../linux/README.md#prerequisites) install documentation, depending
 on the target platform.
 
+First you'll need to install dependencies:
+
+
+#### MacOS
+
+```bash
+brew list cmake || brew install cmake
```
brendan-ward, line 17: Suggestion: remove `brew list <pkg>` from each of following and then use a single `brew install` command so that each package is on a new line:
```bash
brew install \
   cmake \
   ninja \
   ...
```

platform/node/DEVELOPING.md
```diff
@@ -1,19 +1,71 @@
...
+brew list ccache || brew install ccache
+brew list ninja || brew install ninja
+brew list pkgconfig || brew install pkgconfig
+brew list chargepoint/xcparse/xcparse || brew install chargepoint/xcparse/xcparse
+brew list pkg-config || brew install pkg-config
+brew list glfw3 || brew install glfw3
+brew list libuv || brew install libuv
+pip install ansi2html scipy
```
brendan-ward, line 25: Here and for Linux build: it seems odd to install python dependencies to build a NodeJS module; I'd be cautious about including them here unless strictly required, and if so, please add a note indicating why.  

How can you tell that they are required?  I was able to build the libs without these, and I've not had to include them in a Docker container for either building `maplibre-gl-native` or using `mapbox-gl-native` in downstream packages.

platform/node/DEVELOPING.md
```diff
@@ -1,19 +1,71 @@
...
+```
+
+#### Linux
+
+```bash
+sudo apt-get install -y ca-certificates git gzip tar python3-scipy
```
brendan-ward, line 31: Suggestion: please put each package on a newline:

```bash
sudo apt-get update && sudo apt-get install -y \
    ca-certificates \
    gzip \
    ...
```

Also - some of these packages do not seem strictly necessary?  I'd suggest separating this list into those that are required (build tools, mesa, glfw, etc) vs those that are used only in development (e.g., `valgrind`) vs those that are strictly optional (`fonts-noto`?).  There are a bunch here that I've not had to install into a Docker Debian container in order to build `maplibre-gl-native` (though still may be running into segfaults because of missing deps).

It may be good to note that some of these specific packages are dependent on your version of Linux, and may require a different package version (e.g., g++).

platform/node/DEVELOPING.md
```diff
@@ -1,19 +1,71 @@
...
+```
+
+### Compiling
+
 To compile the Node.js bindings and install module dependencies, from the repository root directory, first run:
 
+#### MacOS
+
+```bash
+cmake . -B build -G Ninja -DCMAKE_OSX_DEPLOYMENT_TARGET=12.0 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON
```
brendan-ward, line 46: I've been able to build on MacOS 10.15; I've left off `-DCMAKE_OSX_DEPLOYMENT_TARGET=12.0`.  Maybe include suggested cmake flags separately from this set of commands, so that users don't confuse this as a minimum required version?

I've also always had to hand-edit `macos.cmake` to comment out the line `set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")` in order to get the first `cmake` command to run (without ill effect on the final build, as far as I can tell); I wasn't able to get this step to work then apply the `sed` commands below.

platform/node/README.md
```diff
@@ -14,7 +15,7 @@ By default, installs binaries. On these platforms no additional dependencies are
 Run:
 
 ```
-npm install @mapbox/mapbox-gl-native
+npm install @maplibre/maplibre-gl-native
```
brendan-ward, line 18: this is likely to confuse users until this package is actually available.  Suggest adding a big warning block at the top of this file to indicate that this is planned (I assume?) but does not yet exist.

platform/node/DEVELOPING.md
```diff
@@ -1,19 +1,71 @@
...
+```
+
+#### Linux
+
+```bash
+sudo apt-get install -y ca-certificates git gzip tar python3-scipy
+sudo apt-get install -y ccache cmake doxygen fonts-noto g++-10 libc++-9-dev libc++abi-9-dev mesa-common-dev ninja-build pkg-config python3-bs4 python3-pip python3-requests python3-git python3-github software-properties-common valgrind xvfb zip
+pip3 install cmake-format==0.5.5
+sudo apt-get install -y libcurl4-openssl-dev libgl1-mesa-dev libgles2-mesa-dev libglfw3-dev libicu-dev libjpeg-turbo8-dev libpng-dev libuv1-dev zlib1g-dev g++-10
+git clone --recursive https://github.com/google/bloaty.git /tmp/bloaty && cd /tmp/bloaty && git checkout 3cf5c3feca15 && cmake . && make -j $(nproc) bloaty && cp bloaty /usr/local/bin/ && rm -rf /tmp/bloaty && rm -rf ~/.ccache
```
brendan-ward, line 35: suggest making `cmake` / `make` a separate command from install of deps, so that it is more obvious to users the commands that are used for making the libs.  Also, what is `bloaty` used for here, and is it necessary?

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,136 @@
...
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Install dependencies
+        if: runner.os == 'macOS'
+        env:
+          HOMEBREW_NO_AUTO_UPDATE: 1
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+        run: |
+          brew list cmake || brew install cmake
```
brendan-ward, line 53: Here and for Linux environment, similar to suggestions below, suggest consolidating this to the minimal set of dependencies required to build on Github.  Here, this should include only the strictly required deps and not development or optional deps.

brendan-ward (commented):
Thanks @jutaz and @ntadej for working on this!

I added suggestions below, mostly about the dependencies required.  I'd suggest going as minimal as possible in this PR (rather than too many here than pare it down later).

I'm trying now to figure out how to use the NodeJS binary package created here in `mbgl-renderer` for testing; our test suite should be helpful for verifying that the NodeJS bindings here are working as expected.  Will report back.

brendan-ward:
I was able to successfully test the NodeJS bindings here against `mbgl-renderer`, using NodeJS 16.13, and all our render tests passed!  💥 

I used `npm link` from root directory here and `npm link @maplibre/maplibre-gl-native` in the root directory for `mbgl-renderer`.

I'm not at all familiar with building and deploying binary NodeJS modules, so take the following with that in mind.

It seems like the root `package.json` here includes things no longer needed to distribute this package.  Also, it seems like `gyp` / `node-gyp` may get in the way, given that the build is now done with cmake?  Since the current install specifically disables user-side building of the package (I think?) via `node-pre-gyp install --fallback-to-build=false`, it seems like the only option is to install a pre-built binary.

So it seems that the `dependencies` listed in `package.json` can be removed, and the remaining `devDependencies` be limited to those used for testing the package?  And the `package.json` slimmed down to just what is required to install rather than build this package?


I also tested deploying the package manually by copying a few files to `.../node_modules/@maplibre-gl-native` under `mbgl-renderer` root and modifying:
1. copied `lib/<correct ABI of my NodeJS>/mbgl.node` to directory above (so that it ended up at `.../node_modules/@maplibre-gl-native/mbgl.node`)
2. copied `package.json` to directory above and edited to change `main` to `index.js`
3. copied `platform/node/index.js` to directory above and modified so that it referred to the above node module: `var mbgl = require('./mbgl');`

Then reran my tests, and all passed.  This suggests that there just needs to be a way to get the compiled binaries out to S3, and correct references to those in the `package.json`.  Thus it seems that `.npmignore` can be expanded so that fewer files are inadvertantly included in the package pushed to NPM (maybe just README, LICENSE, and index.js?).  But - a lot of that can get handled in follow-on PRs, just noting it now.

brendan-ward:
How were you able to determine that `libuv` was required for macos builds?  Is it something that will be needed for other targets than just NodeJS builds?

platform/node/DEVELOPING.md
jutaz: I'd keep it this way to be consistent with existing pipelines: https://github.com/maplibre/maplibre-gl-native/blob/master/.github/workflows/ios-ci.yml#L41-L45

IMO if we were to change them, we should do it all at once, with a consolidated pipeline instead of different per-platform ones.

platform/node/DEVELOPING.md
jutaz: I can remove the deployment target completely tbh - AFAIK it's not required.


> I've also always had to hand-edit macos.cmake to comment out the line set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)") in order to get the first cmake command to run (without ill effect on the final build, as far as I can tell); I wasn't able to get this step to work then apply the sed commands below.

What was the error you were seeing that'd cause you to not build it? Curious, as I think that can be fixed.

platform/node/README.md
jutaz: Good call 👍

jutaz:
> How were you able to determine that libuv was required for macos builds? Is it something that will be needed for other targets than just NodeJS builds?

`libuv` is the event loop that nodejs also uses, which is why it's a dependency of any async native  (i.e. c++) code that wants to plug into node and play nice. It's not a requirement for any other builds (such as Android or iOS).

Also, RE: testing and native dependency setup.

That's something that's not fully fleshed out, since deployment isn't sorted. If you really want to test it on your local machine, you can:
`npm pack` and use then extract the resulting archive into a `node_modules` directory under `@maplibre/maplibre-gl-native`. Currently it works rather well, but I'd suggest waiting for the full build (i.e. published to npm) to be able to test it well.

roblabs (approved):
@jutaz — Based on the following battery of tests that I was able to run locally, approved.  Thanks for supporting FOSS4G!


- [x] I successfully built on a local macOS according to `platform/node/DEVELOPING.md`
- [x] `npm test` a success.
```console
# tests 145
# pass  145
```
- [x] Success on an iOS build:  `cd platform/iOS && make ios-uitest`.  
- [x] reviewed `.cpp` changes; either benign or non-breaking changes.
- [x] @brendan-ward node test results
>  *tested.. against mbgl-renderer, using NodeJS 16.13, and all our render tests passed! 💥*

roblabs:
> @brendan-ward: *I was able to successfully test the NodeJS bindings here against `mbgl-renderer`, using NodeJS 16.13, and all our render tests passed!*

Brendan  — do you have any notes on how to replicate your tests against [consbio/mbgl-renderer](https://github.com/consbio/mbgl-renderer) from my local Node JS bindings?  I'd be happy to run/test my locally built bindings.  My motivation is starting using that repo for static map generation.

brendan-ward:
I think this should get you close to what I did:
* use NodeJS 16.13 (I use nvm to manage node version)
* clone https://github.com/consbio/mbgl-renderer to `<something>/mbgl-renderer` and `npm install` or `npm ci` (I used the latter to force use of locked deps), then `npm uninstall @mapbox/mapbox-gl-native` and modify import in `src/render.js` to point to maplibre instead: `import mbgl from '@maplibre/maplibre-gl-native'` (note: I also just pushed a branch to simplify these steps: https://github.com/consbio/mbgl-renderer/tree/test_maplibre-gl-native)
* cloned this repo to `<something>/maplibre-gl-native` and initialized all the git submodules
* built the NodeJS bindings via cmake, these were output into `maplibre-gl-native/lib/node-<ABI version>/mbgl.node`

Then you have 2 options; use `npm link` in maplibre root directory and then `npm link @maplibre/maplibre-gl-native` in mbgl-renderer root directory, OR

From the  `<something>/maplibre-gl-native` directory:
* `mkdir <something>/mbgl-renderer/node_modules/@maplibre`
* `mkdir <something>/mbgl-renderer/node_modules/@maplibre/maplibre-gl-native`
* `cp lib/node-v93/mbgl.node <something>/mbgl-renderer/node_modules/@maplibre/maplibre-gl-native`
* `cp platform/node/index.js <something>/mbgl-renderer/node_modules/@maplibre/maplibre-gl-native`
* `cp package.json <something>/mbgl-renderer/node_modules/@maplibre/maplibre-gl-native`

Modify the copied `index.js`  to point to the NodeJS bindings:
` var mbgl = require('./mbgl');`

Modify the copied `package.json` to point to `index.js`:
`"main": "index.js",`

Most of that was to test out how small the deployed package needed to be.


Then from within `mbgl-renderer` directory:
```bash
npm run build
npm run test
```

brendan-ward:
Thanks for the updates @jutaz , this looks good to me (note that I can't provide feedback on the C++ `Nan` bindings).  🚀

ntadej:
As 3 reviewers are happy, I think we can go ahead and merge.

Thank you @jutaz for the contribution and for the patience!

jutaz:
Appreciate it! I'll follow up w/ build and release setup 🙌

mnutt:
This is great! I'm glad the initial work was helpful, kudos to you for getting it mergeable.

Really looking forward to getting to a place where people can `npm install` it. In inspecting mapbox's npm package from back in the day, one interesting thing I noticed was how they had statically linked a bunch of libraries. It produces huge downloads but makes it oh so much easier for people to just use without worrying about the myriad dependencies involved.

mnutt:
As far as storing binaries themselves, it _looks_ pretty straightforward? `node-pre-gyp` has renamed to `@mapbox/node-pre-gyp`, but I think as long as you have CI zip up the modules for each platform and store them in an s3 bucket or elsewhere it would work. You might also try `https://www.npmjs.com/package/node-pre-gyp-github` to be able to pull from github releases themselves.

mnutt:
Oh, turns out they wrote a whole article on this: https://github.com/mapbox/node-pre-gyp/wiki/External-libraries

brendan-ward:
Is `node-pre-gyp` even needed to install this via NPM?  My understanding is that now the static Node libs are built using CMake; thus I think that distributing these means having binaries available for specific architectures.

If I recall correctly, it used to be that you needed `node-pre-gyp` to actually compile the libraries on your client machine if it couldn't find one of the pre-compiled binaries, but this was always very brittle.  I remember that Mapbox switched to providing pre-compiled binaries via S3, which resolved a lot of these issues.

So if I'm following along correctly (and might not be, binary Node libs are outside my expertise), I think it is primarily a matter of setting up CI jobs to build the static libs for the supported architectures (which is not many?), and then setting up NPM packaging to find and download the correct binary for the client's achitecture / Node version?  My sense from testing locally above is that a whole lot of the NPM related things in `maplibre-gl-native` are no longer necessary; those are superseded by the CMake build.

mnutt:
`node-gyp` is a nodejs build tool for building native addons; `node-pre-gyp` is a tool mapbox built to manage pushing binaries to s3 and retrieving them from s3 when users `npm install`. My recollection is that `node-gyp` was the potentially flaky part and `node-pre-gyp` is just the push/pull tooling.

Is it worth opening a separate issue to track publishing to npm?

acalcutt:
@mnutt

I was trying last night to make this publish to npm and this ended up working for the publishing piece at https://github.com/acalcutt/maplibre-gl-native/
```
      - name: Publish package
        if: github.ref == 'refs/heads/main'
        run: |
          npm config set //registry.npmjs.org/:_authToken ${NPM_TOKEN}
          npm publish --access public
        env:
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
https://www.npmjs.com/package/@acalcutt/maplibre-gl-native

However, when i tried to use it it tries to pull the binaries from the package.json binary host ( https://mapbox-node-binary.s3.amazonaws.com ), which of course doesn't exist and the npm install would fail. I then tried to change the install command to "node-pre-gyp install --fallback-to-build=true" to allow building from source, which got further but node-gyp would always fail.

My understanding above is if the binaries got uploaded somewhere, the package.json host link could be updated to point to them and the install would be less likely to fail. I tried a but to see if the workflow could upload those binaries somewhere, but I was getting a bit lost at that point.

mnutt:
I started looking at it at https://github.com/maplibre/maplibre-gl-native/compare/main...mnutt:publish-github though it’s not working yet. The CI step creates a release and pushes the prebuilt binary to it. It’s only failing because Ubuntu 18.04 and 20.04 produce the same file name and collide, we only need one.(18.04)

acalcutt:
It actually looks like I was wrong, I was able to use the package I created here ( https://www.npmjs.com/package/@acalcutt/maplibre-gl-native ) . 

I was able to use it in an upgraded Tileserver-GL and it seems like it is working ( https://github.com/acalcutt/tileserver-gl ). The only think is I had to manually install some libraries maplibre-gl-native uses, which didn't all exist in Debian 11 I was running ( I had to install some ubuntu 20 packages , like  libjpeg.so.8 and libicui18n.so.66 )

mnutt:
It looks like you're getting close to getting it building. Did you generate a personal access token and add it as an environment variable for the build? I'm also not 100% sure if we even need `{matrix.os}` based on what mapbox did; it looks like having only the oldest publish might be sufficient?

Just getting it published at all would be great, and then we could work on decreasing dependencies: https://github.com/mapbox/node-pre-gyp/wiki/External-libraries

acalcutt:
Once thing I notice with this 'node-pre-gyp-github' upload is it only uploads a file for one version of node, where the npm publish uploads copies for multiple versions.

For example, My github realease has 
node-v93-linux-x64-Release.tar.gz

But my npm publish creates libs for 
node-v57
node-v59
node-v64
node-v67
node-v72
node-v79
node-v83
node-v88
node-v93
node-v102
node-v108

mnutt:
That's interesting, I see that the libs in those directories are different sizes but wasn't aware `npm publish` actually built against a bunch of different node versions? My understanding would be that we'd probably need a matrix build using different versions of node for each one we wanted to support?

mnutt:
I take that back, the build step uses https://github.com/mapbox/cmake-node-module and should build for all modern versions of node. We may want to set `NODE_MODULE_MINIMUM_ABI`.

METACEO:
Would really enjoy getting my hands on this! If there's any need for experimental testers, please count me in!

mnutt:
I have this working in https://github.com/mnutt/maplibre-gl-native/commits/publish-github. It's doing everything but actually publishing the npm module, but you could try it via `npm install mnutt-maplibre-gl-native@https://github.com/mnutt/maplibre-gl-native.git#publish-github`.

Building for MacOS arm64 (M1) was a little tricky and doesn't actually run the tests but I can confirm the result seems to work well. The linux x64 one seems to work just fine on ubuntu 18.04, but requires lots of dependencies installed (opengl, image formats, etc) and I think we can eventually bundle them. CI should build for nodes 14.x, 16.x, and 18.x.

mnutt:
For comparison, mine is 11MB and mapbox's precompiled version is 158MB. Their linked libraries:

```
	linux-vdso.so.1 (0x00007ffd3affc000)
	libEGL.so.1 => /usr/lib/x86_64-linux-gnu/libEGL.so.1 (0x00007f0fad48c000)
	libGLESv2.so.2 => /usr/lib/x86_64-linux-gnu/libGLESv2.so.2 (0x00007f0fad27a000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f0fad05d000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f0faccbf000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f0fac8b2000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f0fac69a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0fac2a9000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0fae1a9000)
	libGLdispatch.so.0 => /usr/lib/x86_64-linux-gnu/libGLdispatch.so.0 (0x00007f0fabff3000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0fabdef000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0fabbd0000)
```

My linked libraries:

```
	linux-vdso.so.1 (0x00007ffff18b7000)
	libGLX.so.0 => /usr/lib/x86_64-linux-gnu/libGLX.so.0 (0x00007f2f9a3f4000)
	libOpenGL.so.0 => /usr/lib/x86_64-linux-gnu/libOpenGL.so.0 (0x00007f2f9a1c6000)
	libcurl.so.4 => /usr/lib/x86_64-linux-gnu/libcurl.so.4 (0x00007f2f99f44000)
	libjpeg.so.8 => /usr/lib/x86_64-linux-gnu/libjpeg.so.8 (0x00007f2f99cdc000)
	libuv.so.1 => /usr/lib/x86_64-linux-gnu/libuv.so.1 (0x00007f2f99ab6000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2f99897000)
	libX11.so.6 => /usr/lib/x86_64-linux-gnu/libX11.so.6 (0x00007f2f9955f000)
	libpng16.so.16 => /usr/lib/x86_64-linux-gnu/libpng16.so.16 (0x00007f2f9932d000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f2f99110000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f2f98d03000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2f98965000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2f9874d000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2f9835c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2f9b089000)
	libGLdispatch.so.0 => /usr/lib/x86_64-linux-gnu/libGLdispatch.so.0 (0x00007f2f980a6000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2f97ea2000)
	libnghttp2.so.14 => /usr/lib/x86_64-linux-gnu/libnghttp2.so.14 (0x00007f2f97c7d000)
	libidn2.so.0 => /usr/lib/x86_64-linux-gnu/libidn2.so.0 (0x00007f2f97a60000)
	librtmp.so.1 => /usr/lib/x86_64-linux-gnu/librtmp.so.1 (0x00007f2f97844000)
	libpsl.so.5 => /usr/lib/x86_64-linux-gnu/libpsl.so.5 (0x00007f2f97636000)
	libssl.so.1.1 => /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007f2f973a9000)
	libcrypto.so.1.1 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007f2f96ede000)
	libgssapi_krb5.so.2 => /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2 (0x00007f2f96c93000)
	libldap_r-2.4.so.2 => /usr/lib/x86_64-linux-gnu/libldap_r-2.4.so.2 (0x00007f2f96a41000)
	liblber-2.4.so.2 => /usr/lib/x86_64-linux-gnu/liblber-2.4.so.2 (0x00007f2f96833000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f2f9662b000)
	libxcb.so.1 => /usr/lib/x86_64-linux-gnu/libxcb.so.1 (0x00007f2f96403000)
	libunistring.so.2 => /usr/lib/x86_64-linux-gnu/libunistring.so.2 (0x00007f2f96085000)
	libgnutls.so.30 => /usr/lib/x86_64-linux-gnu/libgnutls.so.30 (0x00007f2f95d1f000)
	libhogweed.so.4 => /usr/lib/x86_64-linux-gnu/libhogweed.so.4 (0x00007f2f95ae9000)
	libnettle.so.6 => /usr/lib/x86_64-linux-gnu/libnettle.so.6 (0x00007f2f958b3000)
	libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f2f95632000)
	libkrb5.so.3 => /usr/lib/x86_64-linux-gnu/libkrb5.so.3 (0x00007f2f9535c000)
	libk5crypto.so.3 => /usr/lib/x86_64-linux-gnu/libk5crypto.so.3 (0x00007f2f9512a000)
	libcom_err.so.2 => /lib/x86_64-linux-gnu/libcom_err.so.2 (0x00007f2f94f26000)
	libkrb5support.so.0 => /usr/lib/x86_64-linux-gnu/libkrb5support.so.0 (0x00007f2f94d1b000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f2f94b01000)
	libsasl2.so.2 => /usr/lib/x86_64-linux-gnu/libsasl2.so.2 (0x00007f2f948e6000)
	libgssapi.so.3 => /usr/lib/x86_64-linux-gnu/libgssapi.so.3 (0x00007f2f946a5000)
	libXau.so.6 => /usr/lib/x86_64-linux-gnu/libXau.so.6 (0x00007f2f944a1000)
	libXdmcp.so.6 => /usr/lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007f2f9429b000)
	libp11-kit.so.0 => /usr/lib/x86_64-linux-gnu/libp11-kit.so.0 (0x00007f2f93f6c000)
	libtasn1.so.6 => /usr/lib/x86_64-linux-gnu/libtasn1.so.6 (0x00007f2f93d59000)
	libkeyutils.so.1 => /lib/x86_64-linux-gnu/libkeyutils.so.1 (0x00007f2f93b55000)
	libheimntlm.so.0 => /usr/lib/x86_64-linux-gnu/libheimntlm.so.0 (0x00007f2f9394c000)
	libkrb5.so.26 => /usr/lib/x86_64-linux-gnu/libkrb5.so.26 (0x00007f2f936bf000)
	libasn1.so.8 => /usr/lib/x86_64-linux-gnu/libasn1.so.8 (0x00007f2f9341d000)
	libhcrypto.so.4 => /usr/lib/x86_64-linux-gnu/libhcrypto.so.4 (0x00007f2f931e7000)
	libroken.so.18 => /usr/lib/x86_64-linux-gnu/libroken.so.18 (0x00007f2f92fd1000)
	libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007f2f92dbc000)
	libffi.so.6 => /usr/lib/x86_64-linux-gnu/libffi.so.6 (0x00007f2f92bb4000)
	libwind.so.0 => /usr/lib/x86_64-linux-gnu/libwind.so.0 (0x00007f2f9298b000)
	libheimbase.so.1 => /usr/lib/x86_64-linux-gnu/libheimbase.so.1 (0x00007f2f9277c000)
	libhx509.so.5 => /usr/lib/x86_64-linux-gnu/libhx509.so.5 (0x00007f2f92532000)
	libsqlite3.so.0 => /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007f2f92229000)
	libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1 (0x00007f2f91ff1000)
```

So, yeah. Work to do.

acalcutt:
I ended up with this workflow, which has been working pretty well ( https://github.com/acalcutt/maplibre-gl-native/blob/main/.github/workflows/node-ci.yml ). 

Basically I ended up with a matrix for each node version In combination with the node-pre-gyp-github we worked on before. So now it builds for ubuntu and macos for each node version (10-18)
https://github.com/acalcutt/maplibre-gl-native/runs/6459494950
https://github.com/acalcutt/maplibre-gl-native/releases/tag/node_v5.0.15

I also made it only publish the npm package for linux for once, since the build already makes the other versions in the npm publish anf the npm publish fails if the same version is uploaded twice.

I think eventually I would take this out of the node-ci workflow and make it into it's own workflow, since unless the package.json version changes, the npm part of the workflow fails. In my tileserver-gl publish workflow I made it autoincrement the version on push to main, which fixes that but can be a bitt messy commit wise  ( https://github.com/acalcutt/tileserver-gl/blob/main/.github/workflows/Build_Test_Release.yml )

I have been using the linux version with  ( https://github.com/acalcutt/tileserver-gl/ ) , which uses my package https://www.npmjs.com/package/@acalcutt/maplibre-gl-native and so far it has worked pretty well. though I did have to install a few dependencies which I think May have been statically linked before (like you mentioned above)

mnutt:
That makes sense, and while mine kinda cleverly leverages `cmake-node-module` to build all node versions at once, yours is likely better because it actually tests those node versions too. My dev environment is MacOS arm64 so unfortunately I'm stuck with a cross-compiled version that can't easily be tested in Github Actions. 😢 

If we're only building one linux distro version, I might suggest ubuntu 18.04? Older builds work with newer distro versions, but not vice versa.

Mapbox uses EGL while the build now uses GLX; I vaguely recall some discussion of why one was chosen over the other but can't find it now. EGL should in _theory_ eventually allow us to run maplibre-gl-native totally headless without an x server, but seems to be blocked on glfw requiring an x server and maybe no maintainer interest in implementing it. (https://github.com/glfw/glfw/pull/1608 looks reasonable to me, but who knows) I quickly tried building maplibre-gl-native with EGL enabled and it seemed to work but I haven't tested it in depth.

I think in the olden days mapbox would guard the release step on the presence of a git tag, which presumably they'd only apply when bumping the module version.

mnutt:
FWIW you could probably edit the cmake-node-module usage to only build the current version of node under test; right now it downloads node headers and builds for pretty much all versions: https://github.com/acalcutt/maplibre-gl-native/blob/main/platform/node/CMakeLists.txt#L13

Also, awesome to see that you're actively working on tileserver-gl, hopefully all of those changes can get merged in. I have my own fork that has probably diverged too far from maptiler's version to ever be merged.

acalcutt:
I'd love to see some of my tileserver-gl fork get merged back, but I fear I may have diverged to far too since I changed all the requires to imports and reworked it to make that work.

we will see though once I gure out the last label trimming issue I have https://github.com/maplibre/maplibre-gl-native/issues/284

acalcutt:
For EGL vs GLX, there was an old pull request ( https://github.com/mapbox/mapbox-gl-native/pull/5843 ), is that pretty much what you tested?

On the tileserver-gl side they where talking about EGL and using it to run headless here ( https://github.com/maptiler/tileserver-gl/issues/71 )

brendan-ward:
As a point of reference, I'm using EGL rather than GLX for [pymgl](https://github.com/brendan-ward/pymgl) (Python wrapper around maplibre-gl-native for static rendering); this has been working fine for Ubuntu 20.04 but still requires using Xvfb for headless rendering (not enough of an expert on how GLX vs EGL works to know if it _shouldn't_ need Xvfb).  See [GH test action](https://github.com/brendan-ward/pymgl/blob/main/.github/workflows/test.yml) in case it helps.

acalcutt:
I saw that noted in the tileserver-gl thread I linked too.

```
And although the EGL PR (https://github.com/mapbox/mapbox-gl-native/pull/7007) is merged in mapbox-gl-native, you still need an X display:

Though we use software rendering via LLVMpipe, we still need a valid display for EGL to be properly initialized. On Travis CI, like the Qt builds, we still need to start Xvfb to obtain a valid display.
```

acalcutt:
Does the "[MBGL_WITH_EGL](https://github.com/maplibre/maplibre-gl-native/blob/main/CMakeLists.txt#L10)" build option work. Is that switch all thats needed to compile this with EGL instead of GLX?

xabbu42:
We use that option for our intern server side tile renderer to work around a problem we had with the installed xvfb and the default method to get a display. We still need xvfb to start the server but it does use headless_backend_egl instead of headless_backend_glx  on the linux platform which worked around our problem.

mnutt:
Yeah I was thinking we could perhaps compile using `MBGL_WITH_EGL` _and_ statically link a forked glfw in order to have a totally headless server, but doubtful it's worth maintain the fork just to not have to run xvfb.

acalcutt:
I had [this issue](https://github.com/acalcutt/tileserver-gl/issues/2) from [dwoznicki](https://github.com/dwoznicki) where he was having issues with GLX on his amd graphics card trying to use a real X Server instead of xvfb. I was just wondering is maybe MBGL_WITH_EGL would work around his issue. He also had this issue here about it https://github.com/maplibre/maplibre-gl-native/issues/282

mnutt:
@jutaz I was wondering about the motivation behind https://github.com/maplibre/maplibre-gl-native/pull/217/commits/9c85518425bd821fd2c23aec6ca194169cfbecc7 (Simplify node loop). My understanding is that `mbgl-loop-uv` was split out into its own target because mbgl-core needs libuv, and in the standalone case (`mbgl-renderer`, tests, etc) we need to link against libuv to get it, but in the node addon case, node bundles its own statically linked libuv and we'd like to use _that_. I've tried the linux node build and it works, but the libuv dependency seems unnecessary and like it potentially might cause problems at some point?

brendan-ward:
@mnutt thanks for working on this!  I tried to install your version `npm install mnutt-maplibre-gl-native@https://github.com/mnutt/maplibre-gl-native.git#publish-github` for testing `mbgl-renderer`.

It appears to work for nearly all our tests (our label tests fail due to labels being positioned differently now), but I get a ton of errors on my console:

```bash
profiling: /Users/runner/work/maplibre-gl-native/maplibre-gl-native/build/CMakeFiles/mbgl-core.dir/src/mbgl/actor/mailbox.cpp.gcda: cannot open: No such file or directory
profiling: /Users/runner/work/maplibre-gl-native/maplibre-gl-native/build/CMakeFiles/mbgl-core.dir/src/mbgl/actor/scheduler.cpp.gcda: cannot open: No such file or directory
...
```

mnutt:
I saw those too when terminating node. They look like the result of profile-guided optimization; not sure why it’s turned on but should be turned off by some cmake config.

mnutt:
@wipfli if there are docs for the mapbox version they should be the same for this.

acalcutt:
I had a real hard time finding anything documentation wise. For my label issue ( https://github.com/maplibre/maplibre-gl-native/issues/284 ) test application, I basically search until I found some example code that worked, like https://gist.github.com/ezheidtmann/c869f89b6c286082e84c5615311ebd80 . I was also comparing with the tileserver-gl code and a few other example, but this one ended up to be the one that worked for me.

The for that example, I was trying to get two tiles that where side by side. For that I basically had to do manual math to calculate two centerpoints. Basically I used 'Tile width
(° of longitudes)' at the zoom level I wanted from https://wiki.openstreetmap.org/wiki/Zoom_levels and then subtracted that from the other tiles centerpoint longitude.

mnutt:
I believe what you're describing is also what tileserver-gl does: https://github.com/maptiler/tileserver-gl/blob/c89a5ae029f5e9ef0855292cfcce2f72894a8fff/src/serve_rendered.js#L381-L387

acalcutt:
It is, I was trying to show an issue I was having with my tileserver-gl that uses maplibre node

brendan-ward:
> I saw those too when terminating node. They look like the result of profile-guided optimization; not sure why it’s turned on but should be turned off by some cmake config.

I did not observe these when building this PR locally ([see above](https://github.com/maplibre/maplibre-gl-native/pull/217#issuecomment-1021777044)) on macOS 10.15.  Perhaps something additional is turned on in the configuration you used?   I didn't see anything obvious in your [changes](https://github.com/maplibre/maplibre-gl-native/compare/main...mnutt:publish-github) but perhaps the `-DMBGL_WITH_COVERAGE=ON` flag is not needed?

(sidenote: is there a different issue where we should take up NodeJS module publishing related issues, as this PR has been closed a while now, and these are follow-on issues?)

wipfli:
The MapLibre Governing Board would like to give a thank-you payment to the people who helped with bringing back node support. I already reached out to some of you on slack but I could not find @mnutt and @METACEO. If you are interested in receiving a payment, please contact me on Slack. You can join the slack channel here: https://osmus-slack.herokuapp.com/

Thanks for helping with this! I really appreciate it...

Crisfole:
Do the docs need some updating?

https://github.com/maplibre/maplibre-gl-native/tree/main/platform/node

> By default, installs binaries. On these platforms no additional dependencies are needed.

>
> * 64 bit macOS or 64 bit Linux
> * Node.js v10.x
>
> Run:

> 
> `npm install @maplibre/maplibre-gl-native`
> Other platforms will fall back to a source compile with make node; see [DEVELOPING.md](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/DEVELOPING.md) for details on building from source.

If you try and install this in a regular linux container you can install, but you can't require the code:

> require("@maplibre/maplibre-gl-native")
> // Uncaught Error: libGLX.so.0: cannot open shared object file: No such file or directory

We had to install *all* of these libraries to get it functional:

> apt-get install -y -q --no-install-recommends libglx0 libopengl0 libjpeg-turbo8 libuv1 libicu66 libpng16-16 xvfb

We also had to run it with `xvfb-run -a --server-args="-screen 0 1024x768x24" node ./our-script.js`

acalcutt:
I'd say it needs updating.

Currently we are running the node binary builds on node 18 and we are creating binaries for Node 14,16,18 . I'm not sure what the actual minimum is considered, but I'd say at least Node 16 is recommended.

If you are using npmjs package with the pre-built binaries, the binaries are built on Ubuntu 20.04  amd64/arm64 and MacOS 12 amd64/arm64. The linux binaries end up with some hard dependencies that require packages only available in 20.04, so they kind of require Ubuntu 20.04 right now. I'm not sure if MacOS 12 binaries have issues like that, but I assume they won't work on older systems.

To use the package in a workflow, you need it to be ubuntu 20.04 and have installed server prerequisite packages. For that GLX error it would be fix by installing libgles2-mesa-dev. For example, I install these packages in my Tileserver-gl test workflow, which uses maplibre-native node https://github.com/maptiler/tileserver-gl/blob/master/.github/workflows/ct.yml#L34-L36

acalcutt:
By the way, the preconfigured docker image is basically how tileserver-gl does it. I made this workflow to make a x64/arm64 release in a similar way
https://github.com/maptiler/tileserver-gl/blob/master/.github/workflows/release.yml (it is more automated [in my fork](https://github.com/acalcutt/tileserver-gl/blob/main/.github/workflows/release.yml))

PR #220, opened 2022-01-22
[general] Avoid double nested platforms

ntadej:
A necessary cleanup before we start cleaning dependencies. Almost certainly fully broken at this state, but opening this so @roblabs and @birkskyum are aware I'm working on it.

The next step is to cleanup the build infrastructure as the Node MR (#217) uncovered many issues.


wipfli:
This sounds like a great idea, @ntadej For the review it might be useful if you could share the cli commands you used to move the files around...

birkskyum:
@ntadej , is this still releant? It seems reasonable to i.e. move macos from platform/ios/platform/macos to platform/macos, and so on, but is this branch maybe too dated at this?

birkskyum:
I believe this is too far behind at this point, and it would be a lot of work to rebase, but we still have the pr as a reference.

PR #221, opened 2022-01-23
[qt] Initial CI pipeline

ntadej:
Initial CI pipeline for Qt builds. For now macOS + iOS and Windows.


roblabs (approved):
Approved, but there may be slight manual diff integration for some of the commonly changed `*.yml` when merging with #217.

PR #231, opened 2022-02-10
Switch 'vendor/boost' submodule from Mapbox repo to Maplibre repo

kiibimees:
And move super-project-pointer to the new commit with Boost proto fix for MSVC.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,7 +18,7 @@
 	url = https://github.com/okdshin/unique_resource.git
 [submodule "vendor/boost"]
 	path = vendor/boost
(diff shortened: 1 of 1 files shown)


birkskyum:
@kiibimees thanks! 

@ntadej , any idea why qt breaks here? Is it another boost version?

ntadej:
It seems timeouts for mirrors, will retry them a bit later.

ntadej (approved):
All green now. As this only fixes some MSVC workaround it should not affect any other platform, especially not functionality wise (the code would just not compile).

Issue #183, opened 2021-11-08
Attribution not working in iOS sdk


d3mon187:
I built a simple demo app and noticed that on the iOS version I don't have any attribution text or links in the information popup.  Just shows the text for "Maplibre Maps SDK for iOS".  Was this broken, or does iOS require extra code to enable it?  I did try simplifying the attribution field in my style.json, but it doesn't seem like there is an issue there.

atierian:
I noticed the same issue. 

Based on [this](https://github.com/maplibre/maplibre-gl-native/blob/d28e4a040fe22ef42cf23259b165ef726d550253/platform/ios/platform/ios/src/MGLMapView.mm#L2746-L2780), it looks like the attribution text is just being set to the localized string `SDK_NAME`, which is "MapLibre Maps SDK for iOS" for English. I would expect that the MGLMapView pulls the attribution information from the style's source leveraging `MGLVectorTileSource().attributionInfos`.

Could someone confirm that pulling attribution info from the source is expected behavior? If so, I'll try to work on a fix.

PR #236, opened 2022-02-14
[ios] fix missing attribution

atierian:
This fixes the issue of missing style source attribution as noted in #183.


thaddmt (approved):

atierian:
I think the answer to that depends on the release strategy.

There's a pretty straightforward workaround for this issue that involves grabbing the attribution info in the `didFinishLoading` delegate method and creating a custom UI element. Most people have probably implemented this themselves as needed.

That being said, it's my understanding that proper attribution is a contractual obligation between the style source provider and the consumer (developer). Supporting this behavior by default seems pretty important.

The ideal scenario would be fixing this on the Android side as well, and pushing a release(s) that encompasses both for platform parity.

PR #238, opened 2022-02-15
[ios] patch to fix MGLReachability.m compiler error

atierian:
Fixes `Cast to smaller integer type 'unsigned int' from 'MGLReachability *'` compiler error in `darwin/src/MGLReachability.m` existing on `metal-support` branch.


roblabs:
Ahh, sorry you inherited that issue.  It was likely left over from when `metal-support` was being built on Xcode 12.x.

Thank you for fixing!

PR #240, opened 2022-02-20
[qt] Set defaultStyles to const

rinigus:
This sets the method to const as it is expected to be. Please review


ntadej (approved):

ntadej:
Thanks for spotting this!

PR #254, opened 2022-03-11
Implement map projection functionality

OlexandrStepanov:
This PR adds the new public class `MGLMapProjection` which aims to provide functionality for map view camera calculation and screen points <-> geographical coordinates conversion. While the same interface is provided by `MGLMapView` itself, the new class provides the ability to modify the camera or visible coordinates to make the conversion without visual application on the map.

`MGLMapProjection` initializes with the map view instance. It makes the new instance of `mbgl::MapProjection` cpp class, which makes the copy of `mbgl::Transform` out of the map view internal `mbgl::TransformState`. In fact, `MGLMapProjection` is an Objective-C wrapper around `mbgl::MapProjection`. The introduction of the last was required because `mbgl::Transform` is not part of the public interface.

The new class implements next functionality:
* read/modify the current camera with optional edge insets
* modify the camera using the visible coordinate bounds with optional edge insets
* convert screen point to geographical coordinate and vice versa using the current camera
* get the distance in meters spanned by a single screen point for the current camera

The new functionality is covered with unit tests and supplied in this PR.

### DIFF
--- a/platform/ios/platform/ios/src/MGLMapProjection.mm
+++ b/platform/ios/platform/ios/src/MGLMapProjection.mm
@@ -1,11 +1,3 @@
-//
-//  MGLMapProjection.m
-//  ios
-//
-//  Created by Oleksandr Stepanov on 07.12.2021.
-//  Copyright © 2021 Mapbox. All rights reserved.
-//
-
 #include <mbgl/map/map.hpp>
 #include <mbgl/map/map_projection.hpp>
 #include <mbgl/util/geo.hpp>
--- a/platform/ios/platform/ios/src/MGLMapProjection.mm
+++ b/platform/ios/platform/ios/src/MGLMapProjection.mm
@@ -1,0 +1,121 @@
+//
+//  MGLMapProjection.m
+//  ios
+//
+//  Created by Oleksandr Stepanov on 07.12.2021.
+//  Copyright © 2021 Mapbox. All rights reserved.
+//
+
+#include <mbgl/map/map.hpp>
+#include <mbgl/map/map_projection.hpp>
+#include <mbgl/util/geo.hpp>
+#include <mbgl/util/projection.hpp>
+
+#import "MGLMapProjection.h"
+#import "MGLMapView_Private.h"
+#import "MGLGeometry_Private.h"
+
+@interface MGLMapProjection ()
+
+@property (nonatomic) CGSize mapFrameSize;
+
+@end
+
+@implementation MGLMapProjection
+{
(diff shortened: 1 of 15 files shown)


OlexandrStepanov:
> Thanks for contributing to MapLibre @OlexandrStepanov. Are you already in the slack channel?

@wipfli,  no, I'm not. Could you send me an invite on my profile email, please?

wipfli:
https://osmus-slack.herokuapp.com/ should generate an invite. Let me know if that fails...

platform/ios/platform/ios/src/MGLMapProjection.mm
```diff
@@ -0,0 +1,121 @@
+//
+//  MGLMapProjection.m
+//  ios
+//
+//  Created by Oleksandr Stepanov on 07.12.2021.
+//  Copyright © 2021 Mapbox. All rights reserved.
```
roblabs, line 6: Can you review this copyright and verify if this is correct?

platform/ios/platform/ios/src/MGLMapProjection.mm
```diff
@@ -0,0 +1,121 @@
+//
+//  MGLMapProjection.m
+//  ios
+//
+//  Created by Oleksandr Stepanov on 07.12.2021.
```
roblabs, line 5: I leave it up to you if you want to include your name.

roblabs:
I have requested a re-run of the workflow `ios-ci` as it failed for:

* [missing symbol export for class MGLMapProjection in platform/ios/src/MGLMapProjection.h:11:12](https://github.com/maplibre/maplibre-gl-native/runs/5514121746?check_suite_focus=true#step:10:10)

Let's monitor [attempt #2](https://github.com/maplibre/maplibre-gl-native/actions/runs/1969947035) to verify if it's an anomaly or not

platform/ios/platform/ios/src/MGLMapProjection.mm
OlexandrStepanov: I reviewed other .mm files, and removed this copyright header at all, to be aligned with the rest of codebase.

OlexandrStepanov:
@roblabs, there was missing `MGL_EXPORT` next to the `MGLMapProjection` declaration which was failing the CI. Now it's fixed.

I was also thinking about adding and example of usage in the demo app, but found it already overloaded with different options, don't want to overcomplicate it further. I added unit tests for the new functionality, I believe it's enough for showcasing purposes.

roblabs (approved):

PR #256, opened 2022-03-14
[qt] Build with WebAssembly

ntadej:
Build with WebAssembly because why not 😊

Unfortunately it does not run, I suspect I need to build with threading support. Submitting this for now so it's not lost.


birkskyum:
Very nice :) as wasm gets better simd support this even has a potential to disrupt the web version

ntadej:
The binary is still huge compared to the JS one.

Feel free to merge this as is as it should not affect the regular builds.

birkskyum (approved):

PR #262, opened 2022-03-26
Avoid using deprecated std::result_of

ntadej:
Avoid using deprecated `std::result_of`. Use `std::invoke_result` instead.

This fixes building with new compilers as we moved to C++17.


ntadej:
OK, it seems that not all platforms have moved yet.

PR #263, opened 2022-03-26
[qt] Initial import of Qt QML plugin

ntadej:
Initial import of Qt QML plugin. It may be useful to have this in the repo, especially for the Qt 5 version which will never be supported upstream.

For now this is Qt6-only version which is easier to build (as Qt moved to CMake with version 6). I can build it on macOS (both for Intel and M1), but I have some rendering issues:
 - Initial paint seems to work if I add `beginExternalCommands`/`endExternalCommands`, but moving/zooming the map does not work.
 - Resize completely breaks rendering (see attached screenshot).

Screenshot 2022-03-26 at 12 16 14

Tagging @kiibimees if he has any ideas. Note that I use the latest Qt 6.3 beta + `dev` QtLocation built from source on top of it. FYI also to @rinigus.

Keeping as Draft while I provide some documentation and add testing.


rinigus:
Skimming through some parts, looks like you have more complicated renderer method than what I use for Mapbox GL QML plugin: https://github.com/rinigus/mapbox-gl-qml/blob/master/src/qsgmapboxglnode.cpp#L169 . If you see that method, notice `f->glDepthRangef(0, 1);` - that was required for fixing some rendering issue (don't remember which one, though). Maybe it will help you as well.

I presume this is an alternative to current Qt platform and we can choose the one we prefer when building it. 

Also noticed that CMake has some options which were matching the ones we have for Qt. Was that on purpose?

ntadej:
> Skimming through some parts, looks like you have more complicated renderer method than what I use for Mapbox GL QML plugin: https://github.com/rinigus/mapbox-gl-qml/blob/master/src/qsgmapboxglnode.cpp#L169 . If you see that method, notice `f->glDepthRangef(0, 1);` - that was required for fixing some rendering issue (don't remember which one, though). Maybe it will help you as well.
> 
> I presume this is an alternative to current Qt platform and we can choose the one we prefer when building it.
> 
> Also noticed that CMake has some options which were matching the ones we have for Qt. Was that on purpose?

This is the QML plugin yes. It still uses the Qt platform but to avoid potential CMake issues, I decided to make it now as a separate platform as it's built statically.

Qt6 made the rendering a bit trickier, my Qt5 version (https://github.com/ntadej/qt-geoservices-maplibre/blob/f6b4dc3d190a1a3cb6fb2d2301854faa428351ad/src/qsgmapboxglnode.cpp#L92) is also a bit simpler.
I added your extra line and it did not help unfortunately. It seems like Qt is not notified to re-render properly. Maybe it's a bug in the Qt build I'm using.

CMake is mostly copied from upstream Qt patchset (https://codereview.qt-project.org/c/qt/qtlocation/+/382087). The idea is to just remove the ambiguity as basically only one set of flags are supported for the QML plugin.

ntadej:
OK, not sure what was going on. After changing some of the interface in my application it just started working so either it is:
 - a Qt bug
 - something corrupted on the system

kiibimees:
I have to admit that I'm having troubles myself, like:

* Wrong padding/margin offset when Map object is not fullscreen.
* Crash when using QSGMapboxGLRenderNode instead on QSGMapboxGLTextureNode.
* Wrong clipping when placing a QML item on top of map and set "clip: true". It also clips the map, not only it's own content.

So, if you find a good solution for both RenderNode/TextureNode version, I would be happy to fix it also in the Qt plugin :) Otherwite it might take me a while to find solutions myself. It's a bit heavy for me.

ntadej:
Closing this one for now in favour of separate repository.

PR #266, opened 2022-03-31
Fix logging format strings

nvanfleet:
Setting up MGLLogger I noticed that some Log messages contained format strings like "Network error error: @"

I ended up realizing that some of the Log messages in the app are not actually formatted. This change makes it so the strings are now formatted.

Previously: Requesting: Requesting: %@ failed with error: %@ failed with error: <MGLNetworkConfiguration: 0x600001820b40>

Now: Requesting: (null) failed with error: Error Domain=NSURLErrorDomain Code=-1009 \"The Internet connection appears to be offline.\" UserInfo={_kCFStreamErrorCodeKey=50, NSUnderlyingError=0x600003c44420 {Error Domain=kCFErrorDomainCFNetwork Code=-1009 \"(null)\" UserInfo={_NSURLErrorNWPathKey=unsatisfied (No network route), _kCFStreamErrorCodeKey=50, _kCFStreamErrorDomainKey=1}}, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <EC18DF77-C421-456D-AA08-C1A1EDD9A2F3>.<78>, _NSURLErrorRelatedURLSessionTaskErrorKey=(\n \"LocalDataTask <EC18DF77-C421-456D-AA08-C1A1EDD9A2F3>.<78>\"\n), NSLocalizedDescription=The Internet connection appears to be offline., NSErrorFailingURLStringKey=https://d34hkqyc0uusyp.cloudfront.net/tiles/1:1.0.0:qxcrxz/11/323/803.pbf, NSErrorFailingURLKey=https://d34hkqyc0uusyp.cloudfront.net/tiles/1:1.0.0:qxcrxz/11/323/803.pbf, _kCFStreamErrorDomainKey=1}

PR #267, opened 2022-03-31
Fix street label appearance while animating near zoom level threshhold

nvanfleet:
We were seeing some issues while following the users location along a route in the map view that some screen elements (for the most part the labels of street names on the map) would flash in and out.

While it took a long time to determine why this was happening we determined that the issue was simple. We were animating the map camera at a specific zoom level (zoom levels are from 1-25 floats) and the map SDK was sometimes finding us at zoom level 10.0 and sometimes finding us at zoom level 9.9999

This would generally not be a problem except at that particular zoom level we had our map style show the street labels on the map. So when we were at 9.9999 it would hide them, and on the next frame it would have us at 10.0.
The ultimate fix was fairly simple and the zoom level is basically rounded if it is close enough to a rounded zoom level.

This is the issue that was previously reporting this: https://github.com/maplibre/maplibre-gl-native/issues/137

### DIFF
--- a/src/mbgl/map/transform_state.cpp
+++ b/src/mbgl/map/transform_state.cpp
@@ -16,6 +16,17 @@ LatLng latLngFromMercator(Point<double> mercatorCoordinate, LatLng::WrapMode wra
             mercatorCoordinate.x * 360.0 - 180.0,
             wrapMode};
 }
+const double kEpsilon = 1e-9;
+// To avoid flickering issue due to "zoom = 13.9999999..".
+double roundForAccuracy(double x) {
+    double round_x = round(x);
+    double diff = fabs(round_x - x);
+    if (diff < kEpsilon && diff > 0 ){
+        return round_x;
+    } else {
+        return x;
+    }
+}
 } // namespace
 
 TransformState::TransformState(ConstrainMode constrainMode_, ViewportMode viewportMode_)
@@ -639,11 +650,11 @@ bool TransformState::isGestureInProgress() const {
 #pragma mark - Projection
 
 double TransformState::zoomScale(double zoom) const {
-    return std::pow(2.0, zoom);
+    return roundForAccuracy(std::pow(2.0, zoom));
 }
 
 double TransformState::scaleZoom(double s) const {
-    return util::log2(s);
+    return roundForAccuracy(util::log2(s));
 }
 
 ScreenCoordinate TransformState::latLngToScreenCoordinate(const LatLng& latLng) const {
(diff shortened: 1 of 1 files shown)


nvanfleet:
This investigation and fix were actually by @junwu-lyft and I am just creating this for him

nvanfleet:
@junwu-lyft and I are happy to discuss some other solutions to this issue.

ntadej:
Thank you for your contribution!

Wrong commit though? It looks like #266

nvanfleet:
@ntadej welp sorry about that I grabbed the wrong patch somehow and didn't notice 😓

src/mbgl/map/transform_state.cpp
```diff
@@ -16,6 +16,17 @@ LatLng latLngFromMercator(Point<double> mercatorCoordinate, LatLng::WrapMode wra
             mercatorCoordinate.x * 360.0 - 180.0,
             wrapMode};
 }
+const double kEpsilon = 1e-9;
```
ntadej, line 19: Maybe `constexpr double`?

src/mbgl/map/transform_state.cpp
```diff
@@ -16,6 +16,17 @@ LatLng latLngFromMercator(Point<double> mercatorCoordinate, LatLng::WrapMode wra
...
+// To avoid flickering issue due to "zoom = 13.9999999..".
+double roundForAccuracy(double x) {
+    double round_x = round(x);
```
ntadej, line 22: `std::round`

src/mbgl/map/transform_state.cpp
```diff
@@ -16,6 +16,17 @@ LatLng latLngFromMercator(Point<double> mercatorCoordinate, LatLng::WrapMode wra
...
+    double diff = fabs(round_x - x);
```
ntadej, line 23: `std::abs`

ntadej (changes requested):
I like the idea, I only have a few minor comments.

ntadej:
Also to add, I've been recently cleaning the remaining warnings from MSVC and it seems that `float` vs `double` accuracy is quite arbitrary in many cases. It would be interesting to see if it's good enough if we go to single precision in most of the cases, which may also solve such issues.

Of course in the meantime this fix is appreciated and useful.

src/mbgl/map/transform_state.cpp
junwu-lyft: done

src/mbgl/map/transform_state.cpp
junwu-lyft: done

src/mbgl/map/transform_state.cpp
junwu-lyft: done

ntadej (approved):

wipfli:
Thanks for this fix @nvanfleet. Can you add a changelog item?

wipfli:
Ah I see https://github.com/maplibre/maplibre-gl-native/blob/main/CHANGELOG.md has not really been touch since the fork. What is our policy for changelogs?

PR #270, opened 2022-04-03
[msvc] Fix warnings, mainly casting to smaller types

ntadej:
MSVC is quite eager with warnings (which is good in my opinion). This MR cleans most of the core codebase of MSVC warnings:
 - implicit casting to smaller types (e.g. `double` to `float`)
 - some encoding stuff
 - warnings from dependencies

Note that this is mostly silencing warnings (e.g. using `static_cast`) but I also tried to fix some trivial cases and add some template specialisations. This should be purely technical, but let's see what also CI says.

### DIFF
--- a/src/mbgl/text/collision_index.cpp
+++ b/src/mbgl/text/collision_index.cpp
@@ -47,7 +47,7 @@ CollisionIndex::CollisionIndex(const TransformState& transformState_, MapMode ma
       screenBottomBoundary(transformState.getSize().height + viewportPadding),
       gridRightBoundary(transformState.getSize().width + 2 * viewportPadding),
       gridBottomBoundary(transformState.getSize().height + 2 * viewportPadding),
-      pitchFactor(std::cos(transformState.getPitch()) * transformState.getCameraToCenterDistance()) {}
+      pitchFactor(static_cast<float>(std::cos(transformState.getPitch()) * transformState.getCameraToCenterDistance())) {}
 
 float CollisionIndex::approximateTileDistance(const TileDistance& tileDistance,
                                               const float lastSegmentAngle,
@@ -122,13 +122,13 @@ IntersectStatus CollisionIndex::intersectsTileEdges(const CollisionBox& box,
     const float tileY2 = tileEdges[3];
 
     // Check left border
-    int minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
+    float minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
     if (minSectionLength <= 0) { // Check right border
         minSectionLength = std::min(tileX2 - x1, x2 - tileX2);
     }
     if (minSectionLength > 0) {
         result.flags |= IntersectStatus::VerticalBorders;
-        result.minSectionLength = minSectionLength;
+        result.minSectionLength = static_cast<int>(minSectionLength);
     }
     // Check top border
     minSectionLength = std::min(tileY1 - y1, y2 - tileY1);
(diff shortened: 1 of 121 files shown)


ntadej:
@roblabs, @atierian, @birkskyum, is there anything else I should do from my side to proceed?

src/mbgl/text/collision_index.cpp
```diff
@@ -122,13 +122,13 @@ IntersectStatus CollisionIndex::intersectsTileEdges(const CollisionBox& box,
     const float tileY2 = tileEdges[3];
 
     // Check left border
-    int minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
+    float minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
```
wipfli, line 125: Seems to break some render tests...

src/mbgl/text/collision_index.cpp
```diff
@@ -415,13 +415,13 @@ std::pair<Point<float>,float> CollisionIndex::projectAndGetPerspectiveRatio(cons
     auto size = transformState.getSize();
     return std::make_pair(
         Point<float>(
-            (((p[0] / p[3] + 1) / 2) * size.width) + viewportPadding,
-            (((-p[1] / p[3] + 1) / 2) * size.height) + viewportPadding
+            static_cast<float>(((p[0] / p[3] + 1) / 2) * size.width) + viewportPadding,
+            static_cast<float>(((-p[1] / p[3] + 1) / 2) * size.height) + viewportPadding
```
wipfli, line 419: Breaks another render test

src/mbgl/text/shaping.cpp
```diff
@@ -145,7 +145,7 @@ void align(Shaping& shaping,
     if (maxLineHeight != lineHeight) {
         shiftY = -blockHeight * verticalAlign - Shaping::yOffset;
     } else {
-        shiftY = (-verticalAlign * lineCount + 0.5) * lineHeight;
+        shiftY = (-verticalAlign * static_cast<float>(lineCount + 0.5)) * lineHeight;
```
wipfli, line 148: Was fixed in #285

src/mbgl/util/grid_index.cpp
```diff
@@ -263,12 +263,12 @@ void GridIndex<T>::query(const BCircle& queryBCircle, std::function<bool (const
...
 template <class T>
 std::size_t GridIndex<T>::convertToXCellCoord(const float x) const {
-    return util::max(0.0, util::min(xCellCount - 1.0, std::floor(x * xScale)));
+    return util::max(size_t(0), util::min(xCellCount - 1, static_cast<size_t>(std::floor(x * xScale))));
 }
 
 template <class T>
 std::size_t GridIndex<T>::convertToYCellCoord(const float y) const {
-    return util::max(0.0, util::min(yCellCount - 1.0, std::floor(y * yScale)));
+    return util::max(size_t(0), util::min(yCellCount - 1, static_cast<size_t>(std::floor(y * yScale))));
```
wipfli, line 271: Breaks some render tests...

src/mbgl/text/collision_index.cpp
ntadej: Why this happens is obscure to me. If you look closer there is a cast to `int` when this variable is copied over to the `result`.

ntadej: Also it would be useful to actually to paste what tests this breaks to see the effect.

PR #285, opened 2022-05-19
Fix shiftY calculation typo

rmyhal:
In [this PR](https://github.com/maplibre/maplibre-gl-native/pull/270/files#diff-d57cd3143ace69af2f7ae79a917ac1328acc7543ae8a2a15f58b4635b86f1b91R148) during refactoring a typo error was introduced and because of it the calculation of `shiftY` is broken right now. This PR fixes it.
Also, there is [an issue](https://github.com/maplibre/maplibre-gl-native/issues/279) with it.

| Before | After |
| - | - |
| image |  image |

### DIFF
--- a/src/mbgl/text/shaping.cpp
+++ b/src/mbgl/text/shaping.cpp
@@ -145,7 +145,7 @@ void align(Shaping& shaping,
     if (maxLineHeight != lineHeight) {
         shiftY = -blockHeight * verticalAlign - Shaping::yOffset;
     } else {
-        shiftY = (-verticalAlign * static_cast<float>(lineCount + 0.5)) * lineHeight;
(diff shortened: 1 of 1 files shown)


acalcutt:
I tested this in my fork and it worked for me. Thanks for fixing this @rmyhal . Looking forward to an official merge :-)

ntadej (approved):

ntadej:
Thanks for the fix and sorry for causing the issue in the first place!

wipfli:
Are the render tests running in CI? Feels like the sort of typo fixed here would not have passed a render test...

ntadej:
Not sure, but I iOS and Android run a lot of tests.

PR #286, opened 2022-05-19
Fix previous PR that added logging

nvanfleet:
My previous PR #266 did not realize that the networking framework mishandled format strings. This would ultimately cause a crash. This PR corrects that.


ntadej:
Maybe something for @atierian to check?

atierian (approved):
Looks good. Thanks for making this fix, and good catch changing `res` to `req`!

PR #292, opened 2022-05-27
Handle nonexistent mbtiles files, add MBTilesFileSource tests

### LINKED ISSUE #291: Bug: MBTilesFileSource does not correctly handle nonexistent mbtiles files
It looks like the other local asset / file classes check for file existence and return an appropriate error when files don't exist.  `MBTilesFileSource` does not appear to do this; it simply attempts to open the mbtiles SQLite file and then aborts the entire program:

```
libc++abi.dylib: terminating with uncaught exception of type mapbox::sqlite::Exception: unable to open database file
... terminated by signal SIGABRT (Abort)
```

Since there is already a check that the input mbtiles path is on a local absolute path, we should check that the file exists in the `MBTilesFileSource::request` handler.

brendan-ward:
Resolves #291 

This uses the same check for file existence as `local_file_request.cpp`.

This adds basic tests for `MBTilesFileSource` and demonstrates that nonexistent mbtiles files no longer cause an abort.


brendan-ward:
It looks like the node-ci tests failed due to an unrelated reason (X server error); `MBTilesFileSource` isn't used for node.  Perhaps someone with appropriate permissions can trigger the node-ci tests to run again?

roblabs:
> *"Perhaps someone with appropriate permissions can trigger the node-ci tests to run again?"*

I've restarted `node-ci` and there was an option to enable *debug logging*.  Let's if the restart sorts things out.

* https://github.com/maplibre/maplibre-gl-native/actions/runs/2393948381

roblabs (approved):
thanks for adding the TESTs, @brendan-ward

PR #305, opened 2022-06-07
Remove ios mapbox code of conduct

wipfli:


wipfli:
I am converting this to a draft because I think we should also get rid of the `platform/ios/.github` folder...

PR #311, opened 2022-06-08
Cancel previous CI run on new pull request push

wipfli:
Cancels old ci runs.

See https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

### DIFF
--- a/.github/workflows/android-ci-pull.yml
+++ b/.github/workflows/android-ci-pull.yml
@@ -15,6 +15,9 @@ on:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    concurrency: 
+      group: ${{ github.ref }}
+      cancel-in-progress: true
     defaults:
       run:
         working-directory: platform/android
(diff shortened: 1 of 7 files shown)


wipfli:
This seems to not work properly. Workflows get cancelled but no new ones get started. I will revert this commit such that we have functional CI again...

.github/workflows/android-ci-pull.yml
```diff
@@ -15,6 +15,9 @@ on:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    concurrency: 
+      group: ${{ github.ref }}
```
brendan-ward, line 19: I've used `group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}` elsewhere, to prevent this from cancelling things more widely than intended.

wipfli: Cool, let me try this then

wipfli: #330

Issue #288, opened 2022-05-22
FOSS flavor


TheLastProject:
Hey there,

We've seen several Android apps include this library lately, only for developers to be surprised their apps can't be included in F-Droid as MapLibre has proprietary dependencies ([com.google.android.gms:play-services-location](https://mvnrepository.com/artifact/com.google.android.gms/play-services-location)).

Could it be possible to offer a completely libre flavor of MapLibre for Android (that is: a flavor with no proprietary dependencies) for developers who want to offer a 100% open source version of their apps?

Common names for such a flavour would be `foss` or `floss`.

Thanks for considering :)

wipfli:
Thanks for noting this @TheLastProject. Would you have time to work on this and change the dependencies? By the way, I believe @maxammann has a MapLibre-based app in the F-Droid store, right Max?

TheLastProject:
> Would you have time to work on this and change the dependencies?

I have no experience with Android build flavours myself (the apps I write only have a 100% libre version) so I don't think I will be much help there. However, from discussions with the Element team it seems like MapLibre already does check if Google Play Services is actually available so this could be fairly simple, see [these excludes](https://github.com/vector-im/element-android/pull/6136) which have been confirmed to have removed [GMS](https://github.com/vector-im/element-android/pull/6136) completely.

> By the way, I believe @maxammann has a MapLibre-based app in the F-Droid store, right Max?

I asked the team about this and apparently Ehrenamtskarte Bayern [is actually not compliant](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/10991#note_938539763) but it also managed to bypass our scanners until recently. Reading the topic more it seems the developer has been notified of this so I have disabled the app for F-Droid: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/11114 (sidenote: I would love to be able to restore this app, it's a very cool concept even though I don't live in Bayern)

maxammann:
We are currently discussing how we proceed with the Bavarian Volunteering Card on FDroid internally.

Im not sure in which way the maplibre android library depends on properiatry libraries right now. But I think it would be interesting to evaluate this. Maybe we somehow can also elimintate such issues for all one and for all.

TheLastProject:
> Im not sure in which way the maplibre android library depends on properiatry libraries right now. But I think it would be interesting to evaluate this.

See https://github.com/vector-im/element-android/issues/6100

maxammann:
[...]

PR #313, opened 2022-06-08
Document how to avoid non-free dependencies on Android

TheLastProject:
Given #288 was closed with the decision to not provide an explicit libre flavor because excluding the non-free dependencies is very simple, I am hoping documenting it like this should help users be aware that this is an option.

Thanks for considering :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,13 @@ We thank everyone who supported us financially in the past and special thanks to
 
 3. Sync gradle and rebuild your app
 
+*Note: MapLibre by default ships with the proprietary Google Play Location Services. If you want to avoid pulling proprietary dependencies into your project, you can exclude Google Play Location Services as follows:*
+```gradle
+    implementation ('org.maplibre.gl:android-sdk:9.4.0') {
+        exclude group: 'com.google.android.gms'
+    }
+```
+
 ### iOS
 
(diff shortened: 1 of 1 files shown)


wipfli (commented):
Looks like a good idea to me, thanks. @maxammann can you maybe comment if this is correct?

README.md
```diff
@@ -58,6 +58,13 @@ We thank everyone who supported us financially in the past and special thanks to
 
 3. Sync gradle and rebuild your app
 
+*Note: MapLibre by default ships with the proprietary Google Play Location Services. If you want to avoid pulling proprietary dependencies into your project, you can exclude Google Play Location Services as follows:*
+```gradle
+    implementation ('org.maplibre.gl:android-sdk:9.4.0') {
```
wipfli, line 63:
```suggestion
    implementation ('org.maplibre.gl:android-sdk:<version>') {
```

wipfli (approved):

README.md
wipfli: Like that we don't have to update the readme at every version...

TheLastProject: __

TheLastProject: I guess you want that changed on line 54 then too?

wipfli: Yes good catch. Can you update it?

PR #315, opened 2022-06-09
Add `update-gl-js.yml` workflow

wipfli:
Adds a new GitHub workflow which updates the MapLibre GL JS submodule to a user-specified commit, generates code in the native repo, and opens a pull request with these changes.

Here is an example how the automated pull request would look like: https://github.com/wipfli/maplibre-gl-native/pull/7

### DIFF
--- a/.github/workflows/update-gl-js.yml
+++ b/.github/workflows/update-gl-js.yml
@@ -1,0 +1,45 @@
+name: update-gl-js
+
+on: 
+  workflow_dispatch:
+    inputs:
+      commit:
+        description: 'Commit'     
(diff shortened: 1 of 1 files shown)


wipfli:
This workflow runs on a macos GitHub runner which is nice because some code generation scripts in platform/ios only run on mac...

wipfli:
Maybe this workflow also needs to generate more stuff. See https://github.com/maplibre/maplibre-gl-native/issues/310 for a list of generated files.

roblabs:
Thanks for contributing to FOSS4G!

PR #316, opened 2022-06-09
Fix bad endpoint

jwinarske:
-URL was incorrectly pointing to `docs.mapbox.com`


wipfli (approved):
Cool, thanks @jwinarske

wipfli:
Let's wait for ci to pass, but I think this is not a problem...

PR #319, opened 2022-06-09
Add iOS API docs workflow

wipfli:
Adds a workflow which builds the iOS API docs website and publishes it to the GitHub pages branch. The URL would be https://maplibre.org/maplibre-gl-native/ios/api/

For a live demo of how that would look like, see https://wipfli.github.io/maplibre-gl-native/ios/api/


nyurik (approved):
lgtm

PR #320, opened 2022-06-09
Deploy to `android/api` GitHub Pages folder

wipfli:
By default, the github pages deploy action takes a folder and pushes the contents of this folder to the github pages branch.

This deletes all previous files on the github pages branch. This would mean that an existing ios/api folder would be deleted if we push the android/api folder.

With this pull request, we only push the android docs to the android/api folder and not to the root folder.


nyurik (approved):
thx!

PR #330, opened 2022-06-14
Restrict cancel workflow

wipfli:


brendan-ward (commented):
This looks OK.  The real test would be triggering each of the workflows manually (they each have a `workflow_dispatch`), then a few moments later triggering them manually again to see if it cancels the first set.  Also would be a good test to be sure they aren't clobbering each other anymore. (hard to tell here; this only seems to have run the node-ci jobs)

(I'm not able to trigger these; don't have permissions)

wipfli:
@brendan-ward I added you together with @nvanfleet and @atierian to the list of people with write access to MapLibre GL Native. This means you can now approve and merge approved pull requests. So, welcome to the team :)

brendan-ward (approved):

PR #333, opened 2022-06-15
CI: ignore non-target platforms / CI workflows from each CI workflow

brendan-ward:
Many of the CI workflows had blanket exclusions to ignore changes to workflows, unfortunately including their own.  This made it hard to test #330

Provided I grouped the platforms & associated CI workflows correctly, these now should ignore changes to non-target platform directories as well as non-target platform CI workflows (or in a couple cases, other workflows for the same platform).


brendan-ward:
This includes follow-on fixes from #330; the concurrency setting needed to be moved up to the workflow level rather than job level (otherwise the jobs canceled each other within a workflow).  Sorry @wipfli that I missed that during review.

Based on the most recent test of pushing a trivial update while the worflows were still running, this looks to properly cancel running jobs within the PR.  Since there aren't jobs running in other PRs I couldn't very those aren't impacted (something to watch for).

This specifically excludes updates to `main` from canceling running jobs from prior commits on 'main`.  I _think_ that is what we want, but good to get input on that.  Also; it is hard to test impacts to runs from `main` from within this PR.

.github/workflows/android-ci.yml
```diff
@@ -6,18 +6,35 @@ on:
...
+      - ".github/workflows/ios*.yml"
+      - ".github/workflows/macos*.yml"
+      - ".github/workflows/node*.yml"
+      - ".github/workflows/qt*.yml"
+      - ".github/workflows/update-gl-js.yml"
+
+concurrency:
+  # cancel jobs on PRs only
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
wipfli, line 33: Since we have android-ci-pull.yml and android-ci.yaml, and since we want to only cancel workflow runs on pull requests, I wonder if this file should not have concurrency settings at all?

brendan-ward:
Looks like our `paths-ignore` directive isn't working as expected; that last commit should not have triggered all the builds.

wipfli (approved):

wipfli:
Maybe it is easier if we merge this to main and debug from there. The behavior when a workflow file is on main might be different than when it is in a pull request...

PR #345, opened 2022-07-01
Clean-up changelog

wipfli:
The changelog still uses `master`. This pull request replaces `master` with `main` and adds the sections again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Changelog
 
-## master
+## main
 
-- [core] `MaptilerFileSource` renamed to `MBTilesFileSource`.
+### ✨ New features
+
+- *...Add new stuff here...*
+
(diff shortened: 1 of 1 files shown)


ntadej (approved):

wipfli:
Or maybe the intention was that we use the changelogs in the platform folders. For example, here is a pull request that changed something in the core I think and the iOS and Android changelogs were updated:

https://github.com/maplibre/maplibre-gl-native/pull/34/files

ntadej:
I think we need to separate "core" changelog and platform changelogs.

Maybe something to discuss at the next steering meeting.

wipfli:
I added an agenda item.

wipfli:
If the core and platform bindings were versioned independently and lived in separate projects, we would not get into this situation where one fix leads to two changelog updates...

PR #351, opened 2022-07-04
Fix render tests

wipfli:
This pull request reverts some changes of #270 which broke the render tests.

Probably the syntax has to be adapted.

### DIFF
--- a/src/mbgl/text/collision_index.cpp
+++ b/src/mbgl/text/collision_index.cpp
@@ -415,8 +415,8 @@ std::pair<Point<float>,float> CollisionIndex::projectAndGetPerspectiveRatio(cons
     auto size = transformState.getSize();
     return std::make_pair(
         Point<float>(
-            (((p[0] / p[3] + 1) / 2) * size.width) + viewportPadding,
-            (((-p[1] / p[3] + 1) / 2) * size.height) + viewportPadding
+            static_cast<float>(((p[0] / p[3] + 1) / 2) * size.width + viewportPadding),
+            static_cast<float>(((-p[1] / p[3] + 1) / 2) * size.height + viewportPadding)
         ),
         // See perspective ratio comment in symbol_sdf.vertex
         // We're doing collision detection in viewport space so we need
--- a/src/mbgl/text/collision_index.cpp
+++ b/src/mbgl/text/collision_index.cpp
@@ -122,22 +122,22 @@ IntersectStatus CollisionIndex::intersectsTileEdges(const CollisionBox& box,
     const float tileY2 = tileEdges[3];
 
     // Check left border
-    int minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
+    int minSectionLength = static_cast<int>(std::min(tileX1 - x1, x2 - tileX1));
     if (minSectionLength <= 0) { // Check right border
-        minSectionLength = std::min(tileX2 - x1, x2 - tileX2);
+        minSectionLength = static_cast<int>(std::min(tileX2 - x1, x2 - tileX2));
     }
     if (minSectionLength > 0) {
         result.flags |= IntersectStatus::VerticalBorders;
-        result.minSectionLength = static_cast<int>(minSectionLength);
+        result.minSectionLength = minSectionLength;
     }
     // Check top border
-    minSectionLength = std::min(tileY1 - y1, y2 - tileY1);
+    minSectionLength = static_cast<int>(std::min(tileY1 - y1, y2 - tileY1));
     if (minSectionLength <= 0) { // Check bottom border
-        minSectionLength = std::min(tileY2 - y1, y2 - tileY2);
+        minSectionLength = static_cast<int>(std::min(tileY2 - y1, y2 - tileY2));
     }
     if (minSectionLength > 0) {
         result.flags |= IntersectStatus::HorizontalBorders;
-        result.minSectionLength = std::min(result.minSectionLength, static_cast<int>(minSectionLength));
+        result.minSectionLength = static_cast<int>(std::min(result.minSectionLength, minSectionLength));
     }
     return result;
 }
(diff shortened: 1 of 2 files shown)


wipfli:
Instead of running all render tests, one can only run the ones which currently fail on main with a `run-render-tests.sh` script on linux which might look like this:

```sh
#!/bin/bash
cmake --build build -j $(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null)
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "render-tests/symbol-placement/line-overscaled"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "render-tests/symbol-visibility/visible"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "render-tests/text-variable-anchor/rotated-offset"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "render-tests/text-variable-anchor/all-anchors-tile-map-mode"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "query-tests/circle-radius/tile-boundary"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "query-tests/fill-features-in/default"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "query-tests/invisible-features/zero-opacity"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "query-tests/edge-cases/box-cutting-antimeridian-z1"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "query-tests/edge-cases/box-cutting-antimeridian-z0"
./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json --filter "query-tests/edge-cases/null-island"
``` 
This script compiles and then runs the render tests, that were broken by #270. Note that there is one more render test which is broken, something like `enlarge-both`, but this one was already broken before #270...

ntadej (changes requested):
I will object to this right now. Compared to the other fix, none of these changes (especially the `minSectionLength` one look wrong.

Can you please paste the failing test results?

wipfli:
On main, 11 tests fail when running `./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json`

Here is the test output: https://wipfli.github.io/debug-maplibre-gl-native/main-linux-clang8-release-style.html

ntadej:
> On main, 11 tests fail when running `./build/mbgl-render-test-runner --manifestPath metrics/linux-clang8-release-style.json`
> 
> Here is the test output: https://wipfli.github.io/debug-maplibre-gl-native/main-linux-clang8-release-style.html

Thanks. Looking into it.

Note that for posterity it may be useful to attach images to the PR.

wipfli:
Good idea, @ntadej. Here are some screenshots:

image

image

image

src/mbgl/util/grid_index.cpp
```diff
@@ -263,12 +263,12 @@ void GridIndex<T>::query(const BCircle& queryBCircle, std::function<bool (const
 
 template <class T>
 std::size_t GridIndex<T>::convertToXCellCoord(const float x) const {
-    return util::max(size_t(0), util::min(xCellCount - 1, static_cast<size_t>(std::floor(x * xScale))));
```
ntadej, line 266: This one happens if `x` is negative. Thinking of a most elegant fix.

I also found another set of warnings that we can enable which will help us detect signed vs unsigned issues. Will investigate how many warnings there are.

src/mbgl/text/collision_index.cpp
```diff
@@ -415,8 +415,8 @@ std::pair<Point<float>,float> CollisionIndex::projectAndGetPerspectiveRatio(cons
     auto size = transformState.getSize();
     return std::make_pair(
         Point<float>(
-            static_cast<float>(((p[0] / p[3] + 1) / 2) * size.width) + viewportPadding,
-            static_cast<float>(((-p[1] / p[3] + 1) / 2) * size.height) + viewportPadding
+            (((p[0] / p[3] + 1) / 2) * size.width) + viewportPadding,
+            (((-p[1] / p[3] + 1) / 2) * size.height) + viewportPadding
```
ntadej, line 419: We have to drop precision at the end of calculation:
```suggestion
            static_cast<float>(((p[0] / p[3] + 1) / 2) * size.width + viewportPadding),
            static_cast<float>(((-p[1] / p[3] + 1) / 2) * size.height + viewportPadding)
```

src/mbgl/text/collision_index.cpp
```diff
@@ -122,7 +122,7 @@ IntersectStatus CollisionIndex::intersectsTileEdges(const CollisionBox& box,
     const float tileY2 = tileEdges[3];
 
     // Check left border
-    float minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
+    int minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
```
ntadej, line 125: This one does not seem to be needed
```suggestion
    float minSectionLength = std::min(tileX1 - x1, x2 - tileX1);
```

ntadej (changes requested):
Found all the issues. Sorry again for causing them. As I said, will try to enable some more warnings to be able to spot them easier.

src/mbgl/text/collision_index.cpp
ntadej: False alarm. Still investigating this one.

wipfli:
Don't worry about it. I think the changes in #270 are good and probably we can iron out the typos with the render tests now...

ntadej:
OK for the last change. Can you:
 - change the type back to `int`
 - add static casts to `int` to each assignment (when calling `std::min`)
 - remove redundant `static_cast`s below when this variable is used.

ntadej (approved):
Thanks for identifying the problematic files, @wipfli. Looks good to me now.

Note that at least the intersection detection code is very strange and it would be good to understand it.

PR #352, opened 2022-07-04
Only override CMAKE_OSX_ARCHITECTURES for XCode

ntadej:
Only override `CMAKE_OSX_ARCHITECTURES` for `XCode`. This simplifies/fixes node build on macOS.

Looking at `platform/ios/Makefile` we only ever use XCode for building iOS and macOS bindings so this should be a safe change.

### DIFF
--- a/platform/ios/ios.cmake
+++ b/platform/ios/ios.cmake
@@ -7,7 +7,9 @@ endif()
 # Override default CMake NATIVE_ARCH_ACTUAL
 # https://gitlab.kitware.com/cmake/cmake/-/issues/20893
 # https://stackoverflow.com/a/22689917/5531400
-set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+if(CMAKE_GENERATOR STREQUAL Xcode)
+    set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+endif()
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] "YES")
 
 macro(initialize_ios_target target)
--- a/platform/macos/macos.cmake
+++ b/platform/macos/macos.cmake
@@ -4,7 +4,9 @@ set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11")
 # Override default CMake NATIVE_ARCH_ACTUAL
 # https://gitlab.kitware.com/cmake/cmake/-/issues/20893
 # https://stackoverflow.com/a/22689917/5531400
-set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+if(CMAKE_GENERATOR STREQUAL Xcode)
+    set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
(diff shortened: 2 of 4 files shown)


platform/ios/ios.cmake
```diff
@@ -7,7 +7,9 @@ endif()
 # Override default CMake NATIVE_ARCH_ACTUAL
 # https://gitlab.kitware.com/cmake/cmake/-/issues/20893
 # https://stackoverflow.com/a/22689917/5531400
-set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+if(CMAKE_GENERATOR STREQUAL Xcode)
+    set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+endif()
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] "YES")
```
wipfli, line 13: Should the `set_target_properties` also be in the `if` block? I see that the comment on stackoverflow mentions the `set_target_properties` thing... https://stackoverflow.com/a/22689917/5531400

platform/macos/macos.cmake
```diff
@@ -4,7 +4,9 @@ set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11")
 # Override default CMake NATIVE_ARCH_ACTUAL
 # https://gitlab.kitware.com/cmake/cmake/-/issues/20893
 # https://stackoverflow.com/a/22689917/5531400
-set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+if(CMAKE_GENERATOR STREQUAL Xcode)
+    set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+endif()
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] "YES")
```
wipfli, line 10: Same here about `set_target_properties`

wipfli (commented):
Thanks for this suggestion @ntadej. What I don't understand is why does a change in the cmake files in the platform/macos and platform/ios folders have an effect on platform/node... Can you maybe help me here?

platform/ios/ios.cmake
ntadej: `XCODE_ATTRIBUTE_*` only matters for XCode.

platform/macos/macos.cmake
ntadej: Same here.

platform/ios/ios.cmake
wipfli: OK

wipfli (approved):
Since the iOS and macOC CI is green, I think it is safe to merge this. I am setting up OSX in virtualbox on my ubuntu laptop. If I can get that to work, then I can also more directly work with XCode...

ntadej:
Maybe @atierian can also comment if this would have any negative side-effect on the builds?

brendan-ward (approved):
Thanks @ntadej 

I don't do the XCode builds, but this solved core / NodeJS build issues for me on MacOS 10.15.   (was previously commenting out line 7 in `macos.cmake`)

PR #372, opened 2022-07-18
Update how ios and macos CMake files read from the ARCHS_STANDARD environment variable

hy9be:
Using the standard way to access environment variables.

This will avoid the manual workaround like [this](https://github.com/maplibre/maplibre-gl-native/issues/255#issuecomment-1069093435), when trying to build the entire project with CMake by:

``` shell
ARCHS_STANDARD=$(uname -m) cmake . -B build
cmake --build build
```

### DIFF
--- a/platform/ios/ios.cmake
+++ b/platform/ios/ios.cmake
@@ -7,7 +7,7 @@ endif()
 # Override default CMake NATIVE_ARCH_ACTUAL
 # https://gitlab.kitware.com/cmake/cmake/-/issues/20893
 # https://stackoverflow.com/a/22689917/5531400
-set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+set(CMAKE_OSX_ARCHITECTURES $ENV{ARCHS_STANDARD})
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] "YES")
 
 macro(initialize_ios_target target)
--- a/platform/macos/macos.cmake
+++ b/platform/macos/macos.cmake
@@ -4,7 +4,7 @@ set(CMAKE_OSX_DEPLOYMENT_TARGET "10.11")
 # Override default CMake NATIVE_ARCH_ACTUAL
 # https://gitlab.kitware.com/cmake/cmake/-/issues/20893
 # https://stackoverflow.com/a/22689917/5531400
-set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+set(CMAKE_OSX_ARCHITECTURES $ENV{ARCHS_STANDARD})
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] "YES")
 
 set_target_properties(mbgl-core PROPERTIES XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES)


ntadej:
Thanks! See also #352, I prefer that one a bit but I'm waiting for iOS experts to sign-off.

Note that in general it's always recommended to skim through the open PRs to not duplicate your efforts.

hy9be:
Ha. Thanks. I did miss that.

I think the problem this PR and https://github.com/maplibre/maplibre-gl-native/pull/352 trying to solve are different though. This one focus on using ``$EV{xxx}`` instead of ``"$(xxx)"`` when accessing the environment variables in cmake files, as the later could run into issues like this https://github.com/maplibre/maplibre-gl-native/issues/255#issue-1168439113. 

https://github.com/maplibre/maplibre-gl-native/pull/352 seems to be more related to simplifying the build for node and other non-xcode based platforms?

ntadej:
As far as I understand the variable in question is not an evnironment variable but a Xcode variable.

hy9be:
> As far as I understand the variable in question is not an evnironment variable but a Xcode variable.

My understanding is it is a environment variables that cmake uses to generate the ``c++`` or ``g++`` commands in ``compile_commands.json``. And an example of these commands is

``` json
{
  "directory": "/Users/xxx/Codes/github.com/maplibre-gl-native/build/benchmark",
  "command": "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ ... -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk -mmacosx-version-min=10.11 -fPIC -std=c++17 -o CMakeFiles/mbgl-benchmark.dir/function/source_function.benchmark.cpp.o -c /Users/xxx/Codes/github.com/maplibre-gl-native/benchmark/function/source_function.benchmark.cpp",
  "file": "/Users/xxx/Codes/github.com/maplibre-gl-native/benchmark/function/source_function.benchmark.cpp"
}
```

The ``arm64`` value for ``-arch`` is from this ``ARCHS_STANDARD`` environment variable. If I change it to some random string by doing
```
ARCHS_STANDARD=abc cmake . -B build
```
The ``cmake --build build`` command fails with ``clang: error: invalid arch name '-arch abc'``.

ntadej:
The point is, if one does not run Xcode, you should just set `CMAKE_OSX_ARCHITECTURES` directly when running `cmake`.

In any case we need an iOS developer to test both approaches.

wipfli:
I think if our CI should be able to detect problems with build configs. So if this pull request passes CI, then we should be fairly certain that we can merge it. @hy9be you tested this change locally with repect to the iOS test app, right?

hy9be:
@wipfli No worries. We can hold off until we have a conclusion for https://github.com/maplibre/maplibre-gl-native/pull/352 first. I was mainly trying to fix a syntax issue as cmake does not recognize ``$(xxx)``. If we write a ``CMakeLists.txt`` like this

``` cmake
set(CMAKE_OSX_ARCHITECTURES $(ARCHS_STANDARD))
message(STATUS "The value of CMAKE_OSX_ARCHITECTURES is ${CMAKE_OSX_ARCHITECTURES}" )
```

It prints ``The value of CMAKE_OSX_ARCHITECTURES is $(ARCHS_STANDARD)``.

@ntadej 's idea sounds right to me to handle the situation when we do not use xcode. Let's wait until https://github.com/maplibre/maplibre-gl-native/pull/352 being merged first :)

ntadej:
@wipfli, the point is that while the CI may pass we may also disable non-default architectures by accident so we should be a bit careful.

PR #373, opened 2022-07-19
Wrap forward declared Qt classes with Qt namespace macros

tnixeu:
I noticed that the forward declarations of Qt classes are not wrapped with the Qt namespace macros.

Usually the Qt namespace macros are not defined and this is not an issue. 
However, when Qt is compiled to a different namespace then not wrapping the forward declarations breaks the build.


ntadej (approved):

ntadej:
Thanks @tnixeu for the contribution!

Issue #312, opened 2022-06-08
Secrets documentation


wipfli:
MapLibre GL Native has 14 GitHub secrets. Are these documented somewhere?

wipfli:
Who is CR? I see that we use their personal access token in the android ci workflow...

https://github.com/maplibre/maplibre-gl-native/blob/ca5810f61fde26d0fbccbe4d48d4c771d02e460f/.github/workflows/android-ci.yml#L43-L45

wipfli:
The graddle config might be related to

https://github.com/maplibre/maplibre-gl-native/blob/5b3d936521040149e343f81ed4bd2ce24b75e690/platform/android/scripts/create-signing-key.sh#L13-L17

wipfli:
At the moment we have these github secrets:
![image](https://user-images.githubusercontent.com/53421382/178566955-1cfbf7d9-a499-40e3-8b87-976b08f48d79.png)

wipfli:
Lets store all the secrets as clear text in our password manager...

Is in password manager?

* [ ] BINTRAY_API_KEY
* [ ] BINTRAY_USER
* [ ] CR_PAT
* [ ] DIST_GITHUB_TOKEN
* [ ] GLCLOUD_SERVICE_JSON
* [ ] GPG_KEY_CONTENTS
* [ ] GPG_PASSPHRASE
* [ ] MAPBOX_DEVELOPER_CONFIG_XML
* [ ] MAPLIBRE_DEVELOPER_CONFIG_XML
* [ ] NODE_PRE_GYP_GITHUB_TOKEN
* [ ] OSSRH_PASSWORD
* [ ] OSSRH_USERNAME
* [ ] SIGNING_KEY_ID
* [ ] SIGNING_PASSWORD
* [ ] SONATYPE_STAGING_PROFILE_ID

wipfli:
Since `CR_PAT` is used in the container section here:

https://github.com/maplibre/maplibre-gl-native/blob/73973703426ebf11c3c6bea60857644c44d7b5e5/.github/workflows/android-release.yml#L12-L16

It might be that the abbreviation stands for "Container Registry Personal Access Token"...

louwers:
It's also possible to get the plain text of a secret with a hack: https://stackoverflow.com/questions/63003669/how-can-i-see-my-git-secrets-unencrypted

ovivoda:
[...]

PR #378, opened 2022-07-30
Node release workflow

acalcutt:
This is an attempt to create a node release workflow for https://github.com/maplibre/maplibre-gl-native/issues/314 

It is made to publish to github releases and the npm registry. It is based on what I am using for [@acalcutt/maplibre-gl-native](https://www.npmjs.com/package/@acalcutt/maplibre-gl-native) and https://github.com/acalcutt/maplibre-gl-native/releases. It is based on some ideas and conversation from @mnutt in https://github.com/maplibre/maplibre-gl-native/pull/217

**What this needs:**
'Environment Secrets' set in github for 'NODE_PRE_GYP_GITHUB_TOKEN' and 'NPM_TOKEN'

NODE_PRE_GYP_GITHUB_TOKEN (needed to publish to github)
```
Within GitHub, create a new authorization:

go to Settings -> Developer settings
click [Personal access tokens](https://github.com/settings/tokens)
click Generate new token
Select public_repo and repo_deployment
Generate Token
copy the key that's generated and set NODE_PRE_GYP_GITHUB_TOKEN environment variable to it. Within your command prompt:
SET NODE_PRE_GYP_GITHUB_TOKEN=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

NPM_TOKEN (needed to publish to the maplibre npmjs registry)
This needs a maplibre npmjs token which has publish rights...like created here https://docs.npmjs.com/creating-and-viewing-access-tokens

**Other Notes**

- I had to make two changes to node-pre-gyp-github . The first change made it target the 'main' branch instead of 'master'. The second makes the name include 'node-' to fit in with the other maplibre releases (so the name would be like 'node-v5.0.0' instead of 'v5.0.0').
- Because there is no automatic versioning on this project, the version in package.json must be changed before running the node-release workflow.
- I removed 'macos-11.0' and 'ubuntu-18.04' after making the node matrix because it was just to many jobs running. also 'ubuntu-18.04' had trouble building on node 18 due to dependencies not available in that OS. I was planning to use the newer MacOS OS, but was asked to use the older one for compatibility (the older os version would work on the newer od, but not the other way around)


ntadej:
Hi @acalcutt, thank you for preparing this. Do we need any secrets/tokens? I guess yes to publish to npm.

We plan to tag a prerelease version soon for Qt and may as well use that to also publish the node binaries.

wipfli:
I shared the `NPM_ORG_TOKEN` with MapLibre GL Native. This gives you access to publish packages in `@maplibre` on NPM.

image

wipfli:
Stuff to watch out for are pre-releases. These should be explicitly tagged as `next`.

wipfli:
https://github.com/maplibre/maplibre-gl-js/blob/f7043ce3c7e9bf778e9204926d20baff4dea2d1b/.github/workflows/release.yml#L144-L156

wipfli:
@acalcutt it would be great if you could document the release process in a readme.

acalcutt:
I'll update NPM_TOKEN to NPM_ORG_TOKEN, so it can use the one you put in place.

This also needs a github token for the automatic publish to github, which looks like this
image . In my repo this was under my account, but here i'm not sure whatever user we want to publish

wipfli:
Maybe you can use `CR_PAT` for this. @roblabs do you know what this secret does:

https://github.com/maplibre/maplibre-gl-native/blob/fbfff8fbabc1f3f4c390123ca1f721d0e4921433/.github/workflows/android-release.yml#L16

acalcutt:
For the pre-release question, what would determine it was a prerelease? Right now node-pre-gyp-github is set up to publish a release like 'node-v5.0.0' as part of the workflow...which is similar to how the other release are labeled. Right now there is no pre-release

birkskyum:
@acalcutt it's quite subjective, and we generally use prereleases very little, but there are times when it's convenient to gather feedback before pushing a change to everyone. Instead of alpha, beta and rc channels, we have just used pre-release the for now. It's been used only a few times for breaking changes and for big features like 3d terrain, but the criteria is often come down to how comfortable we feel about the quality of a specific change.

acalcutt:
This workflow could be changed so it just makes a pre-release, then someone could just mark it released in github when it is ready. Basically would be removing "--release" from https://github.com/maplibre/maplibre-gl-native/blob/cd692e9fbd113c2dddc2c055249658f484a59586/package.json#L66 . Not sure if we would want to do the same on the npm side though ([looks possible](https://stackoverflow.com/questions/44567010/are-there-npm-version-prerelease-identifiers)) 

I'm sure this needs some work to update the macos runners now, since I see the older macos releases got depreciated

Really what I would love to see is some automated version control. It's really kind of a pain having to manually increase the version number before doing a release. In my [tileserver-gl workflow](https://github.com/acalcutt/tileserver-gl/blob/main/.github/workflows/Build_Test_Release.yml), I am using [phips28/gh-action-bump-version@master](https://github.com/phips28/gh-action-bump-version) and it works pretty well

wipfli:
In maplibre-gl-directions we also bump the version automatically https://github.com/maplibre/maplibre-gl-directions

acalcutt:
Bumping the version in the root package.json wouldn't affect the version in the ios/android releases right? those seem to be separate version numbers?

wipfli:
Yes

acalcutt:
I think I got incrementing the version and publishing pre-releases working. I used some parts of maplibre-directions workflow @wipfli mentioned, but I had to do a lot of trial and error getting it working.

The version incrementing was a bit if a pain because of the way I am using a os/node matirx to publish the binaries to github. the version change had to be done in a separate job before the build matrix ran so they all have the same version. when the second job ran did not pull in the version changes and I had to do an additional "git pull" in the build steps to pull that in.

I also found, doing a pre-release wasn't easy with 'node-pre-gyp-github'. I was saying before I thought I could remove the "--release" switch, but that just made it a draft not a pre-release. I actually had to go make more modifications to node-pre-gyp-github to allow it to submit a pre-release

Also, side note. I accidently merge my other branch which had a lot of history I didn't want in this PR (ugh). I could submit a cleaner PR if it would be easier.

.github/workflows/node-ci.yml
```diff
@@ -74,9 +74,12 @@ jobs:
     strategy:
       matrix:
         os:
-          - [self-hosted, macOS, ARM64]
```
birkskyum, line 77: This is macos-12 as well, and a lot faster - can it not run there?

.github/workflows/node-ci.yml
```diff
@@ -74,9 +74,12 @@ jobs:
...
+          - macos-12
           - ubuntu-20.04
+        node: [ 10, 14, 16, 18 ]
```
birkskyum, line 79: Let's focus on the LTS here. Believe node 10 had EOL 30 Apr 2021, so it's not safe to use.

.github/workflows/node-release.yml
```diff
@@ -0,0 +1,185 @@
...
+          npm version ${{ github.event.inputs.version }} --preid pre -m "Update node version to (${{ github.event.inputs.version }})"
+          git push
+
+  build:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-20.04
+          - macos-12
+        node: [ 10, 14, 16, 18 ]
```
birkskyum, line 49: Same as above comment.

.github/workflows/node-release.yml
```diff
@@ -0,0 +1,185 @@
...
+          git config user.email github-actions@github.com
+          npm version ${{ github.event.inputs.version }} --preid pre -m "Update node version to (${{ github.event.inputs.version }})"
+          git push
+
+  build:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-20.04
+          - macos-12
```
birkskyum, line 48: For macos, this can be run on our faster M1 runner as well, if set to [self-hosted, macOS, ARM64]

package.json
```diff
@@ -15,13 +15,14 @@
...
   "repository": {
     "type": "git",
-    "url": "git://github.com/maplibre/maplibre-gl-native.git"
+    "url": "git+https://github.com/maplibre/maplibre-gl-native.git"
   },
   "license": "BSD-2-Clause",
   "dependencies": {
     "@mapbox/cmake-node-module": "^1.2.0",
+    "@mapbox/node-pre-gyp": "^1.0.9",
+    "@acalcutt/node-pre-gyp-github": "1.4.8",
```
birkskyum, line 24: Did you attemt to upstream these changes, or would they not make sense for anyone but us?

package.json
```diff
@@ -15,13 +15,14 @@
...
     "minimatch": "^3.0.4",
-    "node-pre-gyp": "^0.10.2",
```
birkskyum, line 24: node-pre-gyp was recently bumped from 0.10 -> 0.17, which makes a conflict. It should of course have been replaced by "@mapbox/node-pre-gyp". I'll make a PR to move main to @mapbox/node-pre-gyp so it becomes easier to resolve for this branch.

birkskyum (approved):
Great work! It's getting really close now. I've left a few comments, but nothing major.

birkskyum (dismissed):
See comments

.github/workflows/node-ci.yml
acalcutt: I can change it do that if it works. I was testing on my own fork so I don't think I have that there.

.github/workflows/node-ci.yml
acalcutt: Node 10 was kept because it was the last node version mapbox-gl-native was built for. It made switchover/testing/comparison easier if you could have a release that supports both mapbox and maplibre since there aren't a lot of examples out there.

acalcutt: I'm fine with removing it here, but it was really useful for bug comparison, like the app I made for to compare the label clipping issue ( https://github.com/acalcutt/maplibre-label-clipping-test ), which I could switch between mapbox and maplibre and see the difference in results.

package.json
acalcutt: I remove node-pre-gyp in favor of @mapbox/node-pre-gyp already. This is a removed line right or am I missing something?

package.json
acalcutt: Two changes were specific to us, and may not be useful to others upstream

1.) The original package targets the 'master' branch and has no real way to change or specify a different branch
https://github.com/acalcutt/node-pre-gyp-github/blob/master/index.js#L71

2.) I changed the default name to include 'node-v' instead of just 'v' so it fit in with the rest of our releases
https://github.com/acalcutt/node-pre-gyp-github/blob/master/index.js#L72

The 3rd change could be useful to others, but it specifically tagets the node version preid I set in the workflow. other users of this might want a different preid which would make this not as useful to them.

3.) Make a pre-release on versions that include "pre"
https://github.com/acalcutt/node-pre-gyp-github/blob/master/index.js#L64
https://github.com/acalcutt/node-pre-gyp-github/blob/master/index.js#L75

So ya, I don't know if these are useful upstream without some config options to make it more useful to others

.github/workflows/node-ci.yml
acalcutt: Also, if this is compiled on arm, does this make it not work on other architectures? in the origional bring back node thread it was requested I use the older realease (when it was available) because otherwise it required libs you couldn't get on the older OS versions.

package.json
birkskyum: You're not missing something. The node-pre-gyp was already upgraded to 0.17 in main, making a conflict, but main is on @mapbox/node-pre-gyp now to relieve it. This branch doesn't have to contain that specific change anymore.

.github/workflows/node-ci.yml
birkskyum: Okay, if it's useful to have node 10, and it doesn't add complexity to the codebase moving forward, then let's keep it for now.

acalcutt:
Ok, I think I've made the requested changes other than upstreaming node-pre-gyp-github (which I wouldn't recommend without making it more universal to others).

.github/workflows/node-ci.yml
birkskyum: I believe it's possibly true. The same could happen with these GitHub runners in the near future https://github.com/github/roadmap/issues/528

The thing is that M1's are much more economical to use for runners because they roughly have half the power draw and twice the compute, which is why they are a lot cheaper for us to self-host on aws as well: https://devdosvid.blog/2021/01/19/mac1.metal-and-mac2.metal-ec2-instances-user-experience/

I don't have a good enough overview of our aws budget yet to figure out if we can set up an x86 instance to make this build.

.github/workflows/node-ci.yml
acalcutt: Just saw this comment. i added node 10 back for now. I agree we can remove it soon, but having at least a few releases with it would be good for migrations.

acalcutt:
@birkskyum just a comment on the macos-12 vs self-hosted macos arm instance...

Looking at the package.json and the releases I made here ( https://github.com/acalcutt/maplibre-gl-native/releases/tag/node-v5.1.1 ) it looks like architecture is uses in the binary file name. I'm guessing this would be different on arm?

acalcutt:
I think If we wanted to support both x64 and ARM we would need both a X64 and ARM runner, at least with the way it works with this matrix right now.

birkskyum:
Maybe we can make a universal build like this
https://stackoverflow.com/questions/65157483/macos-build-universal-binary-2-with-cmake

acalcutt:
I think for that though, you would have to switch to something like @mnutt did here
https://github.com/mnutt/maplibre-gl-native/blob/publish-github/platform/node/scripts/publish.sh
and
https://github.com/mnutt/maplibre-gl-native/blob/publish-github/.github/workflows/node-ci.yml#L150-L166

This is somewhat an alternative to this node matrix method. definably an option though. I think i commented somewhere I thought this could be a better option.

acalcutt:
Although looking at it more it isn't much different. it still needs and arm and a x64 runner... it just doesn't need the node matrix

birkskyum:
NodeJS only supports arm from v16. https://doesitarm.com/app/nodejs/

acalcutt:
So i'm guessing the node 10, 14 builds would fail then right.

maybe I'll try to merge in @mnutts publish.js change . I think that will help that issue since it only needs to run on one version of node instead

mnutt:
I didn't know at the time that we had an M1 runner, that's great to hear! A real runner would allow us to build + test it much easier than messing with cross-compilation.

birkskyum:
@mnutt , we had them provided by AWS very recently so that's why. We have one running and I'll set up two more as soon as they are provisioned to unblock our CI which is blocked in queue most of the time.

acalcutt:
I've been testing an alternative workflow based on a mix of this and @mnutt 's version here https://github.com/WifiDB/maplibre-gl-native/tree/node-release2

but the main issue is i cant really test it since i cant seem to get a arm64 macos without doing the self hosting. even if i specificy arm64 it still runs x86_64.

birkskyum:
The github provided macos runners are all virtualized x86_64 instances still.

birkskyum:
I just noticed that it only had the first 3 labels, so I added thee macos-12 just now - if that's what you use to select machines it might help.

Screen Shot 2022-09-10 at 01 53 16

birkskyum:
If it's easier to make a node 16/18 arm/universal release, that would be a great start.

The compatibility version for older node and Intel could be added as a separate release workflow afterwards.

acalcutt:
its not really easier for me since i have no way to test or build an arm build at the moment.

birkskyum:
That makes sense; let's merge an x86 workflow first then. There appear to be some merge conflicts currently, probably due to the package-lock.json, so it's probably easiest to make a new PR with these changes to sort it out.

ntadej:
Just to remind you, there is never a need to open a new PR. You can always force push a fixed version of your branch.

birkskyum:
@ntadej, that's true. Resetting package.json and package-lock.json, rebasing, and reapplying the changes, would do the trick as well.

acalcutt:
I merged in main so I think the merge conflict should be fixed (for now until more package,json changes). I've had to keep package-lock.json updated or the npm ci fails in this workflow.

As for https://github.com/maplibre/maplibre-gl-native/pull/459 I think eventually it could be a be a better approach and it has some work toward arm64 and it doesn't need a node matrix like this version does. The other PR uses the fact that one build generates the binaries for multiple node versions, so it really doesn't have to be run for each node version...it can just run once and package and upload all the versions you want to keep. It also uses some existing build scripts (platform\node\scripts\publish.sh) which might make multiplatform easier.

However if you did approve this one, they other PR could easily be a next step. There's not a lot different between the two (the both publish to the same filenames and use the same node-pre-gyp-github to publish.

birkskyum (approved):
Great! When the CI passes, this seems to be ready to merge.

acalcutt:
can someone rerun those two failed ci tests and see if they work the second time. I have that "Upload render test artifacts" fail before like that once or twitch, but usually it works after a rerun.

acalcutt:
did something change where the regular macos-12 runner is now running on arm64? seems like that is the case. Any way to run an X64 macos-12 now?

birkskyum:
After adding the macos-12 label to the arm runners, the tests seem to run on it by default. I removed this label again and reran the tests.

birkskyum:
Great work!

acalcutt:
Now that it is merged, maybe we can the node-release workflow with the "prerelease" option and see if the "secrets.CR_PAT" and "secrets.NPM_ORG_TOKEN" work for publishing.

birkskyum:
Good idea. It's running now.

birkskyum:
We got this

Screen Shot 2022-09-11 at 13 28 27

ntadej:
OK, we'll have to figure out how to bypass this. Do you have reviews disabled in JS or how do you do it there?

birkskyum:
Review is not disabled in JS. We do manually update the version in a PR, and hit release.

birkskyum:
something like this:
https://github.com/maplibre/maplibre-gl-js/pull/1556

acalcutt:
I tested enabling branch protection in my branch and saw the same thing. It looks like the second option here works, specifying a token with permission on the checkout https://stackoverflow.com/a/63733988

Should I submit the change to this PR or make a new one?

birkskyum:
Because this is merged already, I think you have to make a new one.

ntadej:
What you could also think of is to make an action to make a PR with the changelog updates and version bump and then one to do final processing when the tag is made. This could also work.

acalcutt:
OK I added https://github.com/maplibre/maplibre-gl-native/pull/466

For the tag, that gets created by node-pre-gyp-github which is already using a git token. we will have to see if that has the same issue

acalcutt:
Hmm, that fix didn't seem to work here. wonder ing CR_PAT doesn't have the permission needed

birkskyum:
The PR is not, but the tag is created automatically in the JS repo.

acalcutt:
In my repo I am using the NODE_PRE_GYP_GITHUB_TOKEN secret for this, which seems to be working for me.  That was created like this, with public_repo and repo_deployment permissions

```
Within GitHub, create a new authorization:

go to Settings -> Developer settings
click [Personal access tokens](https://github.com/settings/tokens)
click Generate new token
Select public_repo and repo_deployment
Generate Token
copy the key that's generated and set NODE_PRE_GYP_GITHUB_TOKEN environment variable to it. Within your command prompt:
SET NODE_PRE_GYP_GITHUB_TOKEN=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

Though that token is a personal token and I am the administrator of those repos, so that could make a difference.

birkskyum:
I've tried adding a NODE_PRE_GYP_GITHUB_TOKEN with a key generated like you explained to the repos list of secrets but it still fail. Can we use the same approach as in GL JS repo where PR happens manually, and the release cycle takes it from there?

acalcutt:
This is what I am using in my branch, it is what the [node-pre-gyp-github](https://github.com/acalcutt/node-pre-gyp-github) readme says to use for permissions and it seemed to work. 

These are the only two checkbox checked in permissions
image

I just did a test run of this in my fork and it worked
https://github.com/acalcutt/maplibre-gl-native/actions/runs/3031922268

birkskyum:
Maybe swapping secrets.CR_PAT to secrets.NODE_PRE_GYP_GITHUB_TOKEN will resolve it then.

wipfli:
When adding new secrets, please write down what they do in #312. We could also start adding some documentation about the build processes and secrets used in this repo...

wipfli:
@birkskyum as a Governing Board member you have access to our Bitwarden key manager. Do you think it would make sense to store clear-text passwords and keys there as well? I am asking because maybe it is enough if we just document well how the secrets were generated and who has access to them...

birkskyum:
The release workflow is running now.

acalcutt:
It looks like it all published correctly. One thing I notice though is it shows that it was published by [birkskyum](https://github.com/birkskyum) not github-actions. Most likely because we used your token for the version bump checkout/push

Also, I know you we asking about the security of this. I was thinking about that and I wonder what would could happen if someone moved that into a pr ci script. could they affect something real?

Maybe I should make NODE_PRE_GYP_GITHUB_TOKEN and NPM_TOKEN some type of enterable field so it's not part of the workflow? idk

ntadej:
Maybe we should figure out if there's an action for that so we do not do it manually. It may help.

birkskyum:
This release workflow works quite well for gl-js to npm.
https://github.com/maplibre/maplibre-gl-js/blob/main/.github/workflows/release.yml

I wouldn't mind if we manually had to merge the updated version in package.json / package-lock.json / CHANGELOG.md before it published.

acalcutt:
That looks like it also uses secrets.GITHUB_TOKEN and secrets.NPM_ORG_TOKEN, so possibly the same concern? 

The PR ci workflow runs automatically with whatever workflow someone uploads.... so my thinking would be what happens if someone maliciously threw those secrets in a ci?

birkskyum:
In that workflow, GITHUB_TOKEN is used to make a tag and a release, not push a commit.

ntadej:
Indeed, but something in the current workflow made you Birk the releasing user, not `github-actions`. Maybe `GITHUB_TOKEN` is enough and can replace `NODE_PRE_GYP_GITHUB_TOKEN`?

acalcutt:
The permission we gave NODE_PRE_GYP_GITHUB_TOKEN we so node-pre-gyp-github could release the binaries. It just happended to work for that push too

ntadej:
Also note that in the Qt release workflow I added
```
    permissions:
      contents: write
```
to the job. Not sure where I copied it from but this may add more permissions to `GITHUB_TOKEN`.

birkskyum:
Also, regarding actions automatically running on a PR - it doesn't start workflows automatically for first-time contributors, but it's hardly a guarantee for anything.

acalcutt:
On the plus side I tried testing it in https://github.com/acalcutt/tileserver-gl/tree/maplibre_pre and it seems to work as expected.

PR #380, opened 2022-08-10
Remove observation of key-path

nvanfleet:
This was previously used in an assert when hiding the info view, but that implemenatation was removed but this wasn't.


wipfli:
Could this also be related to #7?

nvanfleet:
>  Could this also be related to https://github.com/maplibre/maplibre-gl-native/pull/7?

It could be, that's a big diff. I saw that it was the assertion for hiding the info view in the Mapbox repo but is now un-used

ntadej (approved):

PR #381, opened 2022-08-10
Move mapview experiment file

nvanfleet:
The implementation of this file appears to be that it's used in the demo app and is not public, but it's still in the path and compiled into the SDK itself. This seems unecessary.


roblabs (approved):
Thank you for contributing to FOSS4G!!

PR #382, opened 2022-08-10
Remove account from xcconfig

nvanfleet:
This was previously used to support the compiled in Mapbox account code. It's not longer in the project but this search path does remain.


wipfli:
Maybe this code is a left-over from #7

nvanfleet:
> Maybe this code is a left-over from #7

possibly

ntadej (approved):

PR #383, opened 2022-08-10
Remove info plist variables

nvanfleet:
*I previously removed both the semantic version and the commit hash but I found that the semantic version is the good version string like "5.12.0" and not linked to the build.*

I could not find any usage of MGLCommitHash anywhere so it seems to be outright dead code. Ultimately there might have been some kind of specific scenario where this was used but from what I can tell it's no longer useful within Maplibre.

### DIFF
--- a/platform/ios/platform/ios/scripts/xcpackage.sh
+++ b/platform/ios/platform/ios/scripts/xcpackage.sh
@@ -42,14 +42,9 @@ mkdir -p ${BINOUT}
 step "Recording library version…"
 VERSION="${OUTPUT}"/version.txt
 echo -n "https://github.com/maplibre/maplibre-gl-native/commit/" > ${VERSION}
-HASH=`git log | head -1 | awk '{ print $2 }' | cut -c 1-10` && true
-echo -n "maplibre-gl-native-ios "
-echo ${HASH}
-echo ${HASH} >> ${VERSION}
 
 PROJ_VERSION=$(git rev-list --count HEAD)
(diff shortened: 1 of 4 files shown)


platform/ios/platform/ios/scripts/xcpackage.sh
```diff
@@ -42,14 +42,9 @@ mkdir -p ${BINOUT}
...
 VERSION="${OUTPUT}"/version.txt
 echo -n "https://github.com/maplibre/maplibre-gl-native/commit/" > ${VERSION}
-HASH=`git log | head -1 | awk '{ print $2 }' | cut -c 1-10` && true
-echo -n "maplibre-gl-native-ios "
-echo ${HASH}
-echo ${HASH} >> ${VERSION}
 
 PROJ_VERSION=$(git rev-list --count HEAD)
 SEM_VERSION=$( git describe --tags --match=ios-v*.*.* --abbrev=0 | sed 's/^ios-v//' )
-SHORT_VERSION=${SEM_VERSION%-*}
```
nvanfleet, line 52: This is also not used within the file.

ntadej:
It may be just used for reference as currently we also just referece the hash from C++ and not the semantic version.

nvanfleet:
> It may be just used for reference as currently we also just referece the hash from C++ and not the semantic version.

I guess you just mean that inspecting the info.plist can tell you the version if you are somehow inspecting it? That seems to be somewhat less useful than something that is accessed within the codebase like the `MGLSemanticVersionString`

nvanfleet:
Okay I think it's fine if this is used by anyone.

PR #384, opened 2022-08-10
Remove duplicate files from ios-darwin and darwin

nvanfleet:
There are duplicated files between platform/iOS/platform/darwin/src and platform/darwin/src

```
src/headless_backend_cgl.mm
src/image.mm
src/string_nsstring.mm
src/headless_backend_eagl.mm
src/number_format.mm
src/local_glyph_rasterizer.mm
src/logging_nslog.mm
src/nsthread.mm
src/collator.mm
src/CFHandle.hpp
src/timer.cpp
src/run_loop.cpp
src/async_task.cpp
src/gl_functions.cpp
```

There are three files which are not the exact same diff even

```
src/local_glyph_rasterizer.mm
src/nsthread.mm
src/collator.mm
```

Uniformly it seems like all platform/iOS/platform/darwin/src files should be removed keeping the platform/darwin/src versions.

I believe these files should have been deleted with a previous change of some kind and are otherwise not referenced.


ntadej:
While this looks fine, we should really remove double nesting of platforms. I started a while ago here #220, but did not continue due to my limited knowledge of iOS/macOS SDK structure.

I'm not sure how we should proceed but maybe doing everything in one go would be even easier to track?

nvanfleet:
> While this looks fine, we should really remove double nesting of platforms. I started a while ago here #220, but did not continue due to my limited knowledge of iOS/macOS SDK structure.
> 
> I'm not sure how we should proceed but maybe doing everything in one go would be even easier to track?

Overall I am not a booster of the double nested structure. Furthermore there are things within platform/darwin and platform/default that appear to not apply to iOS and are presumably only for other platforms which I feel like should also be moved out or at least put in a directory that makes that clearer. 

I'm not really sure if any of that work should block removing untracked and un-used files though.

wipfli:
Thanks @nvanfleet for filing this pull request. How did you find these duplicated files? Is there a tool or something like that which could help us with such tasks?

nvanfleet:
> Thanks @nvanfleet for filing this pull request. How did you find these duplicated files? Is there a tool or something like that which could help us with such tasks?

I was working with my Maplibre Bazel build script and when pulling in whole directories would get issues with duplicate symbols etc. I then had to cross check the cmake/Xcodeproj to figure out the source of truth.

nvanfleet:
I'd still like to merge this given that the files are accidentally duplicated. I think there are definitely some larger-scale things to improve the code organization but I think this is distinct from that.

ntadej (approved):

ntadej:
OK, as it builds I see no blocker.

PR #386, opened 2022-08-11
Dependecy upgrades

birkskyum:
Instead of many small PR's with version updates, this will update most of our dependencies, so if we can test this branch and make it work we are a big step in the right direction. @roblabs  @ntadej 

The CI/CD environments are generally a bit outdated, so let's update those first ( see #387 ).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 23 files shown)


ntadej:
Do not update maplibre-gl-js in this one. It will not work due to shaders.

birkskyum:
Okay, I have reverted the change to submodule.

birkskyum:
I reverted the bump of benchmark from 1.5 to 1.7, because google apparently against semver introduced breaking changed in the 1.6 minor release.

ntadej:
The CI OS versions commit may be more suitable for a separate PR, but it's probably fine here :)

birkskyum:
> The CI OS versions commit may be more suitable for a separate PR, but it's probably fine here :)

I think you are very right about that. There is a PR in here #388

birkskyum:
closed in favour of smaller increments.

PR #387, opened 2022-08-11
Update github actions, and add dependabot to keep track of them

birkskyum:
We use many old versions of github actions, and dependabot can keep them evergreen for us. Most of these changes does that the node environment in the runners is node 16 by default.


ntadej (approved):

ntadej:
Thanks Birk for that!

PR #388, opened 2022-08-11
Explicit versions of CI OS

birkskyum:
While the macos-latest, ubuntu-latest, windows-latest can be a helpful tool to stay up to date, I believe it's more clear for now whats going on if we explicitly state what versions we are using, because we need consistency.

### DIFF
--- a/.github/workflows/android-docker-base.yml
+++ b/.github/workflows/android-docker-base.yml
@@ -4,7 +4,7 @@ on:
 jobs:
   push_to_registry:
     name: Docker image android-base
-    runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


ntadej:
Was this one closed by accident?

birkskyum:
No, it actually wasn't an accident, but we can reopen it when it's clear what OS changes need to happen. It depends on how it influences the other PR. It btw. seemed like the macos-11 didn't comply with installing https://libraries.io/homebrew/sourcekitten in the ios-ci and macos-ci tests.

ntadej:
I propose we reopen with changing latest->fixed but keep the current defaults as they are, what do you think?

PR #389, opened 2022-08-11
update package-lock.json to v2

birkskyum:
Small steps towards updating the dependencies

With Node 16 and executing:
`npm install --ignore-scripts`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,3949 @@
 {
   "name": "@maplibre/maplibre-gl-native",
   "version": "5.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@maplibre/maplibre-gl-native",
+      "version": "5.0.0",
+      "hasInstallScript": true,
+      "license": "BSD-2-Clause",
+      "dependencies": {
+        "@mapbox/cmake-node-module": "^1.2.0",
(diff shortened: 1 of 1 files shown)


wipfli:
Nice thanks @birkskyum! Do you know if we also have to update some readmes or a `.nvmrc` file?

ntadej:
Will this not break old node versions? It seems we still run in version 10 (at least according to #378)!

birkskyum:
I don't think we have to do, as the change is quite minimal - it's basically only adding info to the lockfile. I just noticed that every single github runner on every single build would notice a deprecated lockfile version and update it, so we might as well have it done.

birkskyum:
This lockfile version 2 is supposed to be backwards compatible with version 1. 

https://stackoverflow.com/questions/66150813/is-lockfileversion-2-in-package-lock-json-from-npm-7-compatible-with-older-ve

acalcutt:
Technically we don't have a official node release workflow yet, though I did submit what I am using in my fork in https://github.com/maplibre/maplibre-gl-native/pull/378 . I updated the lockfile to v2 in my fork at https://github.com/acalcutt/maplibre-gl-native and it didn't seem to cause me any issues with the node builds I do there.

Right now I am building in node 10 for tests and the github release, though the npm release builds all version under node 16

wipfli (approved):
Looks good to me, thanks @birkskyum. Note that there are 2 more `package.json` files somewhere in the subfolders...

PR #401, opened 2022-08-15
move ios static analysis to own linux runner

birkskyum:
Right now I added 2 workflows, one temporary one called ios-ci-trimmed that only has the static analysis, and is passing, and then another ios-ci-timmed-ubuntu which is an attempt of a linux port of the same workflow.


birkskyum:
close in favor of #426

PR #412, opened 2022-08-21
Fix typo in geo.cpp

ntadej:
Fix (old!) typo in `geo.cpp` to allow building for Android using new SDKs.


wipfli (approved):
How old is it?

wipfli:
Ah, can you add a changelog item?

ntadej:
5 years or so.

Also do such items really require a changelog item?

wipfli:
I don't know, up to you... We might want to have some automated changlog generation in the future.

ntadej:
Merging for now then 😊

wipfli:
Great thanks, should we update at some point the Android SDK version used in CI?

ntadej:
I'd wait a bit with that until we have better understanding how it works.

I updated the description now, it was actually NDK that was newer (SDK is for java stuff, NDK is for native compilation).

roblabs:
iOS builds did not error/warning when it was either `&` or `&&`.

ntadej:
> iOS builds did not error/warning when it was either `&` or `&&`.

Yes, I assume newer NDKs have stricter warnings/checks enabled.

PR #416, opened 2022-09-01
Migrate examples in android TestApp to Kotlin

birkskyum:
Many of the examples from the TestApp ported 1 to 1 to kotlin.


wipfli:
Thanks @birkskyum. I think Kotlin is the right language for Android. Let's wait with this pull request until we decided on #420.

wipfli:
Do we have some build instructions for this test app? Would like to try it out...

birkskyum:
Yes, it's a matter of opening Android Studio is opened in the platform/android. Create a virtual device and run the MapboxGLAndroidSDKTestApp. Gradle will have to do a lot of work when you open the repo, so it can take a little while, but apart from that, it's quite few steps.

wipfli:
Great, I will give it a go and ping you if I am stuck...

wipfli:
What is the easiest way to review this pull request? The diff seems to be too large for git to recognize that the files were renamed and the code was changed...

birkskyum:
The only way as I see it is to run the app before the change, see how each example works, and then apply the change and validate that the example works the same. Like a practical test.

wipfli:
Sure I can do that. I might also just rename the files back to .java and then do a diff...

birkskyum:
This is like a first round of porting. Many small changes can be done for each file due to the new capabilities of Kotlin, but that'll be a second round.

louwers:
I'm going through all the examples to add to the documentation. Just found out that missing parenthesis caused this funny regression.

https://github.com/user-attachments/assets/c63a2a4b-8f40-4acf-b8f1-eeb58c4b8e2d

PR #420, opened 2022-09-02
Design Proposal - Kotlin

birkskyum:
See proposal [here](https://github.com/maplibre/maplibre-gl-native/pull/420/files#diff-5444c30055b2c9929878e444f8f7858d0a8241fc7292732b5e8420911b388ce5)

### DIFF
--- a/design-proposals/2022-09-02-kotlin.md
+++ b/design-proposals/2022-09-02-kotlin.md
@@ -24,3 +24,4 @@ Full backward compatibility
   
 
 ## Rejected Alternatives
+The alternative is that we go forward with java, meaning that we effectively should port our existing kotlin code to java, which I can't find evidence of any project ever attempting.
--- a/design-proposals/2022-09-02-kotlin.md
+++ b/design-proposals/2022-09-02-kotlin.md
@@ -1,0 +1,26 @@
+# Kotlin Design Proposal
+
+## Motivation
+10 years ago, [JetBrains released Kotlin](https://blog.jetbrains.com/kotlin/2011/07/hello-world-2/)
+5 years ago, [Android got support for Kotlin](https://android-developers.googleblog.com/2017/05/android-announces-support-for-kotlin.html) 
+3 years ago, [Android became Kotlin-first](https://developer.android.com/kotlin/first)
+
+The argument for Kotlin over Java is typically along the lines of Kotlin being an "easier" language than Java. Practically, this shows in Kotlin being less verbose, so fewer lines of code to make errors in. It also has static typing instead of strong typing so many types are inferred, and lastly, it has a Null type which avoids all the common NullPointer exceptions. 
+
+Kotlin seems like a natural successor language to java, because it brings some advantages with no real downsides. It has full interoperability allowing us to port files 1-1.
+
+There are also intersting features being added to the language that we can explore in the future if we decide to invest in kotlin competence now - such as [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform.html#full-stack-web-applications) for cross-platform compilation.
+
+## Proposed Change
+
+I propose that we port all our Android platform code, so we have a language less in our codebase. First the TestApp and then the SDK itself.
+
+
+## API Modifications
+No modifications are planned.
+
+## Migration Plan and Compatibility
+Full backward compatibility
+  
+
+## Rejected Alternatives


lseelenbinder:
I would support this transition. Kotlin is much nicer to use than otherwise. 

My only request would be to ensure bytecode compatibility for older versions, which is likely a few settings we will already need to adjust.

birkskyum:
> bytecode compatibility

Absolutely. My understanding is that the bytecode won't be identical, but it should be possible to make it compatible.

birkskyum:
@wipfli , just putting your [comment](https://github.com/maplibre/maplibre-gl-native/pull/416#issuecomment-1236118879) here to collect feedback in one place.
> Thanks @birkskyum. I think Kotlin is the right language for Android. Let's wait with this pull request until we decided on https://github.com/maplibre/maplibre-gl-native/pull/420.

nvanfleet:
This is interesting, I wonder if we should consider migrating the iOS side to Swift?

birkskyum:
@nvanfleet , if you investigate the topic, and find that it's likely advantageous and have an idea about how we can do it, feel to make a design proposal to get some feedback. It's probably a good idea to cover possible changes in build size, the options for incremental adoption with bridge-headers and [Project]-swift.h, but it would be great to have a proposal to consider.

design-proposals/2022-09-02-kotlin.md
```diff
@@ -0,0 +1,27 @@
...
+
+The argument for Kotlin over Java is typically along the lines of Kotlin being an "easier" language than Java. Practically, this shows in Kotlin being less verbose, so fewer lines of code to make errors in. It also has static typing instead of strong typing so many types are inferred, and lastly, it has a Null type which avoids all the common NullPointer exceptions. 
+
+Kotlin seems like a natural successor language to java, because it brings some advantages with no real downsides. It has full interoperability allowing us to port files 1-1.
+
+There are also intersting features being added to the language that we can explore in the future if we decide to invest in kotlin competence now - such as [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform.html#full-stack-web-applications) for cross-platform compilation.
+
+## Proposed Change
+
+I propose that we port all our Android platform code, so we have a language less in our codebase. First the TestApp and then the SDK itself.
```
roblabs, line 16: Can we `tag` and document the tag for those developers that may be forced to use Java.  We want to make sure MapLibre adopters that may already have a Java codebase for Android can easily adopt before they fully move over to Kotlin.

For example, maybe the tag would be `works-with-java` or whatever.   Then maybe make a note of it in one of the DEVELOPING.md files.

Otherwise, I think the Kotlin proposal is a great idea as "*Kotlin turns `func` into `fun`*".

roblabs (approved):

design-proposals/2022-09-02-kotlin.md
birkskyum: Great idea! The bit of kotlin we have already is mainly in the tests, so we still have the opportunity to make this tag before changes are made to the SDK itself.

atierian:
I'm also in support of this, and would like to see a similar move on the Darwin side to Swift. I'm curious about how this would work technically though. AFAIK Kotlin doesn't have direct interoperability with C++ (please correct me if I'm wrong). That would mean that it would have to through JNI(?) or we'd have to make an ObjC or C wrapper around the C++ code for Kotlin to consume? Or does the Android SDK expose Kotlin APIs that would cover our needs?

There are similar limitations on the Darwin side as well. However Swift <-> C++ interop is actively being designed and implementing by a Swift language workgroup I'm on. Here's the [current status](https://github.com/apple/swift/blob/main/docs/CppInteroperability/CppInteroperabilityStatus.md). This would be a non-issue on the Darwin side if we went to Metal, but that's a larger conversation and I don't want to hijack your proposal here 😄.

birkskyum:
@atierian , it seems like there's a lot of interest in swift, so could you maybe make a PR with design proposal for us to discuss? 

Regarding the C++ interop, we're going to tidy up the C++ core so we can make bindings to a range of targets easier, likely with a tool like SWIG. I imagine that will also be the best approach for Kotlin.

atierian:
Yea, I'll write up a design proposal on it.
@nvanfleet happy to collaborate on it if you're interested.

nvanfleet:
> 



> Yea, I'll write up a design proposal on it.
> @nvanfleet happy to collaborate on it if you're interested.

Sure I'd be happy to discuss it!

PR #425, opened 2022-09-04
Refactor NSExpression Unit Tests

birkskyum:


platform/ios/platform/darwin/test/MGLHeatmapStyleLayerTests.mm
```diff
@@ -189,7 +196,7 @@ - (void)testProperties {
         XCTAssertEqualObjects(layer.heatmapRadius, functionExpression,
                               @"heatmapRadius should round-trip camera expressions.");
 
-        functionExpression = [NSExpression expressionWithFormat:@"mgl_interpolate:withCurveType:parameters:stops:(keyName, 'linear', nil, %@)", @{@18: constantExpression}];
```
birkskyum, line 192: @roblabs What is the correct way to handle this 'keyName'?

birkskyum:
There are just two types of expressions that aren't refactored yet:

functionExpression = [NSExpression expressionWithFormat:@"mgl_interpolate:withCurveType:parameters:stops:(keyName, 'linear', nil, %@)", @{@18: constantExpression}];

It contains the `keyName, and it's not clear from our documentation how to translate it

The other form is: 
constantExpression = [NSExpression expressionWithFormat:@"MGL_FUNCTION('image', 'Icon Image')"];

where MGL_FUNCTION isn't recognized.

Any advice on this is greatly appreciated.

roblabs:
@birkskyum —

I found the expression for `keyName` another one of the curious expressions.  I found that Swift code was easier to understand and see where the original authors were intending.  

The Unit Test `testMigratingToExpressions$Linear()` defined here has Swift source that is compiled and tested in the scheme `CI`.   This Unit Test uses Earthquake magnitude as a `keyName`.  But this was the only example that I found.

https://github.com/maplibre/maplibre-gl-native/blob/901ab18689d96a1804b429c7f505d32d94f417dd/platform/ios/platform/darwin/test/MGLDocumentationGuideTests.swift#L91

--- 

Then see the section in [DEVELOPING.md](https://github.com/maplibre/maplibre-gl-native/blob/7fc8e5948ba3b333f18578dafbc9aa70b3f66442/platform/ios/platform/ios/DEVELOPING.md#generating-documentation) titled "*Generating Documentation*" to run `make idocument` which auto generates the markdown file: 

https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/platform/ios/docs/guides/Migrating%20to%20Expressions.md#linear.

roblabs:
Another comment on how the Unit Tests for Expressions are structured.  I found a single Unit Test to have *many* tests within it.  Ideally, a unit test should have a small bit of code over a few lines.  You'll see that I used `XCTExpectFailure`, which  marks a test as an expected failure from that point in the code on.  If you remove an expected failure, you are likely unearthing a failure somewhere else down in the same Unit Test.  

My longer term recommendation for the Expressions unit tests is:

1. Add more Swift Unit Test for Expressions
2. Refactor the Objective-C Unit Tests for Expressions to have one expression for test (whereas, now we have many per one single Unit Test, and it makes it harder to understand exactly what is failing).
3. Leave the expected failures in place until either 1. or 2. is achieved.

birkskyum:
@roblabs thanks! So a solution could be to port these test files to Swift, right?

roblabs:
From https://github.com/maplibre/maplibre-gl-native/pull/425#issuecomment-1237533943

> My longer term recommendation for the Expressions unit tests is:
> 
> 1. Add more Swift Unit Test for Expressions
> 2. Refactor the Objective-C Unit Tests for Expressions to have one expression for test (whereas, now we have many per one single Unit Test, and it makes it harder to understand exactly what is failing).
> 3. Leave the expected failures in place until either 1. or 2. is achieved.

---

> @birkskyum  *thanks! So is it a solution here to port the test files to Swift?*

Issue #268, opened 2022-04-01
iOS Crash: Simulator Crash (glDrawElements) on custom style when camera is moved


racer1988:
Hello.

I made a small demo app to verify the crash I have in my production app.
I tested this app with the default style and it works successfully.

We have a custom style which is way heavier than the default style, due the feature of the app.
As soon as we move the viewPort with something like:

```
mapView.setCenter(CLLocationCoordinate2D(latitude: 52.414812, longitude: 13.046378), zoomLevel: 16, animated: true)
```

MapLibre crashes (note it crashes even non-animated)

`Thread 1: EXC_BAD_ACCESS (code=1, address=0x7faa2058fd90)`

I would like to know if you could help us out.
We are not able to make the style simpler, but we would like to understand if there is something we can change to avoid the OpenGL Renderer to explode on camera movements.

Some Notes:

- Manual user interactions (Pans, Zoom) are not crashing the map but are extremely laggy and slow
- Programmatical changes are instantly crashing the renderer

Let me know if you want to have the Demo App (it's just a full screen map tbh)

# StackTrace

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x00007fff4020f189 in glvmRasterOpDepthStencilTest ()
#1	0x00007fff4025d195 in glvmInterpretFPTransformFour ()
#2	0x00007fff5bca3f8e in gldLLVMFPTransform ()
#3	0x00007fff5bcb80c4 in gldLLVMVecPolyRender ()
#4	0x00007fff5bca0540 in gldRenderFillPolygonPtr ()
#5	0x000000011d4d616e in 0x11d4d616e ()
#6	0x00007fff400fd6f7 in gleFlushAtomicFunc ()
#7	0x00007fff4014a10d in glDrawElements_IMM_ES2Exec ()
#8	0x0000000110f811fa in mbgl::gl::Context::draw(mbgl::gfx::DrawMode const&, unsigned long, unsigned long) ()
[...]

PR #426, opened 2022-09-06
Use self-hosted AWS runners

birkskyum:
Run macos-ci and ios-ci on dedicated M1 AWS instance, to cut down on CI times.

macos-ci: 20 min -> 2 min.
ios-ci: 108 min -> 22 min.


birkskyum:
@roblabs, @nvanfleet , if anyone of you are familiar with the arch compiler options in xcode, we have a possibility here to dramatically cut ci times for ios/macos builds. There is a arch (x86_64 and arm64) mismatch in our ios-ci that I'm not sure how to resolve. I believe this CI is expecting an arm64 build, where our previous one expected x86_64 builds, but our ios->test target, apparently generates x86_64 builds.

birkskyum:
In the ios-ci runner. https://github.com/maplibre/maplibre-gl-native/runs/8228993960?check_suite_focus=true

Screen Shot 2022-09-07 at 16 06 33

birkskyum:
It's this  part: `(mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))` that seems to make trouble. I've found that if I tick "Enable Bitcode" in the Build Settings of ios->test, it seems like that these libmbgl core dependencies used by ios->test  (on the picture) maybe aren't arm64 compiled at this point.
image

nvanfleet:
> @nvanfleet do you think the bazel build scripts can solve this?

Oh this is really interesting. I haven't covered the test targets for the bazel scripts so far, but it seems feasible that we could select arch for running those in the bazel script. I can look into this.

birkskyum:
@nvanfleet thanks! It would be really helpful.

birkskyum:
It's finally working as expected.

wipfli:
Thanks for fixing the CI on main and getting the self-hosted M1 runners up, Birk. Looks good to me. I guess you merged because main was broken?

wipfli:
Should we open an issue to remember the two disabled tests `"MGLOfflineStorageTests", "MGLMapViewContentInsetTests"`?

birkskyum:
@wipfli , good idea - I made an issue ( #432 ). I actually patched and enabled the MGLOfflineStorageTests just yesterday (#428), but apparently, these CI checks find that there is still something missing which is why I turned it off again.

roblabs:
I would like to request that we re-enable the test from the `iosapp` scheme called `testDebugBoundaryTiles()`

This will help us diagnose #268 (which currently fails on iOS 15 simulators; works on iOS 16 simulators)

* Turns on the red tile boundary
* adds a green polyline at (0,0)
* Measure Frames per second (FPS decoration in the upper left)
* Takes a screen shot and saves it off.

image

PR #411, opened 2022-08-18
Fix for "NSExpression function is forbidden"

roblabs:
* For iOS 15.5, refactor how MGL Expressions are defined. Fix for "NSExpression function is forbidden" #331
* Update Swift & Objective-C sample Code for MGL Expressions.  Use these patterns for fixing unit tests.
 Convert CI Scheme to TestPlan. Add CI-Expressions TestPlan (which skips tests unrelated to `NSExpression`)
* Update Swift sample code & documentation for MGL expressions.  Update API docs (.h) & online docs (HTML) with:  `make darwin-style-code && make idocument`
* Refresh developer documentation for:  build dependencies, running unit tests from the command line, and logging in C++.
* Mark certain unit tests as "Expected Failures":  Awaiting unit test refactoring for https://github.com/maplibre/maplibre-gl-native/issues/331
- [ ] TODO:  The unit tests that are "Expected Failures" will need to be refactored.

Issue #421, opened 2022-09-03
Refactor NSExpression unit tests


birkskyum:
Remove all XCTExpectFailure from macos/ios/darwin test suites introduced in #411  by refactoring the tests to the new format.

birkskyum:
Working in here: https://github.com/maplibre/maplibre-gl-native/tree/refactor-nsexpression-unit-tests

birkskyum:
@roblabs , do you have any idea how to write

```objc
[NSExpression expressionWithFormat:@"MGL_FUNCTION('image', 'Icon Image')"];
```

With the new syntax?

roblabs:
@birkskyum — the expression `MGL_FUNCTION` is curious and still needs attention.  

While I was researching the source for #331 & PR #411, I found documentation for the expression in ["Predicates and Expressions.md"](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/platform/darwin/docs/guides/Predicates%20and%20Expressions.md#mgl_function):

https://maplibre.org/maplibre-gl-native/ios/api/predicates-and-expressions.html#code-mgl_function-code

---

But the Objective-C source mentions that it is not implemented.

https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/platform/darwin/src/NSExpression+MGLAdditions.mm#L256,L259

```objective-C
- (id)MGL_FUNCTION:(id)firstArgument, ... {
    [NSException raise:NSInvalidArgumentException
                format:@"Mapbox GL function expressions lack underlying Objective-C implementations."];
    return nil;
}
```

---

In general, I found that writing the expressions in Swift first was easier to sort out than in Objective-C.

birkskyum:
Because of that, I don't plan to look further into refactoring the obj-c code as it seemed futile.

louwers:
OK thanks for the update.

1ec5:
> But the Objective-C source mentions that it is not implemented.

This just tells you that it’s an abstract class.

`MGL_FUNCTION` is a generic aftermarket expression that developers can use for any style-spec function that isn’t specifically implemented yet as an aftermarket expression.

/ref mapbox/mapbox-gl-native#11472

1ec5:
> The swift syntax is more flexible and we have cases that aren't supported by obj-c anymore.

[...]

PR #427, opened 2022-09-07
Fix macos-ci using darwin-style-code instead of nitpick

birkskyum:
As proposed in here #411 we can run `make darwin-style-code` instead of the `scripts/nitpick/generated-code.js darwin` to check darwin style code.

```diff 
-      - name: Nitpick Darwin code generation
-        run: scripts/nitpick/generated-code.js darwin
+      - name: make darwin-style-code
+        run: make darwin-style-code
```

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -139,5 +139,5 @@ jobs:
       - name: Lint plist files
         run: make macos-lint
 
-      - name: Nitpick Darwin code generation
-        run: scripts/nitpick/generated-code.js darwin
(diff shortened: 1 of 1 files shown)


wipfli:
Should we remove `scripts/nitpick/generated-code.js darwin` now?

birkskyum:
I'd prefer to keep it around for local testing until this #425 #421 refactoring is sorted out.

roblabs (commented):
Thank you for patching this @birkskyum — looks like changes from #411 are now unblocked.

PR #428, opened 2022-09-07
Fix MGLOfflineStorageTests

birkskyum:
This invalidates MGLOfflinePack properly

Screen Shot 2022-09-03 at 19 28 57

### DIFF
--- a/platform/ios/platform/ios/test/CI.xctestplan
+++ b/platform/ios/platform/ios/test/CI.xctestplan
@@ -31,8 +31,7 @@
   "testTargets" : [
     {
       "skippedTests" : [
-        "MGLMapSnapshotterTests\/testLocalGlyphRendering",
-        "MGLOfflineStorageTests\/testAddPackForBounds"
+        "MGLMapSnapshotterTests\/testLocalGlyphRendering"
       ],
       "target" : {
         "containerPath" : "container:ios.xcodeproj",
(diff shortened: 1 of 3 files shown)


roblabs (commented):
There is only a manual test for Offline Packs in the scheme `iosapp`.  Can you verify that it still works?

birkskyum:
It should be able to pass now, so we can turn it back on. Where is it disabled?

roblabs:
It's not disabled, but manually run via the scheme `iosapp`.

> @birkskyum, "*It should be able to pass now, so we can turn it back on to run automatically. Where is it disabled?*"

birkskyum:
From what I can tell it's running just fine now

Screen Shot 2022-09-07 at 16 25 11

platform/ios/platform/ios/test/CI.xctestplan
```diff
@@ -31,8 +31,7 @@
   "testTargets" : [
     {
       "skippedTests" : [
-        "MGLMapSnapshotterTests\/testLocalGlyphRendering",
-        "MGLOfflineStorageTests\/testAddPackForBounds"
+        "MGLMapSnapshotterTests\/testLocalGlyphRendering"
```
roblabs, line 34: Excellent.  Testing in the `CI` scheme works as well.  thanks for enabling the disabled tests.

roblabs (approved):

PR #433, opened 2022-09-08
Java compatibility tag

birkskyum:
Before porting the code to Kotlin, making a clear point in time where the Android SDK is still all Java could become useful later on. The tag is called `before-kotlin-port`, and it was part of the feedback on the Kotlin design proposal. #420


wipfli:
Looks good to me

PR #435, opened 2022-09-08
Add Dependabot for npm packages

birkskyum:
Now that our CI is faster, it's finally feasible to add dependababot for npm packages without blocking the runners for days.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,12 @@
 
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


wipfli (approved):
Looks good to me. I think there are also one or two more `package.json` files in this repo...

birkskyum:
I actually think it can pick up on the nested package-lock.json too because I resolved some of these dependabot-flagged security vulnerabilities that way earlier today.

PR #455, opened 2022-09-09
Change visibility of bazel packages

nvanfleet:
This updates some of the bazel visibility settings for packaging. Settings on Mapbox.static, Mapbox.dynamic and exports_files are all there when integrating with other bazel scripts (but doesn't affect the sh script).

The other changes are just aligning some more narrow visibilities in between packages which actually helps to hint on what depends on each one.


birkskyum (approved):
Thanks!

PR #458, opened 2022-09-10
Configure dependabot for nested package.json files

birkskyum:
Dependabot only covers the root package.json currently. This extend it to the platform directories:

/package.json
platform/ios/package.json
platform/android/package.json

It also set a limit for each of them to 3 PRs simultaneously because the previous limit of 20 wasn't efficient. This was because the packages would keep invalidating each other's CI results.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,11 +5,23 @@
 
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


ntadej (approved):

ntadej:
Looks good to me.

PR #459, opened 2022-09-10
Node release workflow #2 (alternative)

acalcutt:
This is a merge of work done in https://github.com/maplibre/maplibre-gl-native/pull/378 and work done by @mnutt in https://github.com/mnutt/maplibre-gl-native/tree/publish-github

It is still a work in progress but I think this is in a mostly working state.

Notes
- I think something needs to be done with the node-pre-gyp used in this. It is pointing to a repository bring that brings node 18 support. this was added in by @mnutt so I'm not too familiar with it, but one of the build seemed to fail without it.


acalcutt:
I guess the reason node-pre-gyp fails on this PR and not the one we just merged is because this workflow uses the "--target" switch, where the other one uses a node matrix.

When using the target switch, node-pre-gyp relies on https://github.com/mapbox/node-pre-gyp/blob/master/lib/util/abi_crosswalk.json to match version numbers to abi, and right now it has not been updated to support node 18, even though they have a PR for it https://github.com/mapbox/node-pre-gyp/pull/649

Looking more, it looks like there has been some discussion around this, https://github.com/mapbox/node-pre-gyp/issues/647 , but node-pre-gyp seems pretty much unmaintained https://github.com/mapbox/node-pre-gyp/issues/657

I found in another thread how to update abi_crosswalk.json https://github.com/mapbox/node-pre-gyp/blob/master/contributing.md , so for now I ran that an uploaded my own version to @acalcutt/node-pre-gyp

birkskyum:
@acalcutt I think it's a good idea to fork the node-pre-gyp, as Mapbox doesn't maintain most of these and PRs can be there for a very long time - if the repo is somewhat maplibre-only we can also put it under maplibre/*  on npm/github, but that can always happen down the line. Is it ready for review?

ntadej:
Is there any other package that we could use instead (longer term)? We should try to avoid the number of such packages that we maintain under the MapLibre umbrella.

birkskyum:
@ntadej , the other package would be Mapbox's but as I just mentioned, they don't have a good track record of merging PRs in my experience or working on these packages in general, so we might as well have our own as it won't fall behind. I'm all for cutting down on the number of deps wherever possible.

ntadej:
What I mean is, is really Mapbox the only company that set up such a release workflow of pre-building C++ code and packaging it as node package?

birkskyum:
@ntadej , likely not, but they might have been the first to make an open-source library for it that everyone else could use, which would explain the 1k stars on the repo.

birkskyum:
Great to see it passing throughout - what does this need before it's ready?

acalcutt:
Should be ready, I am just testing it now on my fork (minus the arm) to make sure the publish part is working here https://github.com/acalcutt/maplibre-gl-native/actions/runs/3115843886

acalcutt:
Should be ready now. I haven't gotten to test publishing on arm because i don't have it, but the ci tests look good on it here so it should work.

birkskyum (approved):
Thanks for this!

acalcutt:
you have me wondering how the next pre-release goes. I curious if the arm64 build works.

birkskyum:
Lets try itt

ntadej:
I wonder why the builds are separate? Is this the preferred node way? Or should we prepare universal builds?

birkskyum:
@ntadej , I attempted to go the universal build direction, because it then would be faster to run it all on m1 machines, but got a lot of issues with dependencies https://github.com/maplibre/maplibre-gl-native/pull/502

ntadej:
OK, as long as it works, I guess it's fine like that for now. I wonder if we should set up some release testing pipelines.

birkskyum:
It did publish as expected to github, but maybe not to npm. The ubuntu build failed the first time and I had to retry it, so I don't know if that influenced it.

acalcutt:
it looks like i forgot to remove
`&& matrix.node == '18'`
from the npm publish, so thats probably why that didn't work.

birkskyum:
Let's fix it and try again. Would you mind adding a entry in the changelog for the arm support?

PR #474, opened 2022-09-13
Bump to JDK 11 in android CI and generate Gradle Wrapper

carlonzo:
Bumps openjdk-8 to openjdk-11 on docker configuration for CI

Also generates gradle wrapper aligned to the already declared gradle wrapper version (this was probably missed in the past)


ntadej:
I do not like that we have binary files in the repo. Can we avoid it?

carlonzo:
I see what you mean. no, sadly thats the gradle wrapper required to init gradle and the build system. every android project requires it. I am not sure you could init the project without it

ntadej:
> I see what you mean. no, sadly thats the gradle wrapper required to init gradle and the build system. every android project requires it. I am not sure you could init the project without it

But aren't those files project-agnostic? Can't we make a script that a developer runs when cloning the repo but the files stay in gitignore?

carlonzo:
I'm sorry but I am not sure you can avoid that. you could have a script to install gradle on the machine, then generate the wrapper from there, but I've never seen that. if you want I can drop the upgrade of the wrapper (which is already included in the repo) but I cannot remove it. 

as stated here https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html :

> The scripts generated by this task are intended to be committed to your version control system. This task also generates a small gradle-wrapper.jar bootstrap JAR file and properties file which should also be committed to your VCS. The scripts delegates to this JAR.

ntadej:
Fine with me then 😊

carlonzo:
thanks! once this is approved and merged hopefully I will be able to unlock https://github.com/maplibre/maplibre-gl-native/pull/218

birkskyum (approved):
Thanks! I can confirm that it all runs as expected still with JDK 11.

carlonzo:
this PR did not actually change the jdk. the jdk 8 was still installed and used. I can see the logs from Gradle using the jdk8 path `/usr/lib/jvm/java-8-openjdk-amd64/bin/java`

birkskyum:
When I look at the Gradle setting in Android Studio it seems like jdk 11 is in use.
Screen Shot 2022-09-13 at 12 20 01

But is that unrelated to this then?

carlonzo:
no @birkskyum thats because gradle config does not specify a JDK-HOME and Android Studio picks the bundled one in AndroidStudio (JDK 11)

on CI I completely misunderstood how it works. docker files were not pulled by github action (maybe thats something old that can be removed?) but follows the workflow in https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/android-ci.yml

I will create another draft PR and change the jdk in github action, if thats ok with you?

birkskyum:
@carlonzo sounds good

carlonzo:
alright, what I changed was the script to generate a new image for docker. @birkskyum do you guys know if makes sense to publish a new image running the workflow https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/android-docker-ndk-r21b.yml ? otherwise we just change the java version on the workflow itself

ntadej:
We should tag the old version so we do not lose it (just in case). I can look into it later today.

birkskyum:
This docker release workflow hasn't been run for a very long time, but it would be great to have it active again.

carlonzo:
up to you, otherwise this should do the trick https://github.com/maplibre/maplibre-gl-native/pull/475

I leave it there to you to decide

wipfli:
https://github.com/maplibre/maplibre-gl-native/actions/workflows/android-docker-ndk-r21b.yml did not get a trigger it seems.

Does that mean that the current docker image at ghcr.io/maplibre/android-ndk-r21b still uses to previous java verision?

PR #475, opened 2022-09-13
Choose JDK11 in Android workflow

carlonzo:
This setup java 11 in github workflows for Android


birkskyum (approved):
It ran way to many CI tests because android-release.ci hasn't been added to the list of ignored paths from most workflows incl.: node/qt/ios/macos etc.

Let's have this to unblock the work on dependencies - It can readily be removed when the container workflow starts working again.

wipfli:
`make android-javadoc` seems to fail with java 11, see #603

PR #476, opened 2022-09-13
Fix android ci workflows

birkskyum:
- Fixes problem with indentation in android ci
- Avoid android-release.yml triggering too many workflows as seen in #475

### DIFF
--- a/.github/workflows/android-ci-pull.yml
+++ b/.github/workflows/android-ci-pull.yml
@@ -60,8 +60,8 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-        distribution: 'zulu'
-        java-version: '11'
+        distribution: "zulu"
+        java-version: "11"
 
       - name: Cache node modules
         uses: actions/cache@v3
--- a/.github/workflows/android-ci-pull.yml
+++ b/.github/workflows/android-ci-pull.yml
@@ -60,8 +60,8 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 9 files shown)


.github/workflows/android-ci-pull.yml
```diff
@@ -60,8 +60,8 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-        distribution: 'zulu'
-        java-version: '11'
+          distribution: 'zulu'
```
carlonzo, line 63: thanks! sorry by bad!

birkskyum: no worries!

birkskyum:
@carlonzo , can you tell why the ci-pull fails?

carlonzo:
hey @birkskyum the test is failing in https://github.com/maplibre/maplibre-gl-native/actions/runs/3045087247/jobs/4906601925#step:12:6129 

> java.lang.AssertionError: expected:<-85.05112877980659> but was:<-85.0511287798066>

which I "fixed" in https://github.com/maplibre/maplibre-gl-native/pull/218/files#diff-850adbf0a37f67dd3e1127d6779ac5f3fca73c21ac4ee4b090e1a8a0a15a9574L26 reducing the precision of the assertion. probably there is a different implementation of the calculation producing a different output. 

Either we reduce the precision of the assertion as I've done or copy-paste the new calculated value to the expected value asserted

carlonzo (approved):
finally! thanks!

PR #477, opened 2022-09-13
Prepare Qt release workflow

ntadej:
- bump Qt6 to recently released 6.3.2
- setup the release workflow
- update CHANGELOG
- make sure LICENSE.md is distributed with the binaries


ntadej:
We should ignore changelog from the CI 😊

birkskyum (approved):

PR #480, opened 2022-09-13
Do not parallelise Qt release deployment

ntadej:
Do not parallelise Qt release deployment as it may cause multiple draft releases. Force one execution at once. As these are fast jobs (just repacking) it should not matter.


ntadej:
Wow, this was fast! `ccache` really helps!

PR #486, opened 2022-09-18
Fix issues with Qt release tarballs

ntadej:
Make tarball before uploading artifacts.


birkskyum (approved):

wipfli:
Is the artifact attached to the workflow run or to a release?

ntadej:
Now the same is attached to every CI and also to the release.

PR #492, opened 2022-09-19
Revert "[core][tile mode] Make sure only a single tile is used"

acalcutt:
This reverts commit 9ceb619940901d06909f7d6d0fc4a4e64572d228. for https://github.com/maplibre/maplibre-gl-native/issues/479 .

If the tile returned is off center, this causes vector layers to be missing on the tiles that were not included


acalcutt:
I'm still unsure why the render test in my node workflow exit with code 1 after this change, if anyone has any idea. the render test seems to work as it did before with the same number of successes/error. on linux the exist code is 0, but when i tested in macos it did seem to get an exit code of 1 like it does in thi workflow.

acalcutt:
I did some testing on the mac and actually found it's this test that causes the exit code 1 to happen. If I exclude "render-tests/map-mode/tile-avoid-edges" the workflow completes.

It's strange though, the "actual" image generated by the "render-tests/map-mode/tile-avoid-edges" test now looks much more like the expected...even though the test fails

birkskyum:
I, for some reason, get a lot more errors when running the tests - and not just in this PR:

What commands did you run on the mac to build and run the tests? Those from node-ci, or those from DEVELOPING.md?

birkskyum:
I run

`cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON`

`cmake --build build -j $(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null)`

`./build/mbgl-render-test-runner --manifestPath metrics/macos-xcode11-release-style.json`

acalcutt:
I made a few updates to what I remember running. I used the node workflow to get the commands too. One thing I notice is I chose "Release" build type. I wonder if Debug would give more info.

I have been testing the main change in this, removing the single tile restriction and it does for sure fix this issue it tileserver. This is really the one thing holding back a release of that.

I am a bit annoyed the test I used to troubleshoot this had to be disabled though. I tried a few things to update the test, like updating the image and manifest, but the test still failed.

birkskyum:
I noticed the two images you posted look identical, but have much different file sizes. I wonder if there is a hint in that.

acalcutt:
I did notice the size difference. maybe something is ending prematurely and the image isn't quite complete? I tried to update the expected image so it matched but the test still fails.

The other thing is I didn't really get what the diff is showing. Almost like it is a comparing the image with the missing tiles.
diff

birkskyum:
The CI for ios is blocked until we merge this #494

acalcutt:
When I was looking through the commits for this, I noticed they had also updated the expected image and metrics in this commit. https://github.com/WifiDB/maplibre-gl-native/commit/8f8b7afb26dc3ccb587b9af847ab7356892b4846

So it does seem the outcome of a single tile being displayed in tile mode was expected. We don't have that updated image in the maplibre-gl-js, so it is comparing against the original.

In tileserver-gl where I am having this issue, I decided to look again at creating a separate pool of renderers in static mode.  I made this update which didn't require as many changes as I thought it would. https://github.com/maptiler/tileserver-gl/pull/608/ 

Now I am unsure if I could continue this PR or just fix it in tileserver. the original https://github.com/maplibre/maplibre-gl-native/commit/9ceb619940901d06909f7d6d0fc4a4e64572d228 says it was for "various symbol placement problems like, symbol duplications or cut-offs, memory/cpu overhead, so maybe I should just close this and rework the tileserver side.

Issue #485, opened 2022-09-18
Setup M1 build runners


birkskyum:
This is a tracking issue for the current setup of 3 x m1 build runners on AWS, which has the following configuration:

Hardware:
- M1 mac mini
- 256GB EBS storage

Software:
- xcode v14
- tmux
- homebrew
- nvm (with node 18)
- ninja
- xcpretty
- python3 (through homebrew)

Status on machines

- [X] Machine 1 (34-218-210-7)
- [X] Machine 2 (18-236-103-245)
- [X] Machine 3 (3.80.132.4)

ntadej:
Python seems to be needed for Qt install action.

birkskyum:
@ntadej , there are 2 runners running now, both with python

birkskyum:
Awaiting a resolution of this #494 to unblock the CI.

PR #494, opened 2022-09-21
Xcode 14 compatibility

birkskyum:
This unblocks our CI which runs xcode 14, introduced in #485 .

The following simulators with iOS 16 have to be enabled in Xcode 14, as they aren't by default like in Xcode 13.x:

```sh
-destination 'platform=iOS Simulator,OS=latest,name=iPhone 8' \
-destination 'platform=iOS Simulator,OS=latest,name=iPhone Xs Max' \
-destination 'platform=iOS Simulator,OS=latest,name=iPhone Xr' \
-destination 'platform=iOS Simulator,OS=latest,name=iPad Pro (11-inch)'
```

### DIFF
--- a/platform/ios/platform/darwin/test/MGLCoordinateFormatterTests.m
+++ b/platform/ios/platform/darwin/test/MGLCoordinateFormatterTests.m
@@ -24,7 +24,7 @@
     coordinate = CLLocationCoordinate2DMake(38.9131982, -77.0325453144239);
     XCTAssertEqualObjects([shortFormatter stringFromCoordinate:coordinate], @"38°54′48″N, 77°1′57″W");
     XCTAssertEqualObjects([mediumFormatter stringFromCoordinate:coordinate], @"38°54′48″ north, 77°1′57″ west");
-    XCTAssertEqualObjects([longFormatter stringFromCoordinate:coordinate], @"38 degrees, 54 minutes, and 48 seconds north by 77 degrees, 1 minute, and 57 seconds west");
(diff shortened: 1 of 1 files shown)


roblabs:
This may force iOS developers to Xcode 14, which is not a bad thing, but their workflows may not allow it.  

Over in the README.md for [maplibre/maplibre-gl-native-distribution, Line #3](https://github.com/maplibre/maplibre-gl-native-distribution/blob/main/README.md?plain=1#L3), I used an SVG shield to indicate the minimum Xcode & Swift version.

---

Example from that repo:

Requires Xcode 12 

[Requires Swift 5.3](https://swift.org/package-manager/)

birkskyum:
@roblabs , it is a tricky situation. I think Xcode 14 is a solid piece of work, but the change to the long formatting of coordinates surprised me a bit. If a team isn't on xcode 14 yet, don't you think it'll be a priority anyway to get there, or might someone want to stay on 13.5? I think the code can still run on earlier xcode, just not this single ios-test with the coordinate if I'm not mistaken.

birkskyum:
@roblabs , can we do anything to reduce issues for other developers?

birkskyum:
I might be able to add an iPhone 8 target with iOS 16 on the runners - everyone running the test has to do it locally as well.

wipfli (approved):
Looks good to me, thanks Birk. Since this is a change in a test file, I wonder if it is related to some breaking changes in a dependency?

birkskyum:
i believe it's iOS 16 that introduced this change

birkskyum:
this pr bumps the xcode and iOS version, so it's probably related to that, but exactly what cause it is a bit opaque to me as a proper changelog for the software isn't public.

wipfli:
Great, thanks for making CI good again!

PR #495, opened 2022-09-21
Increase minimum node version to 14

birkskyum:
We recently released @maplibre/maplibre-gl-native to npm, with node 10 support (thanks @acalcutt !).

PR here: https://github.com/maplibre/maplibre-gl-native/pull/378 

This can be used as a compatibility version for migration (v5.0.1), but as node 10 passed the end of life a while ago, this PR deprecates it again, raising the minimum supported node version to 14.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -78,7 +78,7 @@ jobs:
         os:
           - macos-12
           - ubuntu-20.04
-        node: [ 10, 14, 16, 18 ]
+        node: [ 14, 16, 18 ]
     name: OS ${{ matrix.os }} Node ${{ matrix.node }}
     runs-on: ${{ matrix.os }}
     continue-on-error: true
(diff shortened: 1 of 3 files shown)


jutaz:
Yeah, I think this should be a major version bump (that way v5 can still receive critical patches, if necessary, without requiring users to upgrade node). Otherwise - great idea, v10 should be mostly deprecated anyway by now.

birkskyum:
awaiting #499

jutaz (approved):

jutaz (approved):

birkskyum:
@acalcutt , would it take a deprecation of node-14 as well to get the arm runners going? and would it in that case be possible with setup to make a universal build compatible with non-arm?

jutaz:
FWIW I think I was able to setup an arm build going w/  node v14 - I'll try to publish a branch for that (or maybe it works already, gotta check). I do know that node v14 _does_ run on arm64, if you really try (as in - compile it on macos)

birkskyum:
@jutaz awesome! because we got much faster m1 build runners waiting idle, while these small github intel-mac runners take all the load.

PR #497, opened 2022-09-22
Add AWS, Meta, MIERUNE to sponsors in README

wipfli:
This pull request adds logos and links to the participants of the MapLibre Sponsorship Program, https://maplibre.org/sponsors

Related to https://github.com/maplibre/maplibre/issues/110

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,26 @@ Beside merging in platform specific SDKs, the following changes were made compar
 
 ## Sponsors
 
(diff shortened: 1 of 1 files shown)


birkskyum:
Thanks for fixing this. We should probably crop these images in their source files so the alignment isn't off in all the repos.

Screen Shot 2022-09-22 at 15 47 09
Screen Shot 2022-09-22 at 15 47 25

birkskyum (approved):

PR #501, opened 2022-09-23
Fixes potential NaN when calling `NativeMapView::nativeMoveBy`

schachi5000:
The code may crash at runtime due to NaN being created when calling `NativeMapView::nativeMoveBy`

I've just applied the same fix, which is already present in the most recent Mapbox codebase.

https://github.com/mapbox/mapbox-gl-native-android/blob/a82dcd9d8aa21dd9a285f319d33755d9333f917a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java#L269


birkskyum:
Thanks @schachi5000 ! As you're likely aware already, this has an open license that complies with ours, but the new [mapbox/mapbox-maps-android](https://github.com/mapbox/mapbox-maps-android) doesn't, so we can't upstream from that.

birkskyum (approved):

ntadej:
The fix is small enough that I do not consider it license violation.

ntadej (approved):

PR #502, opened 2022-09-23
WIP: Arm support for Node

birkskyum:


birkskyum:
@jutaz , this is the problem at hand.

jutaz:
Cool, I’ll take a look over the weekend, hopefully 👍

acalcutt:
I think arm was pretty much working in https://github.com/maplibre/maplibre-gl-native/pull/459

birkskyum:
My understanding of the problem is that it is possible to activate Rosetta 2 on the machine with `arch -x86_64 zsh`, then run `nvm install 14` to install node with Rosetta 2, and that's it. It just doesn't work with the setup-node action.

birkskyum:
@acalcutt , that would be great! What is the advantage of that method compared to the current one?

acalcutt:
One positive is the workflow is less 'heavy'. It only runs on one node version per os instead of 3 node versions. It use the fact that one build makes multiple node abi versions, so it won't need to run on node 14 arm can't build on.

Downside
- when using node-pre-gyp with the --target , it uses a [abi_crosswalk.json](https://github.com/mapbox/node-pre-gyp/blob/master/lib/util/abi_crosswalk.json) file that is out of date. I had to make my own version of the package to support node 18. We may be able to change back to the officially one if they add node 18 support.
- It is not tested for every node version, just the node version it builds on.

birkskyum:
Because these things run in parallel, the total node suite takes 7 minutes right now - how long is the test time of the other branch where only 1 node version is tested?

birkskyum:
Is it possible for an m1 runner to make a universal build of 16 and 18 that can be used on x86 macs as well?

birkskyum:
@acalcutt , why are the ubuntu tests so much faster here than in the other workflow? Caching?

acalcutt:
I tried to make a universal build like you suggested here https://stackoverflow.com/questions/65157483/macos-build-universal-binary-2-with-cmake, with  "CMAKE_OSX_ARCHITECTURES=arm64;x86_64" but it was always missing compenents from the architecture it wasn't running on.

acalcutt:
FYI, I think I sound that command needed quotes like

`-DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"`

PR #504, opened 2022-09-24
Qt query feature

hziya:
I added conversions from mbgl::GeoJsonFeature to QMapLibreQL::Feature. These are necessary if you want to use query features in qt.

### DIFF
--- a/platform/qt/src/utils/geojson.cpp
+++ b/platform/qt/src/utils/geojson.cpp
@@ -159,4 +159,223 @@ mbgl::GeoJSONFeature asFeature(const Feature &feature) {
     }
 };
 
+Coordinate toCoordinate(const mbgl::Point<double> &point)
+{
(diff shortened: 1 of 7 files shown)


platform/qt/src/utils/geojson.cpp
```diff
@@ -159,4 +159,223 @@ mbgl::GeoJSONFeature asFeature(const Feature &feature) {
     }
 };
 
+Coordinate toCoordinate(const mbgl::Point<double> &point)
```
ntadej, line 162: Is there a reason you did not make those public?

ntadej (changes requested):
Looks good overall to me. I only have one question.

Note that compilation needs to be fixed though.

louwers:
I think this PR should be made against https://github.com/maplibre/maplibre-native-qt now

PR #505, opened 2022-09-24
Remove node matix from workflow

acalcutt:
Removed "matrix.node" condintion that is no longer needed in this workflow.
Removed package/release from package.json since it is no longer needed by this workflow.
Added CHANGELOG for node binary release workflow.

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -195,7 +195,7 @@ jobs:
           ./platform/node/scripts/publish.sh --target_arch=arm64
 
       - name: Publish to NPM (release)
-        if: matrix.os == 'ubuntu-20.04' && matrix.node == '18' && github.ref == 'refs/heads/main' && github.event.inputs.version != 'prerelease'
+        if: matrix.os == 'ubuntu-20.04' && github.ref == 'refs/heads/main' && github.event.inputs.version != 'prerelease'
         run: |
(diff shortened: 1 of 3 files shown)


birkskyum (approved):

birkskyum:
It seems like the npm publish step is skipped for arm at least

Screen Shot 2022-09-24 at 20 05 33

acalcutt:
it is skipped in everything but Ubuntu. my understanding is that what the node-pre-gyp binaries are for. when the package is installed it pulls the binary for the platform

birkskyum:
when looking at the running workflow, ubuntu skips it too.

birkskyum:
Screen Shot 2022-09-24 at 20 32 04

acalcutt:
it can only publish once, so it does that on linux.but the binaries and node-pre-gyp allow it to be multiplatform.

birkskyum:
Could it be that this ` if: matrix.os == 'ubuntu-20.04'` always is false, because matrix.os isn't defined?

birkskyum:
@acalcutt , that is really clever with node-pre-gyp - thanks for explaining. I think even the linux build isn't published right now.

acalcutt:
that could make sense. I'm not at a computer right now, but i think it is like, runs-on.name or something like that now.

birkskyum:
Looking at the rest of the code, `matrix.runs-on` and `matrix.arch` seem to be the predence, so it might work.

PR #506, opened 2022-09-24
Enable npm publish for node releases

birkskyum:
Fix for #505 , where the matrix was changed from

      matrix:
        os:
          - macos-12
          - ubuntu-20.04
        node: [ 14, 16, 18 ]
        
to 

      matrix:
        include:
          - runs-on: ubuntu-20.04
            arch: x86_64
          - runs-on: macos-12
            arch: x86_64
          - runs-on: macos-12-arm
            arch: arm64

so the os should be referenced as matrix.runs-on instead of matrix.os

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -195,7 +195,7 @@ jobs:
           ./platform/node/scripts/publish.sh --target_arch=arm64
 
       - name: Publish to NPM (release)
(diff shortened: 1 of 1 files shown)


acalcutt:
I tested this on my branch and the npm publish worked this way. Sorry I missed that, I was so focused on the github publish piece I didn't notice npm publish broke.

You may want to delete the two pre-releases that don't have a node release since I don't think they can be used without the npm publish.

ntadej (approved):

Issue #503, opened 2022-09-23
Refactor CI workflows from paths-ignore -> paths


birkskyum:
All our ci workflows have a list of paths-ignore, where files that shouldn't trigger the workflow are added, and everything else will trigger it. It's problematic because every time a new file is added, ALL workflows are triggered and have to have this file added to the paths-ignore.

I propose we instead of paths-ignore, use paths and add the relevant files so we don't have to ignore everything. It does even support ignoring specific files if we need that.

Affected flies that should be updated:

- [x] android-ci-pull.yml
- [x] android-ci.yml
- [x] ios-ci.yml
- [x] macos-ci.yml
- [x] node-ci.yml
- [x] qt-ci.yml

PR #513, opened 2022-09-27
Refactor qt workflow triggers

birkskyum:
Part of #503, for qt-ci.


birkskyum:
@ntadej , can you see anything missing, or something that doesn't need to be added?

birkskyum:
it seems like the qt workflows are the only that doesn't run `npm install` or `npm ci`, and thus doesn't need to trigger on changes to package.json or package-lock.json.

.github/workflows/qt-ci.yml
```diff
@@ -7,63 +7,33 @@ on:
...
-      - ".github/workflows/node-release.yml"
-      - ".github/workflows/qt-ci-windows.yml"
-      - ".github/workflows/update-gl-js.yml"
-  
+    paths:
+      - "bin/**"
+      - "expression-test/**"
+      - "include/**"
+      - "metrics/**"
+      - "platform/qt/**"
```
ntadej, line 15: `platform/default` is used by all (I think)

ntadej (changes requested):

ntadej:
You need to include the workflow itself 😊

ntadej:
Also the linux builder action...

.github/workflows/qt-ci.yml
```diff
@@ -7,63 +7,41 @@ on:
...
-      - ".github/workflows/update-gl-js.yml"
-  
+    paths:
+      - ".github/actions/qt5-build/**"
+      - ".github/actions/qt6-build/**"
+      - ".github/workflows/qt-ci.yml"
+      - "bin/**"
+      - "expression-test/**"
+      - "include/**"
+      - "metrics/**"
```
ntadej, line 17: I think `maplibre-gl-js` also needs to be included for shaders.

ntadej (changes requested):

.github/workflows/qt-ci.yml
birkskyum: the .gitmodules is added already - isn't that enough?

ntadej: Ah, true 😊

ntadej (approved):

PR #514, opened 2022-09-27
Node workflow - Add version to commit, simplify cmake commands

acalcutt:
The PR does two things discussed in slack.

1.) It adds the package version to the commit message when the version is bumped.

2.) It uses the same cmake command for both macos and linux. It removes DMBGL_WITH_COVERAGE from macos build and adds DCMAKE_BUILD_TYPE to the linux build by using the following command for both
`cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ env.BUILDTYPE }}`

EDIT:
I also added one extra thing. I added the ability to chose a preminor or premajor release type, so we could to something lile use premajor to release v6.0.0-pre.0

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -144,14 +144,8 @@ jobs:
           ccache --show-stats --set-config cache_dir=~/.ccache
 
       - name: CMake
-        if: runner.os == 'MacOS'
         run: |
-          cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ env.BUILDTYPE }} -DMBGL_WITH_COVERAGE=ON
-
-      - name: CMake
-        if: runner.os == 'Linux'
-        run: |
-          cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
+          cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ env.BUILDTYPE }}
 
       - name: Build
         run: |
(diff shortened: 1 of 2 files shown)


birkskyum (approved):
great idea with the pre-x releases

birkskyum:
it seems like "cmake command for both macos and linux." isn't included in this PR

acalcutt:
i must have accidentally wiped it out. I'll fix. it when I'm home later. I was also just thinking a
I need to keep in mind preminor/premajor during the npm publish

birkskyum:
Great, then I'll leave this until. It's probably best if the change to the cmake command is applied both in node-ci and node-release, so they stay consistent.

birkskyum:
Can we change prerelease to prepatch, to make that explicit?

acalcutt:
Apparently I was wrong that the same cmake command works to macos and linux, so I reverted that and just removed DMBGL_WITH_COVERAGE from macos build and added DCMAKE_BUILD_TYPE to the linux build.

I've added prepatch and removed prerelease.

.github/workflows/node-release.yml
```diff
@@ -8,7 +8,10 @@ on:
         required: true
         type: choice
         options:
-          - prerelease 
+          - prerelease
```
birkskyum, line 11: the prerelease doesn't have to be here anymore

acalcutt: In testing I found you still need it after a prepatch, preminor, or premajor

Say you do premajor and get v6.0.0-pre0. 

If you do a prepatch from this release you get v6.0.1-pre0. 
If you do a prerelease from this releaes you get 6.0.0-pre1

birkskyum: okay, so that's how it's operator

acalcutt: ya, honestly I had no idea how the npm versioning worked until I tried it.

If you are at v6.0.0-pre0 and do another premajor, it brings you up to v7.0.0-pre0

acalcutt:
I think this should be ready, once that last MacOS ci completes

PR #517, opened 2022-10-01
Fix bitcode on iOS for nunicode

ntadej:
This dependency had missing `mbgl-compiler-options`.


ntadej:
Hi @birkskyum, this one should go in first.

birkskyum (approved):

PR #518, opened 2022-10-01
Update CI to Qt 6.4.0

ntadej:
Qt 6.4.0 has been release and we should probably just update. The aim is to always build only for the latest version of Qt (but we support older versions if clients want to build themselves).


birkskyum:
@ntadej , this is ready for review right?

birkskyum (approved):

PR #520, opened 2022-10-01
Node workflow - build linux arm64 in a container

acalcutt:
This is an addition to the node release workflow to add linux-arm64 binaries. linux arm64 binaries are needed when running inside docker on a arm64 system. If we could get a Linux arm64 runner here, we could do this simpler because we could just add it into the OS matrix and be done, however I found this way to do it on what we have right now.

To do this build, I made the following

1.) A docker ubuntu focal 20.04.05 image bulder, made to be run inside the maplibre-native root directory, which pulls in all the local repo files, and sets an entrypoint script 'platform\node\scripts\docker_build_entrypoint.sh'

2.) In the 'docker_build_entrypoint.sh' I added all the linux build commands, so when the docker runs it builds all the lib files. when it completes it copies the lib files back to the host.

3.) I made a workflow which uses QEMU to emulate arm64. It first builds the docker image. It then runs the docker image, which generates binaries in an emultated linux\arm64 ubuntu focal, then it copies them back to the local lib fold on the host. It then uses node-pre-gyp to publish the arm64 libs like the normal process does.

This is running on ubuntu-latest right now, so it is a bit slow. It took about 3hr and 20min for a build. I initially was trying to run this on our macos-arm, but I could not get it to run the docker after many attempts (see the commit history)


acalcutt:
@birkskyum I'm wondering what you think about this PR to build a linux arm64 in an emulated container in a new workflow job. It's a bit slow but it works, I've used it to generate a linux arm64 binary in my fork and me and @mnutt have tested it,

Alternately, do you have any idea if we could get a linux arm runner. either as a new aws machine, or possibly a docker on the mac arm we have.

I'd love to get this arm64 linux binary posted officially instead of from my own repo. I've had a few users ask about tileserver-gl arm64 docker support, which needs the linux-arm64 binary.

birkskyum:
I think linux arm64 is a valuable target for us to have, and this approach is a quite interesting way to get that. If you'd like to run a docker container on the m1, I imagine running `brew install docker docker-compose` in the workflow could make it possible. If it requires some permission, I can also install it on the machines so it's part of the list here #485 .  Of course, a cleanup of spawned containers is important to avoid a memory leak.

birkskyum:
My main concern at this point is the 183 minutes that this takes.

acalcutt:
I'm going to do another test on the macos-arm and see if i can get it working with docker-composer added. I did try installing docker with brew already, but maybe not having docker-composer made it fail.

The time right now is mainly the QEMU emulation of arm64....just makes it slow. If he had a real linux arm runner we could just add it as another build in the regular workflow.

birkskyum:
Thanks for trying it out. It's likely required to use sudo to start the docker daemon. I can try to set up the docker service later today and rerun the tests.

shermp:
I was going to recommend cross-compiling using debian multi-arch, but when I tried it, I ran into a screeching halt. The Nodesource binary distribution depends on an architecture specific version of python-minimal, and you can only have one python install at a time. Installing an arm64 version of python basically breaks dpkg :(

It might be possible to compile nodejs and node-gyp separately (and tell node-gyp to use the host python).

acalcutt:
The previous attempt at cross compilation i saw also had issue with glfw, but I never tried that myself

acalcutt:
nodesource does have arm64 packages too, but only when installed on a arm64 platform

shermp:
> nodesource does have arm64 packages too, but only when installed on a arm64 platform. I am using that for my container based build.

Yeah, it was the arm64 packages causing me issues. Because it wants the arm64 version of python. Which then breaks dpkg because it removes the amd64 version of python.

acalcutt:
@birkskyum did you ever get to look at installing that docker service?


On a side note, I was looking a bit at what it would take to make this package build on it's own without needing binaries. I was looking at [cmake-js](https://github.com/cmake-js/cmake-js) but in my test here https://github.com/wifidb/maplibre-gl-native/tree/cmake-js it was having issues with vendor cmake files (mapbox I think). but maybe with some more testing something like this could work so you could build this on any system that had the dependencies installed.

birkskyum:
@acalcutt , unfortunately, I haven't been able to do this yet.

acalcutt:
I still would love to get this linux-arm binary created. On the tileserver-gl side I have already updated the docker images to be compatible with arm, I am just waiting for official binaries so I can make a PR for my release workflow there.

birkskyum (approved):
Great job!

acalcutt:
@birkskyum when do you think we could merge this and try another node pre-release? I think I have most of the groundwork for a arm64 tileserver-gl already, so I would love to see this binary made.

wipfli:
I assume @birkskyum just wanted to let you merge this pull request, but I think that only works if you have write access...

wipfli (approved):
Looks good, thanks

acalcutt:
Nice! glad to see this merged. When someone has a change could the "node-release" workflow be run with another prerelease version.

And no I don't have write access here right now. wouldn't mind having it but I'm fine without.

wipfli:
Granted you write access, @acalcutt. Thanks for all the good work you did on node support!

PR #526, opened 2022-10-07
Design Proposal - Using Conan.io instead of Mason

thehoneymad:
In this pull request, I am contributing a proposal for:

1. Removing usage of Mason.
2. Using conan.io to manage precompiled binary dependencies.
3. Using conan.io to manage mapbox dependencies such as earcut.cpp and geojson-vt-cpp.

### DIFF
--- a/.github/workflows/qt-ci.yml
+++ b/.github/workflows/qt-ci.yml
@@ -61,7 +61,7 @@ jobs:
           - name: Linux
             os: ubuntu-22.04
             build_type: RelWithDebInfo
-            qt_version: 6.3.2
+            qt_version: 6.4.0
             qt_target: desktop
           - name: macOS
             os: macos-12
@@ -73,7 +73,7 @@ jobs:
           - name: macOS
             os: macos-12
             build_type: RelWithDebInfo
-            qt_version: 6.3.2
+            qt_version: 6.4.0
             qt_target: desktop
             deployment_target: 10.14
             deployment_arch: "x86_64;arm64"
@@ -91,7 +91,7 @@ jobs:
             build_type: "RelWithDebInfo;Debug"
             compiler_type: x64
             compiler_version: 14.29
-            qt_version: 6.3.2
+            qt_version: 6.4.0
             qt_target: desktop
             qt_arch: win64_msvc2019_64
             qt_tools: ''
@@ -109,7 +109,7 @@ jobs:
             build_type: RelWithDebInfo
             compiler_type: mingw1120_64
             compiler_version: 11.2.0
-            qt_version: 6.3.2
+            qt_version: 6.4.0
             qt_target: desktop
             qt_arch: win64_mingw
             qt_tools: tools_mingw90
@@ -245,7 +245,7 @@ jobs:
       max-parallel: 1
       matrix:
         name: [Linux, macOS, win64_msvc2019, win64_mingw]
-        qt_version: [5.15.2, 6.3.2]
+        qt_version: [5.15.2, 6.4.0]
     permissions:
       contents: write
 


design-proposals/2022-10-05-conan.md
```diff
@@ -0,0 +1,67 @@
+# Using Conan.io instead of Mason
+MapLibre GL Native stil uses [Mason](https://github.com/mapbox/mason), Mapbox's depcreated cross platform C/C++ package manager to resolve precompiled binaries for third party dependencies.
```
wipfli, line 2: Typo deprec..

wipfli:
Thanks for this nice Design Proposal, @thehoneymad. I think it is a good idea to remove Mason. Conan seems like a good option.

wipfli:
So there will be this conanfile with exact versions if I understand correctly. Is this a bit like an NPM package-lock.json file? And can dependabot make pull requests with updates for us?

wipfli:
Also how is Mason or now conan related to the dependencies we have in the vendor folder with the submodules?

ntadej:
What precompiled dependencies do we even have? Aren't all covered by submodules?

thehoneymad:
Acknowledging comments. Will reply soon.

thehoneymad:
> So there will be this conanfile with exact versions if I understand correctly. Is this a bit like an NPM package-lock.json file? And can dependabot make pull requests with updates for us?

Yes. Conanfile asks for exact version. It does not support version ranges.

On the dependabot question, [sadly both `conan` and `vcpkg` is missing dependabot](https://github.com/dependabot/dependabot-core/issues/2027). But it is a great question. How about if this proposal gets approved, we can add a task to implement it?

wipfli:
Feel free to add the design proposal to the agenda for tomorrows mewting

thehoneymad:
> Also how is Mason or now conan related to the dependencies we have in the vendor folder with the submodules?

This is a great question. We currently use github sudmodules to pull nearly all dependencies. None of them are precompiled binaries. I will double check again.

But we are depedent on Mason who wires key dependencies for us such as Boost, RapidJSON and Mapbox dependencies. 

My high level goal here is to limit the build process as much as possible to only building MapLibre Native and guarantee a confident build on nearly every commit. 

Do you peeps see value in a document that explains how and from where our depdencies are wired?

ntadej:
> > Also how is Mason or now conan related to the dependencies we have in the vendor folder with the submodules?
> 
> This is a great question. We currently use github sudmodules to pull nearly all dependencies. None of them are precompiled binaries. I will double check again.
> 
> But we are depedent on Mason who wires key dependencies for us such as Boost, RapidJSON and Mapbox dependencies.
> 
> My high level goal here is to limit the build process as much as possible to only building MapLibre Native and guarantee a confident build on nearly every commit.
> 
> Do you peeps see value in a document that explains how and from where our depdencies are wired?

I suspect Mason was used to bump the dependencies (or do similar automation), but we never actually call it.

I would propose this actually:
 - remove Mason support
 - evaluate used dependencies per platform
 - then consider which package manager to use (if any)

I'm afraid that it will be hard to support desktop and mobile platforms with compatible pre-built binaries for all platforms.

We can discuss more tomorrow.

thehoneymad:
> What precompiled dependencies do we even have? Aren't all covered by submodules?

Great question.

Yes, all of them are currently covered by the git submodules. But the build process is a bit of a mudball. Unless one deeply knows how Mason wires the dependencies in order we are in risk of:

1. Bad builds due to key dependency being added in wrong order. [Boost, RapidJSON]
2. Bad builds due to Mapbox removing packages from Github. [Mason, geojson-vt, earcut.cpp, protozero, polylabel etc]

Mason can provide static includes and binaries both. :)

thehoneymad:
A great usage of Conan can be seen in [open source routing engine Valhalla](https://github.com/valhalla/valhalla/blob/master/conanfile.txt)

design-proposals/2022-10-05-conan.md
thehoneymad: Ooops, will fix. :)

thehoneymad:
This is not dead yet. I will come back to this promptly after my MapLibre whitepaper work.

PR #535, opened 2022-10-17
Move node build to cmake-js

acalcutt:
This PR allows the node package to build on its own with [cmake-js](https://github.com/cmake-js/cmake-js). This means if a user is on a platform we don't have binaries for, it will build its own binaries (provided they have cmake and other dependencies installed first)

I have also added a 'npm run build' command, which you can run in the package to build the node lib files and os. instead of needing to know all the cmake commands, cmake-js does the build and uses the defaults in CMakeLists.txt. I also used this in the workflow to cut down on the commands needed to run.

Originally I planned to remove node-pre-gyp/node-pre-gyp-github with this, but after testing I think the binaries are good to have since the build of maplibre is kind of slow. I've reworked it, somewhat based on 

EDIT: Updated 10/24/2022 to reflect current state.


acalcutt:
I Finally have this in a working state. 

Moving this package to cmake-js, as a stand alone package, was actually pretty easy. we already have good CMakeLists.txt files that cmake-js was just able to use and compile with. It seems to work well to  checkout the repo, install the build dependencies, and just run 'npm install'. Just doing that builds the binaries for your OS and Node.

It got a bit more complicated when I tried to make the npm package to upload to.
1.) For this to self compile, the npm package has to include the source code, which for this package is a LOT of files. The original package only included the package.json files and licenses. so this change increase the package size quite a bit. The size was actually to big to upload to npmjs, so I had to exclude several big directories in .npmignore.  It could probably be cut down more, but the ones I excluded were the biggest.

2.) This needs the build pre-requests installed on the machine building the package. for example, like pre-reqs at https://github.com/maplibre/maplibre-gl-native/tree/main/platform/linux or the brew commands in macoshttps://github.com/maplibre/maplibre-gl-native/blob/acc00b15813fdca28c4e52e846223040c76ad4b6/.github/workflows/node-ci.yml#L116-L121

3.) 'platform/node/CMakeLists.txt' includes '/node_modules/@mapbox/cmake-node-module/module.cmake', which works fine when you check the package out and build it locally. However if you try to include the package as a submodule in another package, the path to node_modules is no longer inside the maplibre-native directory, it is in the root project's directory. This causes node libs to not get built. To workaround this, I made CMakeLists.txt also look for module.cmake in .'./../../node_modules/@mapbox/cmake-node-module/module.cmake', which is where it end up when the packge is a submodule. (maybe someone know how this could be done better)


I have a working node package here for testing
@acalcutt/maplibre-gl-native-test@5.0.2-pre.2 
```
npm notice === Tarball Details === 
npm notice name:          @acalcutt/maplibre-gl-native-test                
npm notice version:       5.0.2-pre.2                                      
npm notice filename:      @acalcutt/maplibre-gl-native-test-5.0.2-pre.2.tgz
npm notice package size:  82.9 MB                                          
npm notice unpacked size: 204.3 MB                                         
npm notice shasum:        436a81d8c0c8f188888be3476627ff2e4516f335         
npm notice integrity:     sha512-7pWbI19VxV61J[...]WGx5rTm754aKQ==         
npm notice total files:   20150                                            
npm notice 
npm notice Publishing to https://registry.npmjs.org/
+ @acalcutt/maplibre-gl-native-test@5.0.2-pre.2
```

I have tested this in a tileserver-gl branch, which uses @acalcutt/maplibre-gl-native-test@5.0.2-pre.2 that looks like it works
https://github.com/acalcutt/tileserver-gl/tree/cmake

wipfli (commented):
First impression is that it looks much better...

acalcutt:
After some testing and comments from Brendan Ward I decided to add back binary support, but keeping cmake-js for building and fallback if binaries do not exist. This will allow the package to install quickly on platforms with binaries, but still be able to build on it's own on platforms that do not.

So the main thing this PR now does is 
1.) Includes files needed to build in the npm package.
2.) adds cmake-js compile as a fallback is node-pre-gyp does not find a binary
3.) adds a 'build' script that will compile the lib files without trying to pull in the binary.
4.) Simplifies the workflow by using 'npm run build'

.github/workflows/node-ci.yml
```diff
@@ -140,43 +140,8 @@ jobs:
...
-
-      - name: CMake
-        if: runner.os == 'Linux'
-        run: |
-          cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=${{ env.BUILDTYPE }} -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
-
-      - name: Build
-        run: |
-          cmake --build build -j $(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null)
+      - name: npm build
```
ntadej, line 143: Is `ccache` supported by `npm build` (did it actually work now)?

ntadej (changes requested):
I have one minor comment (also you'll probably have to rebase).

.github/workflows/node-ci.yml
acalcutt: I'd say right now this is not using ccache, the cmake-js option it is using looks like this

```
[
  '/root/.nvm/versions/node/v16.15.0/bin/node',
  '/opt/test7/cmake/node_modules/.bin/cmake-js',
  'compile'
]
info TOOL Using Ninja generator, because ninja is available.
info CMD CONFIGURE
info RUN [
info RUN   'cmake',
info RUN   '/opt/test7/cmake',
info RUN   '--no-warn-unused-cli',
info RUN   '-G',
info RUN   'Ninja',
info RUN   '-DCMAKE_JS_VERSION=undefined',
info RUN   '-DCMAKE_BUILD_TYPE=Release',
info RUN   '-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=/opt/test7/cmake/build/Release',
info RUN   '-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded$<$<CONFIG:Debug>:Debug>',
info RUN   '-DCMAKE_JS_INC=/root/.cmake-js/node-x64/v16.15.0/include/node',
info RUN   '-DCMAKE_JS_SRC=',
info RUN   '-DNODE_RUNTIME=node',
info RUN   '-DNODE_RUNTIMEVERSION=16.15.0',
info RUN   '-DNODE_ARCH=x64',
info RUN   '-DCMAKE_CXX_FLAGS=-DBUILDING_NODE_EXTENSION'
info RUN ]
```

I think I could add the ccache commands back to the workflow and set the option to use it in cmake with 
`npm config set cmake_CMAKE_CXX_COMPILER_LAUNCHER ccache`

which adds -DCMAKE_CXX_COMPILER_LAUNCHER=ccache to the options

acalcutt: well i made the changes to include that build option, and can see it is there now, but beyond that I'm not sure how I tell it is being used.

not sure why the macos ci is failing...i don't think it is related to this PR.

ovivoda:
@acalcutt Can you please try and resolve the conflicts and then we can try again to run the CI? Thanks!

.github/workflows/node-ci.yml
ovivoda: @ntadej @acalcutt CI looks like it is passing now. Should we mark this as resolved?

acalcutt:
I'd wait for now. I'd wouldn't mind some more input on it and I was playing a bit with how to add some build options.

I'd put this back to a draft but I don't seem to have that option.

ovivoda:
@acalcutt I drafted it for you. Feel free to ask again for review once you think it is ready for it.

louwers:
Closing because stale, please open a new PR if you want to push this further.

PR #538, opened 2022-10-19
[qt] Add a test for library usage

ntadej:
As discussed at the last technical steering meeting, I'm adding a simple test to run on the tarball contents to test if it configures and compiles.


ntadej:
This one is ready. Node failures are unrelated.

wipfli (approved):
Thanks Tadej

wipfli:
Can you make a follow up pull request with a bug or so such that we can check that it actually detects build problems?

ntadej:
Actually this already fixes a bug or improves the deployment as tarball created on CentOS did not directly work on Ubuntu (due to different library folder). But yes, I can try to break it again. Will you click the merge button?

PR #541, opened 2022-10-21
[qt] Support GeoJSON feature collection as std::list

dpaulat:
The current conversion routine supports GeoJSON feature collections using QList/QVector and QByteArray (JSON string). This adds support for feature collections using std::list.

### DIFF
--- a/platform/qt/CHANGELOG.md
+++ b/platform/qt/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Upcoming
 
+### ✨ New features
+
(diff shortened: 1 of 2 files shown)


ntadej (changes requested):
Thanks! Looks good to me.

Do you mind adding a changelog entry in the `platforms/qt` folder as this seems to be a relevant change for that?

dpaulat:
> Thanks! Looks good to me.
> 
> Do you mind adding a changelog entry in the `platforms/qt` folder as this seems to be a relevant change for that?

Done, thanks for the review!

ntadej (approved):

PR #543, opened 2022-10-21
add missing header guards

tnixeu:
I noticed that in those two files the header guard is missing.


ntadej:
Nice spot! I rebased the PR as the CI should be fixed now.

PR #546, opened 2022-10-27
Remove mapbox tile server support

nvanfleet:
This removes some internal definitions and special cases for Mapbox styles from the code. Mapbox is already basically not supported but this still remains in the codebase.


lseelenbinder (approved):
LGTM. Thanks, @nvanfleet.

(Once we get tests passing, of course.)

ntadej (changes requested):
Let me veto this for a bit while I'm away (until 2nd).

There are still users who use MapLibre Native but Mapbox maps for technical reasons (e.g. Mapbox does not support their platform anymore). There was a discussion here https://github.com/maplibre/maplibre-gl-native/issues/349 that we should not remove support while not providing a mechanism to restore it easily.

louwers:
@nvanfleet Could you fix the tests for the MR?

@ntadej How do you feel about this now?

ntadej:
I still think that we should keep `mapbox` support as long as it works. Maybe we could make it as an optional plugin?

nvanfleet:
IMO I feel like all these special cases should be external to the repo and somehow configured. IMO it's better to have a lot most simple maps codebase. But if there are users who use this I get it, I just didn't think anyone using Maplibre would be using Mapbox since I presumed they'd be using Mapbox 6.x or something since they are presumably paying

wipfli:
I think the next time we release a major version all vendor-specific code should be removed. Such code should not be maintained in the MapLibre Organization. Simultaneously, we should make it easy for vendors to plug in their custom logic for billing.

louwers:
Completely agree @wipfli

lseelenbinder:
Fully aligned @wipfli!

ntadej:
I agree. The main motivation to
keep basic Mapbox support are opensource users on platforms that Mapbox does not support anymore.

PR #561, opened 2022-11-18
Fix compilation issues using newer versions of Xcode

ntadej:


roblabs:
@tadej — Was this a change only for macOS or was it for iOS, as well?  Where was the compilation error?

birkskyum (approved):

birkskyum (approved):

wipfli:
Do we have to update some test fixtures after this change? CI is failing with something which looks like a problem with the image comparison. See #573

wipfli:
Just to understand how the environment is set up, why does a change in how Apple organizes headers make our CI fail, @ntadej?

ntadej:
It is more of a bug in our setup.

In C++ by good practice all headers should be defined by the client. Unfortunately the way how C++ headers work, you always get all headers recursively. This makes it easy to forget headers in the implementation.

In this case what probably happened is Apple stopped using those classes in their code so we did not implicitly get them and this broke our code.

In general if we setup things correctly we should not be sensitive to SDK changes (unless breaking changes), but maybe we
- do not have all headers present
- do not set minimum supported version properly

ntadej:
We always use the latest stable version and that is also the usual policy of Apple.

In the past I never had issues like this with my projects, but they are a bit special. We could also pin a specific version and update it manually each time. I guess this could be done by the maintainer once we get him.

wipfli:
I see, interesting. Yes, probably it would be a good idea to pin the XCode version and update it with something like dependabot...

PR #562, opened 2022-11-19
Create SECURITY_POLICY.txt

maxammann:
This adds a link to the security policy.

Depends on: https://github.com/maplibre/maplibre/pull/141

### DIFF
--- a/SECURITY_POLICY.txt
+++ b/SECURITY_POLICY.txt
@@ -1,0 +1,2 @@
+For an up-to-date policy refer to
+https://github.com/maplibre/maplibre/blob/main/SECURITY_POLICY.txt


maxammann:
@ovivoda @louwers Can you merge this one? The issue in https://github.com/maplibre/maplibre/pull/141 was approved and merged.

louwers:
Could you make it a Markdown file instead?

All other text files in the repo are Markdown.

maxammann:
Hm, the first PRs have already been merged in other repositories. I would say it's too late by now because these files are also referenced by this file: https://github.com/maplibre/maplibre/blob/main/.github/ISSUE_TEMPLATE/new-repo-checklist.md

So changing it now would mean to redo everything.

louwers:
The checklist is for new repos and the other repos can change it to a Markdown file at their own pace if they want. A link in a text file is just not very convenient because it is not clickable.

maxammann:
> We probably also need a workflow for handling security issues specifically for maplibre-gl-native

It is roughly outlined in the document in the maplibre repo. I think as a start it is fine. Though there are Github features which allow the submission of reports I think.

As a start it is good to have at least something. Later we can ofc iterate on it.

louwers:
The security policy states that vulnerabilities of 'services' should be reported to the e-mail provided. I don't know what these 'services' are in the context of MapLibre or MapLibre GL Native.

Vulnerabilities related to 'used open source software' need to be handled through bug reports. Since MapLibre GL Native iself is open source, does that mean that when someone discovers a critical security vulnerability in MapLibre GL Native, it should be reported through a bug report? I am not sure if that is a good idea, we need to think about that and discuss it.

> It is roughly outlined in the document in the maplibre repo. I think as a start it is fine. Though there are Github features which allow the submission of reports I think.

I don't know who can read the e-mails sent to security@maplibre.org, but we (@ovivoda and I) cannot at this point. We also need to think about agility in terms of informing users of MapLibre GL Native when a security vulnerability has been found, and pushing out new releases.

I appreciate the initiative, but instead of just merging this security policy, I want to put some more thought into how to handle secutity vulnerabilities in the context of MapLibre GL Native specifically.

louwers:
We already have the reporting through GitHub enabled for this repo.

image

Let's discuss this tonight during the TSC meeting, but this seems like a much better option than an e-mail address.

maxammann:
Yes indeed, I totally forgot about that feature. Let's change it and use this feature!

maxammann:
Email could be a good fallback though to include.

louwers:
@maxammann I thought about it and I want to include just one way to report an security issue, because otherwise it is more easy to miss a report.

PR #577, opened 2022-12-05
Adding documentation links to MapLibre README

thehoneymad:
Added documentation links for MapLibre GL Native Markdown Book, Android API Reference, and iOS API Reference to README.


wipfli (commented):
I think these links are already there. Maybe we should make them more prominent?

birkskyum:
There is already a docs section in the readme. It can be improved, but this pr seems to be outdated.

PR #587, opened 2022-12-09
Remove requirement to run android release on self-hosted runner

wipfli:
I think the Android release workflow does not work at the moment because it want to run on a self-hosted ubuntu runner, but we don't have any as far as I know.

This pull request makes the android release workflow run on a normal ubuntu github instance, a bit like in the `android-ci.yml` file...


boldtrn (approved):
Lgtm, since we don't have a self-hosted runner 👍

PR #590, opened 2022-12-10
Fix ubuntu arm64 binary getting published to last version

acalcutt:
The first job in this workflow bumps the version number, but the ubuntu arm64 job did not do a 'git pull' to get those changes. This made it publish the binary files to the last release instead of the current one.

This change just adds a 'git pull' like the other jobs in the workflow use to pick up the correct version,

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -249,6 +249,9 @@ jobs:
           submodules: recursive
           fetch-depth: 0
 
+      - name: Get Latest Version
+        run: git pull
+
       - name: Install dependencies macOS
         if: runner.os == 'macOS'
         env:


.github/workflows/node-release.yml
```diff
@@ -249,6 +249,9 @@ jobs:
           submodules: recursive
           fetch-depth: 0
 
+      - name: Get Latest Version
+        run: git pull
+
```
wipfli, line 254: Do you need two separate steps at all?

wipfli: Sorry mean two separat jobs...

acalcutt: ya, the versioning has to be run in a separate job before everything. the build are basically running 3 seperate builds for each OS in the matix, so they each need the same version

acalcutt: you can't really put the version bump in the OS matrix part of the workflow, otherwise it would run once per OS, bumping the version multiple times.

acalcutt:
If you look at  https://github.com/maplibre/maplibre-gl-native/releases/tag/node-v5.0.1-pre.7 , it actually got 'linux-arm64' packages meant for https://github.com/maplibre/maplibre-gl-native/releases/tag/node-v5.1.0-pre.0

wipfli (approved):
OK thanks

wipfli:
Does the release get triggered by tag push or by manual workflow trigger?

acalcutt:
It's a manually triggered worklow. the workflow makes a tag with the version when node-pre-gyp-github publishes the first binary.

acalcutt:
So basically, you run the 'node-release' action, select the kind of release it is (pre/(pre)patch/(pre)minor/(pre)major/. and it does the versioning and tagging.

wipfli:
The author of the releases was apparently always @birkskyum. Do you know why this is the case?

wipfli:
https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/DEVELOPING.md#publishing is this still up-to-date?

Issue #585, opened 2022-12-09
Make an Android release


wipfli:
Similar to #584, we also want an Android release...

wipfli:
The last one was a bit more than 1 year ago: https://github.com/maplibre/maplibre-gl-native/releases/tag/android-v9.5.2

wipfli:
The android release workflow seems to only have a manual trigger:

https://github.com/maplibre/maplibre-gl-native/blob/a405d7a9b33563ae6d3aabd55f075976b0219b24/.github/workflows/android-release.yml#L3-L4

wipfli:
I am tempted to just run this one and see where it crashes...

wipfli:
Hm I did not find any documentation on how to run the release. So let's just try...

wipfli:
![image](https://user-images.githubusercontent.com/53421382/206790057-e41ed290-8ec1-41d6-b49a-655ee1b4e043.png)

wipfli:
https://github.com/maplibre/maplibre-gl-native/actions/runs/3660524523/jobs/6187766501

wipfli:
It seems to be waiting for a self-hosted ubuntu runner. 

```
Requested labels: self-hosted, ubuntu-20.04
Job defined at: maplibre/maplibre-gl-native/.github/workflows/android-release.yml@refs/heads/main
Waiting for a runner to pick up this job...
```

Do we even still have self-hosted ubuntu runners?

wipfli:
#587 makes the android-release.yml workflow run just on the normal github runners.

wipfli:
Next attempt: https://github.com/maplibre/maplibre-gl-native/actions/runs/3660654651/jobs/6188035491

wipfli:
I think this will crash at the firebase test step.

wipfli:
How were the release notes generated in the past? Like here: https://github.com/maplibre/maplibre-gl-native/releases/tag/android-v9.5.2, is this just taken from the platform/android/changelog.md file if that one exists?

wipfli:
Hm that file did not get updated it seems... https://github.com/maplibre/maplibre-gl-native/blob/2fcd4f8dea87a43efd561128396dfec08f463a70/platform/android/CHANGELOG.md?plain=1#L1-L12

wipfli:
`git log --oneline -- platform/android` gives this:

```
d129432ce Bump semver from 7.3.7 to 7.3.8 in /platform/android (#530)
[...]

PR #591, opened 2022-12-10
Fix Maven read timeouts in Android release workflow

wipfli:
The android-release.yml workflow keeps timing out when reading from maven, see #585. This seems to be a known problem and one solution seems to be to change the maven connection settings, see https://github.com/actions/runner-images/issues/1499#issuecomment-718396233.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -19,6 +19,7 @@ jobs:
       JOBS: 8
       BUILDTYPE: Release
       IS_LOCAL_DEVELOPMENT: false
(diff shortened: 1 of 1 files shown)


wipfli:
I had a look at a few of the linked pull requests in https://github.com/actions/runner-images/issues/1499 and lot's of people seem to have taken that approach...

boldtrn (approved):
Weird an cryptic fix, but would be great if it would be fixed by that 👍 

Could you maybe add a comment in the yml to link to this PR so if anyone wonders about it in the future can find this easily?

PR #593, opened 2022-12-12
Fixed a crash in AndroidTestApp because of a parameter is marked non-...

artakka:
…nullable but it can be null. In fact it is guaranteed to be null on the first call.

Also fixed javadoc errors and added a few logs

### DIFF
--- a/platform/linux/src/headless_backend_egl.cpp
+++ b/platform/linux/src/headless_backend_egl.cpp
@@ -68,7 +68,7 @@ public:
     EGLConfig config = 0;
 };
 
-class EGLBackendImpl : public HeadlessBackend::Impl {
+class EGLBackendImpl final : public HeadlessBackend::Impl {
 public:
     EGLBackendImpl() {
         // EGL initializes the context client version to 1 by default. We want to
--- a/platform/android/MapboxGLAndroidSDK/build.gradle
+++ b/platform/android/MapboxGLAndroidSDK/build.gradle
@@ -13,6 +13,7 @@ dependencies {
     implementation dependenciesList.supportFragmentV4
     implementation dependenciesList.okhttp3
     implementation dependenciesList.gmsLocation
+    implementation dependenciesList.gson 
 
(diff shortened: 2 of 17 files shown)


wipfli:
That is really interesting to see. I have a suspicion that the android-release.yml workflow currently fails because we use an old gradle version, see #585. Nice that you update gradle here...

platform/linux/src/headless_backend_egl.cpp
```diff
@@ -68,7 +68,7 @@ class EGLDisplayConfig {
     EGLConfig config = 0;
 };
 
-class EGLBackendImpl : public HeadlessBackend::Impl {
+class EGLBackendImpl final : public HeadlessBackend::Impl {
```
wipfli, line 71: I am surprised that you needed to touch a file in the platform/linux folder since this change is mostly about the android test app. Can you comment?

wipfli:
ios-ci is probably failing for unrelated reasons, see #fix failing ios https://github.com/maplibre/maplibre-gl-native/issues/573

Why android-ci-pull and node-ci are failing, I don't know...

wipfli:
Maybe ci is failing because of the gradle update. Maybe it is better to make a pull request with only the gradle update and afterwards we can add the fix for the test app...

artakka:
@wipfli I will split the changes into 2 pull requests.
The reason Android ci is failing is because it expects the new build tools have their licenses accepted.

Failed to install the following Android SDK packages as some licences have not been accepted.
build-tools;30.0.3 Android SDK Build-Tools 30.0.3
  To build this project, accept the SDK license agreements and install the missing components using the Android Studio SDK Manager.
  All licenses can be accepted using the sdkmanager command line tool:
  sdkmanager.bat --licenses
  Or, to transfer the license agreements from one workstation to another, see https://developer.android.com/studio/intro/update.html#download-with-gradle
  
  We have to look how to do that in an automated build environment.

PR #594, opened 2022-12-13
[iOS]Use custom gesture to solve the map jump on m1 simulator

### LINKED ISSUE #563: Slow on M1 (Grab has bugfix)
## Summary

Fix GrabMap loading laggy and jump after pan on M1 simulator.

## Grab has bugfix

alanchenboy:
This Fixes #563

In Grab, our Apps don't use the arm64 version of iOS simulator SDKs for some reason. That means our code architecture is x86_64 and running under the rosetta stone on the m1 simulator. The project setting is:
Screen Shot 2022-12-13 at 16 44 55

But at this condition, the gesture's velocity will become very large or keep zero, and this also can reproduce with Maplibre, you can see the map jump very far when the pan over:

https://user-images.githubusercontent.com/5135389/207269457-4cf955fb-4320-40f3-9719-1b2b8353455c.mp4

The rotate gesture and the pinch gesture have the same problem.

So I create this fix, which is actually a workaround for m1 simulator users, I create custom gestures to calculate velocity and make the map handle correctly.
And this is the final result of this fix:

https://user-images.githubusercontent.com/5135389/207270549-4a233ca0-8da3-42e1-8935-d12ce876e58a.mp4

I know this may not be necessary for most situations.  But it is a must-have fix or workaround for Grab because a lot of developers running their maps in this situation. So please assess whether this PR needs to be merged into Maplibre.

### DIFF
--- a/platform/ios/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/platform/ios/src/MGLMapView.mm
@@ -74,7 +74,7 @@
 #include <cstdlib>
 #include <map>
 #include <unordered_set>
-#if TARGET_IPHONE_SIMULATOR
+#if TARGET_IPHONE_SIMULATOR & (TARGET_CPU_X86 | TARGET_CPU_X86_64)
 #include <sys/sysctl.h>
 
 // The m1 simulator's gesture's velocity is wrong and here is a workaround to fix it.
@@ -753,7 +753,7 @@ public:
     
     self.anchorRotateOrZoomGesturesToCenterCoordinate = NO;
     
-#if TARGET_IPHONE_SIMULATOR
+#if TARGET_IPHONE_SIMULATOR & (TARGET_CPU_X86 | TARGET_CPU_X86_64)
     bool isM1Simulator = processIsTranslated() > 0;
     if (isM1Simulator) {
         _pan = [[MBGLPanGesture alloc] initWithTarget:self action:@selector(handlePanGesture:)];
@@ -769,7 +769,7 @@ public:
     _scrollEnabled = YES;
     _panScrollingMode = MGLPanScrollingModeDefault;
 
-#if TARGET_IPHONE_SIMULATOR
+#if TARGET_IPHONE_SIMULATOR & (TARGET_CPU_X86 | TARGET_CPU_X86_64)
     if (isM1Simulator) {
         _pinch = [[MBGLPinchGesture alloc] initWithTarget:self action:@selector(handlePinchGesture:)];
     } else {
@@ -782,7 +782,7 @@ public:
     [self addGestureRecognizer:_pinch];
     _zoomEnabled = YES;
 
-#if TARGET_IPHONE_SIMULATOR
+#if TARGET_IPHONE_SIMULATOR & (TARGET_CPU_X86 | TARGET_CPU_X86_64)
     if (isM1Simulator) {
         _rotate = [[MBGLRotationGesture alloc] initWithTarget:self action:@selector(handleRotateGesture:)];
     } else {
--- a/platform/ios/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/platform/ios/src/MGLMapView.mm
@@ -74,6 +74,161 @@
 #include <cstdlib>
 #include <map>
 #include <unordered_set>
+#if TARGET_IPHONE_SIMULATOR
+#include <sys/sysctl.h>
+
+// The m1 simulator's gesture's velocity is wrong and here is a workaround to fix it.
+// These custom gestures will calculate the correct velocity.
+// This may because the system time is incorrect.
+
+@interface MBGLPanGesture : UIPanGestureRecognizer
+@end
+
+@interface MBGLPinchGesture : UIPinchGestureRecognizer
+@end
+
+@interface MBGLRotationGesture : UIRotationGestureRecognizer
+@end
+
+@implementation MBGLPanGesture {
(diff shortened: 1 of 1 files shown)


platform/ios/platform/ios/src/MGLMapView.mm
```diff
@@ -1991,7 +2171,7 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
     else if (pan.state == UIGestureRecognizerStateChanged)
     {
         CGPoint delta = [pan translationInView:pan.view];
-
+        printf("pan delta %f,%f\n", delta.x, delta.y);
```
wipfli, line 2174: Is the print statement still needed?

platform/ios/platform/ios/src/MGLMapView.mm
```diff
@@ -2015,6 +2195,7 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
     else if (pan.state == UIGestureRecognizerStateEnded || pan.state == UIGestureRecognizerStateCancelled)
     {
         CGPoint velocity = [pan velocityInView:pan.view];
+        printf("pan velocity %f,%f\n", velocity.x, velocity.y);
```
wipfli, line 2198: Is the print statement still needed?

wipfli (commented):
Thanks a lot @alanchenboy for sharing this. I don't have a mac at the moment so I cannot test this directly, but since all the code is inside a `#if TARGET_IPHONE_SIMULATOR` guard it will not change the behavior of the main code so we should be able to integrate this I think...

platform/ios/platform/ios/src/MGLMapView.mm
nvanfleet: This should definitely be removed

platform/ios/platform/ios/src/MGLMapView.mm
nvanfleet: same here

platform/ios/platform/ios/src/MGLMapView.mm
```diff
@@ -598,19 +753,44 @@ - (void)commonInit
     
     self.anchorRotateOrZoomGesturesToCenterCoordinate = NO;
     
+#if TARGET_IPHONE_SIMULATOR
+    bool isM1Simulator = processIsTranslated() > 0;
```
nvanfleet, line 757: I suppose it's not just M1 but an M1 using Rosetta (translated)?

platform/ios/platform/ios/src/MGLMapView.mm
alanchenboy: Thanks for review this print is deleted

platform/ios/platform/ios/src/MGLMapView.mm
alanchenboy: Grab doesn't have the arm64 version Map, now I Updated the macro test only effect to x86_64 & x86, I think it's the condition that would happen this issue (An m1 simulator & x86_64 architecture code).

platform/ios/platform/ios/src/MGLMapView.mm
alanchenboy: Thanks for review this print is deleted

platform/ios/platform/ios/src/MGLMapView.mm
alanchenboy: @nvanfleet Do you have any other suggestion to setup the guard?

roblabs:
The GitHub Action that is supposed to run has been [restarted](https://github.com/maplibre/maplibre-gl-native/actions/runs/3690963533).  Please monitor when approving this PR.

Failure is related to file I/O:  ` Permission denied`

* https://github.com/maplibre/maplibre-gl-native/actions/runs/3690963533/jobs/6248519378#step:13:1111

```
2022-12-14 01:48:23.778 xcodebuild[71261:40025363]  IDETestOperationsObserverDebug: Failure collecting logarchive: Error Domain=OSLogErrorDomain Code=13 "Log archive creation failed with error 13, warnings during creation: failed to create directory: /var/tmp/test-session-systemlogs-2022.12.14_01-48-23-+0000.logarchive: Permission denied (13)
[1111](https://github.com/maplibre/maplibre-gl-native/actions/runs/3690963533/jobs/6248519378#step:13:1112)
" UserInfo={NSLocalizedDescription=Log archive creation failed with error 13, warnings during creation: failed to create directory: /var/tmp/test-session-systemlogs-2022.12.14_01-48-23-+0000.logarchive: Permission denied (13)
[1112](https://github.com/maplibre/maplibre-gl-native/actions/runs/3690963533/jobs/6248519378#step:13:1113)
}

```

alanchenboy:
Hi @roblabs, The failing CI is caused by the main branch, and I already fix it https://github.com/maplibre/maplibre-gl-native/pull/599.
After rebasing all Unit tests passed.

platform/ios/platform/ios/src/MGLMapView.mm
birkskyum: @nvanfleet , can this be resolved in your opinion?

birkskyum: These guards are much better, but how do we know that this is an m1 simulator, and not just a developer using an x86_64 mac with a x86_64 simualtor?

birkskyum (approved):

PR #595, opened 2022-12-13
Fixed a crash in MapboxGLAndroidSDKTestApp because of a parameter mar…

artakka:
…ked as non-nullable that is actually guaranteed to be null on the first run.

This broke most of the app functionality.


wipfli:
It looks like we merged this as part of https://github.com/maplibre/maplibre-gl-native/pull/596/

wipfli:
Probably you can close this one, @artakka...

artakka:
You are right.

PR #596, opened 2022-12-13
JavaDoc error fixes

artakka:
Fixed JavaDoc errors
Removed redundant visibility modes
Added a few more logs

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapStrictModeException.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapStrictModeException.java
@@ -1,6 +1,6 @@
 package com.mapbox.mapboxsdk;
 
-class MapStrictModeException extends RuntimeException {
+public class MapStrictModeException extends RuntimeException {
   MapStrictModeException(String message) {
     super(String.format("Map detected an error that would fail silently otherwise: %s", message));
   }
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -12,6 +12,7 @@ import android.view.TextureView;
 import android.view.View;
 import android.widget.FrameLayout;
 import android.widget.ImageView;
+import android.util.Log;
 
 import androidx.annotation.CallSuper;
 import androidx.annotation.NonNull;
@@ -23,6 +24,7 @@ import com.mapbox.android.gestures.AndroidGesturesManager;
 import com.mapbox.mapboxsdk.MapStrictMode;
 import com.mapbox.mapboxsdk.Mapbox;
 import com.mapbox.mapboxsdk.R;
+import com.mapbox.mapboxsdk.WellKnownTileServer;
 import com.mapbox.mapboxsdk.annotations.Annotation;
 import com.mapbox.mapboxsdk.constants.MapboxConstants;
 import com.mapbox.mapboxsdk.exceptions.MapboxConfigurationException;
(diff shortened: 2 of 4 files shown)


wipfli:
There were some lint issues which made `make android-check` fail, I fixed them now...

artakka:
@wipfli  Javadoc errors where there before my changes and I have no idea how to fix them or how the javadoc used to work at some point unfortunately. 

I fixed two lint errors with the latest push.

wipfli:
How can I reproduce the problem which this pull request solves?

platform/android/MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/MapboxTest.java
```diff
@@ -85,8 +85,8 @@ public void testNoInstance() {
 
     expectedException.expect(MapboxConfigurationException.class);
     expectedException.expectMessage(
-      "\nUsing MapView requires calling Mapbox.getInstance(Context context, String apiKey) before "
-      + "inflating or creating the view."
+      "\nUsing MapView requires calling Mapbox.getInstance(Context context, String apiKey," +
+              " WellKnownTileServer wellKnownTileServer) before inflating or creating the view."
```
wipfli, line 89: CI complains that 

```
native/platform/android/MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/MapboxTest.java:88:93: '+' should be on a new line. [OperatorWrap]

```

wipfli: Maybe it is easiest if you run locally `make android-check` to make sure that the linting is fine. Or maybe there are better local options, I don't know...

artakka: Fixed that one. Hardcoded test

artakka:
Almost there, lint wanted + to be on the new line.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapStrictModeException.java
```diff
@@ -1,6 +1,6 @@
 package com.mapbox.mapboxsdk;
 
-class MapStrictModeException extends RuntimeException {
+public class MapStrictModeException extends RuntimeException {
```
boldtrn, line 3: Out of curiosity, why was this needed?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -99,24 +101,28 @@ public class MapView extends FrameLayout implements NativeMapView.ViewCallback {
   @UiThread
   public MapView(@NonNull Context context) {
     super(context);
+    Log.d("MapLibreLibrary", "MapView constructed with context");
```
boldtrn, line 104: Isn't this library using Timber? If yes, could you use Timber instead of Log?

boldtrn (changes requested):
It would be great if you would use Timber instead of Log 👍 

Could you clarify what you mean by redundant visibility mode and why the  visibility changes are necessary? Are some of these related to your other PRs?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
artakka: Good catch. Will do.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapStrictModeException.java
artakka: It was referred from JavaDoc in MapView.java and was erroring.
But also it is thrown in response to a public API call, and being internal does not make sense to me.
Screen Shot 2022-12-14 at 9 47 10 AM

boldtrn: Ok got it, then it's fine :)

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
artakka: Just pushed an update that uses Timber. Had to add timber and plant it on SDK as it was used in TestApp only, but I think it was worth it.

artakka:
@boldtrn By redundant visibility I meant this construct:

 private class LayerListAdapter internal constructor(context: Context?, layers: List<Layer>) :
 
 In this case internal constructor has no effect as the class is private so you cannot have less restrictive constructor. I just removed it for code clarity. It should not affect usage or generated code.

sulewicz (approved):

wipfli:
Welcome back @sulewicz :)

wipfli (approved):

thehoneymad:
I am late to the party. Happy to see the Timber suggestion! :)

PR #601, opened 2022-12-14
Remove unneeded credentials

wipfli:
For read access to the GitHub container registry, we do not need any credentials like the `CR_PAT` secret which I think stands for "container registry personal access token".

This pull request removes the unneeded credentials from the android-release.yml workflow.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -11,9 +11,6 @@ jobs:
         working-directory: platform/android
     container:
       image: ghcr.io/maplibre/android-ndk-r21b
-      credentials:
-        username: ${{ github.repository_owner }}
-        password: ${{ secrets.CR_PAT }}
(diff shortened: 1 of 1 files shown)


thehoneymad:
This still leaves me a little curious. Why do we not need any credentials to access Github container registry?

wipfli:
It is public for read access

wipfli:
You can run `docker run --rm -it ghcr.io/maplibre/android-ndk-r21b`  on your local machine and it will take you into the container.

wipfli:
Only when you want to write to the github docker registry, e.g. in 
https://github.com/maplibre/maplibre-gl-native/blob/2478d54261908711cbea89de4d4212917e021f9f/.github/workflows/android-docker-ndk-r21b.yml#L16-L29
you will need the credentials...

PR #602, opened 2022-12-14
Gradle plugin version and gradle version updates

artakka:
Upgraded gradle plugin from version 3.6.4 to 7.3.1]
Upgraded gradle from 6.4.1 to 7.4
Did the automatic migration steps
Fixed an error with a non final class with final destructor
Fixed other nullability issues


.github/workflows/android-ci.yml
```diff
@@ -58,6 +58,9 @@ jobs:
           submodules: recursive
           fetch-depth: 0
       
+      - name: Setup Android SDK and accept licenses
+        uses: android-actions/setup-android@v2
+
```
wipfli, line 63: We use a docker container to have a reproducible environment. Maybe we should actually change the container? I don't know how this stuff works, but if we use the container, I think we should not change the environment inside it...

wipfli:
The android docker container is documented here: https://github.com/maplibre/maplibre-gl-native/tree/main/platform/android/docker

boldtrn:
Just wondering, why was this PR closed? don't we still want to upgrade gradle and kotlin?

wipfli:
I think @artakka is debugging the license thing in his fork.

artakka:
@boldtrn Almost there. Getting a failure now about submodule commit not matching the main commit I

boldtrn:
Perfect, thanks for keeping me updated 👍

artakka:
@boldtrn I pass the Android test on my fork now with updated gradle, gradle plugin and tools, but fail during later MapboxGLAndroidSDK:androidNitpick

`What went wrong:
Execution failed for task ':MapboxGLAndroidSDK:androidNitpick'.
> maplibre-java vendor repository is not checked out on the consumed binary's tag.
  Expected commit: c328d36ab5cefc33bc7d24285cc6ee1f3dc2f461(v5.9.0 tag).
  Actual commit: 8b4d1d6de6fc7e2766f965edae892c35e11f0235.
  If you've updated the version in the dependencies.gradle file, make sure to bump the submodule pin in the /vendor directory to match the release tag.
  If you've bumped the pin, make sure to verify the version tag prefix in the android-nitpick.gradle file.`

boldtrn:
Interesting, thanks for sharing @artakka. It seems like the `androidNitpick` script is checking out the repo at `https://github.com/maplibre/maplibre-java/commits/main` and checks if the latest commit on main is equal to the commit wit the used tag?

https://github.com/maplibre/maplibre-gl-native/blob/2478d54261908711cbea89de4d4212917e021f9f/platform/android/gradle/android-nitpick.gradle#L46

I am not sure why this check is enforced?

wipfli:
This happens in many places, for example the docs website for MapLibre Gl JS had something similar. It think the idea is just that you always want to be at the latest published version...

wipfli:
You are right @boldtrn

artakka:
Ok, I think I will comment out that test and enter a task to understand and fix it in the future, if you do not object.

We really need to update the build and switch everything to Kotlin which will be my next step.

boldtrn:
Perfectly fine by me 👍, thanks for working on this @artakka.

wipfli:
I think that is a good strategy @artakka. Our goal should just be to do an Android release (#585) as soon as possible because it would be nice to have a published version before larger changes go into `main`. But making a release is blocked currently by the fact that `make android-javadoc` fails in java 11 while it used to work in java 8, see #603.

artakka:
@wipfli @boldtrn Here is the new PR: https://github.com/maplibre/maplibre-gl-native/pull/611
It works well AFAIK and succeeds the automated tests.
I will do more smaller tweaks next.

PR #603, opened 2022-12-15
Run javadoc in android ci pull

wipfli:
Currently I think we are unable to make an Android release because the `make android-javadoc` fails in the `android-release.yml` workflow.

The problem seems to be that `make android-javadoc` does not work with java 11.

This pull request adds a `make android-javadoc` step to the Android pull request workflow such that we detect problems with javadoc earlier in the future.

Draft for now because it will fail...

Steps to reproduce this locally:

```
# get java 11 on ubuntu
cd platform/android
make android-javadoc
```

Getting the environment correct might be tricky. One way to do it is with a docker container, see #604


wipfli:
@artakka do you have a Java 11 development environment?

wipfli:
The javadoc error can be reproduced locally with the steps described in #604

wipfli:
Produces a bunch of errors which look like this:

```
/src/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java:9: error: cannot find symbol
import androidx.annotation.Keep;
                          ^
  symbol:   class Keep
  location: package androidx.annotation

```

wipfli:
Do I need to add the `androidx.annotation` package to a .pom file or something?

wipfli:
The folliwng packages cannot be found:

```
import androidx.annotation.Keep;
import androidx.annotation.NonNull;
import androidx.annotation.WorkerThread;
import androidx.annotation.Nullable;
import androidx.annotation.UiThread;
import androidx.annotation.ColorInt;

import timber.log.Timber;

import com.google.gson.Gson;
import com.google.gson.JsonArray;
import com.google.gson.JsonElement;
import com.google.gson.JsonNull;
import com.google.gson.JsonObject;
import com.google.gson.JsonPrimitive;

import com.mapbox.geojson.GeoJson;
import com.mapbox.geojson.Polygon;
import com.mapbox.geojson.gson.GeometryGeoJson;

import com.mapbox.mapboxsdk.R;

import androidx.core.widget.ImageViewCompat;
```

I wonder if this file is configured correctly?

https://github.com/maplibre/maplibre-gl-native/blob/2478d54261908711cbea89de4d4212917e021f9f/platform/android/gradle/gradle-javadoc.gradle#L1-L18

wipfli:
Commenting out the `classpath += ...` line leads to exactly the same errors as we have now. Does this line the right thing?

artakka:
https://stackoverflow.com/questions/17033878/generate-javadocs-with-android-gradle-plugin

birkskyum:
We would get around this by adopting Dokka, the Kotlin (and Java) generator for mixed codebases that is intended to replace javadoc in situations like ours https://github.com/Kotlin/dokka

birkskyum:
@louwers , I guess it changes the situation around this PR.

PR #613, opened 2022-12-17
Migrate Android test app models to kotlin

birkskyum:
As part of the [kotlin transition](https://github.com/maplibre/maplibre-gl-native/blob/main/design-proposals/2022-09-02-kotlin.md), this PR migrates the Test App models.

Only a few java files are left in the test app at this point.


artakka (changes requested):
There are some nullability failures in the tests. Please fix them and resubmit.

artakka (approved):

PR #615, opened 2022-12-18
Prepare an Android release

wipfli:
Adds some instructions about how to make an android release, updates the changelog in preparation for android-v9.6.0, and adjusts the android-release.yml workflow to something which probably should create a release.

Making a release like proposed here only works if we temprarily revert the gradle update, see #614

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: [self-hosted, Linux]
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


wipfli:
Related to #585

.github/workflows/android-release.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: [self-hosted, Linux]
```
birkskyum, line 8: Is this intended?

birkskyum (approved):

PR #616, opened 2022-12-18
Remove firebase step from android-release workflow

wipfli:
The firebase step currently fails and blocks us from making an android release. I think we should just release what we have now, and then figure out why the test fails and fix it for future releases...

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -104,32 +104,7 @@ jobs:
         shell: bash
       
       - name: Generate javadoc
-        run: make android-javadoc
-
-      - name: Log in to Google Cloud Platform
-        run: |
-          GLCLOUD_SERVICE_JSON='${{ secrets.GLCLOUD_SERVICE_JSON }}'
-          if [[ -n "${GLCLOUD_SERVICE_JSON}" ]]; then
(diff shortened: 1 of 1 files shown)


boldtrn (approved):
Makes sense, we can always add this again when it's fixed.

wipfli:
Added a note about the firebase testing at https://github.com/maplibre/maplibre-gl-native/issues/605#issuecomment-1357608817

wipfli:
I will move the `android-v9.6.0` tag to the latest commit, because the release workflow checks out the repo at the latest tag starting with android.

According to 
https://stackoverflow.com/questions/8044583/how-can-i-move-a-tag-on-a-git-branch-to-a-different-commit

I can do 

Delete the tag on any remote before you push

```
git push --atomic origin main android-v9.6.0
```

Replace the tag to reference the most recent commit

```
git tag -fa android-v9.6.0
```

Push the tag to the remote origin

```
git push origin main --tags
```

wipfli:
I can delete the android-v9.6.0 tag with 
image

at https://github.com/maplibre/maplibre-gl-native/tags

wipfli:
Deleted the android-v9.6.0 tag via web interface

PR #620, opened 2022-12-19
[Android] Update usage of GMS location from deprecated APIs

### LINKED ISSUE #607: [Android] Update, GMS Location
[Currently gms is at 17.0.0](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/android/gradle/dependencies.gradle#L32). We should update this to the latest version. We are seeing weird location behaviour on some Android devices in the wild and believe this might be related to the outdated gms version.

The latest version right now is `[21.0.1](https://mvnrepository.com/artifact/com.google.android.gms/play-services-location/21.0.1)`.

boldtrn:
Fixes #607. The gms dependency was quite outdated. We had user reports where getting the location did not work reliably with the old version.


boldtrn:
Thanks, I updated the PR 👍

wipfli:
Probably this is good, but I have now clue how to review or test this pull request. @artakka what do you think?

boldtrn:
You can test the location itself by using the sample app, I tried several samples, they still work, you just need to set the correct styles (either API key, or you can set a style URL).

I updated a deprecated LocationRequest init.

For this I used the deprecation notice here: https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest

I also checked the changelog of gms location here: https://developers.google.com/android/guides/releases

I haven't seen any other breaking changes.

boldtrn:
With #627 merged, this PR is mostly superseded. Should I still commit the updated LocationRequest init? Or should we close this PR?

wipfli:
Would be good if you could rebase and then we can have a look at what is left...

carlonzo:
What are the reasons we should bump this one here? as far as I understand this is not just for the demo but we are bumping the dependency of the sdk, is it necessary? this will force all the clients to bump as well.
even more, location 21 had some breaking changes that I dont think are necessary for this project. what do you think?

boldtrn:
>  as far as I understand this is not just for the demo but we are bumping the dependency of the sdk, is it necessary? 

Yes this is correct, we are bumping the version of the SDK.

We had several users where it was not possible to get an accurate location, this is now fixed with this update. 

Is there a good reason not to keep dependencies up to date?

You can always overwrite the dependency or completely exclude GMS-Location if there is a particular reason not to update for special scenario?

BTW: the update has been merged with a different PR already, so on the main branch we are already on 21.0.1.

carlonzo:
> 
> We had several users where it was not possible to get an accurate location, this is now fixed with this update.
> 

If the client needs a newer version, they can update on the client side. We dont need to update the library. If your users had issue with the location, bump your location dependency, not the maplibre one.

>
> Is there a good reason not to keep dependencies up to date?
>

I'm all up for upgrading 3rd parties. But upgrading on library side forces the client to make changes that are up them to make or not. A 3rd party should depend on the minimum required 3rd party.

I would recommend to not merge this and downgrade whatever forces the bump of location. We could bump to something lower and stable, but not to the latest.

boldtrn:
IMHO, dependencies should be kept up to date, this helps to improve stability and fix security vulnerabilities. If you are using the default implementation of maplibre, no changes are required. To be honest, I don't really see the issue. I expect the next release of maplibre-gl-native/android to be a major update, if you look at the changes that currently happen in the Android part. All dependencies have been updated to the latest version, not just GMS location.

birkskyum:
Are parts of this PR obsoleted by #627?

birkskyum:
After rebasing this PR, the diff is much smaller. The changelog should be updated to reflect the changes is they are still meaningful.

birkskyum (approved):

boldtrn:
Thanks for rebasing and merging this @birkskyum 👍 Sorry I have been a bit unresponsive the last few days :)

PR #621, opened 2022-12-19
[Android] Catch exception onMove

boldtrn:
This is an exception that showed up in our Crashlytics log.

```
Fatal Exception: java.lang.Error: latitude must not be NaN
       at com.mapbox.mapboxsdk.maps.NativeMapView.nativeMoveBy(NativeMapView.java)
       at com.mapbox.mapboxsdk.maps.NativeMapView.moveBy(NativeMapView.java:248)
       at com.mapbox.mapboxsdk.maps.Transform.moveBy(Transform.java:314)
       at com.mapbox.mapboxsdk.maps.MapGestureDetector$MoveGestureListener.onMove(MapGestureDetector.java:493)
       at com.mapbox.android.gestures.MoveGestureDetector.analyzeMovement(MoveGestureDetector.java:158)
       at com.mapbox.android.gestures.MultiFingerGesture.analyzeEvent(MultiFingerGesture.java:106)
       at com.mapbox.android.gestures.ProgressiveGesture.analyzeEvent(ProgressiveGesture.java:55)
       at com.mapbox.android.gestures.MoveGestureDetector.analyzeEvent(MoveGestureDetector.java:141)
       at com.mapbox.android.gestures.BaseGesture.analyze(BaseGesture.java:61)
       at com.mapbox.android.gestures.BaseGesture.onTouchEvent(BaseGesture.java:39)
       at com.mapbox.android.gestures.AndroidGesturesManager.onTouchEvent(AndroidGesturesManager.java:193)
       at com.mapbox.mapboxsdk.maps.MapGestureDetector.onTouchEvent(MapGestureDetector.java:218)
       at com.mapbox.mapboxsdk.maps.MapView.onTouchEvent(MapView.java:523)
```

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
@@ -480,7 +480,7 @@ final class MapGestureDetector {
     @Override
     public boolean onMove(@NonNull MoveGestureDetector detector, float distanceX, float distanceY) {
       // first move event is often delivered with no displacement
-      if (distanceX != 0 || distanceY != 0) {
+      if (!Float.isNaN(distanceX) && !Float.isNaN(distanceY) && (distanceX != 0 || distanceY != 0)) {
         // dispatching camera start event only when the movement actually occurred
         cameraChangeDispatcher.onCameraMoveStarted(CameraChangeDispatcher.REASON_API_GESTURE);
 
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
@@ -41,6 +41,8 @@ import static com.mapbox.mapboxsdk.constants.MapboxConstants.ZOOM_RATE;
 import static com.mapbox.mapboxsdk.maps.MapboxMap.OnCameraMoveStartedListener.REASON_API_GESTURE;
 import static com.mapbox.mapboxsdk.utils.MathUtils.normalize;
 
+import timber.log.Timber;
+
 /**
  * Manages gestures events on a MapView.
  */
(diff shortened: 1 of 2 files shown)


wipfli:
Thanks for opening this pull request, @boldtrn. Do you know if there are unit tests where we could check that the library does the right thing even if `distanceX` is not a number?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
```diff
@@ -480,7 +480,7 @@ public boolean onMoveBegin(@NonNull MoveGestureDetector detector) {
     @Override
     public boolean onMove(@NonNull MoveGestureDetector detector, float distanceX, float distanceY) {
       // first move event is often delivered with no displacement
-      if (distanceX != 0 || distanceY != 0) {
+      if (!Float.isNaN(distanceX) && !Float.isNaN(distanceY) && (distanceX != 0 || distanceY != 0)) {
```
artakka, line 483: Good find! For short term solution this is a good fix.

This will fix the immediate crash. But there are some questions:

Why is this being called when distances are NAN? Do we call it or is this called externally.

In both cases silently ignoring the call may cause problems later. In some ways I prefer an exception, that will indicator that something is wrong with the code. If we ignore it it may manifest later in a more subtle errors that will be harder to diagnose.

At least we should log the error. because one way or another it is wrong to call this when the distances are not a number.

artakka: We can probably also check / log the onMoveEnd parameters. I suspect they could also be NAN. But the best thing would be to ensure they are not NAN. unless they come from external caller.

boldtrn: So far I haven't been able to reproduce this error, so I am not 100% sure what might have happened here.

My assumption was that this is a similar case to the one mentioned in the comment

`// first move event is often delivered with no displacement`

If you want I can add a logging statement? We can also leave this one open for now, so far there has only been one instance of this error, so it could be that it was some special edge case.

artakka: I say log an error and merge. We can keep an eye on it if it happens more often.

boldtrn: Ok, I added a log :)

artakka (approved):

PR #623, opened 2022-12-20
Make sure the still mode can take snapshot with resource error..

### LINKED ISSUE #564: Snapshot with Resource Errors (Grab has bugfix)
## Summary

Make sure the still mode can take snapshot with resource error.

## Grab has bugfix

alanchenboy:
Remove the throw in `Map::renderStill` to make sure the snapshot could success whit resource error.

Fixes #564


ovivoda:
@alanchenboy I suggest to also fix the test case, in order to have a green build.

louwers (changes requested):
You don't need to use a raw pointer to `std::exception_ptr`, it can be passed it by reference instead.

Although I'm do not think this is idiomatic C++ error handling. Maybe use the `RenderResult` class to indicate errors?

platform/default/src/mbgl/gfx/headless_frontend.cpp
```diff
@@ -140,28 +140,26 @@ PremultipliedImage HeadlessFrontend::readStillImage() {
...
 
-HeadlessFrontend::RenderResult HeadlessFrontend::render(Map& map) {
+HeadlessFrontend::RenderResult HeadlessFrontend::render(Map& map,
+                                                        std::exception_ptr *error) {
     HeadlessFrontend::RenderResult result;
-    std::exception_ptr error;
     gfx::BackendScope guard { *getBackend() };
 
     map.renderStill([&](const std::exception_ptr& e) {
         if (e) {
```
louwers, line 149: What kind of exceptions can `std::exception_ptr` contain? Can all be safely ignored?

platform/default/src/mbgl/gfx/headless_frontend.cpp
```diff
@@ -140,28 +140,26 @@ PremultipliedImage HeadlessFrontend::readStillImage() {
...
 }
 
-HeadlessFrontend::RenderResult HeadlessFrontend::render(Map& map) {
+HeadlessFrontend::RenderResult HeadlessFrontend::render(Map& map,
+                                                        std::exception_ptr *error) {
     HeadlessFrontend::RenderResult result;
-    std::exception_ptr error;
     gfx::BackendScope guard { *getBackend() };
 
     map.renderStill([&](const std::exception_ptr& e) {
```
louwers, line 148: Is the callback passed to `renderStill` called multiple times? In that case the error would get overwritten.

platform/default/src/mbgl/gfx/headless_frontend.cpp
alanchenboy: We ignore the resource error like the font file download failed, you know, HeadlessFrontend::render should handle this kind of error, do you read the unit test it just simulates this condition, the unit test `TEST(Map, ResourceError)` try to take a snapshot, but the font resource download failed at first, and success in the second try.  Before this edition, the snapshot can process success because of the render function abort by resource error(break the render loop and throw the error), and after this edition the snap shot could success.
error not be ignored, SDK(Adr, iOS, etc.) also can capture it via parameters.

platform/default/src/mbgl/gfx/headless_frontend.cpp
alanchenboy: Yes, to prevent lose errors, maybe we could use the callback to handle the errors.

alanchenboy: callback added, all errors could be handled at the endpoint.

PR #631, opened 2022-12-21
Fixed docker build failure

artakka:
The error seems to be related to java11 and how annotations are parsed when generating javadoc. As it is an external error hopefully it will be fixed when we update java.


platform/android/gradle/gradle-javadoc.gradle
```diff
@@ -14,5 +14,6 @@ android.libraryVariants.all { variant ->
         options.overview("src/main/java/overview.html")
         options.group("Mapbox Android SDK", "com.mapbox.*")
         exclude '**/R.java', '**/BuildConfig.java', '**/**.java.ejs', '**/**.html'
+        failOnError false
```
wipfli, line 17: Thanks for looking into this. I think for now this is fine as the warning existed before already I think. Can you isolate this change and make a pull request with only this file here?

wipfli (changes requested):

platform/android/gradle/gradle-javadoc.gradle
artakka: Sure

.github/workflows/node-ci.yml
```diff
@@ -221,6 +221,9 @@ jobs:
         run: |
           brew install docker docker-compose
 
+      - name: Make sure we have enough room
```
wipfli, line 224: I guess this is the relevant change for the docker build failure. If so, can you remove the other files from this pull request?

artakka: @wipfli Let us see if we can merge https://github.com/maplibre/maplibre-gl-native/pull/630 first. With that the files would automatically be in sync.

artakka: Merge did the magic. Now it is one file only!

platform/node/scripts/Dockerfile_Build
```diff
@@ -1,32 +0,0 @@
-FROM ubuntu:focal
```
wipfli, line 1: I think we want to keep this dockerfile

wipfli:
@acalcutt does this look good to you?

platform/node/scripts/Dockerfile_Build
artakka: Absolutely. My plan was to revert it not to delete it 

I will be traveling next 24 hours or so. Could you please unde oth delete?

acalcutt:
Other than putting back the Dockerfile_Build file, I am wondering if a forced system prune would affect ofther instances of the ci running. For example, if you had multiple PRs, the could be multiple instances of this running.

I wonder if we could give the build image a randomized name and then only clean up that image when it is completed.

wipfli:
Another option would be to publish the docker image to our github packages page. We have already containers there to build android. What do you think?

acalcutt:
I don't think that would fix  what this PR is meant to be fixing would it. even if it didn't build the image on the fly, it would still be loading the image to docker when it ran.

I never had this issue of having to clean this up with the github runners, but I'm assuming on our self runner they are building up?

acalcutt:
I wouldn't be opposed to prebuilding that image and putting it on the repo, but one thing to note is it has to be build for the arm64 platform.

wipfli:
The self-hosted runners are different from the github runners, becuase they keep the full state after a workflow run I think. Sad that there is not better isolation and reproducibility in self-hosted github runners...

wipfli:
Maybe the reproducibility and isolation issue could be solved with docker containers

PR #642, opened 2022-12-27
Remove GLKView.snapshot to prevent the app from crash

### LINKED ISSUE #566: Crash Caused by GLKView.snapshot (Grab has bugfix)
## Summary

Delete snapshot when app switches to the background.
## Grab has bugfix

alanchenboy:
GLKView.snapshot can crash the app randomly, GrabMap just removes it to prevent the app from crashing.

Fixes #566

### DIFF
--- a/platform/ios/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/platform/ios/src/MGLMapView.mm
@@ -315,12 +315,6 @@ enum { MGLAnnotationTagNotFound = UINT32_MAX };
 /// The threshold used to consider when a tilt gesture should start.
 const CLLocationDegrees MGLHorizontalTiltToleranceDegrees = 45.0;
 
-/// The time between background snapshot attempts.
-const NSTimeInterval MGLBackgroundSnapshotImageInterval = 60.0;
-
-/// The delay after the map has idled before a background snapshot is attempted.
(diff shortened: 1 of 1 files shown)


birkskyum:
Is it unlikely that this code is used by someone?

alanchenboy:
> Is it unlikely that this code is used by someone?

I think it not execute in SDK. The purpose for these code is: before enter background, the sdk take a snapshot then use a image view cover the whole map, and pause the map rendering to save energy. But GLKView.snapshot just randomly crash, so I think mapbox turn it off by config, and from user view this function is not acceptable, the snapshot is not same with the map when the map bring to front user can get a jump for map because the snapshot view is not as same as the gl view.

ovivoda:
I started by asking a question in maplibre-ios slack channel to double check if anyone is actually using this snapshot feature. 
I suggest we wait a bit, maybe we should define this wait times btw to 1 week or 10 days, before merging this PR's. Just in case someone is using that code part already.

ovivoda:
@alanchenboy can you please help out by resolving the conflicts. Tx!

alanchenboy:
Rebased

ovivoda:
It has been two weeks since I asked this on slack. No response back so I think we are safe to merge it. @birkskyum please advice otherwise.

birkskyum:
I agree with merging

birkskyum (approved):

PR #647, opened 2023-01-03
Replace mapbox watermark on the maps with maplibre #643 and replace mapbox resource names

### LINKED ISSUE #643: Replace mapbox watermark on the maps with maplibre
**Describe the bug**
The watermark icon on the bottom left of the maps says Mapbox

 - OS: Android
Screenshot_20221228-213534

louwers:
Closes #643

You can already disable the icon, but ~@acalcutt~ @ntadej  had the idea to make it configurable which icon is shown. We should create a ticket out of this.


acalcutt:
That was actually @ntadej to make the icon configurable.

I was just saying in maplibre-gl-js it used show based on a mapbox tilejson parameter. it had been changed at one point to look for a maplibre tilejson parameter, which meant it basically never was shown.

I had modified in the js version in https://github.com/maplibre/maplibre-gl-js/pull/786 so it no longer shows based on TileJSON, but instead was a settable map option. The reason was more to repurpose this logo to advertise maplibre.

louwers:
Screenshot_20230103-181606_MapLibre_Android_SDK_TestApp

birkskyum (changes requested):
Transitioning the `resourcePrefix from 'mapbox_'` to `resourcePrefix from 'maplibre_'` almost deserves it's own PR, but in order to do it stepwise like this, I reckon it can be disabled to let the lint pass while it's ongoing.

louwers:
@birkskyum Since you said 'almost'... 

I added `mlns:tools="http://schemas.android.com/tools" tools:ignore="ResourceName"` to  `attrs.xml` and `public.xml` since they are part of the Public API. Maybe we can duplicate them and then phase them out.

birkskyum:
Great! I'd like an Android changelog entry. Also, it's generally preferred to use rebate rather than merge when synching the branch to master as it simplifies the git history.

louwers:
> Also, it's generally preferred to use rebate rather than merge when synching the branch to master

Thanks for the heads-up, I will do that next time. When I squash the commits when I merge later, the git history should be clean in this case right?

birkskyum:
Yes, it'll be clean in the main branch after squashing, so it's only related to the review process

birkskyum (approved):

PR #653, opened 2023-01-04
Redirect users to GitHub Discussions and Slack

louwers:
To make users aware of the fact that questions can be asked on GitHub Discussions and Slack. Looks as follows:

image


wipfli (approved):

birkskyum:
Good idea

birkskyum:
It's merged, but I don't see the new options when opening an issue

louwers:
> It's merged, but I don't see the new options when opening an issue

Typo in the filename, fixed now.

PR #655, opened 2023-01-04
Fixed a crash when native code was accessing the LatLngBounds class

artakka:
Fix for https://github.com/maplibre/maplibre-gl-native/issues/648


birkskyum (approved):
Great! It did the trick.

PR #659, opened 2023-01-05
Add fix for running jazzy on M1/ARM

louwers:
Ran into an issue when running the GitHub action for the iOS release: https://github.com/maplibre/maplibre-gl-native/actions/runs/3848364221/jobs/6556050541

It is a known problem that is caused by running jazzy on M1/ARM chip:

- https://github.com/realm/jazzy/issues/1259
- https://github.com/ffi/ffi/issues/864

### DIFF
--- a/platform/ios/platform/ios/scripts/document.sh
+++ b/platform/ios/platform/ios/scripts/document.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+source ~/.bashrc
+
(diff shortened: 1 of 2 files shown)


birkskyum (changes requested):
How come the build fails?

louwers:
The other gem seems to install just fine though.

Could you give me access so I can check the logs?

birkskyum (approved):

PR #664, opened 2023-01-08
[qt] Fix QThreadStorage cast

ntadej:
Fix `QThreadStorage` cast. This fixes compilation issues in some setups.


birkskyum (approved):

louwers:
@ntadej Do you know why `ThreadLocalBase` doesn't take a template parameter? Seems like that could avoid the need for a `void*` and `reinterpret_cast`.

ntadej:
I think this is because until now this is the only special implementation so it was done quickly. Also I assume this is for compatibility between Qt and MapLibre.

PR #665, opened 2023-01-08
[qt] Force enable moc for Qt6

ntadej:
When I try to build this repo as part of Qt distribution the `moc` tool does not seem to be enabled and needs to be forced. Not 100 % sure this is needed so keeping draft for now to allow more testing.


ntadej:
OK, it's good that I kept this one draft. Need to figure out how to do this properly.

ntadej:
It seems I still need this 🤷‍♂️.

I updated the changes so that they are compatible with standalone builds and builds as part of Qt.

birkskyum (approved):
Let's try it for now.

PR #674, opened 2023-01-10
Enable CPP unit test on macOS CI

alanchenboy:
Enable CPP unit test on macOS CI


ntadej:
We should really move away from this custom `Makefile`...

ntadej:
Well we already have CMake, no reason to hardcode custom makefiles if we can generate them. This is of course out of scope of this MR, it's more important to actually fix the tests.

louwers:
I think the Makefile in this case is just used as an alternative to a glorified script, not as a build tool.

Would you say it's better to use CTest in this particular case?

ntadej:
No, what I mean is that instead of manually making the glorified shell script one could define custom targets in CMake that would be also cross-platform (in this case CTest would probably really be the way forward though).

birkskyum:
If we stay within the scope of the PR for a second, I'd say it's an improvement to run the tests in the CI.

Only thing is we have to make them pass this CI check in order to merge.

ntadej:
Yes, I completely agree! Just I comment so we keep these things in mind.

alanchenboy:
Hi everyone, I create this PR and just want to point out that our CPP unit tests are not enabled.
So if everybody supports turning it on, I will make all unit tests pass(There has a unit test to test the capability of the concurrent requests but the test sever looks can't support it, so I will migrate the test sever to Node or Boost).
And if the make file does not make sense, I could move the building code to the CI file.

louwers:
@alanchenboy Full support for turning these tests on!

> the test sever looks can't support it

These are our self-hosted runners, so if there is anything that needs to be installed you can let me know here.

> And if the make file does not make sense, I could move the building code to the CI file.

For now it is OK in there, restructuring it will be part of another PR.

alanchenboy:
Now the CPP unit test in mac is back. I bring back the node js test server, so it looks a little bit complex now,  the origin server provided by httplib can't handle 50 concurrent requests, it misses some requests.

louwers (changes requested):

test/src/mbgl/test/http_server.cpp
```diff
@@ -205,5 +205,81 @@ HttpServer::~HttpServer() {
     serverThread.join();
 }
 
+// Node Sever
```
louwers, line 208: Is this `NodeHttpServer` really needed? Is it possible to run the Node server before starting the test instead? Passing the Node.js executable and forking does not look very clean.

louwers (changes requested):

test/storage/server.js
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
alanchenboy: test/storage/server.js related to the mbgl-test project, the project base directory is located in test/. And the mbgl-test also exists in Linux,  Node, and QT, Cause I only know the macOS version I only turn the CI test on macOS.

The Sever is as same as the origin HttpServer in the file test/src/mbgl/test/http_server.cpp, because the original one can't handle 50 concurrent HTTP requests, I have to add this workaround.

test/src/mbgl/test/http_server.cpp
alanchenboy: The NodeHttpServer just replace the original HttpSever to support all test finish.

alanchenboy: This implementation is from the history of mapbox-gl-native.

.github/workflows/macos-ci.yml
```diff
@@ -93,7 +93,11 @@ jobs:
           node-version: 18
       
       - name: npm install
-        run: npm ci --ignore-scripts
+        run: |
+          npm ci --ignore-scripts
+          cd ../../
```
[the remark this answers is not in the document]
alanchenboy: Hi, thanks for reviewing, the working directory used.

test/src/mbgl/test/http_server.cpp
alanchenboy: I add a description at NodeHttpServer.

louwers: In my opinion starting the test server should not be done with C++ code, especially not if the server itself is not implemented with C++. If you are able to, please refactor to remove `NodeHttpServer` and instead run it with a shell command before running the test. `fork` and the other system calls you use also not cross-platform.

We can create a ticket to remove the old HttpServer completely after this is merged since it seems redundant now.

test/storage/server.js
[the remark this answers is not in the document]
louwers: Thanks for clarifying.

test/src/mbgl/test/http_server.cpp
louwers: If you disagree, let me know. In any case I would want NodeHttpServer in a separate header / source file and to get rid of the preprocessor macros.

alanchenboy: Another option is using the boost beast, but it's deleted in the Mapbox boost edition. 
You know I am a rendering engineer, and I am not good at socket things, maybe the `httplib` could do some improvement to support the concurrent request unit tests, and we needn't do any changes in `mbgl-test`, but the situation now is the current `HttpSever` can't pass the unit test `HTTPFileSource.Load` and `OnlineFileSource.Load`.
But if I use the `NodeHttpServer` these unit tests can pass without any issues.

So, I could take try to use boost beast to build a new test server but it will take time, as I am not good at this area. Or we can just use the NodeHttpServer(the Mapbox legacy unit test server) to let all unit test pass and do the corresponding improvements in the future.

louwers:
I'm making changes to this PR so that the test is run with CTest and the test server is stared and killed using `FIXTURES_SETUP` and `FIXTURES_SETUP`.

louwers:
> This is a continuation of #674 where @alanchengrab enabled C++ unit tests on macOS.
> 
> Instead of adding another entry to the Makefile I have changed things a bit so that the test can be built and ran through CMake & CTest (thanks @ntadej for the suggestion).
> 
> I removed the C++ wrapper around that forks and runs the Node.js server, and instead made a Node.js wrapper that forks to run the test. This approach is more platform indepentent, which will be helpful if we will use the Node.js based HTTP server for tests of other platforms in the future.

I merged my branch into this one. Hopefully the unit tests will run now. I forgot to install the npm packages in the root directory (which is used as working directory for the test).

louwers:
Node.js test is unfortunetely flakey #720

So merge is possible without it passing.

louwers (approved):
LGTM now, but I made some changes so maybe someone else can have a look. :slightly_smiling_face:

ovivoda:
It looks good to me @louwers ! Go ahead and merge it!

PR #677, opened 2023-01-10
Refactor CI workflows from paths-ignore -> paths #503

### LINKED ISSUE #503: Refactor CI workflows from paths-ignore -> paths

louwers:
Closes #503

I will keep an eye to see if CI is running when it should, to see if I missed a file.

If anyone else sees any glaring omissions, let me know.

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -9,6 +9,7 @@ on:
       - 'ios-*'
     paths:
       - 'platform/ios/**'
+      - 'platform/darwin/**'
       - ".github/workflows/ios-ci.yml"
       - "bin/**"
       - "expression-test/**"
@@ -28,6 +29,7 @@ on:
       - main
     paths:
       - 'platform/ios/**'
+      - 'platform/darwin/**'
       - ".github/workflows/ios-ci.yml"
       - "bin/**"
       - "expression-test/**"
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -3,67 +3,44 @@ name: ios-ci
 on:
   workflow_dispatch:
   push:
-    tags-ignore:
-      - '**'
-    paths-ignore:
-      - ".gitignore"
-      - "**.md"
-      # ignore other platforms
-      - "platform/android/**"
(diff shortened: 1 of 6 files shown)


.github/workflows/ios-ci.yml
```diff
@@ -3,67 +3,44 @@ name: ios-ci
...
-      - ".github/workflows/node-release.yml"
-      - ".github/workflows/qt-ci-windows.yml"
-      - ".github/workflows/qt-ci.yml"
-      - ".github/workflows/update-gl-js.yml"
     branches:
       - main
+    tags:
+      - 'ios-*'
+    paths:
+      - 'platform/ios/**'
```
ntadej, line 11: Needs `darwin` as far as I know.

.github/workflows/macos-ci.yml
```diff
@@ -5,65 +5,42 @@ on:
...
-      - ".github/workflows/macos-release.yml"
-      - ".github/workflows/node-ci.yml"
-      - ".github/workflows/node-release.yml"
-      - ".github/workflows/qt-ci-windows.yml"
-      - ".github/workflows/qt-ci.yml"
-      - ".github/workflows/update-gl-js.yml"
+    tags:
+      - 'macos-*'
+    paths:
+      - 'platform/ios/**'
```
ntadej, line 11: Also here.

ntadej (changes requested):

.github/workflows/ios-ci.yml
louwers: Thanks!

birkskyum (approved):

PR #682, opened 2023-01-11
Remove deprecated `junit.framework.Assert`

louwers:


ntadej:
Just to remind you that you can always reuse existing PR if you mess things up. You do not have to open a new one 😊

louwers:
@ntadej It's not possible to change the source branch, but yes, I could have force pushed the branch.

Thanks, my git-fu is growing stronger. ;-)

PR #683, opened 2023-01-11
Remove deprecated `junit.framework.Assert`

louwers:


birkskyum:
I think there might be some import statements of Assert missing i.e. in LocationCameraControllerTest. The tests fail because of it.

birkskyum:
There are some failing tests in LocationAnimatorCoordinatorTest.kt - issues with assertEquals()

louwers:
`assertEquals` now requires a delta as third argument for floating point numbers. I extracted a custom version with a delta to a package so it does not have to be repeated all the time.

birkskyum (approved):

PR #686, opened 2023-01-11
Port FeatureOverview to Kotlin and remove empty activity tests

birkskyum:
Port the initial screen in the TestApp that lists all the features.

Remove the generate-test-code, because it just generates empty "sanity" tests, one for each activity in the TestApp.


louwers (approved):
👋

PR #694, opened 2023-01-13
Fix issue with iOS network logging

nvanfleet:
A change I had made a while back which integrated networking errors into the logging stack (which was previously just a stub) ended up creating an error. Ultimately the logs were never format strings and were just NSStrings instead but they were being put through a variadic function. This was not a problem _unless_ the string going through had a format in it.

Unfortunately it's very possible for that to occur, specifically if there is a URL that has a space in it which ends up as "%20" which end up being picked up as a format. va_list then assumes that it's a format string and then puts in some kind of garbage information into those parts of the string. This means the URL itself can be completely garbled, often appearing to have random UTF characters (seemingly usually an asian language) and also creates a non-contiguous UTF-8 string which can cause some even greater issues depending on what you're doing with the string.

There is no way to do something like check if there are actual additional arguments provided in a variadic function. I also didn't want to do anything like scan the string itself or sanitize it. So it seems like the best thing to do is to just create a specific pathway for those logs that doesn't use variadic arguments. This ensures the log is still accurate and can't contain mangled information.

### DIFF
--- a/platform/ios/platform/darwin/src/MGLLoggingConfiguration.mm
+++ b/platform/ios/platform/darwin/src/MGLLoggingConfiguration.mm
@@ -76,6 +76,10 @@ public:
     
 }
 
+- (void)logCallingFunction:(const char *)callingFunction functionLine:(NSUInteger)functionLine messageType:(MGLLoggingLevel)type message:(NSString *)message {
+    _handler(type, @(callingFunction), functionLine, message);
+}
+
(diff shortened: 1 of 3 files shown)


ovivoda:
this looks legit to me.

louwers:
@ovivoda Could you approve & merge?

ovivoda (approved):

PR #700, opened 2023-01-16
MapLibre GL Core Build or Pull Request

thehoneymad:
Added a simple Github Workflow that will be triggered if any MapLibre GL Core path is changed. It will ignore all paths for any platform specific code.

test: Have tested with my local fork for include and src directory.

Next change is to deduce how we can export sources and binaries from this build that Android build can use. Or find a way for Android to rely on core build.

### DIFF
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -2,8 +2,9 @@ name: core-build
 
 on:
   workflow_dispatch:
-  push:
   pull_request:
+    branches:
+      - main
     paths:
       - "src/**"
       - "include/**"
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -3,8 +3,6 @@ name: core-build
 on:
   workflow_dispatch:
   pull_request:
-    branches:
-      - main
     paths:
       - "src/**"
       - "include/**"
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -2,6 +2,7 @@ name: core-build
 
 on:
   workflow_dispatch:
+  push:
   pull_request:
     paths:
       - "src/**"
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -1,0 +1,55 @@
+name: core-build
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/core-ci.yml
```diff
@@ -0,0 +1,33 @@
+name: core-build
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "src/**"
```
ntadej, line 9: You have to include the workfow itself I think.

thehoneymad: Ah, right. Good call.

thehoneymad: Added the change as requested.

ntadej:
Also please add `ccache`, it will make subsequent builds much faster.

louwers:
The root `CMakeLists.txt` needs to be added to the paths (this is true for other workflows as well).

thehoneymad:
> Also please add `ccache`, it will make subsequent builds much faster.

Ah, was thinking about it. Wanted to start with something darn simple. Will add it in upcoming commit.

thehoneymad:
Reading up on reusable workflow. The moment I am done, will come back to this. :) Thank you for the quick responses. ❤️

thehoneymad:
> The root `CMakeLists.txt` needs to be added to the paths (this is true for other workflows as well).

This is taken care of.

thehoneymad:
> Also please add `ccache`, it will make subsequent builds much faster.

This is also taken care of.

thehoneymad:
Ready to be reviewed again.

louwers:
I would like to trigger the workflow to make sure it runs through.

louwers (approved):
To build the core independently from the platforms makes sense in my eyes.

How exactly we want to decouple the platforms more would require a design proposal. I am not sure if making core releases and pinning the platforms to a specific core release are a good idea for example. I would rather want to make sure that the platforms are always releasable from the latest (as in `main`) core. But we can spar about this and other topics on during the design proposal discussion.

thehoneymad:
Totally open to. Now that I read your comment, it feels like I didn't land my opinion.

I am saying that we need a strategy for backward incompatible changes in core. Especially the ones that are not trivial to address on the platforms in the single change. :)

PR #707, opened 2023-01-19
Add windows support

tdcosta100:
I have specific needs to run Maplibre GL Native on Windows. So I created a port to Windows platform. It was inspired by Linux platform sources, and works in a similar way, creating some executables and the lib/node-vxx/mbgl.node modules. Therefore, Maplibre GL Native can be used with NodeJS in an Windows environment. I used TileServer-GL to test it, and it worked like a charm, the performance was similar to the Linux version. Build instructions are located at `platform/windows/README.md`. Remember to clone with `--branch windows-support`, because it have a new submodule. Or after changing to `windows-support` branch, you can init the `platform/windows/vendor/vcpkg` submodule manually. Feedback is welcome.


wipfli:
Oh beautiful, thanks @tdcosta100 for implementing windows support. 

If we were to merge this, would you be interested in being the lead-maintainer for the Windows platform binding?

The usual work consists of handling dependency updates, reviewing pull requests, and helping people getting started with the platform binding...

CMakeLists.txt
```diff
@@ -7,12 +7,17 @@ option(MBGL_WITH_SANITIZER "Use [address|thread|undefined] here" OFF)
...
 option(MBGL_WITH_OPENGL "Build with OpenGL renderer" ON)
 option(MBGL_WITH_EGL "Build with EGL renderer" OFF)
+option(MBGL_WITH_OSMESA "Build with OSMesa (Software) renderer" OFF)
 option(MBGL_WITH_WERROR "Make all compilation warnings errors" ON)
 
 if (MBGL_WITH_QT AND NOT CMAKE_OSX_DEPLOYMENT_TARGET)
     set(CMAKE_OSX_DEPLOYMENT_TARGET 13.0)
 endif()
 
+if(WIN32)
```
ntadej, line 17: This looks a bit intrusive. Qt for example should not use this toolchain. Is it too late if you load it in the platform-specific cmake file?

tdcosta100:
> Oh beautiful, thanks @tdcosta100 for implementing windows support.
> 
> If we were to merge this, would you be interested in being the lead-maintainer for the Windows platform binding?
> 
> The usual work consists of handling dependency updates, reviewing pull requests, and helping people getting started with the platform binding...

I can try. I'm just afraid sometimes I will be with lots of demand from my job and not respond so quickly as I would like. Outside these situations, I think I can do the job. Can we make a trial period to see if I am a good fit for it?

benchmark/CMakeLists.txt
```diff
@@ -49,6 +49,19 @@ if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
         LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
         RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
     )
+elseif(MSVC)
+    set_source_files_properties(
+        ${PROJECT_SOURCE_DIR}/benchmark/api/query.benchmark.cpp
+        ${PROJECT_SOURCE_DIR}/benchmark/api/render.benchmark.cpp
+        PRIVATE
+            COMPILE_FLAGS "/wd4244"
```
ntadej, line 57: Please fix the warnings in non-vendor files.

platform/windows/windows.cmake
```diff
@@ -0,0 +1,3 @@
+if(MSVC)
+	include(${CMAKE_CURRENT_LIST_DIR}/windows.msvc.cmake)
```
ntadej, line 2: I would not split the file per compiler.

platform/windows/include/unistd.h
```diff
@@ -0,0 +1,58 @@
+#ifndef _UNISTD_H
```
ntadej, line 1: I would make clear which files are just platform-specific implementations for MapLibre and which are replacements for system headers.

Also I'm not sure if I like we're hacking system-like headers. How many times do we actually need functions from this file?

ntadej (changes requested):
Thank you very much for providing this!

I provided some basic comments, I may look through the new files again later. To get this merged faster I would recommend to split the MR into smaller ones (e.g. fixing errors).

It would also be good to add a CI workflow to test your addition actually works.

tdcosta100:
> Thank you very much for providing this!
> 
> I provided some basic comments, I may look through the new files again later. To get this merged faster I would recommend to split the MR into smaller ones (e.g. fixing errors).
> 
> It would also be good to add a CI workflow to test your addition actually works.

No worries about the changes. I can change everything, even open a new PR if it's better. I would propose this project to be evaluated before doing a PR, but I just didn't know where to do that hahaha. So if it's not mature enough, I will make the necessary changes before the merge.

CMakeLists.txt
tdcosta100: I tried to make every if(WIN32) contained in the windows.cmake file, but this toolchain file needs to be included before the project command, otherwise it will be completely ignored. The idea here is to inject vcpkg if we are going to build in Windows platform. I'm heavily using vcpkg to simplify the dependencies.

benchmark/CMakeLists.txt
tdcosta100: Just to better understand the ability to change files: it's ok to make source code modifications instead supressing warnings? Some places have changes like this piece of code, but I can have the #if _MSVC inside the source code approach as well.

ntadej: I looked up `C4244` which is "floating point truncation". This one can be fixed for all platforms. As such fixes caused regressions in the past, it may be useful to open them in a separate PR.

platform/windows/include/unistd.h
tdcosta100: I can check which functions from this code are used. All I can say by now is some files are including unistd.h. We can trace back these includes and make some deeper changes than this hack.

platform/windows/windows.cmake
tdcosta100: I was hoping to have MingGW support in the future. If so, it would make sense? Or it's better even with MinGW support to have just one file?

ntadej: For Qt the changes between MSVC and MinGW are quite small. Also it will probably cause less duplication of code.

benchmark/CMakeLists.txt
tdcosta100: That's nice, I was hoping this change would be welcome. There are are some of these in several files. Using static_cast would make it better for all platforms. I will provide a separate PR to address this.

ntadej: To clarify, I fixed most of those when I did initial Qt port to Windows, but not all files are used there. We only ignore warnings for external packages where we have no direct control.

platform/windows/windows.cmake
tdcosta100: Great. So I will migrate windows.msvc.cmake to windows.cmake, and when time comes, make the required changes to have it compatible with MinGW.

benchmark/CMakeLists.txt
tdcosta100: Thanks for the clarification. Now I have a better understanding where I should touch the code, and where I souldn't. My primary concern was to be the least intrusive as I could. But if I have the permission to change some pieces of code (nothing too extensive, just some fixes like this one), I'd rather to.

wipfli:
> can I setup a workflow in a non-main branch

Yes, like this for example

```yaml
on:
  workflow_dispatch:
  pull_request:
    branches:
      - main
      - some_other_branch
```

acalcutt:
If you need any help with getting this into the node-ci/build, let me know. It would be nice to add node windows support.

If you modify the node-ci workflow in the PR it is already set to run on pull request, so I don't think you would need to modify the branches it runs on. (at least that was how I was testing it before it got merged)

tdcosta100:
> > can I setup a workflow in a non-main branch
> 
> Yes, like this for example
> 
> ```yaml
> on:
>   workflow_dispatch:
>   pull_request:
>     branches:
>       - main
>       - some_other_branch
> ```

Nice, thanks! I will do that then.

tdcosta100:
Once all pending changes to prepare for this PR are finally done (that is, changes that are not directly related with Windows, like Khronos definitions, and implicit casts), I'm resuming this PR. I converted it to draft in order to test actions for `NodeJS` on Windows platform. I didn't include actions yet, but once all tests passed, I will focus that.

platform/windows/FindOSMesa.cmake
```diff
@@ -0,0 +1,65 @@
...
+# Try to find the header
+find_path(OSMesa_INCLUDE_DIR NAMES GL/osmesa.h)
+
+get_filename_component(OSMesa_DIR ${OSMesa_INCLUDE_DIR}/../ ABSOLUTE)
+
+# Try to find the library
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(_ARCH x64)
+else()
+    set(_ARCH x86)
```
acalcutt, line 15: will the eventually need a arm64 if we want to support that?

acalcutt:
I spent some time working with @tdcosta100 to try and get this integrated into the 'node-ci' and 'node-release' workflows

We were able to get it all working, but I wanted to note a few issues we had along the way.

1.) When checking out the the repository on the github windows runner, I ran into path limit issues with some vendor test files. To get around this, I took a solution from the 'qt-ci', and did a shallow checkout and then pull the submodules after. This gets around the issue, but you can still see some errors relating to the paths that don't affect the build.

2.) To get ccache working in windows, I again pulled from the qt-ci, and use that to simplify our ccache use here (and make it work with windows).

3.) We initially had some issue with build type not being release, which made it this require c++ development dlls. we fixed that issue, but it should still be noted this requires c++ redistributable as a pre-req 

Other than that, I was able to ci/release workflow that works. 

I made a test release at
`NPM: "@acalcutt/maplibre-gl-native": "6.1.0-pre.4"`

and a tileserver-gl that uses it at 
https://github.com/acalcutt/tileserver-gl/tree/windows_test

With that I was able to confirm this works!
image
image

Great work @tdcosta100

tdcosta100:
Regarding the too long path issues, I confirm it's resolved, at least when cloning submodules (recursively!). I don't know how tests will behave because I didn't let a test job finish yet, but I hope in the next few hours we will confirm if it works correctly.

tdcosta100:
Finally got Windows tests working. Thank you @acalcutt again, for helping with the workflow.

platform/windows/FindOSMesa.cmake
tdcosta100: I don't see why not, but I really don't have tools for developing that. But theorically yes, it's possible.

CMakeLists.txt
tdcosta100: Do you think it's better to have the custom toolchain file to be included in cmake command? Something like:

```
cmake . -B build -DCMAKE_TOOLCHAIN_FILE=platform/windows/custom-toolchain.cmake
```

render-test/filesystem.hpp
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#ifdef WIN32
```
louwers, line 3: Can this be set with CMake instead?

platform/default/src/mbgl/storage/http_file_source.cpp
```diff
@@ -280,6 +280,11 @@ HTTPRequest::HTTPRequest(HTTPFileSource::Impl* context_, Resource resource_, Fil
         curl_easy_setopt(handle, CURLOPT_HTTPHEADER, headers);
     }
 
+#ifdef WIN32
+    // Windows have issues with TLSv1.3, so we limit to TLSv1.2. Should be resolved in a later cURL release
```
louwers, line 284:
```suggestion
    // Windows has issues with TLSv1.3, so we limit to TLSv1.2. Should be resolved in a later cURL release
```

expression-test/filesystem.hpp
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#ifdef WIN32
+#define GHC_WIN_DISABLE_WSTRING_STORAGE_TYPE
```
louwers, line 4: Can this `#define` be added with CMake?

louwers:
Added a new label for windows. 😉

louwers:
Could you update the root `README.md`:

```
- [⭐️ Android](platform/android/README.md) 
- [⭐️ iOS](platform/ios/platform/ios/README.md)
- [GLFW](platform/glfw)
- [Linux](platform/linux/README.md)
- [Node.js](platform/node/README.md)
- [Qt](platform/qt/README.md)
- [Windows](platform/windows/README.md)
- [macOS](platform/ios/platform/macos/README.md)
```

etnav:
This is awesome. Thank you @tdcosta100 (and the others of course) for your effort, I can relate to your struggles. Looking forward to seeing this merged and migrating to this (built this on windows about 2 years ago), maybe even contributing to this specific part in the future😄

.github/workflows/node-ci.yml
```diff
@@ -118,63 +123,98 @@ jobs:
...
+          cmake . -B build \
+            -G Ninja \
+            -DCMAKE_BUILD_TYPE=${{ env.BUILDTYPE }} \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_C_COMPILER=gcc-10 \
+            -DCMAKE_CXX_COMPILER=g++-10
+
+      - name: Configure maplibre-gl-native (Windows)
+        if: runner.os == 'Windows'
+        shell: cmd
```
acalcutt, line 170: can this be moved back to bash with the other syntax. the ccache plugin we are using notes

```
Notes on Windows support
Note that using Ccache on Windows is still experimental and https://github.com/ccache/ccache/issues/1023. Use Sccache for stable Windows support.
```
https://github.com/hendrikmuhs/ccache-action#notes-on-windows-support

ovivoda (approved):

CMakeLists.txt
ntadej: Sorry to unresolve this now, but this is not OK for Qt builds, even if it may work by accident.

expression-test/filesystem.hpp
ntadej: +1

platform/node/cmake/module.windows.cmake
```diff
@@ -0,0 +1,239 @@
+# We need CMake 3.9 because the Xcode generator doesn't know about systems include paths before 3.9.
```
ntadej, line 1: This comment does not fit Windows 😊

platform/node/cmake/module.windows.cmake
```diff
@@ -0,0 +1,239 @@
...
+# See https://gitlab.kitware.com/cmake/cmake/issues/16795
+cmake_minimum_required(VERSION 3.9)
+
+if (NOT NODE_MODULE_MINIMUM_ABI)
```
ntadej, line 5: Why do you need so many Node specifics for Windows but not for other platforms?

ntadej (changes requested):

platform/node/cmake/module.windows.cmake
acalcutt: I asked about this file, and the linux/macos equivalent is in the `mapbox/cmake-node-module` package
https://github.com/mapbox/cmake-node-module/blob/master/module.cmake

We talked a bit about if it may make sense to move the one for other platforms out of the cmake-node-module package so we have some control over it.

ntadej: OK, I'm fine with including this for now then.

acalcutt: Just a side fact. The linux version of the file being in a node package actually gave me some trouble in my cmake-js PR, because the path of node_modules changes if maplibre-native is used as a submodule

https://github.com/maplibre/maplibre-gl-native/pull/535/files#diff-6d48556e62e5d1ff41c30848ca089f0bb2fe15ddf3b58bcc1f33e902b7022eddR1-R6

acalcutt: @tdcosta100 just wondering, would this new file also be compatible with the linux/macos builds. like maybe replace module.cmake with this new file all together?

Looking at the difference it seems like mostly a few path changes and sections for WIN32 added.

tdcosta100:
> Could you update the root `README.md`:
> 
> ```
> - [⭐️ Android](platform/android/README.md) 
> - [⭐️ iOS](platform/ios/platform/ios/README.md)
> - [GLFW](platform/glfw)
> - [Linux](platform/linux/README.md)
> - [Node.js](platform/node/README.md)
> - [Qt](platform/qt/README.md)
> - [Windows](platform/windows/README.md)
> - [macOS](platform/ios/platform/macos/README.md)
> ```

Done!

README.md
```diff
@@ -24,10 +24,10 @@ The documentation of MapLibre GL Native is a work in progress. To get an archite
 - [⭐️ iOS](platform/ios/platform/ios/README.md)
 - [GLFW](platform/glfw)
 - [Linux](platform/linux/README.md)
-- [macOS](platform/ios/platform/macos/README.md)
 - [Node.js](platform/node/README.md)
 - [Qt](platform/qt/README.md)
 - [Windows](platform/windows/README.md)
+- [macOS](platform/ios/platform/macos/README.md)
```
acalcutt, line 30: I feel like macOS\Linux\Windows should be together.

platform/node/cmake/module.cmake
```diff
@@ -120,6 +120,22 @@ function(add_node_module NAME)
...
+
+			# Download the win-${_ARCH} libraries if we are compiling on Windows and don't have them yet
+			if(NOT EXISTS "${_CACHE_DIR}/lib/node/${_NODE_VERSION}/win-${_ARCH}/node.lib")
+				_node_module_download(
+					"win-${_ARCH} library for Node ${_NODE_VERSION}"
+					"https://nodejs.org/download/release/${_NODE_VERSION}/win-${_ARCH}/node.lib"
+					"${_CACHE_DIR}/lib/node/${_NODE_VERSION}/win-${_ARCH}/node.lib"
+				)
+			endif()
+		endif()
```
acalcutt, line 138: the indenting in this doesn't seem to match the lines above/below

platform/node/cmake/module.cmake
```diff
@@ -150,6 +166,7 @@ function(add_node_module NAME)
 
         target_include_directories(${_TARGET} SYSTEM PRIVATE
             "${_CACHE_DIR}/node/${_NODE_VERSION}"
+            "${PROJECT_SOURCE_DIR}/platform/windows/include"
```
acalcutt, line 169: should this only apply to WIN32?

CMakeLists.txt
acalcutt: would adding "AND NOT MBGL_WITH_QT" resolve this or would you want something else?

ntadej: Toolchain files should generally not be included by default and can be just provided when you call CMake (this is for example usually done for Android and iOS). Can you rather adapt your Node.js building scripts in such a way?

acalcutt: Is that like @tdcosta100  suggested a few posts above with specifying DCMAKE_TOOLCHAIN_FILE? Can you point to an example from the Android/iOS that we could see what you mean? I looked a bit through those worksflows but I didn't really see anything that looked like what you describe.

acalcutt: really just looking at what can be done to finish this PR off. Maybe another switch that's off by default that we can turn on witht he cmake command.

acalcutt: @ntadej , this has been changed in the current version to not apply to QT. So I'm still looking if this resolve this for you or what you would like instead. This has already been approved and tested, so please this is the one thing in the way of merge I would say.

ntadej: Indeed we can clean it up later.

ntadej (approved):

PR #708, opened 2023-01-19
glGetString signature does not match the pattern of other GL functions

tdcosta100:
The glGetString function's signature does not match other functions's pattern. Other functions are of type (* const functionName), but glGetString is (*glGetString), maybe because the return type of this function is const Glubyte *, but the const in returning type is not related to const in the function pointer. This PR fixes it.


louwers (approved):
Nice catch!

Issue #606, opened 2022-12-15
Java version strategy


wipfli:
In CI, we use java 11 which dates back to 2018 if I am not mistaken, see https://www.oracle.com/java/technologies/java-se-support-roadmap.html

What should be our strategy for the java version? Update as soon as a new one is generally available?

wipfli:
I have java 17 installed on my laptop and it is a bit frustrating for me that basically all commands fail if I do something like `cd platform/android && make android-check` for example. How do people solve this problem usually? Is there something like NVM for java?

birkskyum:
Most of these tools are unfortunately mac/linux only (sdkman / jenv etc.). A common solution in this day and age is using the Java->Kotlin auto-transpiler in android studio to move on.

artakka:
I will switch all Java to  Kotlin soon

louwers:
Java 17 will be used available / used on CI soon.

PR #712, opened 2023-01-20
Android CI based on GitHub Image

### LINKED ISSUE #622: Documenting the android docker build
We have docker containers for building the Android package but I am not sure that we have documentation on how to update and generate the docker containers.

The containers are stored in the github container registry at the MapLibre org level at https://github.com/orgs/maplibre/packages

We have:

* android-ndk-r21b
* android-base

Linux builder is probably for Qt, right @ntadej?

Let's document how we can update the docker images for the android builds...

louwers:
Alternative to the Nix based setup of #702

Closes #622 and #606 as well but uses the default Ubuntu image from GitHub.

> GitHub actually includes many relvant packages into their standard ubuntu images. See here for a list of versions: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md
> 
> Maybe we could just use the same versions as them and use directly their images...


louwers:
- `android-ci.yml`: https://github.com/louwers/maplibre-gl-native/actions/runs/3971563148
- `android-ci-pull.yml`: https://github.com/louwers/maplibre-gl-native/actions/runs/3971584067

thehoneymad:
Woke up in the morning. Got time to review! :D 

*It sounded way cooler in my head.*

Reviewing by EOD

.github/workflows/android-ci-pull.yml
```diff
@@ -28,14 +28,11 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
thehoneymad, line 31: Thank you! Is there a reason we are not using `ubuntu-latest`? Is this due to version pinning?

.github/workflows/android-ci.yml
```diff
@@ -25,17 +25,11 @@ on:
...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     defaults:
       run:
         working-directory: platform/android
-    container:
-      image: ghcr.io/maplibre/android-ndk-r21b
-      credentials:
-        username: ${{ github.repository_owner }}
-        password: ${{ secrets.CR_PAT }}
```
thehoneymad, line 36: What is our plan with this? Should we simply stop publishing this if this approach works out?

.github/workflows/android-release.yml
```diff
@@ -5,14 +5,11 @@ on:
 
 jobs:
   build:
-    runs-on: [self-hosted, Linux]
+    runs-on: ubuntu-22.04
```
thehoneymad, line 8: Awesome! ❤️

.github/workflows/android-ci.yml
```diff
@@ -25,17 +25,11 @@ on:
 
 jobs:
```
thehoneymad, line 26: Should we keep just one Github Action file here? Does not have to be done in this PR. I think having `android-ci.yml` and `android-ci-pull.yml` both for nearly the same outcome is unnecessary. [Reusing GIthub Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) should simplify the amount of copy-pasta 🍝  here.

.github/workflows/android-release.yml
```diff
@@ -44,6 +41,9 @@ jobs:
         run: npm install --ignore-scripts          
         working-directory: platform/android
 
+      - name: Install ccache
+        run: sudo apt-get install -y ccache
+
       - name: Prepare ccache
         run: ccache --clear
```
thehoneymad, line 48: Nit: We can combine the steps here.

.github/workflows/android-ci.yml
thehoneymad: Nevermind. That is what you did anyway. :D

.github/workflows/android-ci.yml
thehoneymad: For example, ccache setup and restore. This can easily be made to a reusable workflow with cache key passed as an input. Tired of seeing that copied and pasted everywhere like a plague.

platform/android/gradle/native-build.gradle
```diff
@@ -22,7 +25,7 @@ ext.nativeBuild = { nativeTargets ->
             externalNativeBuild {
                 cmake {
                     path "../MapboxGLAndroidSDK/src/cpp/CMakeLists.txt"
-                    version "3.10.2"
+                    version androidSdkVersions.cmakeVersion
```
thehoneymad, line 28: Lovely! Thank you!

platform/android/gradle/dependencies.gradle
```diff
@@ -6,6 +6,12 @@ ext {
             compileSdkVersion: 33,
     ]
 
+    androidSdkVersions = [
```
thehoneymad, line 9: Curious, why would we not merge `androidVersions` and `androidSdkVersions` array together? Semantically, what is the significance of keeping them separate?

thehoneymad (commented):
Half of my life went cleaning build systems. Calling this great work out.

Thank you @louwers! 🧹 Most of my comments are nitpicks. Approved.

louwers:
@thehoneymad Thanks for your review.

We can either merge this PR or #702.

Oliver and Ovi will take a look if they can run Nix locally, since one of it's advantages is that you can re-create an exact copy of the CI environment. Which approach has your preference?

thehoneymad:
I am still in favour of merging this. And then reading the Nix PR on top of this. That allows us to have a stable build with pure GitHub Actions first.

And then if Nix pans out, we can use that too. :)

ntadej:
I agree. We should not limit ourselves on Nix. I'm almost sure I'll not use it 😉 (but I also do not plan to develop the Android SDK).

.github/workflows/android-ci.yml
louwers: I agree, but yes let's do a separate PR for that.

.github/workflows/android-ci-pull.yml
louwers: Yes that is the reasoning.

platform/android/gradle/dependencies.gradle
louwers: One is releated to build tools, other to target Android versions. But I can merge them if you want.

.github/workflows/android-ci-pull.yml
wipfli: Dependabot can make pull requests to bump the versions used in github action workflows. Is it turned on at the moment for workflows?

louwers: I may make sense to use Ubuntu 22 for the forseeable future because it is a LTS release.

.github/workflows/android-ci.yml
louwers: Created an issue for it: https://github.com/maplibre/maplibre-gl-native/issues/735

.github/workflows/android-ci-pull.yml
louwers: > Dependabot can make pull requests to bump the versions used in github action workflows. Is it turned on at the moment for workflows?

Yes, I think so.

platform/android/gradle/dependencies.gradle
thehoneymad: If there is a semantic difference, I am okay to have them as is. I was just curious.

.github/workflows/android-release.yml
louwers: Let's do that with the reusable ccache workflow.

thehoneymad (approved):

PR #716, opened 2023-01-22
Avoid implicit casts and make code more portable

tdcosta100:
There are a lot of implicit casts everywhere: from double to float, from double to int, from int to double, etc. This PR changes them to explicit casts as long as they generate warnings like `warning C4244: '=' : conversion from 'double' to 'float', possible loss of data`. Also warnings like `warning C4834: discarding return value of function with 'nodiscard' attribute` are fixed. MSVC won't compile with these warnings.

### DIFF
--- a/benchmark/util/dtoa.benchmark.cpp
+++ b/benchmark/util/dtoa.benchmark.cpp
@@ -28,20 +28,20 @@ static void Util_dtoa(::benchmark::State& state) {
 
 static void Util_standardDtoa(::benchmark::State& state) {
     while (state.KeepRunning()) {
-        std::to_string(0.);
-        std::to_string(M_E);
-        std::to_string(M_LOG2E);
-        std::to_string(M_LOG10E);
-        std::to_string(M_LN2);
-        std::to_string(M_LN10);
-        std::to_string(M_PI);
-        std::to_string(M_PI_2);
-        std::to_string(M_PI_4);
-        std::to_string(M_1_PI);
-        std::to_string(M_2_PI);
-        std::to_string(M_2_SQRTPI);
-        std::to_string(M_SQRT2);
-        std::to_string(M_SQRT1_2);
+        std::ignore = std::to_string(0.);
+        std::ignore = std::to_string(M_E);
+        std::ignore = std::to_string(M_LOG2E);
+        std::ignore = std::to_string(M_LOG10E);
+        std::ignore = std::to_string(M_LN2);
+        std::ignore = std::to_string(M_LN10);
+        std::ignore = std::to_string(M_PI);
+        std::ignore = std::to_string(M_PI_2);
+        std::ignore = std::to_string(M_PI_4);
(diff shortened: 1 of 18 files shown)


platform/node/src/node_expression.cpp
```diff
@@ -79,9 +79,7 @@ void NodeExpression::Parse(const Nan::FunctionCallbackInfo<v8::Value>& info) {
 
     auto success = [&cons, &info](std::unique_ptr<Expression> result) {
         auto nodeExpr = new NodeExpression(std::move(result));
-        const int argc = 0;
-        v8::Local<v8::Value> argv[0] = {};
-        auto wrapped = Nan::NewInstance(cons, argc, argv).ToLocalChecked();
+        auto wrapped = Nan::NewInstance(cons).ToLocalChecked();
```
louwers, line 82: I am not familiar with Nan. Is this the same?

louwers (changes requested):
Thanks for your contribution. Making implicit narrowing explicit is definitely a good idea.

benchmark/util/dtoa.benchmark.cpp
```diff
@@ -28,20 +28,20 @@ static void Util_dtoa(::benchmark::State& state) {
...
-        std::to_string(M_LN10);
-        std::to_string(M_PI);
-        std::to_string(M_PI_2);
-        std::to_string(M_PI_4);
-        std::to_string(M_1_PI);
-        std::to_string(M_2_PI);
-        std::to_string(M_2_SQRTPI);
-        std::to_string(M_SQRT2);
-        std::to_string(M_SQRT1_2);
+        (void)std::to_string(0.);
```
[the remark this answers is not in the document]
tdcosta100: Done.

platform/node/src/node_expression.cpp
tdcosta100: Yes, it is. According to [this](https://github.com/nodejs/nan/blob/main/doc/maybe_types.md#api_nan_new_instance), Nan::NewInstance have 3 overloads. So calling with 0-length arguments is the same than omitting argc and argv parameters.

tdcosta100:
node-ci/test failed for some obscure reason (Failed to open X display, wtf), so I forced a re-run. Now everything is ok to merge, if it's all right with the code.

louwers (approved):
Thanks!

Maybe in the future we can use an implementation of the [Guidelines Support Library](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#gslutil-utilities) with `narrow` and `narrow_cast` utilities to make these narrowing conversions more searchable. But I don't think it's worth it to pull in a whole library for this at this point.

ntadej:
Given the number of dependencies we have GSL would probably be one of the more useful ones, e.g. for `std::span`.

tdcosta100:
I agree, this would improve the code quality a lot. But I fear some dependencies would make it difficult to use widespread. Just a feeling, not anything I really know.

PR #722, opened 2023-01-24
Avoid implicit casts and portable printf with size_t

tdcosta100:
Like the previous PR, this one is meant to avoid those `possible loss of data` errors. Plus, it uses the `z` format for `size_t` instead of `l`, because `size_t` can vary in size (32-bit or 64-bit).


ntadej (approved):

tdcosta100:
Is there anything pending in this PR I need to do?

ntadej:
No, I just thought now that we have more people @ovivoda or @louwers would also have a quick glance and merge it, but I can do it now 😊

louwers:
@ntadej Sorry I completely missed this. Thanks for merging it!

tdcosta100:
No problem, thank you!

PR #723, opened 2023-01-24
use xvfb-action in node-ci

acalcutt:
Test fix for https://github.com/maplibre/maplibre-gl-native/issues/720 . Try using xvfb-action

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -166,12 +166,9 @@ jobs:
           path: metrics/macos-xcode11-release-style.html
 
       - name: Test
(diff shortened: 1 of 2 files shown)


ovivoda:
@acalcutt I would add this one as ready for review. Any other things that you want to push here?

acalcutt:
I think the last ci run confirms this doesn't fix the issue unfortunatly. like I mentioned in https://github.com/maplibre/maplibre-gl-native/issues/720#issuecomment-1402078229 it is basically running the same commmand

image

acalcutt:
Usually with this issue I just re-run the job and the next time it succeeds.

Issue #335, opened 2022-06-16
Post C++ 17 clean-up


wipfli:
If we should migrate the codebase to C++ 17, then we should clean-up some things. This issue is to track where we need to clean-up stuff.

wipfli:
`#if __cplusplus >= 201703L` in `include/mbgl/actor/actor_ref.hpp`

louwers:
I'm closing this as done.

If something else comes to mind that needs updating we will create separate tickets for it.

PR #731, opened 2023-01-25
Use `*_t` and `*_v` trait helpers from C++17 STL

louwers:
Progress towards #335

@wipfli Do you have any other C++17 cleanup ideas in mind?


.gitignore
```diff
@@ -40,3 +40,4 @@ node_modules
 /lib
 
 /bazel-*
+CMakeLists.txt.user
```
ntadej, line 43: This should usually be a separate PR 😊

include/mbgl/actor/actor_ref.hpp
```diff
@@ -38,11 +38,7 @@ class ActorRef {
     template <typename Fn, class... Args>
     auto ask(Fn fn, Args&&... args) const {
         // Result type is deduced from the function's return type
-#if __cplusplus >= 201703L
```
ntadej, line 41: This will certainly break some platforms (unless we fixed them).

ntadej:
I'm not sure we actually moved all the clients to C++17.

We should also remove all the polyfill libraries that were made/added by Mapbox to get C++17 features for older compilers.

include/mbgl/actor/actor_ref.hpp
ntadej: As expected, the macOS build fails.

louwers:
> As expected, the macOS build fails.

Hmm, I will check out why it doesn't use C++17. It is set in the root `CMakeLists.txt`, but it probably needs to be set in the sub project as well.

> We should also remove all the polyfill libraries that were made/added by Mapbox to get C++17 features for older compilers.

@ntadej Thanks, I will have a look! Do you know anything about those polyfill libraries? Are they in `vendor`?

louwers:
Neat, macos passes now.

ntadej:
Great! It may be useful to check for additional warnings though. Not sure if we allow them or not.

louwers:
I had to add CMAKE_CXX_STANDARD and CMAKE_CXX_STANDARD_REQUIRED to platform/android/MapboxGLAndroidSDK/src/cpp/CMakeLists.txt

This file is actually used as a root CMakeLists.txt and uses `add_subdirectory`, which discards the variable scope upon re-entry. It's a bit weird that the root CMakeLists.txt is used both as a child and in other places as a parent.

Another problem is that the CMakeLists.txt under MapboxGLAndroidSDK does not call project() which it should as a root.

ntadej:
Yes, these extra CMake files should go away or be simplified with the reorganisation.

ntadej:
Hi @louwers, what if we merge this now and remove polyfills in a separate one? No need to do everything at once 😊

louwers:
@ntadej Sure!

ntadej (approved):

wipfli:
> @wipfli Do you have any other C++17 cleanup ideas in mind?

I don't know

PR #738, opened 2023-01-26
Fix iOS app signing for hardware and improve docs.

hallahan:
Addresses https://github.com/maplibre/maplibre-gl-native/issues/725

Hopefully this makes getting up and running more clear?

### DIFF
--- a/platform/ios/platform/ios/INSTALL.md
+++ b/platform/ios/platform/ios/INSTALL.md
@@ -25,7 +25,7 @@ Before building, follow these steps to install prerequisites:
 1. Install [Homebrew](http://brew.sh)
 1. Install additional dependencies.
    ```bash
-   brew install cmake ccache pkg-config glfw cmake-docs
+   brew install node cmake ccache pkg-config glfw cmake-docs
    ```
 
 1. Install [xcpretty](https://github.com/supermarin/xcpretty) for prettifying command line builds. (optional):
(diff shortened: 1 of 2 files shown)


louwers (approved):
Great improvement!

platform/ios/platform/ios/INSTALL.md
```diff
@@ -25,7 +25,7 @@ Before building, follow these steps to install prerequisites:
 1. Install [Homebrew](http://brew.sh)
 1. Install additional dependencies.
    ```bash
-   brew install cmake ccache pkg-config glfw cmake-docs
+   brew install node cmake ccache pkg-config glfw cmake-docs
```
roblabs, line 28: 1.  Is node required to build?  I don't see it on the GitHub workflows, but maybe it's used for local development.
2. @louwers — maybe we can have the GitHub Workflows match the README, e.g., we should integrate @hallahan's changes into https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/ios-ci.yml#L70 the dependencies in the iOS CI.

hallahan: I'm not sure if it's required. I think it is for running tests at least.

louwers: The CI dependencies are configured with a separate repo: https://github.com/maplibre/maplibre-ci-runners/blob/main/macos.yaml

So those installation commands should be removed from the workflow.

PR #742, opened 2023-01-27
Improve Kotlinification of LatLng

fynngodau:
This avoids breaking compatibility with existing Kotlin code because the setters are now accessible using `=`, by removing `setLatitute` / `setLongitude` and using normal kotlin setters.


artakka (approved):

louwers:
@fynngodau Thank you for your contribution. The unit tests do not pass, can you have a look?

fynngodau:
@louwers Thanks for your feedback, the tests now compile & pass!

PR #756, opened 2023-01-31
Actually build core in core-build.yml

louwers:
I noticed that the new workflow that builds the core just configures CMake, and doesn't actually build the core.

This should fix it.


thehoneymad (approved):

thehoneymad:
This is my bad. Thank you for fixing it. 🙇

Issue #762, opened 2023-02-01
Run Android Instrumentation Tests on AWS Device Farm


louwers:
We used to run Android Instrumentation tests for Android with Firebase, but we want to move it over to AWS Device Farm (#761).

@birkskyum is able to set up an IAM account that is linked to MapLibre.

When any secrets need to be set up, please send them to me in a secure way (e.g. Bitwarden send via Slack or [PGP](https://keybase.io/louwers)). I can add them to the repo.

---

Inside `platform/android`, the app with the UI tests can be built using

```
make android-ui-test-arm-v8
```

See the [Makefile](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/android/Makefile#L155).

The old workflow (that will be removed soon) may also be helpful:

https://github.com/maplibre/maplibre-gl-native/blob/85e33f88b1386bddc63f340365bbb461e3f8b724/.github/workflows/android-ci.yml#L132

louwers:
Pinging @thehoneymad since he said he knows some at AWS that is willing to jump on this. :rocket: :muscle:

louwers:
https://github.com/realm/aws-devicefarm

louwers:
I tried to run the Android instrumentation tests locally.

I am getting several tests that fail with, with a "Style isn't initialised" assertion failing in`validateTestSetup` and others a failing "MapboxMap isn't initialised" assertion failing in the same method.

The `reopenQueryRendererFeaturesActivity` crashes for me, after which the tests stop.

```
02-18 17:17:06.706 27100 27127 I TestRunner: started: reopenQueryRendererFeaturesActivity(com.mapbox.mapboxsdk.integration.QueryRenderedFeaturesBoxCountTest)
02-18 17:17:06.710  5853  7161 W InputManager-JNI: Input channel object '8f38ee8 com.mapbox.mapboxsdk.testapp/com.mapbox.mapboxsdk.testapp.activity.maplayout.SimpleMapActivity (client)' was disposed without first being removed with the input manager!
02-18 17:17:06.717  5853  7161 W InputManager-JNI: Input channel object 'Letterbox_left_ActivityRecord{8f1cb76 u0 com.mapbox.mapboxsdk.testapp/.activity.maplayout.SimpleMapActivity t139} (client)' was disposed without first being removed with the input manager!
[...]

Issue #761, opened 2023-01-31
Define plan for Android instrumentation tests


louwers:
The Android CI is failing due to a problem with the instrumentation tests: https://github.com/maplibre/maplibre-gl-native/actions/runs/4056188076/jobs/6980347353

We have MapTiler's credentials configured for Firebase Instrumentation tests. @wipfli mentioned that instead of using Google we can use AWS, since we have an AWS account and they have credits for us. I suggest that I disable the instrumentation tests for now, and that I create another issue for moving the instrumentation tests on AWS.

wipfli:
@artakka @thehoneymad are AWS device farm tests of interest to you?

louwers:
@artakka Aready mentioned it's hugely important that MapLibre runs well on older devices as well.

thehoneymad:
@wipfli it does. I have not deep dived on this yet. Would you like to take the lead to propose a design?

thehoneymad:
Looping @roblabs in. Turns out he has tried the device farm path before.

maxammann:
In other open source projects I'm using https://www.browserstack.com/ successfully.

It's free for OSS and very expensive otherwise :D

wipfli:
> Would you like to take the lead to propose a design?

Probably best if Bart does it

ovivoda:
I saw that the test is reporting device not supported. Do we know if this is the reason why it fails? Who can take a look into firebase account? IE has access?

wipfli:
The friebase account is a left-over from MapTiler. They donated the native repo to MapLibre, but we should not use their Google account anymore for CI. We have some donated AWS credits and should use those for testing. Investing time in the google firebase platform does not seem to make sense if we anyway do not plan to use it afterwards.

wipfli:
Or maybe use browserstack as @maxammann suggested. In any case, Firbase can be safely removed I think.

wipfli:
Secrets and credentials should be documented here: https://github.com/maplibre/maplibre/blob/main/SECRETS.md

wipfli:
Like that, we can have one central place where we understand what logins we have and who you have to ask to get the clear text login. It is a bit outdated, but lets aim to collect all relevant info there.

louwers:
> Secrets and credentials should be documented here: https://github.com/maplibre/maplibre/blob/main/SECRETS.md

Done https://github.com/louwers/maplibre/pull/2

PR #763, opened 2023-02-01
Remove Android Instrumentation Tests running on Google

louwers:
See #761.

Result of the short discussion was that we remove the Google-based Android instrumentation tests, and run them on AWS Device Farm (#762).


thehoneymad:
Before we remove the Firebase tests, is there a documentation of what was done here? We probably need to understand it from a high level so we can wire the device farm tests.

Oh, and on that  https://github.com/realm/aws-devicefarm

louwers:
I already copied over the file pinned to a commit to an issue.

Should be enough.

ovivoda (approved):

Issue #737, opened 2023-01-26
Investigate configuring Xcode with pure CMake


louwers:
Investigate if CMake's iOS support is now good enough to configure Xcode with CMake.

Right now we run CMake, but also copy hardcoded xcsettings files and xcconfig around. It's quite hairy and not very transparant for newcomers to the project. See this [Makefile](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/Makefile).

Maybe we can get rid of all the checked in configuration files like [this one](
https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/platform/macos/macos.xcodeproj/project.xcworkspace/contents.xcworkspacedata#L5).

References:

- [Comment from @hallahan](https://github.com/maplibre/maplibre-gl-native/discussions/699#discussioncomment-4791544) on this
- [Inveresting thread on Slack about this topic](https://osmus.slack.com/archives/C04K0NBUEEM/p1674662156251219)

ntadej:
Note that I do that for my project, but it is Qt-based, so maybe it's simpler in that sense.

louwers:
First potential blocker https://gitlab.kitware.com/cmake/cmake/-/issues/20160

louwers:
> we need to figure out what are the actual limitations of CMake

@ntadej As an example, setting up the iosapp target: https://github.com/maplibre/maplibre-gl-native/tree/main/platform/ios/platform/ios/app

I don't think it is possible with CMake. Even the [ios-cmake](https://github.com/leetal/ios-cmake/tree/master/example/example-app) project has a checked in `.xcodeproj` directory alongside their example app.

louwers:
Closing because #840 is implemented.

PR #765, opened 2023-02-02
Enable C++ unit tests on macOS

louwers:
This is a continuation of #674 where @alanchengrab enabled C++ unit tests on macOS.

Instead of adding another entry to the Makefile I have changed things a bit so that the test can be built and ran through CMake & CTest (thanks @ntadej for the suggestion).

I removed the C++ wrapper around that forks and runs the Node.js server, and instead made a Node.js wrapper that forks to run the test. This approach is more platform indepentent, which will be helpful if we will use the Node.js based HTTP server for tests of other platforms in the future.


ovivoda:
hmm @louwers This looks odd:

Screenshot 2023-02-02 at 21 24 51

louwers:
@ovivoda Yup, I'll have a look.

alanchenboy:
Thanks for pick this up.

My origin pr runs the server at the root dir because the CPP unit test is not a macOS CPP unit test, just the mbgl-test project of macOS version. So if it's not running at the root dir the resources in the test directory will not be loaded successfully. and that's why the unit test can't pass.

BTW some objects in the macOS targets can't work, the node objects, the `test-render-runner`. 

this unit test is quite like the `Run render tests on *`,  in node-ci.

louwers:
@alanchenboy The root dir is still used as working directory, but I just made an out of source build.

As part of #737 I am investigating if we can move more stuff out of the custom Makefile and into CMake.

@ovivoda Turns out I forgot to push a file.

ntadej:
Thanks for working on this! I'll look at it in detail tomorrow.

For reworking the scripts, we may want to wait for platforms reorganisation. I should have a draft of the proposal this weekend.

louwers:
I'm not sure if the tests are being run:

image

alanchenboy:
`mbgl-test-runner` cost 0.06 seconds I think this means no unit tests runed

PR #766, opened 2023-02-03
Add typings for Node Platform

KiwiKilian:
I augmented the module as far as I could figure out the types from the docs and usage. As I see there are a few more hidden function from the native module. I'm not sure if these should also be augmented and if so how:

```js
  setBackendType: [Function: setBackendType],
  ParseError: [class ParseError extends Error],
  Map: [Function: Map], // this is alreade augmented
  Expression: [Function: Expression] { parse: [Function: parse] },
  Resource: {
    Unknown: 0,
    Style: 1,
    Source: 2,
    Tile: 3,
    Glyphs: 4,
    SpriteImage: 5,
    SpriteJSON: 6
  },
```

### DIFF
--- a/platform/node/index.d.ts
+++ b/platform/node/index.d.ts
@@ -1,0 +1,39 @@
+declare module '@maplibre/maplibre-gl-native' {
+  export type RequestResponse = {
+    data: ArrayBuffer;
+    modified?: Date;
+    expires?: Date;
+    etag?: string;
+  };
+
+  export type MapOptions = {
+    request: (
+      request: { url: string; kind: number },
+      callback: (error?: Error, response?: RequestResponse) => void,
+    ) => void;
+    ration?: number;
+  };
+
+  export type RenderOptions = {
+    zoom?: number; // defaults to 0
+    width?: number; // px, defaults to 512
+    height?: number; // px, defaults to 512
+    center?: [number, number]; // coordinates, defaults to [0,0]
+    bearing?: number; // degrees, counter-clockwise from north, defaults to 0
+    pitch?: number; // degrees, arcing towards the horizon, defaults to 0
+    classes?: string[]; // array of strings
+  };
+
+  export class Map {
+    constructor(mapOptions: MapOptions);
+
+    load: (style: any) => void;
+
+    render: (
+      renderOptions: RenderOptions,
+      callback: (...args: [error: Error, buffer: undefined] | [error: undefined, buffer: Uint8Array]) => void,
+    ) => void;
+
+    release: () => void;
+  }
+}
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "lib"
   ],
   "main": "platform/node/index.js",
+  "types": "platform/node/index.d.ts",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/maplibre/maplibre-gl-native.git"


platform/node/index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  export class Map {
+    constructor(mapOptions: MapOptions);
+
+    load: (style: any) => void;
+
+    render: (
+      renderOptions: RenderOptions,
+      callback: (...args: [error: Error, buffer: undefined] | [error: undefined, buffer: Uint8Array]) => void,
```
KiwiKilian, line 34: I assume it's either error without buffer OR buffer without error – or could at some point both be available at the same time?

KiwiKilian:
For documentation purposes, I'm successfully using this augmentation like this:

```ts
render({ center, zoom, width, height }: { center: [number, number]; zoom: number; width: number; height: number }) {
    return new Promise((resolve, reject) => {
      const options: mbgl.MapOptions = {
        request(req, callback) {
          request(
            {
              url: req.url,
              encoding: null,
              gzip: true,
            },
            (error, response, body) => {
              if (error) {
                callback(error);
              } else if (response.statusCode === 200) {
                const result: mbgl.RequestResponse = { data: body, modified: undefined, expires: undefined };

                if (typeof response.headers.modified === 'string') {
                  result.modified = new Date(response.headers.modified);
                }
                if (response.headers.expires) {
                  result.expires = new Date(response.headers.expires);
                }
                if (response.headers.etag) {
                  result.etag = response.headers.etag;
                }

                callback(undefined, result);
              } else if (response.statusCode === 404) {
                // Some tiles might be missing...
                callback();
              } else {
                callback(new Error(JSON.parse(body).message));
              }
            },
          );
        },
      };

      const map = new mbgl.Map(options);
      map.load(style);

      map.render(
        {
          width,
          height,
          zoom,
          center,
        },
        (err, buffer) => {
          if (err) {
            reject(err);
          }

          map.release();

          const image = sharp(buffer, {
            raw: {
              width,
              height,
              channels: 4,
            },
          });

          resolve(image.toFormat('jpg').toBuffer());
        },
      );
    });
  }
  ```

**Comment by ovivoda** [comment]
@KiwiKilian can you please add a bit more context to the change? Why it is needed and how went wrong before?

**Comment by KiwiKilian** [comment]
@ovivoda Those typings improve the usage of the Node package in combintation with TypeScript or IDEs capable of supplying hints like JSDoc. Other Maplibre packages also supply those typings (e.g. [Maplibre GL JS](https://github.com/maplibre/maplibre-gl-js/blob/main/package.json#L13), https://github.com/maplibre/maplibre-gl-js/issues/32). I would consider it best practice, to ship TypeScript typings with the package, for this the `types` field in `package.json` is used.

File: platform/node/index.d.ts [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+  type MapOptions = {
+    request: (
+      request: { url: string; kind: Resource },
+      callback: (error?: Error, response?: RequestResponse) => void,
+    ) => void;
+    ratio?: number;
+  };
+
+  type RenderOptions = {
+    zoom?: number; // defaults to 0
```
HarelM, line 28: Please use jsdoc style comments, and also add comments to the classes and other types.

HarelM:
I'm not super familiar with the project, but can you make sure this new file will be picked up when using npm publish?

platform/node/index.d.ts
```diff
@@ -0,0 +1,49 @@
...
+    expires?: Date;
+    etag?: string;
+  };
+
+  type MapOptions = {
+    request: (
+      request: { url: string; kind: Resource },
+      callback: (error?: Error, response?: RequestResponse) => void,
+    ) => void;
+    ratio?: number;
```
acalcutt, line 24: shouldn't '[mode](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L1402-L1409)' also be in map options?

platform/node/index.d.ts
acalcutt: unfortunately no, I'm not familiar with typescript types. from https://github.com/maptiler/tileserver-gl/blob/master/src/serve_rendered.js#L1202-L1226 it seems like return data (not always an error)

platform/node/index.d.ts
KiwiKilian: I'm not really good with C – possible enum values can be `tile` or `static`?

KiwiKilian:
> I'm not super familiar with the project, but can you make sure this new file will be picked up when using npm publish?

Thanks for pointing out, I thought the `types` field would be enough to publish. I added the `index.d.ts` to `files`, now it should be published:
```
npm publish --dry-run
npm notice 
npm notice 📦  @maplibre/maplibre-gl-native@5.1.1
npm notice === Tarball Contents === 
npm notice 30.2kB LICENSE.mbgl-core.md    
npm notice 1.4kB  LICENSE.md              
npm notice 4.0kB  README.md               
npm notice 2.2kB  package.json            
npm notice 1.2kB  platform/node/index.d.ts
npm notice 1.8kB  platform/node/index.js  
npm notice 7.0kB  platform/node/README.md 
```

platform/node/index.d.ts
acalcutt: ya, setting mode to 'tile' puts it into tile mode. I think anything else does static mode, but I usually do put 'static'

KiwiKilian: What's the difference between the modes? Are their any docs/examples?

KiwiKilian: From a quick test I assume only a single tile will be fetched and rendered.

platform/node/index.d.ts
KiwiKilian: Thanks for this proposal. Would be awesome if you could have another look!

platform/node/index.d.ts
acalcutt: From what I have read here, the purpose of tile mode is to return a single tile. However, from what I learned in https://github.com/maplibre/maplibre-gl-native/issues/284 it seems to help with labels that cross tile borders. 

For example, in tile mode it uses more padding for collision index ( https://github.com/maplibre/maplibre-gl-native/blob/main/src/mbgl/text/collision_index.cpp#L26-L50 ), so it is able to consider labels from the next tiles (at least that's my understanding)

HarelM (approved):

HarelM:
LGTM in terms of syntax. Thanks!

acalcutt:
@KiwiKilian is there anything else you think this needs or do you feel this is ready? I'm not sure we had a good answer for you on the one unresolved conversation on callback.

KiwiKilian:
@acalcutt I think this is a good to go. The events listening is not typed yet, but I think this could be added at any later point. This lays out the base functionality to make the TypeScript dev experience much better.

etnav:
Just checked the usage of the typings on my end. Now only getting `render`, `release` and `load`, what about the rest, such as the ones below?

```ts
        static void New(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void Loaded(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void Cancel(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void AddSource(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void RemoveSource(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void AddLayer(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void RemoveLayer(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void AddImage(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void RemoveImage(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetLayerZoomRange(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetLayerProperty(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetFilter(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetCenter(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetZoom(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetBearing(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetPitch(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetLight(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetAxonometric(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetXSkew(const Nan::FunctionCallbackInfo<v8::Value>&);
        static void SetYSkew(const Nan::FunctionCallbackInfo<v8::Value>&);
```

Is something along the lines of:

```ts
setLayoutProperty: (layer: any, propertyName: string, propertyValue: string) => void;
```

correct?

KiwiKilian:
I don't know anything about those functions, but if they are available through the node binding, feel free to open another PR.

PR #767, opened 2023-02-03
Git bisect until maplibre-gl-js v1.14.0

louwers:
First making sure v1.14.0 is bad.

Then I will do a `git bisect`.


louwers:
I'm using [act](https://github.com/nektos/act) to run the workflows locally because the built went through for me on Arch Linux.

PR #786, opened 2023-02-08
Improve node docs with available platforms

KiwiKilian:


platform/node/README.md
```diff
@@ -1,26 +1,22 @@
...
 ## Installing
 
-#### :warn: This package isn't live yet.
-
-Requires a modern C++ runtime that supports C++14.
-
-By default, installs binaries. On these platforms no additional dependencies are needed.
+Binaries are available and downloaded during install for the following platforms:
 
 - 64 bit macOS or 64 bit Linux
```
acalcutt, line 10: I would change this to something like

`MacOS 12 (amd64/arm64) or Ubuntu 20.04 (amd64/arm64)`

KiwiKilian: Clarified. Also added a link to the debian instructions.

platform/node/README.md
```diff
@@ -1,26 +1,24 @@
...
+- Node.js 14, 16, 18
 
 Run:
 
 ```
 npm install @maplibre/maplibre-gl-native
 ```
 
-Other platforms will fall back to a source compile with `make node`; see [DEVELOPING.md](DEVELOPING.md) for details on
+Further platforms might work [with additional libraries installed](https://github.com/maplibre/maplibre-gl-native/tree/main/platform/linux#prerequisites). Otherwise, the package will fall back to a source compile with `make node`. See [DEVELOPING.md](DEVELOPING.md) for details on
```
acalcutt, line 21: right now there is no fallback to `make node`. I think there used to be based on my research for https://github.com/maplibre/maplibre-gl-native/pull/535 .

For now I would probably just remove 
`Otherwise, the package will fall back to a source compile with `make node``

acalcutt: removed in https://github.com/mapbox/mapbox-gl-native/pull/15748 fyi. My cmake-js PR https://github.com/maplibre/maplibre-gl-native/pull/535 adds something similar back.

acalcutt:
I think this looks better. 

One thing I am thinking about is the dependencies information. this mentions the dependencies for other platforms, but really even the platforms we support need a few things installed since we don't bundle dependencies into the binary like the mapbox version did.

Also, maybe a future change on the linux docs, it might be good to differentiate the 'packages needed to build' vs 'packages needed to run'. for example, in tileserver-gl, we are using the [non-dev versions of packages](https://github.com/maptiler/tileserver-gl/blob/master/Dockerfile#L66-L76) when running and it doesn't need build components.

acalcutt (approved):

PR #796, opened 2023-02-11
Doxygen-based documentation for C++ API

### LINKED ISSUE #759: Publish C++ API and inline documentation with Doxygen
The C++ API and inline documentation should be published to make it more easliy accessible.

Doxygen (still) seems to be king for C++, but if anyone knows any alternatives I am happy to take a look.

louwers:
Closes #759.

Please see the included `README.md` to try it out locally.

The Core of MapLibre GL Native is not a typical C++ library in that there's only a handful of platforms that use it. It's neat to have hosted C++ documentation, but I think that most platform authors will want to read the source code itself for the most part. So for now I think it's best to keep it simple. That is why I decided use a Doxygen-only setup, but with a bit more pleasant theme than the default one.

I had to change a lot of documentation comments, Doxygen supports [a variety of different styles](https://www.doxygen.nl/manual/docblocks.html). I just went with the minimal changes to get the documentation to appear, so I changed double slash comments into tripple slash `///` where applicable and `/**` into `/***` in other places. I think it is OK to use both styles, the second is a bit more readable for longer comments.

I also moved the `docs` directory in the root to `docs/mdbook` and added `docs/doxygen` so that the root directory of the project is not cluttered with more directories or files.

Once this has been merged I will create a PR to update maplibre.org.


birkskyum:
What is the part to generate and show the docs locally?

birkskyum:
Will it look like this?
https://jothepro.github.io/doxygen-awesome-css/index.html#autotoc_md8

src/mbgl/renderer/buckets/fill_extrusion_bucket.cpp
```diff
@@ -120,8 +120,9 @@ void FillExtrusionBucket::addFeature(const GeometryTileFeature& feature,
                     const auto d2 = convertPoint<double>(p2);
 
                     const Point<double> perp = util::unit(util::perp(d1 - d2));
-                    const auto dist = util::dist<int16_t>(d1, d2);
-                    if (edgeDistance + dist > std::numeric_limits<int16_t>::max()) {
+                    const size_t dist = util::dist<int16_t>(d1, d2);
+                    if (edgeDistance + dist >
+                        static_cast<size_t>(std::numeric_limits<int16_t>::max())) {
```
birkskyum, line 125: Is this related to doxygen?

louwers: The `core-build` failed. It is not related but I need to fix this to make the checks pass.

louwers:
@birkskyum 

```
cd docs/doxygen
doxygen
```

birkskyum:
And how do I apply the theme?

image

louwers:
> And how do I apply the theme?
> 
> image

`git submodule update --init --recursive`

birkskyum (approved):
Well done!

image

PR #804, opened 2023-02-15
Update node (ubuntu-20.04, arm64) docker build workflow

acalcutt:
This PR is an attempt to fix the node (ubuntu-20.04, arm64) docker build space issues.

1.) I've made some changes so the docker images the workflow creates get cleaned up when it finishes, whether it finishes successfully or not. I did the by giving each build a unique name, and then removing those images after the build is complete. I set the cleanup to use "always()" so even if steps fail, this cleanup still runs.

2.) I've updated 'Dockerfile_Build' so it no longer copies all the source code inside the docker image. It instead just copies 'docker_build_entrypoint.sh', and then does the build inside the mounted volume. since it builds inside the mounted /data/ volume, it does not add space inside the docker image.

3.) I've added in an alternative config to run this on an ubuntu-latest runner. if the runner is switched to that, the build will use qemu. This is mainly so it is easier to switch if needed, like when someone tries to run this on their own fork and doesn't have the mac-os arm runner (like me in my fork)

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -199,6 +199,9 @@ jobs:
         run: |
           brew install docker docker-compose
 
+      - name: Docker Prune
+        run: docker system prune --force
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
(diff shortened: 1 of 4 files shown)


acalcutt:
One thing I do notice about not copying all files into the docker image like I was before is it seems like the build is much slower, which is unfortunate. I wonder if it was working in memory before?

louwers:
Thanks for taking a look at this.

Maybe instead of the comment you can use:

```
${{ (github.action_repository == 'maplibre/maplibre-gl-native' && 'macos-12-arm') || 'ubuntu-latest' }}
```

(I hope the syntax is correct).

Then you don't have to change this file in a fork.

louwers:
You could set `node-build-${{ github.run_id }}-${{ github.run_attempt }}` to an environment variable at the top of the workflow so you don't have to repeat it.

Otherwise LGTM.

louwers (approved):

PR #821, opened 2023-02-18
Move module.cmake out of mapbox/cmake-node-module

acalcutt:
This PR moves module.cmake into a location we can control. It relates to https://github.com/maplibre/maplibre-gl-native/pull/707  which needs changes to support windows and had to suppliment it's own cmake file.

module.cmake is the only file we need form mapbox/cmake-node-module and I think it makes sense to remove dependency on this package. The original file seems to be ISC license on npm (unspecifed on github)

Note, this is also a precursor to changes I talked about with @tdcosta100 to try and make one file that works with all the platforms.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
       "dependencies": {
         "@acalcutt/node-pre-gyp": "^1.0.11",
         "@acalcutt/node-pre-gyp-github": "1.4.8",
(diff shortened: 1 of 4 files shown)


louwers (approved):
I did not look too closely at the CMake code.

But since we are already using it and it is unmaintained, makes a lot of sense to move it into the repo!

PR #826, opened 2023-02-19
Add mbgl-compiler-options to Node targets

tdcosta100:
Node targets didn't inherit global compiler settings defined in `mbgl-compiler-options`. This PR adds this target to every Node target. But doing so revealed a problem in `platform/node/src/node_map.cpp` caused by a unused parameter in a lambda function. So this PR also correct that, allowing Node targets to be compiled without any error.

### DIFF
--- a/platform/node/cmake/module.cmake
+++ b/platform/node/cmake/module.cmake
@@ -160,7 +160,7 @@ function(add_node_module NAME)
             )
         endif()
 
-        target_link_libraries(${_TARGET} PRIVATE ${NAME})
+        target_link_libraries(${_TARGET} PRIVATE ${NAME} mbgl-compiler-options)
 
         if(APPLE)
             # Ensures that linked symbols are loaded when the module is loaded instead of causing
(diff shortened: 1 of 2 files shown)


platform/node/cmake/module.cmake
```diff
@@ -160,7 +160,7 @@ function(add_node_module NAME)
             )
         endif()
 
-        target_link_libraries(${_TARGET} PRIVATE ${NAME})
+        target_link_libraries(${_TARGET} PRIVATE ${NAME} mbgl-compiler-options)
```
acalcutt, line 163: What is the benefit of passing this on?

tdcosta100: The main benefit is the consistency in compile flags being shared with all targets. This ensures, for example, the code requirements are the same in all source files. Without this option enabled, some things may pass in one target but not in another, like C4244 warnings (in current compille options it should not pass).

tdcosta100: `mbgl-compiler-options` target is defined here:

https://github.com/maplibre/maplibre-gl-native/blob/1341f33ba425b4115028a8e54d337c432eb46a33/CMakeLists.txt#L38

acalcutt: Thanks

acalcutt (approved):

PR #830, opened 2023-02-20
Fixed an issue with native code throwing an exception expecting bound…

artakka:
…s to be a float array

Fixed same issue for center

Now both work:
       val bounds = floatArrayOf(-180.0f, -85.05113f, 180.0f, 85.05113f)
        tileSet.setBounds(*bounds) // or like tileSet.setBounds(-180.0f, -85.05113f, 180.0f, 85.05113f) doesn't matter
        val otherBounds : Array<Float> = bounds.toTypedArray()
        tileSet.setBounds(otherBounds)


louwers:
@ovivoda I added a test and updated the changelog. Could you have a look and review this PR?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
```diff
@@ -200,12 +205,14 @@ class TileSet(val tilejson: String, vararg tiles: String) {
...
      * @param center the Float array to set
      */
+    @Suppress("unused")
     fun setCenter(@Size(value = 2) vararg center: Float) {
-        this.center = center
+        val latLng = LatLng(center[1].toDouble(), center[0].toDouble())
+        setCenter(latLng)
     }
 
     fun setCenter(center: LatLng) {
```
ovivoda, line 214: let's also add kotlin style comments/docs for this one as well

ovivoda (dismissed):
It looks good. Nothing major to change. Tx @louwers !

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
```diff
@@ -200,12 +205,14 @@ class TileSet(val tilejson: String, vararg tiles: String) {
      *
      * @param center the Float array to set
```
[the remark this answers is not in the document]
louwers: Why?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
louwers: There is no need to duplicate the documentation. The overloads will share the same documentation:

https://maplibre.org/maplibre-gl-native/android/api/-mapbox-g-l-android-s-d-k/com.mapbox.mapboxsdk.style.sources/-tile-set/index.html

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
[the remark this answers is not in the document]
ovivoda: just for the user of the method to know what to expect.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
ovivoda: This is true, but for consistency I rather add the documentation here as well. I think, not adding it will encourage future PR's to not add documentation and it will be hard to follow every time if the method is overload or not.

louwers: I disagree, since the behavior of both the overloads is the same. Duplicating the documentation will just make it harder to update and will only add noise. I think it will also show the documentation double, which is really not what we want in this case I think.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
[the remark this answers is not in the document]
louwers: Isn't that an implementation detail? Since a double is wider it should not affect the user.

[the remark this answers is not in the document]
artakka: Yes, it does not change anything for the users.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/TileSet.kt
artakka: Will add a comment. I think we should deprecate setCenter that take two floats. It is not strictly typed and very easy to confuse which parameter is which. It does take longitude first, which is very confusing.  @ovivoda @louwers what do you think about marking as deprecated and steering them toward using LatLng or shall I leave it as is?

artakka: I wish there was strongly typed alternative to setting the bounds too.

louwers: > what do you think about marking as deprecated and steering them toward using LatLng 

@artakka Sounds good to me!

artakka: @louwers @ovivoda Deprecated the unsafe setBounds and added a strongly typed alternative. Also added Doc comments to all of them.

Please check again.

louwers (approved):

Issue #825, opened 2023-02-19
Cannot access '<init>': it is package-private in 'LocationEngineProxy


ravenfeld:
Hello, I am trying to upgrade to version 10.0.0 but I have errors 

Cannot access '<init>': it is package-private in 'LocationEngineProxy

I can't use fused location from google which gives very good results I think and having the play service I would still like to use it but the migration example in the change log does not work.

Thank you in advance for your help

louwers:
Hmm I thought the visibility of constructors was public by default with Kotlin if the class is public.

I'll have a look. Just in case, could you clear the cache of your IDE?

PR #832, opened 2023-02-20
make LocationEngineProxy constructor public

jDilshodbek:
LocationEngineProxy constructor class access modifier changed to public in order to make custom LocationEngine from it


louwers:
Closes #825

louwers (approved):

Issue #322, opened 2022-06-09
Repository automation


wipfli:
This repository has automation scripts to build, test, and release the bindings for different platforms. It seems that these automation scripts come in a variety of languages such as:

- GitHub action workflows
- makefiles
- shell scripts
- python scripts
- javascript scripts

Maybe all of these implementations are needed for some technical reasons but maybe we could also remove one or the other and use a more unified approach to automation scripting.

What would be great in any case is if we could document a bit more what the different scripts do and how one can achieve certain tasks like updating the GL JS submodule for example. This is something that also gets mentioned in https://github.com/maplibre/maplibre-gl-native/blob/main/DEVELOPING.md

louwers:
We only use the Makefile as a convenient way to give easier to type commands to long Gradle commands. We don't use Makefiles for building.

There are only two Python scripts and it's not really worth it to rewrite them.

Shell scripts are sometimes more convenient than using a scripting language, especially when it's a short script that mainly ties different commands together.

For more complicated scripting I think we should default to JavaScript with JSDoc comments in the future:

- Most of our existing scripts (e.g. for shader code generation) are already written in JavaScript.
- Our community is much more familiar with JavaScript than Python thanks to MapLibre GL JS.
- I'd rather only add Node.js as a dependency for running scripts, and not TypeScript as well (although we can still use TypeScript for type checking on CI).

New scripts should be added to `/scripts` and have an `*.mjs` extension, then they will be automatically type checked on CI.

I'll put this on the agenda for the TSC Meeting today.

PR #833, opened 2023-02-20
Remove duplicated scripts

louwers:
Related ticket #322

PR to remove some duplicated scripts in `/platform/ios/scripts` that are also in `/scripts` . Most are byte for byte the same, some have minor differences. I did some checking, some of these might be referenced from somewhere under `ios`, but we will find out soon enough and a fix would be easy.


typebrook:
Upload my work, then I can use `user-images.githubusercontent.com` to add images into `README.md`

birkskyum:
@typebrook , I don't understand what's being referred to - can you elaborate?

typebrook:
>  I don't understand what's being referred to

Oh, I mean here I upload images to this issue, then permalinks are available for images in newly add markdown file.

Because I see #939 also use github as image gallery in this way:
https://github.com/maplibre/maplibre-native/blob/72a1f7864711ab4fa3fa59818b29ac4cb3d4b293/docs/mdbook/src/android/getting-started-guide.md?plain=1#L111-L113

PR #837, opened 2023-02-21
Enable Render test CI

alanchenboy:
Enable Render test CI


render-test/runner.cpp
```diff
@@ -629,7 +629,7 @@ TestOperations getAfterOperations(const Manifest& manifest) {
         }
         if (networkProbeOp == probe) {
             result.emplace_back([](TestContext& ctx) {
-                assert(ProxyFileSource::isTrackingActive());
+//                assert(ProxyFileSource::isTrackingActive());
```
louwers, line 632:
```suggestion
```

render-test/parser.cpp
```diff
@@ -1018,7 +1024,7 @@ TestOperations parseTestOperations(TestMetadata& metadata) {
         } else if (operationArray[0].GetString() == networkProbeEndOp) {
             // probeNetworkEnd
             result.emplace_back([](TestContext&) {
-                assert(mbgl::ProxyFileSource::isTrackingActive());
+//                assert(mbgl::ProxyFileSource::isTrackingActive());
```
louwers, line 1027:
```suggestion
```

render-test/parser.cpp
```diff
@@ -1007,7 +1013,7 @@ TestOperations parseTestOperations(TestMetadata& metadata) {
             assert(operationArray[1].IsString());
             std::string mark = std::string(operationArray[1].GetString(), operationArray[1].GetStringLength());
             result.emplace_back([mark](TestContext& ctx) {
-                assert(mbgl::ProxyFileSource::isTrackingActive());
+//                assert(mbgl::ProxyFileSource::isTrackingActive());
```
louwers, line 1016:
```suggestion
```

render-test/parser.cpp
```diff
@@ -997,7 +1003,7 @@ TestOperations parseTestOperations(TestMetadata& metadata) {
         } else if (operationArray[0].GetString() == networkProbeStartOp) {
             // probeNetworkStart
             result.emplace_back([](TestContext&) {
-                assert(!mbgl::ProxyFileSource::isTrackingActive());
+                //assert(!mbgl::ProxyFileSource::isTrackingActive());
```
louwers, line 1006:
```suggestion
```

louwers (changes requested):
- The commented out code should be removed.
- Could you document somewhere what `ignoreProbing` does? Maybe as a comment.
- Do you think the name of this workflow (node-ci) accurately reflects what it does? Because the render test that runs on macOS has little to do with Node.js, right?

render-test/runner.cpp
alanchenboy: Thanks for review，the comment is deleted

alanchenboy:
@louwers 
The commented out code was removed.
ignoreProbing has added a comment.
Render tests don't belong to the node.js, maybe we can move it to macOS-ci, right?

louwers (approved):
I am not 100% clear about the specifics of these render tests, but it looks good to me.

Let's move the render test to its own workflow in a separate PR.

Issue #310, opened 2022-06-07
Remove generated files from source control


wipfli:
I think generated files should not be part of git. Instead, we should have something like `npm run generate-stuff` which generates all the required files.

wipfli:
`scripts/nitpick/generated-code.js` probably tells us what code is generated:

```js
if (!mode || mode == 'sources') {
  checkGeneratedFiles('Source list', ['scripts/generate-file-lists.js']);
}
if (!mode || mode == 'shader') {
  checkGeneratedFiles('shader', ['scripts/generate-shaders.js']);
}
if (!mode || mode == 'style') {
  checkGeneratedFiles('style', ['scripts/generate-style-code.js']);
}
if (!mode || mode == 'android') {
  checkGeneratedFiles('Android', ['platform/android/scripts/generate-style-code.js']);
}
if ((!mode || mode == 'darwin') && os.platform() == 'darwin') {
  checkGeneratedFiles('Darwin', ['platform/darwin/scripts/generate-style-code.js', 'platform/darwin/scripts/update-examples.js']);
}
```

If we remove generated files from source control, we can also get rid of this nitpicking.

louwers:
@wipfli I am in favor of deleting generated files but only if they are generated automatically as part of a build step, and if it is not something that you have to run manually yourself.

louwers:
For reference:

> For now I'm going to go with running the generator manually when you make a change. I looked at including it as part of the build step but outputting to the src directory (which I want to do) didn't fit in cleanly with our use of gyp.

https://github.com/mapbox/mapbox-gl-native/pull/4839#issuecomment-215546791

louwers:
@wipfli The build system we use should be responsible for generating these files. It should also keep track of when dependencies change and automatically regenerate.

fynngodau:
I think it's okay to have the generated files in the source, but the version in the repository should always match the generated files exactly. Right now `pre-commit.ci` auto-reindents the files (diff: https://github.com/maplibre/maplibre-native/pull/1201/commits/d487434f919adee70b62713c4cfbabf518662593), so when `make style-code` is called locally afterwards, it undoes these changes again. Instead the CI should ensure that running `make style-code` leads to no changes.

louwers:
@fynngodau That is annoying. I'll make sure those files are not included.

PR #844, opened 2023-02-22
Remove files generated by `platform/ios/platform/darwin/scripts/generate-style-code.js`

louwers:
Related issue #310


louwers:
https://crascit.com/2017/04/18/generated-sources-in-cmake-builds/

platform/ios/Makefile
```diff
@@ -323,8 +330,8 @@ endif
 
 .PHONY: style-code
 style-code:
-	node scripts/generate-style-code.js
-	node scripts/generate-shaders.js
+	node ../../scripts/generate-style-code.js
+	# node ../../scripts/generate-shaders.js
```
birkskyum, line 334: how come its disabled?

louwers: I am getting an error when I run it, could you try it out?

louwers: Removing the generated shaders is also out of scope for this PR, but I will get to that next.

birkskyum (approved):

platform/ios/Makefile
birkskyum: Great

platform/ios/.gitignore
```diff
@@ -47,3 +47,54 @@ mapbox-gl-native.code-workspace
 /scripts/generate-shaders.list
 /scripts/generate-style-code.list
 *.patch
+
+
+# node platform/darwin/scripts/generate-style-code.js
+# generated file list will be in:
+# platform/ios/platform/darwin/scripts/generate-style-code.list
+
+platform/darwin/src/MGLLight.h
```
1ec5, line 56: This list would need to be updated any time the codegen scripts change, such as any time there’s a new layer type. It would be more maintainable to move these generated files into a generated/ subdirectory.

platform/ios/Makefile
```diff
@@ -182,7 +189,7 @@ ios-uitest: $(IOS_PROJ_PATH)
 	echo "You can review uitest results:  open $(IOS_OUTPUT_PATH)/Logs/Test"
 
 .PHONY: ios-test
-ios-test: $(IOS_PROJ_PATH)
+ios-test: $(IOS_PROJ_PATH) style-code
```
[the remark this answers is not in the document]
1ec5: This also needs to be added to `document`. Otherwise, the generated jazzy docset will omit all the runtime styling classes that developers should have access to.

PR #859, opened 2023-02-26
Remove reference to qmake

louwers:
@ntadej qmake is not used anymore, correct?


ntadej:
Correct! Thanks for spotting this!

ARCHITECTURE.md
```diff
@@ -76,7 +76,7 @@ Outside of the core C++ static library, platform SDKs typically rely on platform
 
 * For iOS and macOS this means Xcode and the xcodebuild command line tool.
 * For Android, Gradle and Android Studio.
-* For Qt, `qmake`.
+* For Qt, plain CMake is used to generate the build system.
```
ntadej, line 79: Maybe the wording is not ideal, CMake just generates the makefiles here.

ntadej (changes requested):

ARCHITECTURE.md
louwers: In that case the Makefiles are the build system, right?

louwers:
> gyp is a build system originally created for Chromium and since adopted by Node.js. In MapLibre GL Native it's used to build the core C++ static library, the Node.js platform module, and the shared JNI module for Android.

This is also outdated, right?

acalcutt:
> > gyp is a build system originally created for Chromium and since adopted by Node.js. In MapLibre GL Native it's used to build the core C++ static library, the Node.js platform module, and the shared JNI module for Android.
> 
> This is also outdated, right?

We don't use gyp build system, but we do still use node-pre-gyp for the node binaries. Which reminds me, on https://github.com/maplibre/maplibre-gl-native/blob/main/package.json#L60 , fallback-to-build should be false. This is something I changed early on in getting node process working, but we don't have things set up to build that way,

ARCHITECTURE.md
louwers: @ntadej Could you suggest a different wording?

ntadej: Let's just keep it as is.

ntadej (approved):

PR #867, opened 2023-02-27
First step towards DocC iOS docs

### LINKED ISSUE #676: Evaluate using DocC for iOS documentation
The Apple recommended way to provide documentation is [DocC](https://developer.apple.com/documentation/docc).

Worth taking a look at I think, especially since the iOS docs are in need of a refresh anyway.

louwers:
Closes #676

After this is merged I will create some separate issues to polish the documentation comments throughout the codebase to take avail of DocC specific features (e.g. cross-linking). We also need to move the guides that are part of the Jazzy-based doc site. Before 'going live' and publicly advertising the DocC based docs we 

image

Something that still needs to be done / figured out is creating a Document Catalogue and making sure DocC picks it up. This can be used to add the additional articles and tutorials, as well as for extensions that for example set a display name of the framework (so it does not show as Mapbox).


roblabs:
How MapLibre DocC looks from within Developer Documentation when importing the output file `build/ios/.../Mapbox.doccarchive`

roblabs (approved):

PR #868, opened 2023-02-28
put back docker process changed in #804

acalcutt:
Fixes https://github.com/maplibre/maplibre-gl-native/issues/860 . This is needed to publish a new node release, since the current workflow is failing to publish the arm64 binary.

In #804 I tried to change the linux arm64 docker build process so it worked inside the current directory instead of copying files inside the docker image. This worked fine in the ci tests, but in the real publish it adds extra step to install node-pre-gyp and publish. Unfortunately, when it tries to does "npm ci" to install node-pre-gyp it fails due to the node_modules directory already existing from the docker build (which is for a different platform). 

At first I tried several ideas to remove the node_modules folder to get around this. Trying to remove it in the workflow fails with permission issues. trying to remove it in the entrypoint file kind of worked, but then the publish still failed due to a permission issues.

This PR reverts changes in  https://github.com/maplibre/maplibre-gl-native/pull/804 which made the build happen in the current directory. It still keeps some changes, like running in linux and qemu, which has been helping this workflow run successfully.

I've tested this change on my own fork at https://github.com/WifiDB/maplibre-gl-native/actions/runs/4284198145/jobs/7460830106

### DIFF
--- a/platform/node/scripts/Dockerfile_Build
+++ b/platform/node/scripts/Dockerfile_Build
@@ -27,6 +27,6 @@ RUN set -ex; \
 VOLUME /data
 
 RUN mkdir -p /usr/src/app
-COPY /platform/node/scripts/docker_build_entrypoint.sh /usr/src/app/docker_build_entrypoint.sh
-ENTRYPOINT ["/usr/src/app/docker_build_entrypoint.sh"]
-RUN ["chmod", "+x", "/usr/src/app/docker_build_entrypoint.sh"]
+COPY / /usr/src/app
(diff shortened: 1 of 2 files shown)


louwers:
Hello @acalcutt, shall I instead set up a self-hosted ARM64 runner?

acalcutt:
I would love that, but can we merge this first so i can finish the node pre-release I tried to do the other day.

acalcutt:
If we do set up a runner do we have a choice of OS? the other linux runner is running ubunto 20.04 (but eventually it may be food to bump to 22.04)

louwers:
> If we do set up a runner do we have a choice of OS?

Yes.

PR #871, opened 2023-03-02
Adjust Typings for Node Platform

etnav:
Adjusted the existing typings and expanded it with some new functions.

### DIFF
--- a/platform/node/index.d.ts
+++ b/platform/node/index.d.ts
@@ -60,6 +60,8 @@ declare module '@maplibre/maplibre-gl-native' {
    */
   type RenderOptions = {
     /**
+     * Zoom level
+     *
      * @default 0
      */
     zoom?: number;
@@ -111,6 +113,9 @@ declare module '@maplibre/maplibre-gl-native' {
   class Map {
     constructor(mapOptions: MapOptions);
 
+    /**
+     * Load a style into a map
+     */
     load: (style: any) => void;
 
     /**
@@ -125,5 +130,90 @@ declare module '@maplibre/maplibre-gl-native' {
      * Call to permanently dispose the internal map resources, instance can't be used for further render calls
      */
     release: () => void;
+
+    /**
+     * Add source to map's style
+     */
+    addSource: (sourceId: string, source: {}) => void
+
+    /**
+     * Remove source from map's style
+     */
+    removeSource: (sourceId: string) => void
+
+    /**
+     * Add layer to map's style
+     */
+    addLayer: (layer: {}, beforeId?: string) => void
+
+    /**
+     * Remove layer from map's style
+     */
+    removeLayer: (layerId: string) => void
+
+    /**
+     * Add image to map's style
+     */
+    addImage: (imageId: string, image: any) => void
+
+    /**
+     * Remove image from map's style
+     */
+    removeImage: (imageId: string) => void
+
+    /**
+     * Set the extent of the zoom for a specified layer
+     */
+    setLayerZoomRange: (layerId: string, minzoom: number, maxzoom: number) => void;
+
+    /**
+     * Set the value for a layer's property
+     */
+    setLayoutProperty: (layerId: string, name: string, value: string) => void;
+
+    /**
+     * Set filter for specified style layer
+     */
+    setFilter: (layerId: string, filter: [] | null | undefined) => void;
+
+    /**
+     * Set the center of the map
+     */
+    setCenter: (center: any) => void;
+
+    /**
+     * Set zoom of the map
+     */
+    setZoom: (zoom: number) => void;
+
+    /**
+     * Set bearing (rotation) of map
+     */
+    setBearing: (bearing: number) => void;
+
+    /**
+     * Set pitch (tilt angle) of map
+     */
+    setPitch: (pitch: number) => void;
+
+    /**
+     * Set light value of map
+     */
+    setLight: (light: any) => void
+
+    /**
+     * Set axonometric view of map
+     */
+    setAxonometric: (state: boolean) => void
+
+    /**
+     * Set X skew of map
+     */
+    setXSkew: (x: number) => void
+
+    /**
+     * Set Y skew of map
+     */
+    setYSkew: (y: number) => void
   }
 }


acalcutt:
@KiwiKilian , since you made the original typings PR https://github.com/maplibre/maplibre-gl-native/pull/766, I was wondering what you think of these changes.

Layout wise this looks good to me I think.

louwers:
Thanks for your contribution! @etnav 🚀

platform/node/index.d.ts
```diff
@@ -125,5 +130,90 @@ declare module '@maplibre/maplibre-gl-native' {
      * Call to permanently dispose the internal map resources, instance can't be used for further render calls
      */
     release: () => void;
+
+    /**
+     * Add source to map's style
+     */
+    addSource: (sourceId: string, source: {}) => void
```
KiwiKilian, line 137:
```suggestion
    addSource: (sourceId: string, source: object) => void
```

See the [`object` definition](https://www.typescriptlang.org/docs/handbook/2/functions.html#object):

> The special type `object` refers to any value that isn’t a primitive (`string`, `number`, `bigint`, `boolean`, `symbol`, `null`, or `undefined`). This is different from the empty object type `{ }`, and also different from the global type `Object`. It’s very likely you will never use `Object`.

Or would any of those other primitives be desired? Then `any` would be better readable?

platform/node/index.d.ts
```diff
@@ -125,5 +130,90 @@ declare module '@maplibre/maplibre-gl-native' {
...
+
+    /**
+     * Remove source from map's style
+     */
+    removeSource: (sourceId: string) => void
+
+    /**
+     * Add layer to map's style
+     */
+    addLayer: (layer: {}, beforeId?: string) => void
```
KiwiKilian, line 147:
```suggestion
    addLayer: (layer: object, beforeId?: string) => void
```

platform/node/index.d.ts
```diff
@@ -125,5 +130,90 @@ declare module '@maplibre/maplibre-gl-native' {
...
+
+    /**
+     * Remove image from map's style
+     */
+    removeImage: (imageId: string) => void
+
+    /**
+     * Set the extent of the zoom for a specified layer
+     */
+    setLayerZoomRange: (layerId: string, minzoom: number, maxzoom: number) => void;
```
KiwiKilian, line 167:
```suggestion
    setLayerZoomRange: (layerId: string, minZoom: number, maxZoom: number) => void;
```

platform/node/index.d.ts
```diff
@@ -125,5 +130,90 @@ declare module '@maplibre/maplibre-gl-native' {
...
+
+    /**
+     * Set filter for specified style layer
+     */
+    setFilter: (layerId: string, filter: [] | null | undefined) => void;
+
+    /**
+     * Set the center of the map
+     */
+    setCenter: (center: any) => void;
```
KiwiKilian, line 182:
```suggestion
    setCenter: (center: [number, number]) => void;
```

https://github.com/maplibre/maplibre-gl-native/blob/02c196a4721ba0c0a72a60597854f2bf26fe64a7/render-test/parser.cpp#L768-L771

KiwiKilian (changes requested):
Looks good! Just some small recommendations from my side.

I struggle to call these functions to validate if it works. For example, when to call `setBearing`, `setPitch` or `setCenter`? My best guess was to call between `map.load` and `map.render`, but I can't seem to make a difference in the rendered output image. Do you have an example/use case where those functions are used?

KiwiKilian:
@etnav thanks for applying the improvements. Could you provide some usage example?

> I struggle to call these functions to validate if it works. For example, when to call setBearing, setPitch or setCenter? My best guess was to call between map.load and map.render, but I can't seem to make a difference in the rendered output image. Do you have an example/use case where those functions are used?

acalcutt:
> @etnav thanks for applying the improvements. Could you provide some usage example?
> 
> > I struggle to call these functions to validate if it works. For example, when to call setBearing, setPitch or setCenter? My best guess was to call between map.load and map.render, but I can't seem to make a difference in the rendered output image. Do you have an example/use case where those functions are used?
> 
> If it helps, here is a minimal usage example repo: https://github.com/KiwiKilian/maplibre-gl-native-node-example/blob/main/index.ts

I was trying myself to see if I could get any of these functions to work with https://github.com/acalcutt/maplibre-node-test/blob/main/maplibre-node-test/app.js#L46-L47 , but I had no luck getting them to do anything there either.

Not sure I really get the need, wouldn't you have to call render, which has most of these options already?
https://github.com/maplibre/maplibre-gl-native/tree/main/platform/node#rendering-a-map-tile

etnav:
> @etnav thanks for applying the improvements. Could you provide some usage example?
> 
> > I struggle to call these functions to validate if it works. For example, when to call setBearing, setPitch or setCenter? My best guess was to call between map.load and map.render, but I can't seem to make a difference in the rendered output image. Do you have an example/use case where those functions are used?
> 
> If it helps, here is a minimal usage example repo: https://github.com/KiwiKilian/maplibre-gl-native-node-example/blob/main/index.ts

@KiwiKilian @acalcutt If it helps, I have only used `map.setLayoutProperty(layerId, "visibility", "visible")` to change a property of a specific layer, and that works. I figured that adding all of the functions would make sense also, but it seems that they do absolutely nothing indeed. Maybe someone with more knowledge could shed some light here?

acalcutt:
I'm not really a c++ programmer, but to me it looks like render (center, zoom, bearing pitch) always gets [set based on RenderOptions](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L463-L471)

However, when you call SetCenter/SetZoom it doesn't update [RenderOptions](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L37-L50), instead it tries to set the [Camera directrly](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L943), which I think later gets overwitten by whats in RenderOptions when render runs.

Sees like it would be pretty simple to update RenderOptions when SetCenter/SetZoom gets called, kind of like how those options get updates when you supply options to render https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L304-L314

KiwiKilian:
> Seems like it would be pretty simple to update RenderOptions when SetCenter/SetZoom gets called, kind of like how those options get updated when you [supply options to render](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L304-L314)

Question would be, what's the benefit of setting those render options another way?

acalcutt:
I was talking to @tdcosta100 about this yesterday and he found a solution to make this work.

If I understand correctly, mbgl core has two methods for render. One version takes camera options, and the [other takes no options](https://github.com/maplibre/maplibre-gl-native/blob/main/include/mbgl/map/map.hpp#L46). In the node version, only the render that takes camera options was implemented, so like I found this overwrites the view settings.

@tdcosta100 was able to implement the version of render that takes no options, With the second method available, you can call a render with no options, and it just renders whatever view is showing. This allows it to work with the additional functions we couldn't get to work.

Right now you can see the changes he made here, but we should have a PR to discuss it more soon 
https://github.com/maplibre/maplibre-gl-native/compare/main...tdcosta100:maplibre-gl-native:mbgl-node-enhancements

etnav:
> I was talking to @tdcosta100 about this yesterday and he found a solution to make those functions work.
> 
> If I understand correctly, mbgl core has two methods for render. One version takes camera options, and the [other takes no options](https://github.com/maplibre/maplibre-gl-native/blob/main/include/mbgl/map/map.hpp#L46). In the node version, only the render that takes camera options was implemented, so like I found this overwrites the view settings.
> 
> @tdcosta100 was able to implement the version of render that takes no options, With the second method available, you can call a render with no options, and it just renders whatever view is showing. This allows it to work with the additional functions we couldn't get to work here.
> 
> Right now you can see the changes he made here, but we should have a PR to discuss it more soon [main...tdcosta100:maplibre-gl-native:mbgl-node-enhancements](https://github.com/maplibre/maplibre-gl-native/compare/main...tdcosta100:maplibre-gl-native:mbgl-node-enhancements)

Sounds and looks good. Interesting to see what is actually going on in the background. 

What do you think about the typings changes regarding this PR, do we wait for the new PR or implement only the working parts now?

tdcosta100:
PR #891 is alive, and if you accept my suggestion: I think all mapped methods can be enriched by their documentation here:

https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp

acalcutt:
I merged https://github.com/maplibre/maplibre-gl-native/pull/891 . Could Typing for SetSize, and possibly a new one for render without the RenderOptions be added?

etnav:
> I merged #891 . Could Typing for SetSize, and possibly a new one for render without the RenderOptions be added?

Great, I added `SetSize` and `render` without RenderOptions. I tested it briefly on my end and it worked with both `render` calls. Would be great if someone could also test it.

acalcutt:
I Tested here in Visual Studio 2022 Community and this all looks nice there

image
image
image
image
image
image

acalcutt (approved):

acalcutt:
Before this is merged, can you add a entry to the node changelog, under 5.2.0
https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/CHANGELOG.md

Once this is merged I will do another 5.2.0 pre-release so we can test this all together better.

PR #885, opened 2023-03-05
Update missing history in node CHANGELOG.md

acalcutt:
Add missing history to the node changelog based on commits.

### DIFF
--- a/platform/node/CHANGELOG.md
+++ b/platform/node/CHANGELOG.md
@@ -2,8 +2,9 @@
 # main
 
 ### ✨ New features
-
 - *...Add new stuff here...*
+
+# 5.2.0
 * Move node ci+release to self hosted Ubuntu arm64 by @acalcutt in https://github.com/maplibre/maplibre-gl-native/pull/873
 * Add windows support by @tdcosta100 in https://github.com/maplibre/maplibre-gl-native/pull/707
 * Add Typings for Node Platform by @KiwiKilian in https://github.com/maplibre/maplibre-gl-native/pull/766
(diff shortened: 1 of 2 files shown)


ovivoda (approved):

louwers (changes requested):
The CI is not relevant for end-users of the library, should it be in the changelog?

acalcutt:
> The CI is not relevant for end-users of the library, should it be in the changelog?

A lot of the CI is binary file generation for the npm package, so I debated that a bit. Early in the pre 5.0.1 version that was the main changes happening (workflow and new binary support, which affected who could use the package)

louwers (approved):

ntadej:
> > The CI is not relevant for end-users of the library, should it be in the changelog?
> 
> A lot of the workflow is binary file generation for the npm package, so I debated that a bit. Early in the pre 5.0.1 version that was the main changes happening (workflow and new binary support, which affected who could use the package). I tried to bundle the workflow changes together under one entry when I could so they weren't too overbearing.

Just to add that we may want to move away from purely MR-based changelog. For end-users all build-related changes could be summarised in one bullet.

But we can try that for the next release.

acalcutt:
Right now the changelog and change on release for node needs work. Right now, node-pre-gyp-github makes the node release with no release notes at all. I have been manually going and trying to add them in, which I someday hope to find something better for. I think starting to log better in this CHANGELOG would be a good start. I have accidently been directing users to update the changelog in the root directory, but that seems to be wrong (it is maps related)

PR #888, opened 2023-03-07
Fix Makefile `style-code` target

louwers:
Now the `style-code` target really will install the node modules in the root.


ovivoda (approved):

louwers:
@ovivoda Do you know why `make ios-test` does not run `style-code`?

PR #891, opened 2023-03-08
Node platform improvements

tdcosta100:
This PR brings some improvements to the Node platform:

* Allows calling Map.render without an options object (just with a callback function), mapping to the [`void mbgl::Map::renderStill(StillImageCallback)`](https://github.com/maplibre/maplibre-gl-native/blob/main/include/mbgl/map/map.hpp#L46) method
* Implements Map.setSize, which was missing
* Changes the default map size from 256x256 to 512x512 (which is the default size elsewhere, including  [`NodeMap::RenderOptions`](https://github.com/tdcosta100/maplibre-gl-native/blob/f8c1fd2194d92b284ce7781c76868066e27901b3/platform/node/src/node_map.cpp#L44))

Bonus: a little addition to the code example at README.md when the status code is 204 (empty response).

### DIFF
--- a/platform/node/CHANGELOG.md
+++ b/platform/node/CHANGELOG.md
@@ -5,6 +5,7 @@
 - *...Add new stuff here...*
 
 # 5.2.0
+* Node platform improvements (added setSize and a new render call without render options object) by @tdcosta100 in https://github.com/maplibre/maplibre-gl-native/pull/891
 * Move node ci+release to self hosted Ubuntu arm64 by @acalcutt in https://github.com/maplibre/maplibre-gl-native/pull/873
 * Add windows support by @tdcosta100 in https://github.com/maplibre/maplibre-gl-native/pull/707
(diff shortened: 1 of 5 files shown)


acalcutt (approved):
This help with several functions exposed by https://github.com/maplibre/maplibre-gl-native/pull/871 , like SetCenter, SetZoom, SetBearing, not working because because there was no way to call render without changing the view. I think this is an improvement and consistent with other plaform syntax.

I have worked with @tdcosta100 in testing this and it seems to work as expected, the SetCenter, SetZoom, SetBearing functions can now be used.

Tests have been added that show this syntax works.

acalcutt:
@tdcosta100 before we merge this can you add a entry to the readme 

https://github.com/maplibre/maplibre-gl-native/blob/node-v5.2.0-pre.0/platform/node/CHANGELOG.md

PR #899, opened 2023-03-14
Trying out C++20

louwers:


louwers:
Nice https://developer.apple.com/forums/thread/706486

ntadej:
The tests confirm what I mentioned in the meeting. While for mobile devices we may be OK, the desktop platforms oldest supported versions do not support C++20.

louwers:
@ntadej Android looks fine, for iOS and macOS the Objective-C++ looks broken, for Qt it might just be that the CMake version is outdated.

PR #904, opened 2023-03-16
Make Node Map object options "request" property optional

tdcosta100:
The current Node Map requires you to implement a `request` function, so everything Node Map needs, it asks to this function. This PR allows you to choose if you really want this, or if it's ok to let Node Map use its internal loaders.


acalcutt:
Whats the difference between the internal loaders and the one you implement youself? Would it work with http and local files or would it depend what you wanted to use this with if you need to implement your own.

tdcosta100:
The internal loaders implement their own logic, you can learn about them [here](https://github.com/maplibre/maplibre-gl-native/tree/main/platform/default/src/mbgl/storage). But they work well for the basics. There is a provider for local files (`local_file_source.cpp`), or online files (`http_file_source.cpp`), and a dedicated provider for MBTiles as well. So unless they don't do something specific you need, they are a good fit, so there's no point in reimplementing something they already do, like dealing with all those HTTP status codes and stuff. But of course, if your logic is more complex, like doing specific things depending on the resource requested, you still can implement your request function.

tdcosta100:
An addendum: if you implement your request function, then the internal loaders are disabled and all requests are sent to your function, exactly like it was before this PR.

louwers (changes requested):
Typescript `index.d.ts` needs to be updated to reflect this.

tdcosta100:
Done!

acalcutt (approved):
I tested this and it seems to work. I was able to use options like was required in the past, or go without the options all together.

Can you add an entry to the node readme for this.

tdcosta100:
Nice. Please check if the `README.md` changes are reflecting all those changes.

PR #905, opened 2023-03-16
Add Windows support to MBTiles file source

tdcosta100:
MBTiles file source requires absolute URIs using the scheme `mbtiles://`. In Linux it works with something like `mbtiles:///home/user/map/tiles.mbtiles`, but in Windows it would be something like `mbtiles:///C:/Users/user/map/tiles.mbtiles` (local files) or `mbtiles:////computer/share/map/tiles.mbtiles` (UNC paths), but it does not work, because the path does not start with a slash. This PR aims to correct this situation allowing validation of Windows absolute paths.

### DIFF
--- a/platform/android/android.cmake
+++ b/platform/android/android.cmake
@@ -66,6 +66,7 @@ target_sources(
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/storage/sqlite3.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/text/bidi.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/compression.cpp
+        ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/filesystem.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/monotonic_timer.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/png_writer.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/thread_local.cpp
--- a/platform/default/BUILD.bazel
+++ b/platform/default/BUILD.bazel
@@ -50,6 +50,7 @@ cc_library(
         "src/mbgl/storage/sqlite3.cpp",
         "src/mbgl/text/bidi.cpp",
         "src/mbgl/util/compression.cpp",
+        "src/mbgl/util/filesystem.cpp",
         "src/mbgl/util/monotonic_timer.cpp",
         "src/mbgl/util/png_writer.cpp",
         "src/mbgl/util/thread_local.cpp",
--- a/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
+++ b/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
@@ -14,6 +14,7 @@
 #include <mbgl/util/url.hpp>
 #include <mbgl/util/chrono.hpp>
 #include <mbgl/util/compression.hpp>
+#include <mbgl/util/filesystem.hpp>
 
 #include <mbgl/storage/sqlite3.hpp>
 
@@ -283,7 +284,8 @@ std::unique_ptr<AsyncRequest> MBTilesFileSource::request(const Resource &resourc
         return req;
     }
 
-    if (resource.url.find(":///") == std::string::npos) {
+    if (resource.url.find("://") == std::string::npos ||
+        !util::is_absolute_path(resource.url.substr(resource.url.find("://") + 3))) {
         Response response;
         response.noContent = true;
         response.error = std::make_unique<Response::Error>(Response::Error::Reason::Other,
--- a/platform/default/src/mbgl/util/filesystem.cpp
+++ b/platform/default/src/mbgl/util/filesystem.cpp
@@ -1,0 +1,13 @@
+#include <mbgl/util/filesystem.hpp>
+
+#if defined(USE_STD_FILESYSTEM)
+#include <filesystem>
+
+bool mbgl::util::is_absolute_path(std::string path) {
+    return std::filesystem::path(path).is_absolute();
+}
+#else
+bool mbgl::util::is_absolute_path(std::string path) {
+    return path.at(0) == '/';
+}
+#endif
--- a/platform/ios/ios.cmake
+++ b/platform/ios/ios.cmake
@@ -73,6 +73,7 @@ list(APPEND
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/storage/sqlite3.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/text/bidi.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/compression.cpp
+        ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/filesystem.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/monotonic_timer.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/png_writer.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/thread_local.cpp
--- a/platform/linux/linux.cmake
+++ b/platform/linux/linux.cmake
@@ -42,6 +42,7 @@ target_sources(
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/text/local_glyph_rasterizer.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/async_task.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/compression.cpp
+        ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/filesystem.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/image.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/jpeg_reader.cpp
         ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/util/logging_stderr.cpp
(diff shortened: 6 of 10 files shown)


louwers (changes requested):
Please extract this logic in a function in `platform/default` which is overriden in `platform/windows` similarly as for other platform-specific implementations.

Edit: I looked into the Windows PR now and I realize quite a few platform-specific `#define` statements were included with that as well. As I see it we should avoid this as much as possible going forward and keep the core as platform-agnostic as possible.

tdcosta100:
Okay, I can make a `mbtiles_file_source.cpp` for Windows. It's a little overkill, since my only change is the validation of the URL, the rest of the code will remain untouched. But like you said, doing this will allow it to be platform-agnostic.

tdcosta100:
I think yes, the code can be reworked to extract this as a function. Something like `is_absolute_uri(const std::string&)`.

It checks against the resource URL to guarantee the path to the .mbtiles file is absolute. For example, in the URI `mbtiles:///home/user/tiles.mbtiles`, the part of the path is `/home/user/tiles.mbtiles`. This is an absolute path. In other hand, the URI `mbtiles://user/tiles.mbtiles` have the path `user/tiles.mbtiles`, that is, a relative path. Checking against `:///` ensures the path starts with `/`, so it's an absolute path.

In Windows, absolute paths must have a drive letter (C:, D:, X:, etc), or be UNC paths (starting with `\\` or `//`). So we cannot check against `:///`, because we must accept something like `mbtiles://C:/Users/user/tiles.mbtiles`.

louwers:
That sounds good!

tdcosta100:
The only problem is: even if we extract a function to check if the URI contains an absolute path, where can I override it in Windows platform? This class doesn't have a `Impl` internal class, unfortunately. But for sure extracting this as a method is the best way, I will commit this change.

tdcosta100:
Okay, forget what I said. It has a `Impl` class, so I can customize it.

tdcosta100:
So I moved this piece of code to an external function. But how to implement it in Windows? Should I keep it as a prototype, then I implement it in every platform? I need guidance in this refactor.

tdcosta100:
I think we could use `[std::filesystem::path::is_absolute()](https://en.cppreference.com/w/cpp/filesystem/path/is_absrel)`, but only if we use C++17. Is this our case?

louwers:
> I think we could use `[std::filesystem::path::is_absolute()](https://en.cppreference.com/w/cpp/filesystem/path/is_absrel)`, but only if we use C++17. Is this our case?

Yes it is!

tdcosta100:
Nice. This is already a platform-indepent way to do things, so it will be a nice approach.

louwers:
> Nice. This is already a platform-indepent way to do things, so it will be a nice approach.

Neat. Let's use it then.

tdcosta100:
Done, hope the tests will pass. I think it's even nicer than before.

tdcosta100:
Oh no, apparently C++17 is not fully supported in MacOS builds, unless you put a `-std=c++17` switch when configuring. I will investigate further and do a separate PR to test the workflows with this option.

louwers:
> Oh no, apparently C++17 is not fully supported in MacOS builds, unless you put a `-std=c++17` switch when configuring. I will investigate further and do a separate PR to test the workflows with this option.

Let me fix that.

tdcosta100:
Thanks! I'm really trying where to change that, but I don't have the knowledge to do that yet.

louwers:
> Thanks! I'm really trying where to change that, but I don't have the knowledge to do that yet.

We are already compiling with the C++17 flag, but this particular function has only been implemented on macOS 10.15 and iOS 13.0.

Since we want to support iOS 12.0 for the time being, I suggest the following mitigation:

```
// remove the following #ifdef once iOS 13.0+ is set as deployment target
// https://github.com/maplibre/maplibre-gl-native/pull/905
#ifdef __APPLE__
old if statement
#elif
new if statement
#endif
```

This is acceptable in my eyes because it is temporary.

tdcosta100:
Okay, I can do that.

louwers:
I *think* the header is available even though the functions aren't.

tdcosta100:
Nice, it will make the code cleaner.

louwers:
Is it OK if we await the result of the following discussion?

https://github.com/maplibre/maplibre-gl-native/discussions/915

tdcosta100:
Yes, I think is good to wait for the #915 discussion. If it doesn't last too much, of course hahaha.

tdcosta100:
It's a reasonable time, but I just remembered I can skip QT as well, so I will keep it out of this change (using the `__QT__` define check).

tdcosta100:
Yay, everything passed. I also tested skip just Qt 5 and allow this change on Qt 6, but a lot of errors succeeded, so I will keep it only in other platforms (except, of course, Apple).

louwers:
> Yay, everything passed. I also tested skip just Qt 5 and allow this change on Qt 6, but a lot of errors succeeded, so I will keep it only in other platforms (except, of course, Apple).

OK, but I assume an MBTiles file source on Qt + Windows will not work then?

tdcosta100:
> > Yay, everything passed. I also tested skip just Qt 5 and allow this change on Qt 6, but a lot of errors succeeded, so I will keep it only in other platforms (except, of course, Apple).
> 
> OK, but I assume an MBTiles file source on Qt + Windows will not work then? Do you know which Qt build was giving errors?
> 
> One more annoying question: can we add a test?

At most, it will work as always have been since it was implemented, I don't know how it was working before. You can see the failed workflow [here](https://github.com/maplibre/maplibre-gl-native/actions/runs/4452955656/jobs/7821074283). The most strange part is: it has the `-std=c++17` switch, but apparently this is not enough. Maybe it is using an outdated compiler, something like that. But it needs further investigation, I can try to reproduce it in my computer, if I know how to download Qt without installing it. I think I can do that, just need to try it.

Regarding the test: if it doesn't exist yet, we can add one, I just need to understand where these test types are.

tdcosta100:
Okay, so I studied the Qt buildsystem a little more, and managed to enable `#include <filesystem>` in Qt >= 6.2.2 (which is the first version you can use MinGW >= 9.1.0, versions below that doesn't support `filesystem` well).

But this doesn't mean you cannot use absolute paths in Windows in versions below 6.2.2. You can use [DOS device paths](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#dos-device-paths) to achieve that. They work ok, you just need to use forward slashes (`//`) instead backslashes (`\\`). I tested with Qt 5.15.2 and Qt 6.4.2, and both worked (6.4.2 allows using standard Windows absolute paths, in addition to DOS device paths)

tdcosta100:
Added a use to `std::filesystem::temp_directory_path()` in Qt test app, since it defines the temporary database path as a hardcoded path. This will be future-proof, since it leaves to the platform implementation to decide where the temporary directory is.

tdcosta100:
@louwers, I think it's better now. I dropped the changes for the Qt test app, since I'm not touching Qt anymore.

louwers (changes requested):
@tdcosta100 Duplicated code is a no go in my book. My idea was to just create a new file for just this (maybe platform specific) is absolute path function, sorry if I was being unclear. If it works for you we can schedule a 30 minute call and get this sorted: https://calendly.com/bart-louwers/30min

tdcosta100:
I think we can think how to rework `MBTilesFileSource` class. Maybe I can contact you in Slack? I think it's better to communicate with.

tdcosta100:
@louwers, please check if doing this way makes it better. I can, of course, move the entire `#if defined(USE_STD_FILESYSTEM)` to the Windows section, if you think it's better. `std::filesystem::path` will become more available in other systems when upgrading the minimum platform versions.

ntadej:
Which platform is missing `std::filesystem`? This should be part of C++17.

tdcosta100:
@ntadej, unfortunately even with C++17, some build systems fail: [MacOS, old buid but same error](https://github.com/maplibre/maplibre-gl-native/actions/runs/4451303053/jobs/7817798284), [iOS](https://github.com/maplibre/maplibre-gl-native/actions/runs/4768658715/jobs/8478219496) and [MinGW 8.1.0 (Qt 5)](https://github.com/maplibre/maplibre-gl-native/actions/runs/4768658720/jobs/8478221471).

ntadej:
Thanks for clarification. Your implementation makes sense.

ntadej (approved):

tdcosta100:
When in the future minimum supported versions increase, we can use `std::filesystem::path::is_absolute()` for everyone. Maybe in 1 year or 2.

louwers (approved):

donno:
Is the syntax suppose to be: 
`mbtiles:///C:/Users/user/map/tiles.mbtiles`
or 
`mbtiles://C:/Users/user/map/tiles.mbtiles`

The original comment here suggests the former, however the merged code change suggests the latter  because it looks for :// then passes the resulting path to essentially `std::filesystem::path(path).is_absolute();`. which won't return true if there is a / at the start of the path as it will expect it to start with a drive letter not a slash. A UNC style path with \\ or // will be considered absolute.

Perhaps the correct form could be added to the README.md for Windows.

tdcosta100:
The original comment suggests a way that not touches the original check, hence the first syntax. But the idea evolved since then, so the check was changed and now it accepts non-slash absolute paths (mandatory in Windows). A more recent comment about this is the one below:

> I think yes, the code can be reworked to extract this as a function. Something like `is_absolute_uri(const std::string&)`.
> 
> It checks against the resource URL to guarantee the path to the .mbtiles file is absolute. For example, in the URI `mbtiles:///home/user/tiles.mbtiles`, the part of the path is `/home/user/tiles.mbtiles`. This is an absolute path. In other hand, the URI `mbtiles://user/tiles.mbtiles` have the path `user/tiles.mbtiles`, that is, a relative path, so it does not accept this URI. Checking against `:///` ensures the path starts with `/`, so it's an absolute path.
> 
> In Windows, absolute paths must have a drive letter (C:, D:, X:, etc), or be UNC paths (starting with `\\` or `//`). So we cannot check against `:///`, because we must accept something like `mbtiles://C:/Users/user/tiles.mbtiles`.

You are right, a tip on `mbtiles://` URL syntax can be added in README.md for Windows.

PR #911, opened 2023-03-16
Use `std::optional` instead of `mbgl::optional`

louwers:
- Remove `optional.hpp`. Replace usage of `mbgl::optional` with `std::optional` and `mbgl::nullopt` with `std::nullopt` throughout the code base.
  - ~One point of discussion can be whether to use `{}` or `std::nullopt` when the former is possible.~ I changed it to `std::nullopt` because then it is a bit more clear that an optional is involved.
- Remove two overloads of [record](https://github.com/maplibre/maplibre-gl-native/pull/911/files#diff-0f0970c6c08a84fe2e6127e0b7d3c31dcd41263c759df985cf75a760e48b7ddbL78) used for logging. This made use of the [type-unsafe sprintf and vargs](https://github.com/maplibre/maplibre-gl-native/pull/911/files#diff-c2ce0a629dd89b46734646d7ef3e021b30aefaabd32a7fb1a41f89f064fbea55L67). std::optional has an implicit constructor which was causing ambiguity with the overloads.
- Set macOS deployment target to 10.13 using Xcode. Surpress or fix deprecated AppKit constructs. This was required for `std::optional<T>::value` support.
- Set iOS deployment target to 11.0 using Xcode. Surpress or fix deprecated UIKit constructs. This was required for `std::optional<T>::value` support.

When I started with this I thought it would be relatively uncomplicated, but the change is a bit unwieldy since so many files needed to be changed. In retrospect splitting this up would probably have been better.


render-test/runner.cpp
```diff
@@ -76,7 +76,7 @@ std::string simpleDiff(const Value& result, const Value& expected) {
             break;
         }
 
-        if (!deepEqual(resultTokens[i], expectedTokens[i])) {
+        if (resultTokens[i] != expectedTokens[i]) {
```
louwers, line 79: These are `std::string`s. I was getting an error that the call to `deepEqual` was ambigious (probably because `std::optional` can be implicitly constructed).

platform/default/src/mbgl/storage/online_file_source.cpp
```diff
@@ -70,7 +70,7 @@ struct OnlineFileRequest {
     // backoff when retrying requests.
     uint32_t failedRequests = 0;
     Response::Error::Reason failedRequestReason = Response::Error::Reason::Success;
-    optional<Timestamp> retryAfter;
+		std::optional<Timestamp> retryAfter;
```
louwers, line 73:
```suggestion
    std::optional<Timestamp> retryAfter;
```

platform/default/src/mbgl/storage/online_file_source.cpp
```diff
@@ -267,7 +267,7 @@ class OnlineFileSourceThread {
             }
         }
 
-        optional<OnlineFileRequest*> pop() {
+            std::optional<OnlineFileRequest*> pop() {
```
louwers, line 270:
```suggestion
            std::optional<OnlineFileRequest*> pop() {
```

src/mbgl/util/url.hpp
```diff
@@ -23,7 +23,7 @@ class URL {
 };
 
 // Class that holds position + lenth pairs for directory, extension, and filename of a path.
-// The extension will contain the preceding ., and optionally a preceding @2x specifier.
+// The extension will contain the preceding ., and std::optionally a preceding @2x specifier.
```
louwers, line 26:
```suggestion
// The extension will contain the preceding ., and optionally a preceding @2x specifier.
```

test/style/style_layer.test.cpp
```diff
@@ -361,7 +361,7 @@ void testHasOverrides(LayoutType& layout) {
     // Nested expressions, overridden text-color.
     auto formattedExpr1 = format("first paragraph");
     FormatExpressionSection secondParagraph(literal("second paragraph"));
-    secondParagraph.setTextSectionOptions(nullopt, nullopt, toColor(literal("blue")));
+    secondParagraph.setTextSectionOptions({}, {}, toColor(literal("blue")));
```
louwers, line 364:
```suggestion
    secondParagraph.setTextSectionOptions(std::nullopt, std::nullopt, toColor(literal("blue")));
```

test/text/cross_tile_symbol_index.test.cpp
```diff
@@ -29,7 +29,7 @@ SymbolInstance makeSymbolInstance(float x, float y, std::u16string key) {
                                                                  SymbolContent::IconSDF,
                                                                  false,
                                                                  false);
-    return SymbolInstance(anchor, std::move(sharedData), shaping, nullopt, nullopt, 0, 0, placementType, textOffset, 0, 0, iconOffset, subfeature, 0, 0, key, 0.0f, 0.0f, 0.0f, variableTextOffset, false);
+    return SymbolInstance(anchor, std::move(sharedData), shaping, {}, {}, 0, 0, placementType, textOffset, 0, 0, iconOffset, subfeature, 0, 0, key, 0.0f, 0.0f, 0.0f, variableTextOffset, false);
```
louwers, line 32:
```suggestion
    return SymbolInstance(anchor, std::move(sharedData), shaping, std::nullopt, std::nullopt, 0, 0, placementType, textOffset, 0, 0, iconOffset, subfeature, 0, 0, key, 0.0f, 0.0f, 0.0f, variableTextOffset, false);
```

test/text/formatted.test.cpp
```diff
@@ -19,15 +18,15 @@ TEST(Formatted, Empty) {
     Formatted emptyFormatted{""};
     EXPECT_TRUE(emptyFormatted.empty());
 
-    auto emptyText = FormattedSection{"", nullopt, nullopt, nullopt};
+    auto emptyText = FormattedSection{"", {}, {}, {}};
```
louwers, line 21:
```suggestion
    auto emptyText = FormattedSection{"", std::nullopt, std::nullopt, std::nullopt};
```

test/text/formatted.test.cpp
```diff
@@ -41,7 +40,7 @@ TEST(Formatted, ToString) {
     Formatted emptyFormatted{""};
     EXPECT_EQ(emptyFormatted.toString(), "");
 
-    auto text = FormattedSection{"Formatted", nullopt, nullopt, nullopt};
+    auto text = FormattedSection{"Formatted", {}, {}, {}};
```
louwers, line 43:
```suggestion
    auto text = FormattedSection{"Formatted", std::nullopt, std::nullopt, std::nullopt};
```

test/text/quads.test.cpp
```diff
@@ -50,7 +50,7 @@ TEST(getIconQuads, style) {
     // shapedText.positionedGlyphs.emplace_back(PositionedGlyph(32, 0.0f, 0.0f, false, 0, 1.0));
     shapedText.positionedLines.emplace_back();
     shapedText.positionedLines.back().positionedGlyphs.emplace_back(
-        PositionedGlyph(32, 0.0f, 0.0f, false, 0, 1.0, /*texRect*/ {}, /*metrics*/ {}, /*imageID*/ nullopt));
+        PositionedGlyph(32, 0.0f, 0.0f, false, 0, 1.0, /*texRect*/ {}, /*metrics*/ {}, /*imageID*/ {}));
```
louwers, line 53:
```suggestion
        PositionedGlyph(32, 0.0f, 0.0f, false, 0, 1.0, /*texRect*/ {}, /*metrics*/ {}, /*imageID*/ std::nullopt));
```

src/mbgl/renderer/buckets/symbol_bucket.hpp
```diff
@@ -100,8 +100,8 @@ class SymbolBucket final : public Bucket {
     // Returns references to the `symbolInstances` items, sorted by viewport Y.
     SymbolInstanceReferences getSortedSymbols(float angle) const;
     // Returns references to the `symbolInstances` items, which belong to the `sortKeyRange` range;
-    // returns references to all the symbols if |sortKeyRange| is `nullopt`.
-    SymbolInstanceReferences getSymbols(const optional<SortKeyRange>& sortKeyRange = nullopt) const;
+    // returns references to all the symbols if |sortKeyRange| is `{}`.
```
louwers, line 103:
```suggestion
    // returns references to all the symbols if |sortKeyRange| is `std::nullopt`.
```

src/mbgl/renderer/image_manager.cpp
```diff
@@ -270,7 +272,9 @@ void ImageManager::notify(ImageRequestor& requestor, const ImageRequestPair& pai
 }
 
 void ImageManager::dumpDebugLogs() const {
-    Log::Info(Event::General, "ImageManager::loaded: %d", loaded);
+    std::ostringstream ss;  // TODO: replace with std::format
```
louwers, line 275:
```suggestion
    std::ostringstream ss;
```

src/mbgl/tile/tile_loader_impl.hpp
```diff
@@ -37,11 +37,11 @@ TileLoader<T>::TileLoader(T& tile_,
     }
 
     if (fileSource->supportsCacheOnlyRequests()) {
-        // When supported, the first request is always optional, even if the TileLoader
-        // is marked as required. That way, we can let the first optional request continue
-        // to load when the TileLoader is later changed from required to optional. If we
+        // When supported, the first request is always std::optional, even if the TileLoader
```
louwers, line 40:
```suggestion
        // When supported, the first request is always optional, even if the TileLoader
```

src/mbgl/tile/tile_loader_impl.hpp
```diff
@@ -37,11 +37,11 @@ TileLoader<T>::TileLoader(T& tile_,
...
+        // is marked as required. That way, we can let the first std::optional request continue
```
louwers, line 41:
```suggestion
        // is marked as required. That way, we can let the first optional request continue
```

src/mbgl/tile/tile_loader_impl.hpp
```diff
@@ -37,11 +37,11 @@ TileLoader<T>::TileLoader(T& tile_,
...
+        // to load when the TileLoader is later changed from required to std::optional. If we
```
louwers, line 42:
```suggestion
        // to load when the TileLoader is later changed from required to optional. If we
```

src/mbgl/text/collision_feature.hpp
```diff
@@ -86,7 +86,7 @@ class CollisionFeature {
                            shapedText.bottom,
                            shapedText.left,
                            shapedText.right,
-                           nullopt,
+                           {},
```
louwers, line 89:
```suggestion
                           std::nullopt,
```

louwers:
Qt issue looks like a compiler bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86465

Can we update it?

ovivoda (approved):
This is a big one, code is looking good. Also judging from the CI standpoint, we should be good to go. Great work @louwers !

include/mbgl/util/logging.hpp
```diff
@@ -75,13 +75,12 @@ class Log {
...
     static void record(EventSeverity severity, Event event, const std::string &msg);
-    static void record(EventSeverity severity, Event event, const char* format = "", ...);
-    static void record(EventSeverity severity, Event event, int64_t code, const char* format = "", ...);
+    static void record(EventSeverity severity, Event event, int64_t code, const std::string &msg);
     static void record(EventSeverity severity,
                        Event event,
                        int64_t code,
                        const std::string& msg,
-                       const optional<std::string>& threadName);
+                       const std::optional<std::string>& threadName);
```
etnav, line 83: Build is failing (Node on Windows with `MBGL_EGL = ON`), seems to be affected by these changes, any ideas?

```
C:\builds\maplibre-gl-native\include\mbgl/util/logging.hpp(70): error C2665: 'mbgl::Log::record': none of the 3 overloads could convert all the argument types
C:\builds\maplibre-gl-native\include\mbgl/util/logging.hpp(78): note: could be 'void mbgl::Log::record(mbgl::EventSeverity,mbgl::Event,int64_t,const std::string &)'
C:\builds\maplibre-gl-native\include\mbgl/util/logging.hpp(66): note: while trying to match the argument list '(mbgl::EventSeverity, mbgl::Event, const char [48], _Ty)'
        with
        [
            _Ty=EGLint
        ]
C:\builds\maplibre-gl-native\include\mbgl/util/logging.hpp(61): note: see reference to function template instantiation 'void mbgl::Log::Record<const char(&)[48],_Ty>(mbgl::EventSeverity,mbgl::Event,const char (&)[48],_Ty &&)' being compiled
        with
        [
            _Ty=EGLint
        ]
C:\builds\maplibre-gl-native\platform\windows\src\headless_backend_egl.cpp(34): note: see reference to function template instantiation 'void mbgl::Log::Error<const char(&)[48],EGLint>(mbgl::Event,const char (&)[48],EGLint &&)' being compiled
```

louwers: We need to update CI so that this can be catched.

There is a logging statement there that uses sprintf. Needs to be rewritten to pass a std::string instead.

louwers: @etnav 

Could you try this and maybe submit a PR if you get it to build? I am not on Windows.

```
            mbgl::Log::Error(mbgl::Event::OpenGL, "eglBindAPI(EGL_OPENGL_ES_API) returned error " + std::to_string(eglGetError()));
```

PR #919, opened 2023-03-20
Files prefix changed from MGL to MLN

alexcristici:
Prefix MGL was removed in favor of new prefix MLN.


ntadej:
Why not only ML?

louwers:
Sorry, it is a bit unclear but we have multiple changelogs in the repo.

@alexcristici the changelog for iOS is here: https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/platform/ios/CHANGELOG.md

Could you also add (Breaking) so we know to bump a major version, and a mitigation guide (in this case, it's just "change all your MGL prefixes to MLN").

mwilsnd (approved):

platform/ios/platform/darwin/scripts/style-spec-overrides-v8.json
sjg-wdw, line 1: Is this going to effectively change the name of the extension feature in the style spec?
Like an MGLLight is now an MLNLight.

Are these sorts of extended features in common use in the style spec?

alexcristici: It will be loaded in a `MLNLight` object instead of `MGLLight` object, but the json key will be the same `light`.

platform/ios/platform/ios/CHANGELOG.md
sjg-wdw, line 1: Just a nitpick, but does it make sense to rename the objects in here?  They were called that back when those changes were made.  But now they're called the new thing.  I can see an argument for either.

sjg-wdw (approved):

platform/ios/platform/ios/CHANGELOG.md
louwers: I think it is a bit clearer for people reading the changelog what is being meant. We still have the old names in the git history. ;-)

PR #921, opened 2023-03-21
Better MinGW build support

tdcosta100:
If you ever tried to build the entire project with MinGW (specially Qt and the test targets) instead only the core projects or using Microsoft Visual Studio, you will discover it's painful. This PR allows to build with MinGW a smoother process.

### DIFF
--- a/platform/qt/qt.cmake
+++ b/platform/qt/qt.cmake
@@ -366,6 +366,13 @@ if(NOT MBGL_QT_LIBRARY_ONLY)
         PRIVATE WORK_DIRECTORY=${PROJECT_SOURCE_DIR}
     )
 
+    if(WIN32)
+        target_compile_definitions(
+            mbgl-test-runner
+            PRIVATE MBGL_BUILDING_LIB
+        )
+    endif()
+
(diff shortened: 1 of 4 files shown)


ntadej:
Thanks for looking into this!

As I mentioned in the discussion, I'm not sure how many people actually use MinGW. In my experience MSVC always built faster and gave smaller binary sizes (plus it's easier to use the Windows APIs).

tdcosta100:
Yeah, I use MSVC too, it's so much better. But since there is MinGW in CI, so it's fair to provide it better support, I guess. Some tests failed because it was a connection error while checking out the repository. Could you please re-run the failed workflows?

ntadej:
If we ever want to be included back in the Qt distribution, we need to support it at least for latest Qt 6.

tdcosta100:
I agree with you, if MinGW support is dropped, I won't complain.

tdcosta100:
node-ci is failing with Mac OS 12, but I think with #926 merged, it won't fail anymore.

louwers (approved):

PR #922, opened 2023-03-21
Run node-ci on changes to macOS, Linux & Windows platforms

louwers:


ovivoda (approved):
Looks legit!

acalcutt:
If you are going to add windows, doesn't it make sense to include macos and linux too?

louwers:
> If you are going to add windows, doesn't it make sense to include macos and linux too?

I'm not sure, are there files of the macOS or linux platforms that do not get compiled on CI?

acalcutt:
I don't think it makes sense to run it on all iOS/Android/QT changes. that ci still has that annoying xfrb crash. If you look at the cmakelists file, linux/macos/windows only get run in a core build or node build.

PR #924, opened 2023-03-21
Add shader registry

mwilsnd:
This PR adds the initial foundation for a runtime shader registry. The registry is implemented side-by-side with the existing `mbgl::Programs` object in `gfx::RenderStaticData` and exposed to layer renderers in the same way via `PaintParameters`. Additionally, the ability to re-write shader source has been introduced via a new method in `ProgramParameters`. A new map observer event has been added to notify users after shader initialization to permit mutation of the registry.

In following PRs:
* Built-in programs will be installed in the registry via the new stub method `Programs::registerWith`
* Shader source for each program will be lifted out of the compressed, code generated blob
* With the implementation of drawables & builders, outright replacement of programs in the registry will then become practical

These changes were tested using the desktop GLFW platform application under MSVC.


include/mbgl/gfx/shader_registry.hpp
```diff
@@ -0,0 +1,78 @@
...
+                return nullptr;
+            }
+            return std::dynamic_pointer_cast<T>(shader);
+        }
+
+    private:
+        std::unordered_map<
+            std::string,
+            std::shared_ptr<gfx::Shader>
+        > programs;
```
stefankarschti, line 73: could 'shaders' be more appropriate for member name?

ovivoda (approved):
It looks good!

Really nice to also see the TODO's.

louwers (changes requested):
I love your code. Very clean and taking full advantage of C++17.

Could you add a few tests?

- Make sure `onRegisterShaders` is called.
- Calling isShader with a shader registered / not registered
- Making sure replacing a shader does indeed replace it
- Maybe pull the vertexSource and fragmentSource concatenation in a separate function, and test that?

mwilsnd:
> Could you add a few tests?

I've added a few tests and revised the base shader & registry to function without requiring RTTI. Vertex and fragment source are the only directly un-tested feature however there is coverage in some of the existing tests. Currently that functionality is private to the core, I'll have some more changes for it in following PRs.

include/mbgl/gfx/shader.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <type_traits>
+
+namespace mbgl {
+namespace gfx {
+
+class Shader;
+
+// Populate type info for use with downcasting shaders.
+// This is done to ensure -frtti is not needed.
+#define DECLARE_SHADER_TYPEINFO(ClassName)                  \
```
louwers, line 13: Between this and enabling `-frtti`, I would personally prefer the latter.

louwers: This may be a stupid question, but why does the shader name need to be available at compile time?

sjg-wdw: RTTI added a bunch of size.

louwers (changes requested):
Please see my comment above.

Would using a `dynamic_cast` be possible?

include/mbgl/gfx/shader.hpp
louwers: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c146-use-dynamic_cast-where-class-hierarchy-navigation-is-unavoidable

mwilsnd: It's included at compile time so the conversion methods can do the type check (Shader::to, Registry::get).

I decided on this method for a few reasons:
* Building mbgl-test-runner with MinSizeRel, -frtti: 10,880,248B, -fno-rtti: 9,747,944B. Requiring those that used to get by without RTTI to suddenly require it would bring a jump in binary size.
* The CMake scripts already default to RTTI off so a hard requirement would be an annoying regression.
* The existing solution was static polymorphism using CRTP, also likely chosen to avoid requiring RTTI. While either solution can achieve this, compile time polymorphism doesn't solve the fundamental problem of down-casting from a non-template base without some form of type info.
* static_cast doesn't perform any sanity checking beyond if types at compile time are related. That means it's possible to treat any program in the registry as any other program which isn't great. The original implementation stepped around the down-casting issue by just never doing it. If we want a runtime registry however, we kind of have to.

The registry can be made to support both options however. An ifdef could relax the rules around final and available type info and switch to dynamic_pointer_cast. Then at least it's opt-in. This is a compromise solution that I'm hoping gets easier to improve (or delete outright) as the refactor progresses.

louwers: Does `dynamic_cast` need RTTI?

mwilsnd: Yeah, down-casting with dynamic_cast (from base to derived) requires type information. Up-casting (going from derived to base, yes confusing terms) does not need RTTI.

louwers: Thanks! Given the binary size bump a 'home grow RTTI' seems warranted. Two more questions:

1. I am sure you thought of this, but is there any way to avoid the macro?

2. Instead of `InternalTypeInfoDeclared`, can we do something like...

```
static constexpr std::string_view shaderName = "...";
```

```
...
std::is_same_v<typename T::shaderName, std::string_view> &&
...
```

Coming back to (1), is manually adding

```
  static constexpr std::string_view shaderName = "MyShader";                \
  const auto name() const noexcept override { \
      return shaderName;                                        \
  }
```

to every shader, acceptable?

mwilsnd: 1. The macro is there just to avoid that boiler plate, manually defining the member and overload works as well. While I'm not usually a fan of macros, this is one case where they're still useful. If the use of macros in this way is objectionable we can change to requiring the manual implementation in every class.
2. Yeah we can switch to looking for that instead.

louwers (approved):

PR #925, opened 2023-03-22
Fix "Filename too long" when cloning the repository in Qt CI

tdcosta100:
Qt CI currently generate a lot of `Filename too long` errors when cloning the repository, because filepaths have more than 256 characters. Adding the `-c core.longpaths=true` switch in git command fixes this.


acalcutt (approved):
we fixed this in the [node-ci](https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/node-ci.yml#L99) (which was based on this qt-ci) and it fixed the checkout errors. I think it is worth getting rid of the checkout error on windows.

ntadej:
Yes, sorry, also fine with me!

PR #928, opened 2023-03-22
Update Node CHANGELOG.md

tdcosta100:


ntadej:
Not sure this is really a feature. It's more of a bugfix actually. We were otherwise C++17 for a while now (maybe that should actually be the feature).

tdcosta100:
Ah yes, it's not a feature, I'm just following the former notes, since a PR was approved and I forgot to mention it in the CHANGELOG.md.

ntadej:
OK, maybe something for @louwers and @ovivoda to try to improve in the future.

louwers:
Our main focus is Core, Android and iOS and in principle we don't spend too much time on other platforms. We also don't make releases for them.

ntadej:
What I meant is we may want to define general guidelines on changelogs and release notes.

wipfli:
Thanks for clarifying this @louwers. People are welcome to contribute to all platforms and it is great if they handle releases. The paid maintainers focus is on iOS and Android.

ntadej:
Maybe I should have tagged you Oliver then 😊

acalcutt:
Usually I do the node releases, so a appreciate the changelog being uodated sinc eit helps me make the release

I agree though that this should maybe be under bug fixes. maybe something like "fixed -std=c++11 when building node release", or something like that.

Could you add a heading 
### 🐞 Bug fixes
and add this under it, like https://github.com/maplibre/maplibre-gl-native/edit/main/CHANGELOG.md

tdcosta100:
Done!

platform/node/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 
 ### ✨ New features
 - *...Add new stuff here...*
+* Compile Node targets without -std=c++11 option by @tdcosta100 in https://github.com/maplibre/maplibre-gl-native/pull/926
```
acalcutt, line 6: Could you put this one under a new heading,
```
### 🐞 Bug fixes
*...Add new stuff here...*
```

tdcosta100: Yes, of course! Nice to have this section here.

acalcutt (approved):

PR #937, opened 2023-03-23
Fixed and enabled iOS render tests

alexcristici:
Fixed broken iOS render tests.
Now iOS render tests will be included in the iOS demo app and would run successfully from Xcode.


sjg-wdw (approved):
I'm not the expert on MapLibre tests, but this looks fine to me.

louwers:
What command did you use to run them?
Do they run in the simulator as well?

I could enable them on CI.

What command did you use to update the metrics?

Will you still try the newest version of the render tests from maplibre-gl-js?

alexcristici:
> OK thanks for clarifying.
> 
> What do you think, should we pull the style.json and expected image from maplibre-gl-js into this repo? And if yes, should be try to use the most recent version? Or should they live in a shared repository?

We should reorganize the render tests and copy them into this repo. This will be my next PR. In this PR I just re-enabled iOS render tests.

alexcristici:
> Just an FYI, I did test moving the integration tests a bit in #855 . I had put the styles and expect files into metrics/integration . I originally was trying to move them into metrics/tests/integration, since there are other style/expect files there, but I didn't want to change the path for the other tests in metrics/tests.
> 
> Feel free to ignore this test, I just thought I would mention it. To run the integration tests you use json files in the metrics folder, so i though it made sense to have the test results there too.

Sounds like a good location. I will check it. Thank you! 👍

mwilsnd (approved):
Nice!

PR #941, opened 2023-03-25
Fix/crash in map snapshotter activity

ovivoda:
Fix MapSnapshotterActivity.kt crash if there is IllegalStateException exception thrown from native code.

However I think we will need to dig in a bit more to find out the real problem. 

cc @artakka @birkskyum Any clues on the root cause?


alanchenboy:
Just curious, can be captured in native code? there  may throw exceptions in `mbgl::map::renderStill`

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/snapshotter/MapSnapshotter.kt
```diff
@@ -652,22 +666,26 @@ open class MapSnapshotter(context: Context, options: Options) {
...
+                            addLayerBelow(layerWrapper.getLayer(), layerWrapper.belowLayer)
+                        } else {
+                            addLayerBelow(layerWrapper.layer, MapboxConstants.LAYER_ID_ANNOTATIONS)
+                        }
+                    }
+                    for (image in builder.images) {
+                        addImage(image.id, image.bitmap, image.isSdf)
+                    }
+                } catch (e: IllegalStateException) {
+                    e.printStackTrace()
```
louwers, line 688: Did you left this in on purpose?

ovivoda: Yes. Just fixing the crash not the cause. Once the cause is fixed we can reuse fully previous code.

ovivoda:
This is a temporary fix for showcasing the MapLibre features. We should investigate more the native cause of the crash.

ovivoda:
> mbgl::map::renderStill

Indeed this should be the next step. To investigate more on this part of the native code. Not trivial to debug though but I will give it a try.

ovivoda:
linking initial reported issue to this for reference: https://github.com/maplibre/maplibre-gl-native/issues/903

louwers (changes requested):
Any quick fixes need to be documented with comments.

Can we reproduce the issue with a test?

louwers:
Added a test but it's passing, so need to tweak it a bit more.

louwers:
OK now the test reproduces the error (if the exception is not caught).

louwers:
I am digging into JNI to be able debug this. But it might take a while.

We should add a comment to the catch block if this quick fix is urgent.

ovivoda:
Actually I think I would go for another approach. I would leave `MapSnapshotter` in place as it was and create a failing test with the loaded style from JSON and fix the problem for that one.

louwers:
> Actually I think I would go for another approach. I would leave `MapSnapshotterTest` in place as it was and create a failing named something like`MapSnapshotterLocalStyleLoadTest` test with the loaded style from JSON and fix the problem for that one. 

I can use another name but I would use the same file. The code for running the test is identical, it is just the `MapSnapshotter.Options` that is different.

ovivoda:
I am closing this one as the main problem relies somewhere in the native code.

Issue #927, opened 2023-03-22
Improve fling animation


Helium314:
The current fling animation feels very awkward, and is different to basically all/most other map engines and online maps.
It appears more like an easeCamera animation than a slowly decelerating continuation of the previous movement. After some testing I think the issue mainly comes from
* the camera at some point moving faster than it started (this seems to be tunable, see below)
* a relatively high velocity is required to start the fling animation, but if the animation starts the move is relatively far
* the camera stopping relatively quickly

I tried using a custom move gesture detector (see below) to implement a slightly modified [fling gesture](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java#L408). This feels much better than current behavior, but a) still a little awkward (due to the sudden stop) and b) requires re-implementing move and pinch-zoom gestures.
A first and simple step could be allowing to choose the fling velocity threshold and animation base time, but ideally the camera velocity behavior during the movement would be adjusted.

```kotlin
            override fun onMoveEnd(p0: MoveGestureDetector, velocityX: Float, velocityY: Float) {
                // maybe fling camera using somewhat different settings
                val minVelocity = 200 // default is 1000
                val animationVelocityFactor = 1.0 / 7 // default is 1 / 7
                val animationDurationOffset = 500 // default is 150
                // without moveFactor (or setting to 1)
                //  the move feels too far
                //  the move starts too fast (animation should not start faster than the move before lifting finger)
                val moveFactor = 0.5

                // reproduce roughly what maplibre does
                val screenDensity = mapView.pixelRatio
                val velocityXY = hypot(velocityX / screenDensity, velocityY / screenDensity)
                if (velocityXY < minVelocity) return
                val tiltFactor = 1.5 // assume tilt = 0 for simplicity
                // default way of determining animation time
                val animationTime = (velocityXY * animationVelocityFactor / tiltFactor + animationDurationOffset).toInt()
                // determine target position
                val offsetX = velocityX / tiltFactor / screenDensity
                val offsetY = velocityY / tiltFactor / screenDensity
                // below is mostly guess, but seems to fit at least roughly
                val c = mapboxMap.cameraPosition
                val p = c.target ?: return
                val metersPerPixel = mapboxMap.projection.getMetersPerPixelAtLatitude(p.latitude)
                val degreesPerMeterLat = 1 / 111225.0 // only approximate, but good enough
[...]

PR #963, opened 2023-03-30
Improve fling animation

Helium314:
PR for #927

This improves the fling animation with the main goal of the animation being similar to other maps.
See https://github.com/maplibre/maplibre-gl-native/issues/927#issuecomment-1486700986 for the reasons.
Short:
* UnitBezier parameters are adjusted for a less sudden stop of the fling animation
* Offset is now velocity / animationTime (with some mysterious factor, chosen so that animation starts at same velocity as the user moved the map)
* Animation base time and fling threshold can be set in UiSettings. In my opinion default threshold is too high, and animation time too short, making the map feel "sticky".

### DIFF
--- a/platform/android/CHANGELOG.md
+++ b/platform/android/CHANGELOG.md
@@ -6,6 +6,8 @@ MapLibre welcomes participation and contributions from everyone. Please read [`C
 
 ### ✨ Features and improvements
 
+* Change to a more natural fling animation and allow setting `flingThreshold` and `flingAnimationBaseTime` in `UiSettings` ([#963](https://github.com/maplibre/maplibre-gl-native/pull/963))
+
 ### 🐞 Bug fixes
 
 ### ⛵ Dependencies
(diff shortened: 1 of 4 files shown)


louwers:
Neat! Could you update the android `CHANGELOG.md`?

Helium314:
Sure!

Did you test it? I am not 100% sure the factor 0.28 will work on all screenDensities / tilts.
Especially with tilt testing is necessary (wasn't able to tilt, as mentioned in https://github.com/maplibre/maplibre-gl-native/issues/927#issuecomment-1486700986

What else is to be done?
* javadoc for changes in UI settings
* moving UI settings changes to "correct" position (not sure if there is any ordering / grouping of settings)
* maybe increasing `ANIMATION_DURATION_FLING_BASE` from `ANIMATION_DURATION_SHORT`to `ANIMATION_DURATION` (150 ms  is much too short in my opinion)

ovivoda (approved):
This looks good! Are you able to provide a video of the change in action on one of the native platforms?

Helium314:
> This looks good! Are you able to provide a video of the change in action on one of the native platforms?

Same app on Android as in https://github.com/maplibre/maplibre-gl-native/issues/927#issuecomment-1480170974, but with the updated library with flingAnimationBaseTime set to 300 ms and flingThreshold set to 300:

https://user-images.githubusercontent.com/43007630/229335908-6a03dfa7-9c74-4126-990f-297d60070a7f.mp4

ysmaherzi:
Hi,
This fix was published in the version 10.1.0 and is still present to this day on master but it's unavaiable on the latest stable 10.2.0.
Does someone know why ?

Thank you,

louwers:
@ysmaherzi 10.2.0 reverted 10.1.0 because it contained a breaking change and should have been a major release.

Sorry for the confusion. This change will be included in the 11.0.0 release.

ysmaherzi:
Thank you @louwers :)

PR #964, opened 2023-03-31
Migate shaders to the registry

mwilsnd:
Following the shader registry, this PR does a number of things:
* Moves shader source out of the compressed blob to human-readable locations
  * Deletes old shader source (blob, generated code)
  * Introduces new scheme for selecting shader source based on a built-in program enum + graphics backend enum
  * Shader source is header only and publicly exposed outside the toolkit (via include/)
  * These changes were code-generated using a utility program authored in D (see branch [here](https://github.com/mwilsnd/maplibre-gl-native/tree/shader-tools)), we can either merge this tool or write something similar in another language later
* Makes `ProgramParameters` backend-aware and introduces the concept of 'default' shader source
  * `gl::Program` has been updated to refer to the new shader source lookup & provide it via default source of `ProgramParameters`
  * `mbgl::Program` has been modified to pass the built-in shader enum as an additional template parameter, decoupling the CRTP pattern from shader source selection
* All existing programs have been migrated to the new shader source selection and gfx::Shader base for registry compatibility
* New tests:
  * GLSL Replace no-op: Simply replaces a shader instance without shader source mutation
  * GLSL Replace 1: Replaces a shader instance with a bare-bones fragment shader
  * GLSL Replace 2: Replaces a shader instance with a modified fragment shader
  
 Passing render tests, Debian 11 + llvmpipe
 ```
1222 passed (93.1%)
10 passed but were ignored (0.8%)
80 ignored (6.1%)
Results at: /home/mwilsnd/git/maplibre-gl-native/metrics/linux-clang8-release-style.html
```

### CODE REVIEW

**Comment by TimSylvester** [comment]
> Shader source is header only and publicly exposed outside the toolkit

The strings aren't duplicated across compilation units, right?

**Comment by mwilsnd** [comment]
> > Shader source is header only and publicly exposed outside the toolkit
> 
> The strings aren't duplicated across compilation units, right?

They're static constexpr, made accessible through a struct template at compile time. While the template specializations are available in compilation units that include the shader manifest, it is only in TLUs that instantiate one of these struct templates that the compile-time strings become directly accessible. This happens inside `mbgl/gl/program.hpp` for accessing the prelude and `mbgl/programs/program.hpp` for setting the default shader source of a program. It should also be noted, every modern compiler in use today will perform string pooling (de-duping) optimization for placement in the .rdata/const segment, so in the case where multiple shaders access the same built-in source in separate TLUs, you shouldn't see duplicate strings emitted in the binary (if that was your concern).

**Review by TimSylvester — APPROVED** [review]

**Review by sjg-wdw — APPROVED** [review]
Comments in source code!  My goodness!

**Review by ovivoda — APPROVED** [review]
nice to see the reorg taking shape

File: include/mbgl/gfx/backend.hpp [hunk 1]
```diff
@@ -11,6 +11,7 @@ class Backend {
 public:
     enum class Type : uint8_t {
         OpenGL,
+        TYPE_MAX,
```
louwers, line 14: What's this?

louwers (changes requested):
We should pull in the shader generation code + build & run instructions or forgo generating the shaders from now on.

Generation code should _probably_ be based on JavaScript / TypeScript (like other scripts in the repo) but if we may want to forgo generating shader code later, it's probably not worth it to rewrite your D code for now.

include/mbgl/shaders/shader_source.hpp
```diff
@@ -0,0 +1,47 @@
...
+    LineSDFProgram,
+    LineProgram,
+    RasterProgram,
+    SymbolIconProgram,
+    SymbolSDFTextProgram,
+    SymbolSDFIconProgram,
+    SymbolTextAndIconProgram
+};
+
+template <BuiltIn T, gfx::Backend::Type> struct ShaderSource;
```
louwers, line 39: Could you explain the purpose of this template (maybe in a comment)?

src/mbgl/programs/programs.cpp
```diff
@@ -10,79 +10,60 @@
...
+{
+    ( [](bool expr) {
+        if (!expr) {
+            throw std::runtime_error(
+                "Failed to register " +
+                std::string(T::Name) +
+                " with shader registry!");
+        }
+    }( registry.registerShader(
+        std::make_shared<T>(programParameters_))), ... );
```
louwers, line 35: How `...` work again? 😅

Care to add a comment what this lambda + call to registerShader is doing?

mwilsnd: Yeah this does need a comment - this is a fold expression over a function call. Template varargs (a parameter pack) are expanded at compile time type-wise over another expression (function call in this case). It's like a compile time loop-unroll.

include/mbgl/gfx/backend.hpp
mwilsnd: The max type is used to determine the number of back-ends available. ProgramParameters uses this to size arrays for indexing shader source by back-end type.

louwers: Ah so this one needs to be at the end when adding new backends.

alexcristici (approved):
👍

include/mbgl/gfx/backend.hpp
```diff
@@ -9,8 +9,11 @@ namespace gfx {
 
 class Backend {
 public:
+    /// @brief The active graphics API/backend type.
     enum class Type : uint8_t {
-        OpenGL,
+        OpenGL,     // The OpenGL API backend
```
louwers, line 14:
```suggestion
        OpenGL,     ///< The OpenGL API backend
```

include/mbgl/gfx/backend.hpp
```diff
@@ -9,8 +9,11 @@ namespace gfx {
...
+        TYPE_MAX,   // Not a valid backend type, used to determine the number
```
louwers, line 15:
```suggestion
        TYPE_MAX,   ///< Not a valid backend type, used to determine the number
```

include/mbgl/gfx/backend.hpp
```diff
@@ -9,8 +9,11 @@ namespace gfx {
...
+                    // of available backends (ie for array allocation).
```
louwers, line 16:
```suggestion
                    ///< of available backends (ie for array allocation).
```

Issue #949, opened 2023-03-27
Render Performance Tests


sjg-wdw:
Startup time
How long until the first frame
Min max frame latency
Average frame time

Uniforms
Buffer changes
draw() calls

sjg-wdw:
Feel free to break these issues out into pieces as that makes sense.  We've got a bunch of different tests in the list.

sjg-wdw:
This might be a duplicate or it may just be too broad and ongoing.

PR #965, opened 2023-03-31
Get the benchmark UI test working again

TimSylvester:
Fix a missing escape character in glyph downloading that kept URL-decoding from working fully.
Add missing glyph resources that kept `[mapView isFullyLoaded]` from ever returning true.
Make the Benchmarking UI Test finish as soon as the benchmarking is complete.

#949

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ node_modules
 
 /bazel-*
 CMakeLists.txt.user
+CMakeFiles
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -42,3 +42,4 @@ node_modules
 
 /bazel-*
 CMakeLists.txt.user
+CMakeFiles
```
mwilsnd, line 45: This looks fine to me, but you can also specify a sub-directory for cmake to place artifacts. ex: `cmake . -B build -G Xcode` to place CMakeFiles, generated Xcode project and all other artifacts in /build (which is already under the gitignore).

mwilsnd (approved):

.gitignore
TimSylvester: Ah, right, those are probably just left over from messing around with CMake previously.

louwers (approved):

PR #966, opened 2023-03-31
Add clang-format

mwilsnd:
Open to any style alterations, but I noticed there isn't any formatting rule for the main repo so I'd like to add one.


louwers:
@mwilsnd Can we run it, or it messing up the commit history not worth it?

ntadej:
MapBox had a config: https://github.com/mapbox/mapbox-gl-native/blob/master/.clang-format

I guess it was accidentally removed.

ntadej (changes requested):
I think it's better to use the original MapBox config, I expect less unwanted changes (if any).

louwers:
I like the suggestion from @ntadej.

Could you try running it @mwilsnd to see if the amount of changes are manageable?

We should also add a note about it in the root `CONTRIBUTING.md`.

In addition, maybe you can add a verifies the formatting to one of the CI workflows, otherwise we will quickly diverge again.

mwilsnd:
> I think it's better to use the original MapBox config, I expect less unwanted changes (if any).

I'll pull it in to this PR

mwilsnd:
> Could you try running it @mwilsnd to see if the amount of changes are manageable?

Looks like mapbox didn't enforce their existing format either. I ran a bulk format on the contents of src/mbgl/** which generated 403 changes:
`403 files changed, 7793 insertions(+), 8136 deletions(-)`

Looks like most of them are breaking to stay under the column limit and some indentation alignment & whitespace rules.

Also interesting, I had to shut off include sorting as it caused a build failure. Another thing to investigate.

louwers:
> Also interesting, I had to shut off include sorting as it caused a build failure. Another thing to investigate.

That's interesting. I had that problem when playing around with GLFW + GLAD.
Hopefully you can nail down the problem.

> Looks like mapbox didn't enforce their existing format either.

🤷 Let's make sure we strictly enforce it!

mwilsnd:
> @mwilsnd Do you still want to enable clang-format?
> 
> I could take this over as well.

I'd like to see it enabled, yes, You're more than welcome to pick this up if you have the bandwidth, otherwise I'll be revisiting it later this week.

SiarheiFedartsou:
> > Could you try running it @mwilsnd to see if the amount of changes are manageable?
> 
> Looks like mapbox didn't enforce their existing format either. I ran a bulk format on the contents of src/mbgl/** which generated 403 changes: `403 files changed, 7793 insertions(+), 8136 deletions(-)`
> 
> Looks like most of them are breaking to stay under the column limit and some indentation alignment & whitespace rules.
> 
> Also interesting, I had to shut off include sorting as it caused a build failure. Another thing to investigate.

Heads up that even with the same .clang-format file different versions of clang-format can give different output. Probably they simply used different version of clang-format…

At glance they used clang-format-8 

https://github.com/mapbox/mapbox-gl-native/blob/f9b8ddb804246a1b40cb4056b50761b32f901434/circle.yml#L403


Btw it complicates things a bit: you have to have particular version of clang-format somehow “pinned” locally, it can be not so easy to achieve. But I would have automatic formatting anyway - it is always better to have it, than not :)

ovivoda:
I think we definitely need this going forward. We should also enforce it for each commit by adding a pre-commit hook. This way we can keep the consistency maybe? 

I would use the less painful one, but if both are creating lots of changes then I'd go with the most modern one.

louwers:
@ovivoda Yes that is good idea. I'll add it to this PR.

test/util/token.test.cpp
```diff
@@ -7,47 +7,46 @@ using namespace mbgl;
 
 TEST(Token, replaceTokens) {
     EXPECT_EQ("literal", mbgl::util::replaceTokens("literal", [](const std::string& token) -> std::string {
-        if (token == "name") return "14th St NW";
-        return "";
-    }));
+                  if (token == "name") return "14th St NW";
```
louwers, line 10: This does not look correct.

louwers:
I blasted the source tree with `clang-format`.

Try it out yourself:

```
pre-commit install
pre-commit run --all-files
```

Using pre-commit has the advantage that it installs a specific version of clang-format for different platforms and we don't need to manage that ourselves. As @SiarheiFedartsou  mentioned different versions of clang-format can give different results, so we can't just use the clang-format that happens to be installed on someone's system. It needs to be the same everywhere otherwise making a PR will be frustrating.

This also looks cool, I want to enable it after this is merged: https://pre-commit.ci/ It runs clang-format and updates PRs after running the pre-commit hooks. This saves frustration if (new) contributors did not run the pre-commit hooks themselves.

platform/qt/src/utils/conversion.hpp
```diff
@@ -31,13 +31,9 @@ class ConversionTraits<QVariant> {
...
 
-    static std::size_t arrayLength(const QVariant& value) {
-        return value.toList().size();
-    }
+    static std::size_t arrayLength(const QVariant& value) { return value.toList().size(); }
 
-    static QVariant arrayMember(const QVariant& value, std::size_t i) {
-        return value.toList()[static_cast<int>(i)];
-    }
+    static QVariant arrayMember(const QVariant& value, std::size_t i) { return value.toList()[static_cast<int>(i)]; }
```
louwers, line 36: Let's ditch `AllowShortFunctionsOnASingleLine`?

With the 120 Column Limit this leads to pretty unreadable one liners.

ntadej:
Is the situation really that bad? I thought Mapbox used some config before?

PR #972, opened 2023-04-03
Bump iOS target to 12 for `std::optional::value` error

TimSylvester:
Add `GLES_SILENCE_DEPRECATION` for `EAGLContext`.


louwers (approved):

alexcristici:
@TimSylvester Let's merge this PR first - https://github.com/maplibre/maplibre-gl-native/pull/973. Then make sure to change also min version manually for all targets, some targets are not coming from cmake scripts.

PR #979, opened 2023-04-04
Fix concurrency crash happened with location update.

alanchenboy:
This is a problem we have met on the Android side,  I think our Android engineer forget to add it to the bug list -_-.

what happens here is:
`SymbolLocationLayerRenderer` the location render source wrap will constantly send new `locationFeature` to the map, and the Map will parse the java object in the worker threads, and meanwhile `SymbolLocationLayerRenderer` are operating the same `locationFeatue` object at the main thread. So the concurrency happened, and this issue is hard to detect, the crash rate is ... OK I forgot... I think it's about <0.1%.

And this demo just show how to reproduce it with some magic hardcode, we continue to operate the `locationFeature` after we sent it to the map:

```
      locationSource.setGeoJson(locationFeature);
      for (int i=0; i<1000;i++) {
        locationFeature.removeProperty(i+"");
        locationFeature.addStringProperty(i+"", count+"");
        count++;
      }
```
Then the crash will be easily reproduced:
Screenshot 2023-04-04 at 15 56 57

The map crashed at the worker thread when it want to read the json java object.

So let's make a small discussion here, we can quickly fix it with `setGeoJson(String json)`:
```
      // This workaround will fix the problem, the ideal is that the locationFeature is not pass
      // to the worker thread, we convert it to string, and the worker thread will parse the json string
      // so there is no chance to trigger concurrency with the location Feature
      locationSource.setGeoJson(locationFeature.toJson());
```

Or we can send a copy to `setGeoJson` to prevent the use of the same java object:
```
      // Send a copy to the worker thread also work. because the feature passed to the worker thread is
      // a new object, not the feature in the main thread. The question here: what is the GeoJsonSource.setGeoJson
      // can handle this problem, it can make a copy for the feature or just convert the object locally to
      // prevent the concurrency between the main thread and worker threads.
     Feature locationFeatureCpy = Feature.fromJson(locationFeature.toJson());
     locationSource.setGeoJson(locationFeatureCpy);
```
Or directly make the copy operation inside the `setGeoJson`
```
 fun setGeoJson(feature: Feature?) {
        if (detached) {
            return
        }
        checkThread()
        if (feature == null) {
            nativeSetFeature(null)
        } else {
            nativeSetFeature(Feature.fromJson(feature.toJson()))
        }
    }
```

In GrabMap we change the convert operation from async to sync. but the fix also increases the cost at the main thread and the cost I think is equal with make a copy in setGeoJson.


louwers:
@alanchenboy I turned this into a Draft PR, feel free to change remove the draft status when you feel it is ready.

alanchenboy:
Now I demonstrate the best way that I think.

alanchenboy:
Hi everyone if you think it's too heavy to add safeguard at GeoJsonSource, we can change `SymbolLocationLayerRenderer` from `setGeoJson(feature)` to `setGeoJson(string)`, and add warnings comments at  `setGeoJson(feature)` that shouldn't operate the feature after it was sent to the map.

ovivoda:
@alanchenboy can we use flags to allow/disallow this? instead of warnings comments?

alanchenboy:
OK. let's turn feature update with flag, and location update with safe mode always.

alanchenboy:
Add flag safeSetGeoJson fore GeoJsonSource.

ovivoda:
@alanchenboy it looks good! Can you please also write a test for it? Then I think we are good to merge it! 

Thanks a lot for making the time to fix this!

alanchenboy:
I've added a unit test to demonstrate how the crash happens,  It's a Java exception but thrown at the map core worker thread, so it's impossible to capture the exception in the Unit test, so I write this unit test, if you uncomment `//source.setSafeSetGeoJson(false);` the crash can be reproduced.

alanchenboy:
All thread Pass, not it ready for review.

artakka:
So making a copy of the array does not help? That would have been the preferred solution probably.

alanchenboy:
@artakka Now I just make the copy of properties, it also can solve this problem.

artakka:
Great work. Do we still need the flag?  @ovivoda what do you think? If it still works as before and does not crash I suggest to just fix it without the flag.

ovivoda:
@artakka @alanchenboy if this fixes the initial problem we should remove the flag. 

@alanchenboy can you confirm that this fixed the initial issue and we do not have a crash if we remove the flag?

alanchenboy:
Yes, it can fix the initial problem, keeping the flag can give us a little benefit with creating a feature without modification after it is created.
 ```
      GeoJsonSource source = new GeoJsonSource("source");
      source.setSafeSetGeoJson(false);

      mapboxMap.getStyle().addSource(source);
      mapboxMap.getStyle().addLayer(new CircleLayer("layer", source.getId()));

      // Create a Feature and set it to the source 
      Feature feature = Feature.fromGeometry(Point.fromLngLat(20, 55));
      source.setGeoJson(feature);  // no copy for feature
      feature = null; // feature not modified anymore
```

alanchenboy:
Flag removed.

ovivoda:
@alanchenboy can you please fix the merge conflicts? Then I think this can be merged.

alanchenboy:
Rebased, but does anybody know why the iOS pipeline is broken? it looks not a random failure.

alanchenboy:
@ovivoda after rebased pipelines passed. please help review, thanks.

louwers (changes requested):
The test seems to include duplicated code and outdated comments.

alanchenboy:
duplicated unit test removed.

alanchenboy:
Now all ios and linux render test are passed. android is broken by system?

we could we update qt pipeline cmake to 3.15?

ntadej:
> we could we update qt pipeline cmake to 3.15?

I'll try to have a look this weekend (although I'm puzzled why it affects this PR).

alanchenboy:
@ntadej I replied in the wrong PR, I should reply in this PR https://github.com/maplibre/maplibre-native/pull/1439.

PR #980, opened 2023-04-04
Increase timeout Android instrumentation test

louwers:


louwers:
It does not time out any more at least. https://github.com/maplibre/maplibre-gl-native/actions/runs/4609128594/jobs/8146492289

Issue #974, opened 2023-04-03
Drawable Implementation: First Pass at Design


sjg-wdw:
Build enough of a drawable to support the simplest layer we can find.  Or the most basic, like Fill.

ovivoda:
@TimSylvester please update the issue description once you decide how to proceed. 
Eg. What will you build in the drawable and which layer will you support first.

TimSylvester:
Still in the very early stages, trying to establish where things are defined and used.  I will shortly create a draft PR for feedback on an initial stab at it, along with some discussion points.

sjg-wdw:
I'm turning the architectural decisions into Issues, so you can close this Issue out.

PR #983, opened 2023-04-04
First cut at Drawable fundamentals

TimSylvester:
I have created the generic `Drawable` along with `DrawableGL` and `DrawableGL::Impl`, trying to follow existing patterns.

I have drawables being instantiated by `RenderBackgroundLayer`, collected from the layers by `RenderTreeImpl`, and rendered alongside the layer render items in `Renderer::Impl`.  

I pulled in a bunch of the things that are passed along in the rendering, but it's by no means a correct or complete list.

I think much of the private implementation detail, e.g., `mat4` and `PaintParameters`, will need to be made public for API clients to be able to implement their own classes.  If so, that's going to be a sweeping change, and I'd prefer to handle it in a separate PR that doesn't have any practical effect on functionality.

#974


sjg-wdw:
The Class/Impl dichotomy makes sense because so many of those necessary objects are hidden.  After we make the Drawable transition, I'd like to see all that opened up at the API level.  Not necessarily all the way up to the Java and Swift, but at the C++ level at least.

I'm curious how other people feel about this.

sjg-wdw:
The drawable refers to things like ShaderIDs and RenderTargetIDs and other resources.  Looking at their implementation, they tend to be simple wrappers around GL IDs.

That worries me a bit because it means they have to be set up with OpenGL before we can refer to them.  It also means we might end up referring to a new one if we add, then remove, then add another one with the same ID.

In WhirlyGlobe we make use of Identifiables for a simpler purpose, but they're not tied to an SDK.  Might be good here.

sjg-wdw:
In our conversation, you mentioned the Uniform logic is a bit of hairball (my term).  I'm wondering if we want to come up with something new for that we can share between the new Shader subclass and a Drawable.

src/mbgl/renderer/render_tree.hpp
```diff
@@ -61,6 +62,7 @@ class RenderTree {
     // Render items
     virtual RenderItems getLayerRenderItems() const = 0;
     virtual RenderItems getSourceRenderItems() const = 0;
+    virtual std::vector<std::shared_ptr<gfx::Drawable>> getDrawables() const = 0;
```
mwilsnd, line 65: Can you return a const reference here? That way callers can decide if they want to make a copy or not.

src/mbgl/renderer/render_orchestrator.cpp
```diff
@@ -64,10 +64,12 @@ class RenderTreeImpl final : public RenderTree {
                    PatternAtlas& patternAtlas_,
                    RenderLayerReferences layersNeedPlacement_,
                    Immutable<Placement> placement_,
-                   bool updateSymbolOpacities_)
+                   bool updateSymbolOpacities_,
+                   std::vector<std::shared_ptr<gfx::Drawable>> drawables_ = {})
```
mwilsnd, line 68: This should probably be an rvalue reference if you're moving it

src/mbgl/renderer/render_layer.hpp
```diff
@@ -123,6 +124,8 @@ class RenderLayer {
     // TODO: Only for background layers.
     virtual std::optional<Color> getSolidBackground() const;
 
+    std::vector<std::shared_ptr<gfx::Drawable>> getDrawables() const { return drawables; }
```
mwilsnd, line 127: Ditto below, return a const ref

src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -140,6 +141,13 @@ void Renderer::Impl::render(const RenderTree& renderTree) {
 
     parameters.depthRangeSize = 1 - (layerRenderItems.size() + 2) * parameters.numSublayers * parameters.depthEpsilon;
 
+    // Drawables
+    {
+        for (const auto &drawable : layerDrawables) {
```
mwilsnd, line 146: And here iterate on renderTree.getDrawables() directly to avoid a copy of the render tree's vector

mwilsnd (changes requested):

mwilsnd:
> In our conversation, you mentioned the Uniform logic is a bit of hairball (my term). I'm wondering if we want to come up with something new for that we can share between the new Shader subclass and a Drawable.

Uniforms, attributes and textures are all specified per-program at compile time via templates. We'll need to move away from that as part of porting each render layer to drawables. That means adding a runtime path to configure programs with the texture+attribute+uniform parameters.

I'm thinking the best approach here is implementing a new version of mbgl::gl::Program/mbgl::Program, or just dropping that cruft all together. Currently programs configure the pipeline state and dispatch the draw call, along with being a container for the shader. A new class extending gfx::Shader can just deal with uniforms/textures/attributes and everything else should be the drawable's responsibility.

mwilsnd:
> The drawable refers to things like ShaderIDs and RenderTargetIDs and other resources. Looking at their implementation, they tend to be simple wrappers around GL IDs.
> 
> That worries me a bit because it means they have to be set up with OpenGL before we can refer to them. It also means we might end up referring to a new one if we add, then remove, then add another one with the same ID.
> 
> In WhirlyGlobe we make use of Identifiables for a simpler purpose, but they're not tied to an SDK. Might be good here.

The IDs are a custom implementation of a proposed standard addition, unique resources. Think unique_ptr, but for trivial types like integer. It's just a move only wrapper around GLuint. The shader will own it's set of GLuints for the various GL shaders it owns, with handles for other things like textures coming from other parts of the toolkit.

sjg-wdw:
> I'm thinking the best approach here is implementing a new version of mbgl::gl::Program/mbgl::Program, or just dropping that cruft all together. Currently programs configure the pipeline state and dispatch the draw call, along with being a container for the shader. A new class extending gfx::Shader can just deal with uniforms/textures/attributes and everything else should be the drawable's responsibility.

Yeah, I think that makes sense.  The draw call will belong to the Drawable, but the Shader will have a lot of its own logic and should have ways of linking up all that various information.

I wonder if we should just move to UBOs for the new Drawable approach.  The advantage would be we could skip all the uniform variants and go straight to buffers.  We can also pass these around a bit easier.  

We've found it's useful to put the equivalent of UBOs on shaders directly (for single-user shaders) and it's useful to allow them to be passed in from the outside.

In the first case imagine you're applying a color map to data.  So a data value corresponds to a color value.  It's useful to just set that color map on the shader, rather than having to associate it with every Drawable that you might feed in for the geometry.

src/mbgl/renderer/renderer_impl.cpp
TimSylvester: Some projects avoid that due to the possibility of the caller keeping the reference too long or otherwise abusing it, but if you're okay with it returning a const ref is my preference as well.

mwilsnd: At least in this case, the lifetime of the render tree is coupled directly with the renderer, so temporary references here are safe. Most important, making copies on the hot-path should only be done in cases where concurrency is required and atomics or locks are a worse option compared to the copy.

mwilsnd:
> > I'm thinking the best approach here is implementing a new version of mbgl::gl::Program/mbgl::Program, or just dropping that cruft all together. Currently programs configure the pipeline state and dispatch the draw call, along with being a container for the shader. A new class extending gfx::Shader can just deal with uniforms/textures/attributes and everything else should be the drawable's responsibility.
> 
> Yeah, I think that makes sense. The draw call will belong to the Drawable, but the Shader will have a lot of its own logic and should have ways of linking up all that various information.
> 
> I wonder if we should just move to UBOs for the new Drawable approach. The advantage would be we could skip all the uniform variants and go straight to buffers. We can also pass these around a bit easier.
> 
> We've found it's useful to put the equivalent of UBOs on shaders directly (for single-user shaders) and it's useful to allow them to be passed in from the outside.
> 
> In the first case imagine you're applying a color map to data. So a data value corresponds to a color value. It's useful to just set that color map on the shader, rather than having to associate it with every Drawable that you might feed in for the geometry.

Yeah, I'm working on the ES 3 upgrade right now and figure UBOs are going to make our lives a bit easier. I can see 1 global UBO for per-frame data, and each shader has it's own UBO for per-draw stuff. We still have to unwind [attributes](https://github.com/maplibre/maplibre-gl-native/blob/main/src/mbgl/gl/attribute.hpp) and [textures](https://github.com/maplibre/maplibre-gl-native/blob/main/src/mbgl/gl/texture.hpp) though which are specified at compile time on the program object.

sjg-wdw:
> > > I'm thinking the best approach here is implementing a new version of mbgl::gl::Program/mbgl::Program, or just dropping that cruft all together. Currently programs configure the pipeline state and dispatch the draw call, along with being a container for the shader. A new class extending gfx::Shader can just deal with uniforms/textures/attributes and everything else should be the drawable's responsibility.
> > 
> > 
> > Yeah, I think that makes sense. The draw call will belong to the Drawable, but the Shader will have a lot of its own logic and should have ways of linking up all that various information.
> > I wonder if we should just move to UBOs for the new Drawable approach. The advantage would be we could skip all the uniform variants and go straight to buffers. We can also pass these around a bit easier.
> > We've found it's useful to put the equivalent of UBOs on shaders directly (for single-user shaders) and it's useful to allow them to be passed in from the outside.
> > In the first case imagine you're applying a color map to data. So a data value corresponds to a color value. It's useful to just set that color map on the shader, rather than having to associate it with every Drawable that you might feed in for the geometry.
> 
> Yeah, I'm working on the ES 3 upgrade right now and figure UBOs are going to make our lives a bit easier. I can see 1 global UBO for per-frame data, and each shader has it's own UBO for per-draw stuff. We still have to unwind [attributes](https://github.com/maplibre/maplibre-gl-native/blob/main/src/mbgl/gl/attribute.hpp) and [textures](https://github.com/maplibre/maplibre-gl-native/blob/main/src/mbgl/gl/texture.hpp) though which are specified at compile time on the program object.

Definitely one global UBO one passed to everything.
But beyond that I wonder if we can name them and it might be useful to have a more flexible number.
One for the Layer which generates the Drawable.
One potentially for Shader-wide values, but we could allow that to be attached to a Drawable too.
One optional one per Texture.  

As to why textures, we do some trickery during our own tile loading that substitutes a lower res texture for one we're trying to load.  It's just a few offset and scale values, but I was considering how you'd implement that without the renderer understanding what you're up to.

But yeah, we still have to address vertex attributes which are annoyingly complex due to how they can be interleaved or not and they have types, some of which are very similar.  All have important use cases.

TimSylvester:
If drawables are to be created/updated/deleted by serialized change requests, there will need to be a master list somewhere rather than pulling them out of the layers on each pass.  The `RenderTree` is transient, so that seems to leave `RenderOrchestrator`, or is there a better place for that to live?

sjg-wdw:
> If drawables are to be created/updated/deleted by serialized change requests, there will need to be a master list somewhere rather than pulling them out of the layers on each pass. The `RenderTree` is transient, so that seems to leave `RenderOrchestrator`, or is there a better place for that to live?

That spot may not yet exist.  We do it with an explicit Scene.  Perhaps we want to move that direction with MapLibre, but one step at a time.

mwilsnd:
> If drawables are to be created/updated/deleted by serialized change requests, there will need to be a master list somewhere rather than pulling them out of the layers on each pass. The `RenderTree` is transient, so that seems to leave `RenderOrchestrator`, or is there a better place for that to live?

The orchestrator sounds like a good spot for now. That whole createRenderTree method is going to need an overhaul at some point regardless. We can look at moving to an abstract scene concept later down the line.

TimSylvester:
How would you feel about managing this as a branch here so that I can do incremental PRs into it, rebase it, etc., rather than continuing to expand this one?  Then we can merge it to main with a PR when that makes sense.

sjg-wdw:
> How would you feel about managing this as a branch here so that I can do incremental PRs into it, rebase it, etc., rather than continuing to expand this one? Then we can merge it to main with a PR when that makes sense.

I think that approach is appropriate for something this complex.

include/mbgl/util/identity.hpp
sjg-wdw, line 1: Just "id" might cause some problems with Objective-C++

ovivoda:
> The Class/Impl dichotomy makes sense because so many of those necessary objects are hidden. After we make the Drawable transition, I'd like to see all that opened up at the API level. Not necessarily all the way up to the Java and Swift, but at the C++ level at least.
> 
> I'm curious how other people feel about this.

I totally support this initiative. This will make the render part very flexible.

sjg-wdw:
Getting close to a generic vertex attribute handling system I think.

src/mbgl/renderer/layers/render_background_layer.cpp
```diff
@@ -69,6 +124,43 @@ void RenderBackgroundLayer::render(PaintParameters& parameters) {
...
+            //#version 300 es
+            precision highp float;
+            //out vec4 color;
+            void main() {
+                //color = vec4(1.0,0.0,0.0,1.0);
+                gl_FragColor = vec4(1.0,0.0,0.0,1.0);
+            })";
+
+        try {
+            shader = gl::ShaderProgramGL::create((gl::Context&)parameters.context, shaderName, vert, frag);
```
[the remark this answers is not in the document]
TimSylvester: Agreed, though when this code is moved to where it belongs, no cast should be necessary.

sjg-wdw:
Looks like you're getting close to rendering the background layer.
Shall we set up a meeting to go over the structure of things after that's working?

TimSylvester:
I don't have the same feeling that it's close, but I'm open to a discussion.

sjg-wdw:
> I don't have the same feeling that it's close, but I'm open to a discussion.

You're closer to it than I.  Keep on and let me know if I can do anything.

TimSylvester:
I think I'm on the right track, because a bunch of stuff I was doing collapsed into nothing when I started using methods on `UploadPass`, and I only had to adjust visibility a little to expose some things.  But I have some questions:

I think [gl::UploadPass::buildAttributeBindings](https://github.com/maplibre/maplibre-gl-native/pull/983/files#diff-8d5b129968f78724fb7596e0b4430cb29958fedd8e1de6266201d0743ca6ea51) needs to return the buffer because it needs to live at least as long as the bindings, right?

And the renderer, or wherever this ends up living, needs to store that buffer along with the VAO on the drawable, and keep track of whether it's "stale" because something changed, right?

[render_impl](https://github.com/maplibre/maplibre-gl-native/pull/983/files#diff-17f957df4b134f724f7b7afb4fb6a169c8240a1580722e8fcc203491e59e40c8)

sjg-wdw:
I would think those would live in the gl Drawable and get rebuilt as necessary based on the dirty flag being set, yeah.
I don't know if buffer management needs to be particularly general.  The concepts are the similar between SDKs, but specifics can get a little wonky.

TimSylvester:
Reset as #1026

PR #984, opened 2023-04-04
iOS size test

alexcristici:
Building the app for 2 configurations: Release and MinSizeRel, and comparing the MapLibre dynamic library size with saved sizes before starting the modularization and migration to Metal.

### DIFF
--- a/platform/ios/Makefile
+++ b/platform/ios/Makefile
@@ -2,6 +2,7 @@ export BUILDTYPE ?= Debug
 export IS_LOCAL_DEVELOPMENT ?= true
 export TARGET_BRANCH ?= main
 export CI ?= 0
+export DEV_TEAM ?= ""
 
 # When only the dependencies in the package.json are installed
 # Node.js will error when running a script from the root scripts directory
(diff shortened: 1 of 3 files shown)


sjg-wdw:
Over the years Xcode has moved its build around, changed its arguments and such.
I wonder if you could put some checks around the xcode invocation and double check that the build directory exists.

TimSylvester (approved):
We're depending on the signing key being set up in the test runner, right?  I don't see anything about it here, which is good.

alexcristici:
> We're depending on the signing key being set up in the test runner, right? I don't see anything about it here, which is good.

Yes, the command line for running the size test is: `make ios-sizetest DEV_TEAM=[TEAM_ID]`. Maybe we will change to use the prov and signing certificate directly if needed. We will see when we will reach the point to add it on GitHub CI.

mwilsnd (approved):

sjg-wdw:
> Over the years Xcode has moved its build around, changed its arguments and such. I wonder if you could put some checks around the xcode invocation and double check that the build directory exists.

To provide a bit more context, I've had framework building scripts that have broken several times over the years and it's never clear why.  It would be nice to provide whomever deals with this in 3 years a starting point for which part of it is failing.

sjg-wdw:
> > > Over the years Xcode has moved its build around, changed its arguments and such. I wonder if you could put some checks around the xcode invocation and double check that the build directory exists.
> > 
> > 
> > To provide a bit more context, I've had framework building scripts that have broken several times over the years and it's never clear why. It would be nice to provide whomever deals with this in 3 years a starting point for which part of it is failing.
> 
> Yes, I added checks after each step. Let me know if I missed something.

Looks great!

sjg-wdw (approved):

louwers (approved):
I can’t try this out yet but LGTM.

Feel free to merge.

PR #985, opened 2023-04-04
Fix macOS CI by running tests in graphical env

louwers:
This PR fixes the macOS CI.

Turns out Apple does not like people running tests in an headless environment (that is not Xcode server). Some people are [reporting this](https://stackoverflow.com/a/71417206) at least. So using this [horrible hack](https://aahlenst.dev/blog/accessing-the-macos-gui-in-automation-contexts/#continuous-integration-with-ssh) I am spawning a Terminal.app window and running the tests in there.


louwers:
Fails with 

```
The application /System/Applications/Utilities/Terminal.app cannot be opened for an unexpected reason, error=Error Domain=RBSRequestErrorDomain Code=5 "Launch failed." UserInfo={NSLocalizedFailureReason=Launch failed., NSUnderlyingError=0x60000300dc20 {Error Domain=OSLaunchdErrorDomain Code=125 "Domain does not support specified action" UserInfo={NSLocalizedFailureReason=Domain does not support specified action}}}
```

louwers:
Enabling Automatic Login worked. 🎉

birkskyum:
For linux we do something like this:

https://github.com/maplibre/maplibre-gl-js/blob/1b3242280da112118e6ec9ec603c2dccd94d1741/.github/workflows/test-all.yml#L57-L61

Don't know if there is a macos counterpart

louwers:
Someone else with the same problem:

https://developer.apple.com/forums/thread/727333

@birkskyum macOS works completely different from what I understand. From the blog post:

> If no user has logged into the machine by using the GUI, there is no GUI. And you cannot do anything about it except logging in by using the GUI, either manually or by [enabling automatic login](https://aahlenst.dev/blog/accessing-the-macos-gui-in-automation-contexts/#enabling-automatic-login-on-macos).

birkskyum (approved):
It seems like the only way to do it then. Weird decision of apple

louwers:
> It seems like the only way to do it then. Weird decision of apple

@birkskyum Not so weird if you look at these prices. 😅 https://developer.apple.com/xcode-cloud/

PR #987, opened 2023-04-05
Organize source code based on tree folders

alexcristici:
I organized only the core files and iOS files.
For organizing the other platform specific files, just follow the same logic as in `ios.cmake`.


sjg-wdw:
I'm excited for this one, but I'm curious how the maintainers feel.

sjg-wdw (approved):

TimSylvester (approved):
So much nicer, thank you!

ntadej (approved):
No objections for me if it helps how Xcode project is structured. I suspect not many other platforms are affected by external project organisation so it's fine to only do it for iOS at this point.

mwilsnd (approved):
This also works when generating a Visual Studio solution. We can change the Windows CMake later though, no rush!

PR #989, opened 2023-04-06
Add custom name support to shader registry

mwilsnd:
This PR introduces support for shader registry names distinct from gfx::Shader-derived type names. This enables registration of many instances of the same shader type, all under different names, to support runtime-configurable shaders.

Adds new tests and changes the type name method `name` to `typeName` to better communicate intent.

### DIFF
--- a/include/mbgl/gfx/shader.hpp
+++ b/include/mbgl/gfx/shader.hpp
@@ -11,7 +11,7 @@ class Shader;
 // Assert that a type is a valid shader for downcasting.
 // A valid shader must:
 //   * Inherit gfx::Shader
-//   * Have properly called DECLARE_SHADER_TYPEINFO in the class body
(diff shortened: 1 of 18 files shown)


sjg-wdw (approved):
Makes sense.

alexcristici (approved):

TimSylvester (approved):

TimSylvester:
Does the shader object need to have the name by which it's known?  It could be simpler if it were only in the registry.

mwilsnd:
> Does the shader object need to have the name by which it's known?  It could be simpler if it were only in the registry.

Shader objects are only required to provide a type name for checked casting. The optional shader name is up to you, store it how you see fit. Refer to the added tests for an example use. If you don't provide a custom name via the new overloads the registry will fall back to the type name.

ovivoda (approved):

Issue #893, opened 2023-03-12
Replace custom variant with `std::variant`


louwers:
The several implementations of [variant](https://github.com/search?q=repo%3Amaplibre%2Fmaplibre-gl-native%20variant&type=code) can be replaced by `std:variant`. Thanks @ntadej for pointing it out.

bencsikandrei:
Hello,

I want to look into this one if there isn't already an ongoing effort.

louwers:
@bencsikandrei Sorry for getting back to you late.

Would be great! Thanks.

louwers:
```
mapbox variant: 12561us
boost variant: 14392us
std variant: 10113us
mapbox variant: 7604us
boost variant: 10809us
std variant: 8073us
mapbox variant: 6388us
boost variant: 9566us
std variant: 7497us
```

Benchmark mapbox variant vs std::variant

louwers:
`std::variant` is bloated.

PR #991, opened 2023-04-07
[WIP] replace mbgl::variant with std::variant

bencsikandrei:
This is part 1/n of a series of commits (or PRs?) towards solving: #893 

My idea for this PR is to not aim at fully resolving this issue, as I believe it would be too much for one PR, but rather make small increments towards that goal (solve the simple things first, etc...)

My decision was to split into at a couple commits since I have some concerns regarding replacing some of the functionality of mbgl::variant with std::variant equivalents (.match and some of its operator== semantics come to mind)

Please let me know if my decision suits what you have envisioned for this task

Thanks!

### DIFF
--- a/expression-test/expression_test_parser.cpp
+++ b/expression-test/expression_test_parser.cpp
@@ -6,7 +6,6 @@
 
 #include <mbgl/util/io.hpp>
 #include <mbgl/util/logging.hpp>
-#include <mbgl/util/variant.hpp>
 #include <mbgl/style/conversion/function.hpp>
 #include <mbgl/style/rapidjson_conversion.hpp>
 #include <mbgl/style/expression/parsing_context.hpp>
@@ -19,6 +18,7 @@
 #include <args.hxx>
 
 #include <regex>
+#include <variant>
 
 using namespace mbgl;
 using namespace mbgl::style;
@@ -52,7 +52,7 @@ void writeJSON(rapidjson::PrettyWriter<rapidjson::StringBuffer>& writer, const V
 }
 
 using ErrorMessage = std::string;
-using JSONReply = variant<JSDocument, ErrorMessage>;
+using JSONReply = std::variant<JSDocument, ErrorMessage>;
 JSONReply readJson(const filesystem::path& jsonPath) {
     auto maybeJSON = util::readFile(jsonPath);
     if (!maybeJSON) {
@@ -384,11 +384,11 @@ Ignores parseExpressionIgnores() {
     Ignores ignores;
     const auto mainIgnoresPath = filesystem::path(TEST_RUNNER_ROOT_PATH).append("metrics/ignores/platform-all.json");
     auto maybeIgnores = readJson(mainIgnoresPath);
-    if (!maybeIgnores.is<JSDocument>()) { // NOLINT
+    if (!std::holds_alternative<JSDocument>(maybeIgnores)) { // NOLINT
         return {};
     }
 
-    for (const auto& property : maybeIgnores.get<JSDocument>().GetObject()) {
+    for (const auto& property : std::get<JSDocument>(maybeIgnores).GetObject()) {
         std::string id{toString(property.name)};
         // Keep only expression-test ignores
         if (id.rfind("expression-tests", 0) != 0) {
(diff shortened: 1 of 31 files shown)


expression-test/expression_test_parser.cpp
```diff
@@ -407,12 +407,12 @@ Ignores parseExpressionIgnores() {
 std::optional<TestData> parseTestData(const filesystem::path& path) {
     TestData data;
     auto maybeJson = readJson(path.string());
-    if (!maybeJson.is<JSDocument>()) { // NOLINT
+    if (!std::holds_alternative<JSDocument>(maybeJson)) { // NOLINT
```
bencsikandrei, line 410: I'm not sure what was linted here, if someone could help me with that one I'd happily check and maybe remove the NOLINT comment

I'll do some searching in the meanting to try to figure out what analyzer I should run to test this

include/mbgl/util/overloaded.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+namespace mbgl
+{
+namespace util
+{
+    
+template <typename... Ts>
```
louwers, line 8: Everything in `include` will become part of our public API. Is this intended? Can you add some triple slash comments to indicate what this template struct is for?

louwers (changes requested):
Looks great overall but I would like to know why the `Overload` helper is needed.

expression-test/expression_test_parser.cpp
louwers: Could you try removing it? If the CI passes we can remove it.

I think Mapbox used some kind of linter in the past, but we are not (yet).

bencsikandrei:
> Looks great overall but I would like to know why the `Overload` helper is needed.

the API for `std::visit` is quite different from the one in `mapbox::variant::match`, in that it favors applying the same visitor to multiple variant objects (via variadic template arguments) as opposed to multiple callables on the same variant object

the API is as follows:
`template <class Visitor, class... Variants>
constexpr auto visit( Visitor&& vis, Variants&&... vars );`

which means we've got to pass 1 visitor for N (which could also be 0) variants

in order to obtain the same behavior as `mapbox::variant::match` (which AFAIK takes a variadic template argument pack for multiple function like objects) we need to pass some sort of function like object which contains all the visit functions we need

one way to do so is to manually create a visitor with multiple overloads for operator()(...)

e.g.: `struct Visitor { void operator()(int); void operator()(double); ..};` which we could apply to a `variant<int, double>`, but that's tedious and we already have the lambdas from the old calls to `match` so here's where the `Overloaded` thing comes in

it effectively inherits from a bunch of lambdas (so it has all their operator()(...)) in a single place (it, unfortunately, also needs a deduction guide for c++17 to allow for {..} initialization from the lambdas)

maybe I made the explanation way too long sorry for that!

include/mbgl/util/overloaded.hpp
bencsikandrei: agreed! I believe there's not need for it to be public but I have to check

bencsikandrei: done, I moved it to src/mbgl/util

bencsikandrei:
Hi!

I've been working on this on and off but have struggled with a couple things in the public headers (notably changes involving some stuff that mbgl::variant and std::variant handle differently - mostly related to operator== and co.).

In any case since this has been a long PR and has been lingering I've had to rebase several times.

I believe the changes that exist now (that don't touch public headers) could be merged in a first step.

Then we can move to the other (fewer) changes to the public headers and a couple platform things. Those will, hopefully, be faster to do than this one.

I'd rather have everything done in one PR but I've not had the time to figure out what happens to a couple headers when I changed to std::variant (the compiler cries for help with errors).

Keeping this up to date will become increasingly harder if it lingers and I'd rather at least make a small step.

What do you think?

louwers:
@bencsikandrei Sorry for never replying, I am afraid I don't understand that error either.

Let's see if we can get this in a mergable state.

louwers:
Closing because stale, I have opened https://github.com/maplibre/maplibre-native/pull/2946

PR #994, opened 2023-04-10
Fix for recent breaking changes in conversion of Java to Kotlin 

stevehayles:


louwers (changes requested):
Please give a bit more context what this PR is trying to accomplish.

artakka:
Looks good to me but also do not understand the need for AndroidXRuntime

stevehayles:
> Please give a bit more context what this PR is trying to accomplish.

The recent conversion of the Java code to Kotlin introduces breaking changes to any calling code (including other Maplibre libraries like flutter-malibre-gl) due to the difference in Kotlin's non-handling of static methods and properties.  This fix attempts to remain both forward and backward compatible but introducing delegated properties and the JvmStatic compiler annotation to avoid Java code having to refer to the singleton instance via the 'companion' object. 

> Does it build without the `androidxRuntime` dependency?

It does not build for me without the  `androidxRuntime` dependency.  It's introduction can be avoided by changing the property names but that will break any forward compatibility

louwers:
@stevehayles Makes sense. Thank you for clarifying.

@artakka What would you like to do? Do you know if AndroidXRuntime increases the size of the library?

artakka:
Should be good

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.kt
```diff
@@ -5,6 +5,7 @@ import android.os.Handler
 import android.os.Looper
 import androidx.annotation.IntDef
 import androidx.annotation.Keep
+import androidx.compose.runtime.getValue
```
boldtrn, line 8: Why was this import required? It's a bit weird that we have to add a new dependency just for for that?

stevehayles: The Kotlin rewrite changed the property name from 'GetID' to 'GetId' and to maintain forward and backward compatibility I have introduced a a delegated property `val ID: Long by this::id` which requires this import.  If a decision is made to target backward compatibility only then this can obviously be dropped

ovivoda: @boldtrn I managed to build and run even if I commented out this line.

boldtrn: >  'GetID' to 'GetId' 

Since this was changed in a major release it should be fine to rename this? But maybe we should add it to the Changelog to make it clear to everyone that it was changed? IMHO backwards compatibility for major releases can break.

Thanks for verifying @ovivoda.

stevehayles: As you point out, it was changed in a major release but it still seems a shame to break existing code when the intention of a rewrite from Java to Kotlin was not to introduce any new functionality.  I am just a user of the library, not my call on how best to proceed, I'm happy to close the PR and fix it in [flutter-maplibre-gl](https://github.com/maplibre/flutter-maplibre-gl). Thanks for the amazing library!

PR #995, opened 2023-04-10
Upgrade OpenGL backend to ES 3.0

mwilsnd:
This PR upgrades the shaders and GL contexts of all platforms (minus darwin) to OpenGL ES 3.0. A large chunk of changes here are in the shaders & generated shader code (mbgl/shaders/**, shaders/**) where the following changes were made:
* `#version 300 es` support
  * attribute -> `layout (location = n) in`
  * varying -> in/out
  * texture2D -> texture
  * gl_FragColor -> explicit output color
  * The shader code generator was upgraded to emit vertex attribute layout qualifiers

All instances of MBGL_USE_GLES2 have been removed, with older pathways for ES 1 stuff deleted. All context creation methods have been updated to request ES 3 or ES 3 compatible (windows) contexts, save for macOS's CoreGL backend (support here is pending, may be done via ANGLE in the future).


include/mbgl/shaders/gl/circle.hpp
```diff
@@ -15,56 +15,55 @@ uniform vec2 u_extrude_scale;
 uniform lowp float u_device_pixel_ratio;
 uniform highp float u_camera_to_center_distance;
 
-attribute vec2 a_pos;
-
-varying vec3 v_data;
+layout (location = 0) in vec2 a_pos;
```
alexcristici, line 18: I'm not sure, but `layout (location = 0)` is mandatory?

mwilsnd: Coming from desktop GL I'm assuming it is, I can see what happens without layout qualifiers though

mwilsnd: [Looks like you're probably right](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf) reading the spec (pg. 40).

alexcristici: From what I remember I know that I didn't use it in past project on Android. I know that layout location is needed only if we don't what to call the `glGetVertexAttribLocation`.

sjg-wdw: They're not required, but also not a bad idea.

sjg-wdw (approved):
Nice to see the 3.0 syntax in the shaders.

TimSylvester (approved):
Merges and works fine with my pending changes 👍

alexcristici (approved):

ntadej:
No, Qt does not work anymore on macOS. I guess then Node.js would also be broken. What's the motivation to move to GLES 3.0?

mwilsnd:
> No, Qt does not work anymore on macOS. I guess then Node.js would also be broken. What's the motivation to move to GLES 3.0?

Any platform built for macOS is going to fail with this PR - Qt, node, render-test-runner and so on. Moving to ES 3 gives us a few important features, the two big ones are uniform buffer objects (UBOs) and Vertex array objects (VAOs). Using these, we can significantly simplify the existing GL rendering code and bring it more inline with what we anticipate the coming Metal backend will look like. ES 3 also brings increased minimums on certain implementation limits like number of vertex attributes, which increases to 16 with ES 3, removing one of the reasons for keeping the existing shader permutation system around.

The current plan then, is merge this PR and shut off macOS on CI. Existing users on macOS will have to remain on the last stable release until the Metal platform is ready later this year. Once Metal is ready, the darwin platform (and all dependents like glfw, Qt, node) can be moved to the new native backend and revived.

The other option is to bridge the gap with ANGLE on mac, but this is extra temporary work until Metal is implemented. This option should only be entertained if there is significant enough demand for it.

ntadej:
OK, then we should cut releases before this is merged. Also this motivates more repo split as I effectively can't develop anymore mac being my primary platform.

To be safe, we should also manually test other platforms I guess (Windows is another platform I'm worried about).

Let's discuss this tomorrow how to best proceed (and if you're willing to eventually help supporting Metal in Qt 😉).

sjg-wdw:
Does Qt not support Metal?

ntadej:
Qt6 fully supports Metal (Qt5 I'd have to check). The problem is I barely know bits of OpenGL so I'll need help gluing things together. Unfortunately they're missing some Metal examples that they have for Vulcan and D3D11. It will probably be minimal work for an expert though.

TimSylvester:
It looks like there are new definitions that should be added to `mbgl/gl/defines.hpp`, e.g., `GL_UNSIGNED_INT_VEC2` from [https://docs.gl/es3/glGetActiveAttrib](https://docs.gl/es3/glGetActiveAttrib).

mwilsnd:
> It looks like there are new definitions that should be added to `mbgl/gl/defines.hpp`, e.g., `GL_UNSIGNED_INT_VEC2` from https://docs.gl/es3/glGetActiveAttrib.

Being addressed in #1007

louwers (approved):

louwers:
@ntadej OK to merge?

ntadej:
First the CI should be green (so failing tests disabled) 😊

We may want to tag some releases before we merge this though. I'll be travelling tomorrow but I will be available on the weekend. I think @acalcutt also wanted to tag a release before this is merged.

Can this wait a few days? I can also merge it when green when I'm ready (and the CI is green).

louwers:
> Can this wait a few days?

Not really I think. This is a blocker for quite a lot of other things. (cc @sjg-wdw)

If you want to make a release, it's possible to do it from the `opengl-2` branch?

acalcutt:
I tried to do the last node release i wanted to do but it failed to bump the version. wonder if that node-pregyp-githib secret expired @birkskyum ?

ntadej:
OK, I can also tag from some older commit the way how the release is set-up (only the changelog may not be OK). So from my side you're OK to proceed.

birkskyum:
@acalcutt  it probably did

birkskyum:
@acalcutt , unless github has made some updated to their personal token settings, my best guess is that it's caused by the branch protection in this repo. Currently the branch protection setting for `main` has enabled "Do not allow bypassing the above settings", which is good practice, but if I don't remember wrong the node deployment rely on auto-pushing a commit directly to main effectively bypassing the PR requirement.

acalcutt:
Before we were using that secret to get passed the branch protection on the version bump, since we checked out with that user also. must not be working anymore. ill have to look into it and see if i have the same issue on my fork. not sure the workflow is set up to build a release for a branch..hmm.

probably wont be able to look until later

louwers:
@ntadej We had a discussion about keeping macOS CI running on `main`.

In the end @mwilsnd, @alexcristici and me decided against it. We can apply some tricks to make the project build (but not run) for macOS on `main`, but this creates the false expectation that macOS will work on `main`. As we decided during the TSC meeting, until the Metal migration is complete macOS will be broken on `main` and only functioning on the `opengl-2` branch. Since macOS support is one of the [north stars](https://github.com/maplibre/maplibre-gl-native/blob/main/design-proposals/2022-11-29-metal-port.md#north-stars) of the design proposal and thus one of the deliverables of the team, I think we can be confident that the team will get to bringing back macOS support in due time.

birkskyum:
@acalcutt , the token used on the fork, can you see what permissions it has? The one used here has only `public_repo, repo_deployment`, and I actually wouldn't expect that to work because they are read-only... I wonder if they changed this list of permissions. Would expect something like `write:repo_hook` to be needed

tdcosta100:
Hi guys. I know it's been a while since this PR was merged, but [`platform/windows/FindOSMesa.cmake`](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/windows/FindOSMesa.cmake#L19) doesn't fit these changes, because Mesa3D doesn't support GLES 3.0, only [GLES 1.1 and GLES 2.0](https://docs.mesa3d.org/opengles.html).

I'm planning to submit a PR to fix this, please validate my plans:

1. Drop Mesa3D support in `main` branch, since the minimum supported version is 3.0
1. Move Mesa3D support to `opengl-2` branch

Is this the correct way of doing this change?

Thank you in advance.

acalcutt:
@tdcosta100 , does that depend on the driver used? from this page it seems like mesa does support ES 3.0

https://mesamatrix.net/

tdcosta100:
Nice, I will try to test it creating a GLES 3 surface. If it works, I will just adjust it (Mesa3D doesn't have a libGLESv3.dll file).

tdcosta100:
Yeah, it works, but interesting is: the DLL name is libGLESv2.dll, even it holds all the ES 3.0 functions. I will submit a new PR correcting this.

PR #1007, opened 2023-04-11
API links for OpenGL ES 3.0

alexcristici:
OpenGL ES 3.0 API links for iOS, Android, macOS, Linux, Windows and Qt.
Removed vertex array object as extension, now it's available in GLES 3.0.
Also updated GLFunctions test.

Blocked by https://github.com/maplibre/maplibre-gl-native/pull/995, which needs to be merged first.


sjg-wdw (approved):

sjg-wdw (approved):

src/mbgl/gl/context.cpp
```diff
@@ -98,7 +98,7 @@ void Context::initializeExtensions(const std::function<gl::ProcAddress(const cha
         // Block Adreno 4xx as it crashes in a driver when VBOs are destructed (Android 5.1.1)
         // Block ARM Mali-T720 (in some MT8163 chipsets) as it crashes on glBindVertexArray
         // Block ANGLE on Direct3D as the combination of Qt + Windows + ANGLE leads to crashes
-        if (renderer.find("Adreno (TM) 2") == std::string::npos &&
+        /*if (renderer.find("Adreno (TM) 2") == std::string::npos &&
```
louwers, line 101: Why comment this instead of removing it?

louwers (changes requested):
Can the dead code be removed instead of commented out?

mwilsnd (approved):

src/mbgl/gl/context.cpp
alexcristici: Should be removed, not sure why remained commented here. I will fix it now.

TimSylvester (approved):

louwers (approved):

platform/darwin/src/gl_functions.cpp
```diff
@@ -36,11 +38,7 @@ void (* const glBufferSubData)(GLenum, GLintptr, GLsizeiptr, const void *) = ::g
 GLenum (* const glCheckFramebufferStatus)(GLenum) = ::glCheckFramebufferStatus;
 void (* const glClear)(GLbitfield) = ::glClear;
 void (* const glClearColor)(GLfloat, GLfloat, GLfloat, GLfloat) = ::glClearColor;
-#if !TARGET_OS_IPHONE && !TARGET_OS_SIMULATOR && TARGET_OS_MAC
```
ovivoda, line 39: @alexcristici what was this used for? why you removed it?

src/mbgl/gl/context.cpp
```diff
@@ -93,23 +93,6 @@ void Context::initializeExtensions(const std::function<gl::ProcAddress(const cha
               && renderer.find("Direct3D") != std::string::npos)) {
             debugging = std::make_unique<extension::Debugging>(fn);
         }
-
```
ovivoda, line 96: these are removed because are not supported anymore right?

test/map/map.test.cpp
```diff
@@ -727,21 +727,6 @@ TEST(Map, AddLayer) {
     test::checkImage("test/fixtures/map/add_layer", test.frontend.render(test.map).image);
 }
 
-TEST(Map, WithoutVAOExtension) {
```
ovivoda, line 730: can you please elaborate on why was this removed?

louwers: From what I understand since we are on OpenGL ES 3.0 now so we have VAO Extensions, so we don't need this test anymore.

louwers: https://www.khronos.org/opengl/wiki/Vertex_Specification#Vertex_Array_Object

PR #1012, opened 2023-04-13
Run macOS workflows only on `opengl-2` branch

louwers:
I created a new branch `opengl-2` that will stay on OpenGL 2.0 while the renderer modularization and Metal backend implementation are underway. This branch can still be used with macOS. The `main` branch will start using OpenGL ES 3.0 after https://github.com/maplibre/maplibre-gl-native/pull/995 is merged.

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - gles-2
+      - opengl-2
     tags:
       - 'macos-*'
     paths:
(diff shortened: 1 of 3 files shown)


ntadej:
We should still build on main to make sure the code compiles. Only tests should be disabled.

louwers:
@ntadej Do you think we should disable the tests or still run them for the `opengl-2` branch?

ntadej:
In `main`, everythng should run but (render) tests for macOS.

In `opengl-2`, everything should run as now.

louwers:
@ntadej I enabled the macOS Objective-C tests on `main`.

The Node CI fails to build due to a missing header on macOS. 

> In `opengl-2`, everything should run as now.

This is the case.

mwilsnd:
Let me see what all is involved in keeping mac building on main - we probably need to create stub gl* functions to satisfy the linker.

alexcristici (approved):

PR #1018, opened 2023-04-15
Wayland Support

jwinarske:
To build on Linux pass `-DMBGL_WITH_EGL=ON`

-expects GLFW built for Wayland
-defaults to Wayland over X11

### DIFF
--- a/platform/glfw/glfw_view.cpp
+++ b/platform/glfw/glfw_view.cpp
@@ -559,9 +559,9 @@ void GLFWView::updateFreeCameraDemo() {
 }
 
 mbgl::Color GLFWView::makeRandomColor() const {
-    const auto r = static_cast<float>(std::rand()) / RAND_MAX;
-    const auto g = static_cast<float>(std::rand()) / RAND_MAX;
-    const auto b = static_cast<float>(std::rand()) / RAND_MAX;
+    const auto r = static_cast<float>(std::rand()) / static_cast<float>(RAND_MAX);
+    const auto g = static_cast<float>(std::rand()) / static_cast<float>(RAND_MAX);
+    const auto b = static_cast<float>(std::rand()) / static_cast<float>(RAND_MAX);
     return { r, g, b, 1.0f };
 }
 
--- a/platform/linux/linux.cmake
+++ b/platform/linux/linux.cmake
@@ -1,10 +1,15 @@
+option(MBGL_WITH_X11 "Build with X11 Support" ON)
+option(MBGL_WITH_WAYLAND "Build with Wayland Support" OFF)
+
 find_package(CURL REQUIRED)
 find_package(ICU OPTIONAL_COMPONENTS i18n)
 find_package(ICU OPTIONAL_COMPONENTS uc)
 find_package(JPEG REQUIRED)
 find_package(PNG REQUIRED)
 find_package(PkgConfig REQUIRED)
-find_package(X11 REQUIRED)
+if (MBGL_WITH_X11)
+    find_package(X11 REQUIRED)
+endif ()
 find_package(Threads REQUIRED)
 
 pkg_search_module(LIBUV libuv REQUIRED)
@@ -64,6 +69,13 @@ if(MBGL_WITH_EGL)
         PRIVATE
             OpenGL::EGL
     )
+    if (MBGL_WITH_WAYLAND)
+        target_compile_definitions(mbgl-core PUBLIC
+                EGL_NO_X11
+                MESA_EGL_NO_X11_HEADERS
+                WL_EGL_PLATFORM
+        )
+    endif()
 else()
     find_package(OpenGL REQUIRED GLX)
     target_sources(
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -8,6 +8,7 @@ on:
     tags:
       - 'node-*'
     paths:
+      - "platform/linux/**"
       - "platform/default/**"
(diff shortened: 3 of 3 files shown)


CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ option(MBGL_WITH_OPENGL "Build with OpenGL renderer" ON)
 option(MBGL_WITH_EGL "Build with EGL renderer" OFF)
 option(MBGL_WITH_OSMESA "Build with OSMesa (Software) renderer" OFF)
 option(MBGL_WITH_WERROR "Make all compilation warnings errors" ON)
+option(MBGL_WITH_X11_SUPPORT "Build with X11 Support" OFF)
```
ntadej, line 12: I wonder if it should be just `MBGL_WITH_X11` like other options?

Also should these be only in `linux.cmake` as they do not seem to be relevant to other platforms?

platform/linux/linux.cmake
```diff
@@ -4,11 +4,21 @@ find_package(ICU OPTIONAL_COMPONENTS uc)
...
-find_package(X11 REQUIRED)
+if (MBGL_WITH_X11_SUPPORT)
+    find_package(X11 REQUIRED)
+endif ()
 find_package(Threads REQUIRED)
 
 pkg_search_module(LIBUV libuv REQUIRED)
 
+if (MBGL_WITH_WAYLAND_SUPPORT)
+    add_compile_definitions(
```
ntadej, line 15: We try to avoid global definitions. Are they really needed or can they be added to the relevant target (maybe even `mbgl-compile-options` would do)?

ntadej (changes requested):

CMakeLists.txt
jwinarske: good point. will move

platform/linux/linux.cmake
jwinarske: I was avoiding having to define them twice.  Switched to target_compile_definitions for each

ntadej: If you define them as `PUBLIC` they will get propagated with dependencies.

jwinarske: yes I'm very familiar with CMake

jwinarske:
Pushed changes

platform/linux/linux.cmake
ntadej: Sorry, I just wanted to be helpful 😊

ntadej (approved):
Thank you for the contribution! This is OK from my side now.

ntadej:
Actually I see no CI pipeline spawned so I'll tag @acalcutt and @louwers for feedback before merging.

jwinarske:
Build passed
https://github.com/maplibre/maplibre-gl-native/actions/runs/4707278115/jobs/8349053136

Let me know if a Linux CI job is desired

ntadej:
I'm more worried that this may affect Node.js as Wayland is the default now (but maybe it does not matter there).

louwers (changes requested):
We use `xvfb-run` to run the render tests on Linux (not on CI right now).

So I suggest we keep X11 as the default for now.

louwers:
I cannot push to your branch. Could you add `"platform/linux/**" on these two places so the `node-ci` workflow runs?

https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/node-ci.yml#L10

https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/node-ci.yml#L34

ovivoda (commented):
Looks legit! Let's check if CI/CD passes then we can merge.

jwinarske:
@louwers pushed

ovivoda (approved):

platform/glfw/glfw_view.cpp
```diff
@@ -559,9 +559,9 @@ void GLFWView::updateFreeCameraDemo() {
 }
 
 mbgl::Color GLFWView::makeRandomColor() const {
-    const auto r = static_cast<float>(std::rand()) / RAND_MAX;
-    const auto g = static_cast<float>(std::rand()) / RAND_MAX;
-    const auto b = static_cast<float>(std::rand()) / RAND_MAX;
+    const auto r = static_cast<float>(std::rand()) / static_cast <float> (RAND_MAX);
```
louwers, line 562:
```suggestion
    const auto r = static_cast<float>(std::rand()) / static_cast<float>(RAND_MAX);
```

platform/glfw/glfw_view.cpp
```diff
@@ -559,9 +559,9 @@ void GLFWView::updateFreeCameraDemo() {
...
+    const auto g = static_cast<float>(std::rand()) / static_cast <float> (RAND_MAX);
```
louwers, line 563:
```suggestion
    const auto g = static_cast<float>(std::rand()) / static_cast<float>(RAND_MAX);
```

platform/glfw/glfw_view.cpp
```diff
@@ -559,9 +559,9 @@ void GLFWView::updateFreeCameraDemo() {
...
+    const auto b = static_cast<float>(std::rand()) / static_cast <float> (RAND_MAX);
```
louwers, line 564:
```suggestion
    const auto b = static_cast<float>(std::rand()) / static_cast<float>(RAND_MAX);
```

louwers (approved):

louwers:
@jwinarske Thanks for contributing.

PR #1022, opened 2023-04-16
[ci] Cleanup caches when PR is closed

ntadej:
Cleanup caches when PR is closed to avoid `main` caches to be obsoleted too fast due to storage constraints.

Based on official documentation here: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#force-deletion-of-caches-overriding-default-cache-eviction-policy


louwers (approved):
Neat!

PR #1026, opened 2023-04-17
Drawable fundamentals 2

TimSylvester:
I tried to rebase on the latest main with the GLES3 work, but there were an enormous number of conflicts, so I'm starting fresh with a squash-merge into a new branch.

- `Drawable` objects generated by `Layer` using `DrawableBuilder` and processed in `RenderOrchestrator`
- Dynamic VAOs with defaults on `ShaderProgram` and overrides on `Drawable`

Next steps:

- Use `Context` to activate shaders and VAOs, actually draw vertexes
- Some texture support, render drawables to texture targets
- Refactor GL-specific stuff into renderer backend


src/mbgl/shaders/gl/shader_program_gl.cpp
```diff
@@ -0,0 +1,148 @@
...
+        GLenum glType = 0;
+        MBGL_CHECK_ERROR(glGetActiveUniform(program, index, maxLength, &length, &size, &glType, name.get()));
+        addAttr(uniforms, name.get(), index, length, size, glType);
+    }
+
+    VertexAttributeArrayGL attrs;
+
+    count = 0;
+    MBGL_CHECK_ERROR(glGetProgramiv(program, GL_ACTIVE_ATTRIBUTES, &count));
+    MBGL_CHECK_ERROR(glGetProgramiv(program, GL_ACTIVE_ATTRIBUTE_MAX_LENGTH, &maxLength));
```
mwilsnd, line 135: Reallocate `name` here, otherwise when GL_ACTIVE_ATTRIBUTE_MAX_LENGTH returns a size greater than GL_ACTIVE_UNIFORM_MAX_LENGTH you're corrupting the heap.

mwilsnd (changes requested):

src/mbgl/gl/renderer_backend.cpp
```diff
@@ -64,5 +67,72 @@ void RendererBackend::setScissorTest(bool enabled) {
...
+    constexpr auto shaderName = "background_generic";
+
+    gl::Context& glContext = static_cast<gl::Context&>(*context);
+
+//    "layout (location = 0) in vec2 a_pos;\nuniform mat4 u_matrix;\n\nvoid main() {\n
+    // gl_Position = u_matrix * vec4(a_pos, 0, 1);\n}\n", fragment = "uniform vec4 u_color;\nuniform float u_opacity;\n\nvoid main() {\n    fragColor = u_color * u_opacity;\n\n#ifdef OVERDRAW_INSPECTOR\n    fragColor = vec4(1.0);\n#endif\n}\n"
+    
+    auto shader = shaders.get<gl::ShaderProgramGL>(shaderName);
+    if (!shader) {
+        constexpr auto vert = R"(
```
mwilsnd, line 80: On more picky compilers (like my windows desktop), `#version 300 es` must literally be the first thing, it even complains on the line break here.

src/mbgl/gl/renderer_backend.cpp
```diff
@@ -64,5 +67,72 @@ void RendererBackend::setScissorTest(bool enabled) {
...
+                switch (gl_VertexID) {  // fake it for now
+                case 0: gl_Position = vec4(-1.0, -1.0, 1.0, 1.0); break;
+                case 1: gl_Position = vec4( 1.0, -1.0, 1.0, 1.0); break;
+                case 2: gl_Position = vec4( 1.0,  1.0, 1.0, 1.0); break;
+                case 3: gl_Position = vec4(-1.0,  1.0, 1.0, 1.0); break;
+                }
+                gl_Position = u_matrix * vec4(gl_Position.xy * 8192.0, 0, 1);
+                //gl_Position = u_matrix * vec4(a_pos, 0, 1);
+            })";
+        constexpr auto frag = R"(
```
mwilsnd, line 99: ditto

mwilsnd (changes requested):

TimSylvester:
I've reached the point where the background is rendering using drawables instead of the render layer, though it's only supporting solid colors, no patterns or transitions yet.

https://user-images.githubusercontent.com/71895881/235247411-08286abd-4b52-4ec0-905a-9e7d0a733fc3.mp4

It should be reasonable for other people to begin working on some other functionality.

This may also be a good point to isolate the new behavior with preprocessor blocks and merge it.

I started on moving the shader init to a GL-specific derived class, but doing that in `MLNMapViewOpenGLImpl` which derives from `mbgl::gl::RendererBackend` requires moving a lot of stuff from `src/` to `include/`, and also ends up requiring, e.g., `vendor/unique_resource`.   Even if we're going to move all that stuff, it might be better to do it separately, though it may not be immediately clear what needs to move.

I'm going to look into creating a new GL-specific intermediate between those two in hopes that it requires fewer changes.

sjg-wdw:
Exciting!

sjg-wdw:
I favor the #ifdef approach to see if we can merge this into main turned off.
Then do the code reorg to expose more of it as a separate PR.
But we can discuss more on Monday.

include/mbgl/gfx/drawable_builder.hpp
```diff
@@ -60,6 +60,11 @@ class DrawableBuilder {
     const Color& getColor() const;
     void setColor(const Color& value);
 
+    enum class ColorMode { PerDrawable, PerVertex };
+    /// Set how the color value is used
+    /// This should not be changed while a build is in progress
+    void setColorMode(ColorMode mode) { colorMode = mode; }
```
mwilsnd, line 66: Does it make sense to have the concept of color in the builder like this? There are so many paint properties and they vary per layer, it feels like it would make more sense to have stuff like this in the respective layer code using the builder and decouple the builder from properties.

Take a look [here](https://maplibre.org/maplibre-style-spec/layers/#layout-property), note the difference between "layout" and "paint" properties. Properties with the palette icon are paint properties.

TimSylvester: I don't know, it's a placeholder that will be replaced, but I'm not sure by what.
Should the drawables carry along the evaluated paint properties?  That stuff all seems to be templated, so isn't easy to pass around generically.
I don't think we want to create new ways of representing all the properties that can be dynamic, though.

sjg-wdw: I guess there are other things that can be per-vertex vs per-drawable, so that could be more generic.
As for changing these only the fly, I like how you're putting that logic into the Tweaker.  But the Drawable should probably have a way of setting new values and then updating when dirty.

TimSylvester:
I was able to remove some of the GL dependencies from the background layer code, but only by creating an ugly hack to get around `is_shader_v`.  We need some way of getting the intermediate, generic shader without reference to the final type, I think.  Or maybe it would make more sense to have a GL-specific `LayerFactory` work out the shaders and make the generic references available to, e.g., `BackgroundLayerFactory`.

sjg-wdw:
> I was able to remove some of the GL dependencies from the background layer code, but only by creating an ugly hack to get around `is_shader_v`. We need some way of getting the intermediate, generic shader without reference to the final type, I think. Or maybe it would make more sense to have a GL-specific `LayerFactory` work out the shaders and make the generic references available to, e.g., `BackgroundLayerFactory`.

At this point it's probably fine to be there mostly in spirit.  The rest could be done when we get to Metal.
And I'd think once we switch over completely to the new style shader that might be a bit simpler.

include/mbgl/gfx/drawable_builder.hpp
TimSylvester: WhirlyGlobe/Maply translates the style paint properties into `Expression` objects which are further translated into shader inputs to be evaluated in the shaders on Metal.  But it only supports a subset of the available properties.
For anything we want to be evaluated that dynamically, obviously we can't avoid at least one additional representation.
We probably also want to be careful to avoid re-evaluating any layer properties for each drawable.

mwilsnd:
> I was able to remove some of the GL dependencies from the background layer code, but only by creating an ugly hack to get around `is_shader_v`. We need some way of getting the intermediate, generic shader without reference to the final type, I think. Or maybe it would make more sense to have a GL-specific `LayerFactory` work out the shaders and make the generic references available to, e.g., `BackgroundLayerFactory`.

Yeah, I see the issue you're having. Casting to a type that is not the most derived or the base `gfx::Shader` in a safe way is going to require more involved RTTI. Let me think about it and we can discuss at our next meeting.

include/mbgl/gfx/drawable_builder.hpp
sjg-wdw: In Metal we're going to do most if not all of this calculation in the shaders so it makes sense to turn it into a generic representation which can then either be run in a Tweaker or translated to data which is passed to a shader.

TimSylvester: Currently, drawables have tweakers that get run, but I think we'll have tweakers that apply their results to a collection of drawables, so it might make sense to invert that relation, or add a new mechanism for it.

sjg-wdw: I'd say do it naively to start and then see how it looks.
Rather than putting effort into a CPU side solution it might just make more sense to put it into the shaders, even for OpenGL.

TimSylvester:
Getting the generic shaders through a virtual method on `Context` seems work pretty well.

include/mbgl/gfx/drawable_builder.hpp
TimSylvester: Looking into this more, it doesn't really seem practical to pass the paint parameters around through generic drawables/tweakers, but I don't think we need to.  The layer update method that generates new tile drawables when necessary is also the right place to update those drawables with new results for evaluated properties.

sjg-wdw: That makes a lot of sense for this phase.  I think we'll want to revisit that approach in the next phase when we get to the multi-threading exploration.  If we have time to move some of this logic into the Metal shaders that will help some too.

sjg-wdw: I do want to emphasize you shouldn't worry about multi-threading quite yet.
I can see we're going to do a (mostly) single threaded implementation for Metal at first.  Once that's working, then we can come back to some of these issues and figure out how to make the toolkit truly multi-threaded.

TimSylvester: Agreed.  If we can make the property evaluation async I think we'd also do the translation of the evaluated properties into attribute values at the same time, and then apply them via change requests, so we just need to be able to pass attribute values/arrays through generically.

sjg-wdw: That makes sense.

src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -139,31 +201,86 @@ void Renderer::Impl::render(const RenderTree& renderTree) {
...
+
+            mat4 matrix = drawable.getMatrix();
+            if (drawable.getTileID()) {
+                const UnwrappedTileID tileID = drawable.getTileID()->toUnwrapped();
+                const auto tileMat = parameters.matrixForTile(tileID);
+                matrix::multiply(matrix, drawable.getMatrix(), tileMat);
+                matrix = tileMat;
+            }
+
+            if (auto& shader = drawable.getShader()) {
```
mwilsnd, line 237: Add `[[maybe_unused]]` here to suppress GCC warnings

mwilsnd (changes requested):

include/mbgl/shaders/gl/drawable_background.hpp
```diff
@@ -13,6 +13,8 @@ struct ShaderSource<BuiltIn::BackgroundShader, gfx::Backend::Type::OpenGL> {
     static constexpr const char* vertex = R"(layout (location = 0) in vec2 a_pos;
 layout (std140) uniform DrawableUBO {
     mat4 u_matrix;
+    vec2 u_world;
+    vec2 pad;
```
mwilsnd, line 17: Don't let these diverge from the source GLSL if you're skipping the generator

TimSylvester: Good catch, I forgot, thanks!

src/mbgl/gl/layer_group_gl.cpp
```diff
@@ -87,8 +89,11 @@ void TileLayerGroupGL::render(RenderOrchestrator&, PaintParameters& parameters)
             matrix = tileMat;
         }
 
+        const auto renderableSize = parameters.backend.getDefaultRenderable().getSize();
+        
         gfx::DrawableUBO drawableUBO;
         drawableUBO.matrix = util::cast<float>(matrix);
+        drawableUBO.world = {(float)renderableSize.width, (float)renderableSize.height};
```
stefankarschti, line 96: does this apply to all render layers / groups or is it FillLayer specific?

TimSylvester:
Oops, I had a local branch tracking this one and let my client propagate deleting the local branch to the remote.  Hopefully didn't cause any issues...

alexcristici:
> Oops, I had a local branch tracking this one and let my client propagate deleting the local branch to the remote. Hopefully didn't cause any issues...

@mwilsnd and @stefankarschti Your PRs were closed because the base branch was removed. Please reopen your closed PRs.

PR #1028, opened 2023-04-17
Configure clang-tidy CI check

SiarheiFedartsou:


.clang-tidy
```diff
@@ -0,0 +1,99 @@
+---
+Checks: >
+    bugprone-*,
```
SiarheiFedartsou, line 3: Many of these checks are useful and to be enabled in the future, but I disabled everything what was non-trivial to resolve at glance. 
Based on .clang-tidy from OSRM https://github.com/Project-OSRM/osrm-backend/blob/master/.clang-tidy

.gitignore
```diff
@@ -20,6 +20,7 @@
 *.code-workspace
 
 .DS_Store
+.cache
```
SiarheiFedartsou, line 23: This `.cache` is generated by VS Code plugin for C++ I am using - I hope it is not a problem :)

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
 option(MBGL_WITH_CORE_ONLY "Build only the core bits, no platform code" OFF)
+option(MBGL_WITH_CLANG_TIDY "Build with clang-tidy checks enabled" OFF)
```
SiarheiFedartsou, line 4: Should we still use `MBGL_` prefix? :)

CMakeLists.txt
```diff
@@ -966,6 +967,18 @@ target_sources(
...
+
+if (MBGL_WITH_CLANG_TIDY)
+    find_program(CLANG_TIDY_COMMAND NAMES clang-tidy)
+    if(NOT CLANG_TIDY_COMMAND)
+        message(FATAL_ERROR "ENABLE_CLANG_TIDY is ON but clang-tidy is not found!")
+    else()
+        message(STATUS "Found clang-tidy at ${CLANG_TIDY_COMMAND}")
+    endif()
+
+    set_target_properties(mbgl-core PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_COMMAND};--extra-arg=-Wno-error=unknown-warning-option")
```
ntadej, line 979: Which warning option is uknown?

.github/workflows/core-build.yml
```diff
@@ -53,5 +53,5 @@ jobs:
       - name: Build Maplibre GL Core
         run: |
           cmake --version
-          cmake -B build -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_CORE_ONLY=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DMBGL_WITH_COVERAGE=ON
+          cmake -B build -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_CLANG_TIDY=ON -DMBGL_WITH_CORE_ONLY=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DMBGL_WITH_COVERAGE=ON
```
SiarheiFedartsou, line 56: I decided that it is the good place to run clang-tidy checks on CI.

ntadej (changes requested):

.clang-tidy
ntadej: We overlapped, see my comment regarding what MapBox used to have enabled and the files that we have in the repo (can probably also wait for a follow-up PR though).

.gitignore
ntadej: Yes, seems reasonable.

.clang-tidy
```diff
@@ -0,0 +1,99 @@
+---
```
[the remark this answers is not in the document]
SiarheiFedartsou: Yes, will do (not today though) :+1:

CMakeLists.txt
ntadej: For now, yes. We first need to cleanup the configuration a bit and then we may do one big rename.

CMakeLists.txt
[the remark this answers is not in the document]
ntadej: I think it's a good start, yes.

SiarheiFedartsou: `error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'?`

I guess this option is only available in GCC. There is a way to check that flag is supported by compiler before adding it in CMake(i.e. we could fix that), but let's do things step by step please :) 

https://github.com/Project-OSRM/osrm-backend/blob/0ca913132acea46e0c56e4665af50d68f53eec68/cmake/warnings.cmake#L5

louwers (commented):
This is nice but I would like to see a comment about it in [DEVELOPING.md](https://github.com/maplibre/maplibre-gl-native/blob/main/DEVELOPING.md).

Suggestion:

- Could you name the option `MLN_REQUIRE_CLANG_TIDY`?
- Enable by default
- When set but clang-tidy is not found exit with FATAL error
- When not set but clang-tidy is not found, show warning

I think enabling it by default makes sense since PRs won't pass without passing the clang tidy check anyway.

CMakeLists.txt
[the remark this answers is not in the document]
louwers: Agreed this is a good start.

.github/workflows/core-build.yml
louwers: Good!

ntadej:
I think we should have consistent flag naming. We can rename all at once in a separate PR, but I would not like to have some prefixed `MGL` and some `MLN`.

.clang-tidy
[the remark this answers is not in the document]
SiarheiFedartsou: > Can you check the old https://github.com/mapbox/mapbox-gl-native/blob/master/.clang-tidy and existing .clang-tidy files under test and platform/ios if they can be unified?

- I took old `.clang-tidy`, added some checks which were not enabled there and also suppressed which is difficult to fix right now.  
- `platform/ios/.clang-tidy` was removed - now we use the same checks as in `.clang-tidy` there (`.clang-tidy` files in nested folders just override checks in root folder)
- enabled clang-tidy for mbgl-test target - now `platform/.clang-tidy` is used (I decided that it is better to have separate list of checks for tests)

CMakeLists.txt
SiarheiFedartsou: Sure, done.

SiarheiFedartsou:
> This is nice but I would like to see a comment about it in [DEVELOPING.md](https://github.com/maplibre/maplibre-gl-native/blob/main/DEVELOPING.md).

Added, please review - not sure what worth adding there.

> Suggestion:
> 
> * Could you name the option `MLN_REQUIRE_CLANG_TIDY`?
> * Enable by default


@louwers I can rename `MLN_REQUIRE_CLANG_TIDY` now(but lets have an agreement with @ntadej first - I don't have strong opinion here, from my perspective both name options are okay).

When it comes to enabling it by default - I am not sure it will be easy, e.g. I guess we would have to disable it for Android/iOS builds, because clang-tidy can be used there. Don't you mind to do that later in separate PR? 

> * When set but clang-tidy is not found exit with FATAL error

It already works like this:
```
message(FATAL_ERROR "ENABLE_CLANG_TIDY is ON but clang-tidy is not found!")
```

 > * When not set but clang-tidy is not found, show warning

Not sure I understand what benefits it would bring. Could you elaborate why do you think it would be useful?

CMakeLists.txt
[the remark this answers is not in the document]
SiarheiFedartsou: Also added it for mbgl-test, but for limited number of checks (using slightly modified `test/.clang-tidy` which we inherited from original Mapbox repo).

ntadej:
We will soon rename everything to `MLN_*` so you can also change now if you want.

For now I'd keep `clang-tidy` enabled for Linux. The main issue is it's often much slower than regular build (which is already slow on some platforms). Also core has large overlap between platforms.

SiarheiFedartsou:
> For now I'd keep clang-tidy enabled for Linux. The main issue is it's often much slower than regular build (which is already slow on some platforms). Also core has large overlap between platforms.

Do you mean to modify CMakeLists to make MLN_REQUIRE_CLANG_TIDY=ON by default if current system is Linux? Or you mean keep it enabled on Linux's CI?

ntadej:
Sorry for confusion, `clang-tidy` is slow so it should always be disabled by default. For now only Linux CI (the one where you added it already) should have it enabled.

SiarheiFedartsou:
Okay, let me please know if there is something to do to make it merged then 😀

ntadej (approved):
Looks OK to me, but maybe Bart should also review before we merge it (especially as it conflicts with his PR, so he may decide which goes in first).

louwers (approved):

ntadej (changes requested):
The pipeline is very slow (reaching 1 hour). Not sure this is feasible...

Will un-approve for now to allow a bit of discussion.

SiarheiFedartsou:
> The pipeline is very slow (reaching 1 hour). Not sure this is feasible...
> 
> Will un-approve for now to allow a bit of discussion.

Hm, it was faster before?

SiarheiFedartsou:
> > The pipeline is very slow (reaching 1 hour). Not sure this is feasible...
> > Will un-approve for now to allow a bit of discussion.
> 
> Hm, it was faster before?

I see it is about 30m for other PRs. What I am thinking about is that may be it is because addition of clang-tidy changed compiler flags being passed and this kind of invalidated existing ccache’’s cache, but it should back to normal with a next build. I will check it later.

ntadej:
I can retry if you want, but as far as I know `clang-tidy` is not cached (or is it?).

SiarheiFedartsou:
> I can retry if you want, but as far as I know `clang-tidy` is not cached (or is it?).

Yes… ccache does not support clang-tidy https://gitlab.kitware.com/cmake/cmake/-/issues/21092

Well, any thoughts then? 

Btw I am not sure that all cores are used here to build(from cmake docs it is not clear what is default value) 🤔 https://github.com/maplibre/maplibre-gl-native/blob/d81257b8ad52dab32ef7dc98455bf4a4aa392ff6/.github/workflows/core-build.yml#L57

ntadej:
Indeed ninja is not used. Can you add `-GNinja` somewhere at the start of the CMake command line?

SiarheiFedartsou:
> Indeed ninja is not used. Can you add `-GNinja` somewhere at the start of the CMake command line?

Will do, but unfortunately not today - I am already not near my laptop :) 

Also it seems separate clang-tidy checks can have performance issues and there is a way to profile it https://discourse.llvm.org/t/clang-tidy-appears-to-be-significantly-slower-from-clang-15-to-clang-16/68717

louwers (approved):
48 minutes. Not the fastest but acceptable.

We can consider setting up a more powerful runner.

One question: why are there two `.clang-tidy` files? Can we consolidate them at some point?

ntadej:
> 48 minutes. Not the fastest but acceptable.
> 
> We can consider setting up a more powerful runner.
> 
> One question: why are there two `.clang-tidy` files? Can we consolidate them at some point?

As Siarhei said, tests may not need to be as strict (as often one must do "weird" things in tests). Also at the moment tests are not built in the CI so we may expect a slightly longer build once this happens.

ntadej (approved):

louwers:
Thanks, I missed that.

@SiarheiFedartsou Thanks for your contribution. 🚀 🙂

PR #1033, opened 2023-04-19
Linux C++ test & render tests on CI

louwers:
- I renamed the `core-build` workflow to `linux-ci` and repurposed it.
- Added two jobs to it `render-test` and `test`.
- `render-test` uploads the render test output to S3 when it fails and leaves a comment where you can inspect the HTML output
- Both tests types have some failures, but that is not a reason not to merge this. The tests themselves run fine, except for the fact that xvfb-run is a bit flakey.
- I made inclusion of the Node.js CMake files non-default behind an option. These CMake files download Node.js headers so I don't want to include them by default.


platform/node/DEVELOPING.md
```diff
@@ -48,13 +48,13 @@ To compile the Node.js bindings and install module dependencies, from the reposi
 #### MacOS
 
 ```bash
-cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON
+cmake . -B build -G Ninja -DMLN_WITH_NODE -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON
```
acalcutt, line 51: Shouldn't this be -DMLN_WITH_NODE=ON ? same with line 57

ntadej: I think all should stay `MGL` for now and then we rename all at once. Otherwise it's just confusing.

louwers: I will keep it, but make a separate PR to rename the other prefixes.

I don't want to introduce new `MBGL_*` options.

ntadej: Can you also pick these PRs to the `opengl-2` branch (before I setup automatic picking)?

louwers: Sure.

.github/workflows/linux-ci.yml
```diff
@@ -0,0 +1,144 @@
...
+      - main
+    paths:
+      - "src/**"
+      - "include/**"
+      - ".github/workflows/linux-ci.yml"
+      - "vendor/**"
+      - "CMakeLists.txt"
+      - metrics/linux-gcc8-release-style.json
+
+permissions:
```
ntadej, line 16: Where do you use the token? This is not clear why it's needed

platform/linux/linux.cmake
```diff
@@ -137,7 +137,9 @@ target_link_libraries(
 add_subdirectory(${PROJECT_SOURCE_DIR}/bin)
 add_subdirectory(${PROJECT_SOURCE_DIR}/expression-test)
 add_subdirectory(${PROJECT_SOURCE_DIR}/platform/glfw)
```
ntadej, line 139: Why do we include also `glfw` by default? I think this should also be excluded.

platform/node/DEVELOPING.md
```diff
@@ -48,13 +48,13 @@ To compile the Node.js bindings and install module dependencies, from the reposi
...
 ```bash
-cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON
+cmake . -B build -G Ninja -DMLN_WITH_NODE=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DMBGL_WITH_COVERAGE=ON
 ```
 
 #### Linux
 
 ```bash
-cmake . -B build -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
+cmake . -B build -G Ninja -DMLN_WITH_NODE -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER=gcc-10 -DCMAKE_CXX_COMPILER=g++-10
```
ntadej, line 57: Typo - missing `ON`

ntadej (changes requested):

.github/workflows/linux-ci.yml
louwers: `aws-actions/configure-aws-credentials@v2`

platform/linux/linux.cmake
louwers: Let's do that in a separate PR?

.github/workflows/linux-ci.yml
louwers: I added a link

platform/linux/linux.cmake
ntadej: OK 😊

platform/node/DEVELOPING.md
ntadej: Still missing 😊

louwers:
I fixed the merge conflict.

@ntadej I will fix change the `MBGL_` prefix in a separate PR. Do you think it is OK to merge?

.github/workflows/linux-ci.yml
```diff
@@ -0,0 +1,145 @@
...
+            libcurl4-openssl-dev \
+            libuv1-dev \
+            libjpeg-dev \
+            libpng-dev \
+            libglfw3-dev
+
+      - name: Build Maplibre GL Core
+        run: |
+          cmake --version
+          cmake -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DMBGL_WITH_COVERAGE=ON
```
ntadej, line 56: You should recover Ninja and clang-tidy I guess?

platform/node/DEVELOPING.md
ntadej: Readme still not correct!

ntadej (dismissed):
Unfortunately you reverted some changes when solving conflicts. Also you keep resolving one typo which you do not fix 😉

.github/workflows/linux-ci.yml
louwers: Oops. I did that, but something went wrong during the rebase. 🤦‍♂️

louwers: Thanks for catching this. I added it back.

louwers:
@ntadej @ovivoda OK to merge?

Test & render-test fail but fixing them is out of scope for this PR.

ntadej:
> @ntadej @ovivoda OK to merge? I may have to tweak uploading the render test results a bit, right now forks don’t have the right permissions.
> 
> Test & render-test fail but fixing them is out of scope for this PR.

Then disable them. The CI should not be red 😊

louwers:
> > @ntadej @ovivoda OK to merge? I may have to tweak uploading the render test results a bit, right now forks don’t have the right permissions.
> > Test & render-test fail but fixing them is out of scope for this PR.
> 
> Then disable them. The CI should not be red 😊

I disagree. CI is there to indicate a problem. We need to enable CI to make the problem apparent.

The 'build' step is green, the two steps that this PR add are red.

ntadej:
Yes, but how does the author of another PR know? Also then we need to force-merge every PR.

louwers:
@ntadej We don't need to force merge because requiring CI to be green is not a requirement for merging.

People can see the history of workflows and see that they are already failing on `main`. It is not ideal and I agree we need to fix these problems ASAP, but my reasoning is if these checks were already part of the CI we would not disable them just because they are red.

louwers:
@ntadej There is a team of engineers working on the core... I would not worry about CI staying red too long... They are actively resolving the problems that appear. The reason that the render tests fail for example is that we did not catch a regression in a recent PR.

ntadej:
I leave it up to Ovi.

louwers:
> I leave it up to Ovi.

Alright! Thanks for your critical eye. 🙂

ovivoda:
@louwers @ntadej Nice to read the debate here! I think we should merge this but also open an issue in the repo to address the missing parts. 

I was never a fan of big PR's.

ovivoda (approved):

PR #1035, opened 2023-04-19
Fix CameraUpdateFactory#zoomOut

### LINKED ISSUE #1017: Android CamerUpdateFacotry#zoomOut regression
**Describe the bug**
Zoom out leads to zooming to z0.

I think this should be max not min: https://github.com/maplibre/maplibre-gl-native/blob/4828cc7a8be58579b4a2beed8033fefc2fda5170/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraUpdateFactory.kt#L451

Quickfix for users would be: `CameraUpdateFactory.zoomBy(-1.0)`

boldtrn:
Fixes #1017.

I added a @JvmStatic to the OfflineManager#getInstance method as this requires to use the Companion object in Java, which is not that nice :).

### DIFF
--- a/platform/android/CHANGELOG.md
+++ b/platform/android/CHANGELOG.md
@@ -10,6 +10,8 @@ MapLibre welcomes participation and contributions from everyone. Please read [`C
 
 ### 🐞 Bug fixes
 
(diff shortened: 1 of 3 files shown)


louwers (approved):

louwers:
Thanks!

PR #1042, opened 2023-04-19
[qt] Make unit tests build and run

ntadej:
Make unit tests build and run:
 - if Qt GL context is invalid return default (tagging @alexcristici as he last touched the code if this makes sense)
 - use C++ http server
 - update references
 - disable some DB tests
 - disable bidi tests

Note I plan to enable the test in a separate PR as I have to figure out how to get it run on Linux (macOS obviously does not work in `main`).


platform/qt/src/mbgl/gl_functions.cpp
```diff
@@ -10,278 +10,416 @@ namespace platform {
 /* OpenGL ES 2.0 */
 
 void (* const glActiveTexture)(GLenum) = [](auto... args) {
+    if (!QOpenGLContext::currentContext())
+        return;
```
louwers, line 14: Maybe it's better to let it crash here instead? If these functions are not initialized MapLibre will not be able to render anything.

test/src/mbgl/test/test.cpp
```diff
@@ -7,6 +7,10 @@
 namespace mbgl {
 
 int runTests(int argc, char *argv[]) {
+#if TEST_HAS_SERVER && defined(__QT__)
```
louwers, line 10: Maybe it's better to run the Node.js server.

I think we want to get rid of the C++ server because of it's limitations.

platform/qt/src/mbgl/gl_functions.cpp
ntadej: Then we can not run unit tests. It seems that some GL calls are done in core after context is cleaned and this is a nice workaround. I'd wait for Metal work to finish to properly fix this.

test/src/mbgl/test/test.cpp
ntadej: What limitations exactly? I can run unit tests just fine and I do not want to depend on Node.js.

louwers: Limited concurrent connections apparantly https://github.com/maplibre/maplibre-gl-native/pull/674#discussion_r1089674583

test/storage/offline_database.test.cpp
```diff
@@ -990,7 +1000,9 @@ TEST(OfflineDatabase, TEST_REQUIRES_WRITE(ClearAmbientCache)) {
         db.clearAmbientCache();
     }
 
+#ifndef __QT__ // Qt doesn't decrease the size of the database file.
```
louwers, line 1003: > Qt doesn't decrease the size of the database file.

Should it?

test/src/mbgl/test/test.cpp
ntadej: I'd keep this one for Qt if I may until we clarify why we even need that many connections for testing? I guess to run tests in parallel (although it's not obvious to me if we use that)?

test/storage/offline_database.test.cpp
ntadej: If it should, it's a Qt bug so out of scope. All commands are just passed-through to a Qt DB engine, there's no MapLibre implementation specifics.

platform/qt/src/mbgl/gl_functions.cpp
louwers: Maybe you could wrap the check in a function that returns a bool and logs a warning?

Your call, but it may save someone a frustrating debugging session.

ntadej: We can't change return type, but I could add logging, although I'm a bit hesitant. Even this `if` may slow down the code actually.

Let me see if I can hack the tests then...

ovivoda:
@alexcristici would be nice to get your opinion on this PR if you have a few mins to spare. Tx!

platform/qt/src/mbgl/gl_functions.cpp
ntadej: OK, I managed to add a functionality to `headless_backend.cpp` to restore the context when doing cleanup so we do not crash. This is a more proper solution.

platform/default/src/mbgl/gl/headless_backend.cpp
```diff
@@ -43,10 +43,19 @@ HeadlessBackend::HeadlessBackend(const Size size_,
 
 HeadlessBackend::~HeadlessBackend() {
     gfx::BackendScope guard{ *this, gfx::BackendScope::ScopeType::Implicit };
+    // Some implementations require active context for GL functions to work
+    if (impl->glNeedsActiveContext()) {
+        impl->activateContext();
+    }
```
alexcristici, line 49: We should check if `impl` is valid.
The name `glNeedsActiveContext` is too generic, maybe we should be more specific in naming this.
But it will be better if possible to fix it outside of this class. What is the callstack, from where the destructor is called?

ntadej: Will add the check, nice spot!

In the MBGL context below some GL functions are called on destruction (when `context.reset()` is called). The problem is that Qt GL context is not valid when deactivated so it segfaults.

One option was to add if clauses to every GL call in the wrapper, but maybe that's an overkill (and also potential slowdown), so by advice of Bart I found an easier workaround.

Regarding the naming I think the name fits quite well, maybe `glCallsNeedActiveContext`?

alexcristici: Regarding the naming, I mean `glNeedsActiveContextOnDestruction`.

This workaround adds more code to core which is needed on single platform, so makes the code harder to read. It assumes also the `context.reset` calls some gl functions, which can change in the future and this check will remain here.

I was thinking more from where the destructor is called, maybe there is the issue, why the destructor is called and context is no more active?

ntadej: I see your concerns. Maybe we can open a ticket to try to improve it in the future.

It's called in the HeadlessBackend destructor.

My bigger question is why platform contexts are not inside the core MBGL context but somehow separate. This does not make much sense to me, but it would be a larger development to improve (and probably something to consider with your modularisation work).

Also I'm not sure what happens in other platforms. Do GL calls just take some default context then?

alexcristici: I mean from where the `HeadlessBackend destructor` is called.

ntadej: Each unit test that renders basically.

alexcristici: So maybe there the fix should happen. Maybe there is a good place to make sure the destructor is called with an active context, or a default one. When I will have time I will check on iOS how the context is handled in this scenario.

ntadej: Yes, I think the whole headless rendering needs a bit of rework. Let me open a ticket and we merge it as is (will rename the property and add impl check first).

louwers (changes requested):
Not sure what the status of this PR is, but I don't think I like the `#ifndef`s sprinkled throughout the tests. Especially when more platforms only have certain features implemented, the tests would become quite unreadable.

Here is a suggestion how I would like to handle this and similar situations; feel free to let me know what you think. I suggest that we refactor each test so that it only tests for one thing, then move the the tests with functionality that Qt does not support to its own file, and then conditionally add that file to the `mbgl-test` target if it is not explicity excluded by some platform-independent variable.

ntadej:
I agree, but I propose we do it gradually. The defines are already there in the first place, I'm just adding a few more.

I'll push the final changes probably on the weekend.

ntadej:
Hi @ovivoda, @louwers, can this one please be reviewed as it's picking up conflicts?

louwers:
Could you make an issue to remove the #ifdefs from the tests again and link to this PR?

louwers (commented):
@ovivoda What do you think? I think which tests are run should be defined in the platforms.

@ntadej As a temporary solution, instead of adding `#ifdefs`, I would prefer not running the tests that fail at all for Qt. Is that a workable solution?

ntadej:
> @ovivoda What do you think? I think which tests are run should be defined in the platforms.
> 
> @ntadej As a temporary solution, instead of adding `#ifdefs`, I would prefer not running the tests that fail at all for Qt.

Opened a ticket now: https://github.com/maplibre/maplibre-native/issues/1116

Note that these are not new, and probably not all are Qt specific. In interest of proceeding, I'd merge as is and cleanup later, similar as we did with the failing linux tests 😊

louwers (approved):

ovivoda:
> @ovivoda What do you think? I think which tests are run should be defined in the platforms.
> 
> @ntadej As a temporary solution, instead of adding `#ifdefs`, I would prefer not running the tests that fail at all for Qt. Is that a workable solution?

@louwers I agree that we should do it per platform

PR #1049, opened 2023-04-21
Use pull_request_target instead of pull_request

louwers:
We need access from the iOS workflow to the repository secrets. Otherwise we cannot authenticate with AWS to upload the tests artifacts. If a pull request is made from a fork, the workflows are run in the context of the forked repo. This PR ensures that the workflow is run in the context of this repo. See: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

This is a security risk, because a bad actor could extract the secrets with a malicious PR. Since our website is hosted on GitHub Pages, it could be defaced for example. The AWS role set up for GitHub Actions is quite limited so there would not be an issue there.

Right now if someone is a collaborator in the past, workflows will automatically run. We can also enable requiring approval before any run from people without any repository access, but I'd like to take the slight risk and keep things as they are now.


louwers:
> Thanks for looking into this @louwers. What alternatives did you consider?

@wipfli We can only run these tests after they have been merged to main. But then we don't know if a PR broke it of course.

We can also use a safe to test label: https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci

wipfli:
> Otherwise we cannot authenticate with AWS to upload the tests artifacts. 

Does this mean that the test result cannot be stored on AWS, or does it mean that we cannot run the tests at all?

wipfli:
So the only alternative to sharing secrets with everyone is to run the iOS tests from a branch in maplibre-gl-native?

wipfli:
Given that we are allowing people to run tests on our self-hosted runner, how different is this new situtation from what we have already now?

louwers:
> So the only alternative to sharing secrets with everyone is to run the iOS tests from a branch in maplibre-gl-native?

We will not be sharing secrets but a malicious PR can steal and abuse them.

> Given that we are allowing people to run tests on our self-hosted runner, how different is this new situtation from what we have already now?

I don't think it stores secrets across runs. But technically someone could install malicious code on the self hosted runner that sends secrets to subsequent runs.

We have a few options as I see it:

- Require manual workflow run approval for every PR for external contributions.
- Run the instrumentation tests and render tests (which require secrets) only on approval.
- Trust people that have already made contributions in the past (current situation).

acalcutt:
On the question of workflow stealing secrets... Once we have these tests the way we want them, could we doing something like was done in the tileserver-gl pipeline and specify running the tests from the master branch like https://github.com/maptiler/tileserver-gl/blob/master/.github/workflows/pipeline.yml#L29 . I feel like this could make it so a user couldn't throw something malicious into the workflow, since it would always run the version in master.

I does make it a bit harder to test changes to the workflow though.

louwers:
@acalcutt This is already the behavior of `pull_request_target` 🙂 

https://github.com/orgs/community/discussions/27084#discussioncomment-3254544


```
With on: pull_request_target as trigger, secrets can be accessed, but the workflow of the base branch is used (so the version trusted by you). Any changes done to .github/workflows/ by the PR are ignored, which should prevent malicious users from leaking secrets. Note that also the code of the base branch is used unless you explicitly checkout the PR code.
```

ovivoda:
@louwers I am not an expert in this field, but I took some time to read a bit about it and here is my conclusion:

I would only do the switch if we will make this bullet proof. From my perspective, a security breach is not something worth risking for a small gain as `access from the iOS workflow to the repository secrets`.  

Some things we can try maybe is what is suggested in this post. Even though I do not think it is applicable in our case exactly as mentioned in the example: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ 

A combination of `pull_request_target`, and `workflow_run` might worth investigating more though.

PR #1050, opened 2023-04-21
Remove Mapbox::Base::optional dependency from CMake

SiarheiFedartsou:
We moved to std::optional some time ago and this is no longer needed.


ntadej:
Hmm, why did only core CI run. This is not ideal...

louwers:
@ntadej https://github.com/maplibre/maplibre-gl-native/pull/1051

louwers:
@SiarheiFedartsou Nice catch, thanks.

louwers (approved):

PR #1079, opened 2023-04-27
Revert GLES3 changes for OSMesa in Windows

tdcosta100:
The PR #995 bring us the long awaited support for OpenGL ES 3.0. But for OSMesa, which does have support for OpenGL ES 3.0, a reference to the file `libGLESv2.dll` was changed to `libGLESv3.dll`, but the latter file doesn't exist. Support for OpenGL ES 3.0 is provided by `libGLESv2.dll`, so I'm reverting these changes for the reference to this file. The other changes are good and work well, so I kept all of them.

### DIFF
--- a/platform/windows/windows.cmake
+++ b/platform/windows/windows.cmake
@@ -91,7 +91,7 @@ if(MLN_WITH_EGL)
         mbgl-core
         PRIVATE
             unofficial::angle::libEGL
-            unofficial::angle::libGLESv3
+            unofficial::angle::libGLESv2
     )
 elseif(MLN_WITH_OSMESA)
     list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
@@ -114,7 +114,7 @@ elseif(MLN_WITH_OSMESA)
         mbgl-core
         PRIVATE
             OSMesa::osmesa
-            OSMesa::libGLESv3
+            OSMesa::libGLESv2
     )
 else()
     find_package(OpenGL REQUIRED)
--- a/platform/node/cmake/module.cmake
+++ b/platform/node/cmake/module.cmake
@@ -218,7 +218,7 @@ function(add_node_module NAME)
                     TARGET ${_TARGET}
                     POST_BUILD
                     COMMAND ${CMAKE_COMMAND} -E copy "${PROJECT_SOURCE_DIR}/platform/windows/vendor/mesa3d/${_ARCH}/libglapi.dll"  "${_OUTPUT_PATH}"
-                    COMMAND ${CMAKE_COMMAND} -E copy "${PROJECT_SOURCE_DIR}/platform/windows/vendor/mesa3d/${_ARCH}/libGLESv3.dll" "${_OUTPUT_PATH}"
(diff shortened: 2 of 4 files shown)


acalcutt (approved):
From the information at https://github.com/pal1000/mesa-dist-win/blob/master/readme.md#opengl-es-drivers-and-egl-library it does confirm this. 

There is no libGLESv3.dll and between libGLESv1_CM.dll and libGLESv2.dll there is support for OpenGL ES 1.x, 2.x and 3.x

mwilsnd (approved):

platform/windows/windows.cmake
```diff
@@ -91,7 +91,7 @@ if(MLN_WITH_EGL)
         mbgl-core
         PRIVATE
             unofficial::angle::libEGL
-            unofficial::angle::libGLESv3
+            unofficial::angle::libGLESv2
```
etnav, line 94: Was this an intended change for `if(MLN_WITH_EGL)`?

etnav: @tdcosta100 It seems to be causing issues when using the `MLN_WITH_EGL` flag.

`"GLES3/gl2.h": No such file or directory` -> from `gl_functions.cpp`

tdcosta100: > Was this an intended change for `if(MLN_WITH_EGL)`?

Yes, because there is not a `libGLESv3.dll` even with ANGLE.

> @tdcosta100 It seems to be causing issues when using the `MLN_WITH_EGL` flag.
> 
> `"GLES3/gl3ext.h": No such file or directory` -> from `gl_functions.cpp`

I will check this. Also, apparently the ANGLE port in vcpkg is broken too, so probably I will need to submit a PR there too. As soon as possible I will rework this part.

etnav: Have you by any chance had any success with this?

tdcosta100: Hi, @etnav. I did some changes to the ANGLE package and submitted to the vcpkg project. Then things went bad:

https://github.com/microsoft/vcpkg/pull/31254

I can point the vcpkg subdirectory to my branch, so ANGLE works again. But I'm really tired of trying to submit this change to their repository and they are demanding me to do things like I was the package developer. I'm not, I'm just trying to fix something broken in it.

etnav: Had a glance, that's very annoying. After such a long drag and waste of time. 😠 

Thanks for your effort though, I will try to check this again when I got some spare time and get back to the issue.

tdcosta100: 2 weeks of work, and it wasn't good enough. But ok to stay broken. I will change the ref to my branch, for now.

tdcosta100: Good news, they finally merged the change, made by someone else they like more. I will change the ref now.

tdcosta100: Finally fixed!

xiaozhuai: > made by someone else they like more

LOL, they didn't like me either.
I often get into arguments with them, but I get used.
Whatever, thank you again for your efforts!

tdcosta100: Oh my, looks like my PR. They seem to not like others's contribution, but as they are on GitHub, they are obbligated to.

PR #1080, opened 2023-04-27
Create security policy

louwers:
I named it `SECURITY.md`  because it is the default filename GitHub expects for a security policy.

It is a bit simpeler than the 'default' security policy because things like CSRF are not relevant in the context of MapLibre Native.

Instead of reporting by e-mail we should use the built-in functionality of GitHub, because it is more secure, will reach the right persons quicker, and allows for better collaboration with the reporter to fix the issue.


ovivoda (approved):
Thanks for the initiative @louwers !

PR #1085, opened 2023-04-28
Made AndroidLocationEngineImpl public

boldtrn:
Since we have the FusedLocationEngine public, we should also have the AndroidLocationEngine public, so implementers can choose the engine freely.


ovivoda (approved):
For consistency I think we should do this.

PR #1087, opened 2023-04-30
Fix MapTiler Satellite Tile Source

dpaulat:
When using MapTiler as the tile server, satellite map tiles do not display properly (using mbgl-qt as a reference):

2023-04-29 09_01_57-MapLibre GL_ Satellite Hybrid

When updating the source template:
```
- .withSourceTemplate("/tiles{path}/tiles.json", "sources", {})
+ .withSourceTemplate("/maps{path}/tiles.json", "sources", {})
```

The satellite map tiles display as expected.

2023-04-29 22_24_27-MapLibre GL_ Satellite Hybrid

From what I can tell, everything *seems* to work, but I'm concerned I'm missing something.  Some documents reference tiles.json at the old path (e.g., standard tiles), and some at the new path (e.g., raster tiles for standard styles).

Standard tiles with the old path: https://cloud.maptiler.com/tiles/
Standard styles with the new path: https://cloud.maptiler.com/maps/

Any thoughts?


dpaulat:
It looks like the MapTiler.SourceURL test failed as a result. The existing functionality is not clear to me at this stage, but I suspect the change required is going to be more complicated.

louwers:
@dpaulat Thanks for this PR.

It's not a priority for us to fix this. We can merge a fix, but over time we want to move away from including vendor-specific code.

If you are a MapTiler customer it is best to contact them directly. (cc @klokan)

ntadej:
It seems that production MapTiler paths still use `/tiles`. I think we should probably close the PR.

@dpaulat, did you manage to solve the issue in the meantime?

dpaulat:
I agree, changing from tiles causes some other issues. I still have issues with satellite maps, so I think there's still some improvement that can be made, although I don't think my solution is the correct one. I've worked around the issue on a fork. Let's go ahead and close this PR without merge.

Issue #1092, opened 2023-05-01
Run iOS size test on CI


louwers:
@alexcristici created an iOS size test.

https://github.com/maplibre/maplibre-native/issues/943

We have a developer team now so this can be run on CI.

louwers:
I am having a lot of challenges with this.

I spend some time fiddling with manually generated provisioning profiles and certificates, but I would always get errors about the right profiles and certificates not being available. So I set up a new macOS runner with macOS 13.3 so I could login in and pull the provisioning profile automatically.

Right now due to #788 only CMake 3.24 works, otherwise `xcodebuild` fails with the `-lmbgl-core` linking error... I didn't want to downgrade CMake but I found out that adding

```
LIBRARY_SEARCH_PATHS='$(LIBRARY_SEARCH_PATHS) '"$(PWD)/build/ios/$BUILD_CONFIGURATION-iphoneos"
```

as an argument to `xcodebuild` makes it link again.

Now I get _another_ code signing related error

```
/Users/ec2-user/maplibre-native/platform/ios/build/ios/ArchiveIntermediates/iosapp/InstallationBuildProductsLocation/Applications/MapLibre GL.app/Frameworks/Mapbox.framework: errSecInternalComponent
```

A [common problem](https://stackoverflow.com/questions/24023639/xcode-command-usr-bin-codesign-failed-with-exit-code-1-errsecinternalcomponen) which I have not found a solution for yet.

Will let this rest for a bit but maybe I should look into [fastlane](https://fastlane.tools/).

PR #1097, opened 2023-05-02
Run iOS size test on CI

louwers:
Runs iOS size test on CI (#1092).

When this is merged I need to install the provisioning profile on the CI runners and make another PR to comment the size test results on the PR.

A `.size-test-result` file is generated that looks like this:

```
BASELINE_Release=5504352
CURRENT_Release=5654368
BASELINE_MinSizeRel=4249952
CURRENT_MinSizeRel=4383408
```

I am uploading this file as an artifact. I want to comment the change compared to the original baseline and the latest recorded value for `main`.


ovivoda (approved):
looks legit @louwers !

PR #1102, opened 2023-05-03
Get rid of mapbox-gl-native-curl

SiarheiFedartsou:
I don't think we need this dependency now: it is used only in Android tests. At the same time we already have platform implementation of HTTP client and don't have to use any external dependencies.


platform/android/android.cmake
```diff
@@ -139,24 +139,11 @@ target_link_libraries(
         -Wl,--no-whole-archive
 )
 
-if(ANDROID_NATIVE_API_LEVEL VERSION_LESS 24)
-    target_sources(
-        mbgl-test-runner
-        PRIVATE ${PROJECT_SOURCE_DIR}/platform/android/src/test/http_file_source_test_stub.cpp
```
SiarheiFedartsou, line 145: Tbh I don't know well this part, but implementation in http_file_source_test_stub.cpp seem to do nothing, i.e. probably it is not even important if we have some implementation here or not (probably it should just compile).

louwers:
Neat! 😎

platform/windows/vendor/vcpkg
SiarheiFedartsou, line 1: not sure why I have it change here...

platform/android/android.cmake
```diff
@@ -189,7 +176,7 @@ add_library(
     ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/text/local_glyph_rasterizer.cpp
     ${PROJECT_SOURCE_DIR}/platform/android/src/test/collator_test_stub.cpp
     ${PROJECT_SOURCE_DIR}/platform/android/src/test/number_format_test_stub.cpp
-    ${PROJECT_SOURCE_DIR}/platform/android/src/test/http_file_source_test_stub.cpp
```
SiarheiFedartsou, line 192: Not sure about this place - probably it is better to leave this `http_file_source_test_stub.cpp` here which basically does nothing.

SiarheiFedartsou:
I don’t think that failing tests on Linux are caused by this change 🤔

louwers (approved):

Issue #840, opened 2023-02-21
Investigate configuring Xcode with Bazel


louwers:
Anagolous to #737, I would like to investigate if configuring Xcode with Bazel is a vaiable option.

The Bazel setup donated by Lyft & @nvanfleet seems a lot more solid than the CMake based iOS build setup in its current form.

There is also a quite widely used ruleset for generating Xcode projects with Bazel: https://github.com/buildbuddy-io/rules_xcodeproj

louwers:
I am hitting this issue when trying to write a config to build the app

https://github.com/bazelbuild/bazel/issues/9429

And this issue when generating an Xcode config for the framework:

https://github.com/buildbuddy-io/rules_xcodeproj/issues/1768

ntadej:
Should we maybe have a technical chat at some point, because we need to figure out what are the actual limitations of CMake.

What I would actually propose is to use CMake to generate the Xcode project, but then some other higher-level tool to use it to build the frameworks (potentially also Bazel).

louwers:
I replied in the other thread.

`@import`s within Objective-C [are not yet supported by Bazel](https://github.com/bazelbuild/bazel/issues/9429), which the iosapp target uses:

keith:
From the Lyft perspective the value of using bazel wasn't because cmake was bad but just because that made it easier to integrate with our internal app's build which uses bazel.

I do think you should be able to get rules_xcodeproj working and you could use that locally. Given the very few uses of `@import` I would just swap them with the old style. There shouldn't be a practical downside for this use case and IIRC cmake doesn't support those by default either.

louwers:
@keith Thanks for pitching in. Switching to the angled bracket notation for the test app is a good idea, although importing an Objective-C library from another also seems problematic: https://github.com/bazelbuild/bazel/issues/9461

keith:
it should be ok with the right `includes`

PR #1108, opened 2023-05-04
Xcode project generation with bazel

mwilsnd:
This PR uses rules_xcodeproj to create an Xcode project for the MapLibre core and iosapp example.

Currently an experimental draft, project can be generated which contains the core but no iosapp code. Navigate to `platform/ios/platform/ios/scripts` and run `bazel-xcodeproj.sh` to generate the Xcode project under `platform/ios`.


louwers:
Would close #840

I will have a look as well because the iOS CMake setup has issues with the latest CMake, with multiple people having looked at it without much success.

louwers (approved):

Issue #1071, opened 2023-04-25
Test `MainResourceLoader.CachedResourceLowPriority` (sometimes) failing


louwers:
The `MainResourceLoader.CachedResourceLowPriority` is failing. Maybe a race condition according to @mwilsnd.

To reproduce, build the `mbgl-test-runner` target and run it.

Issue #1070, opened 2023-04-25
Test `Projection.MetersPerPixelAtLatitude` failing


louwers:
C++ unit tests were enabled on Linux recently. The `Projection.MetersPerPixelAtLatitude` test is failing on Linux.

rpitre-wetdog:
I just looked at the code and it involves testing equality of doubles.  For me any equality test with double/float types is asking for trouble. Test will pass many many times and then mysteriously fail

PR #1109, opened 2023-05-05
Configure expression tests on CI

SiarheiFedartsou:
While working on another PR I noticed that we do not run it on CI, so I propose to start doing it.

### DIFF
--- a/expression-test/expression_test_parser.cpp
+++ b/expression-test/expression_test_parser.cpp
@@ -438,7 +438,6 @@ std::optional<TestData> parseTestData(const filesystem::path& path) {
         Log::Error(Event::General, std::string("Cannot load test data from '" + path.string() + "': " + ex.what()));
         return std::nullopt;
     }
-
 }
 
 std::string toJSON(const Value& value, unsigned indent, bool singleLine) {
--- a/expression-test/expression_test_parser.cpp
+++ b/expression-test/expression_test_parser.cpp
@@ -402,37 +402,43 @@ Ignores parseExpressionIgnores() {
 }
 
 std::optional<TestData> parseTestData(const filesystem::path& path) {
-    TestData data;
-    auto maybeJson = readJson(path.string());
-    if (!maybeJson.is<JSDocument>()) { // NOLINT
-        mbgl::Log::Error(mbgl::Event::General, "Cannot parse test '" + path.string() + "'.");
+    try {
+        TestData data;
+        auto maybeJson = readJson(path.string());
+        if (!maybeJson.is<JSDocument>()) { // NOLINT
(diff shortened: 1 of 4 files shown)


metrics/ignores/platform-all.json
```diff
@@ -16,6 +16,7 @@
     "expression-tests/legacy/interval/composite-default": "https://github.com/mapbox/mapbox-gl-native/issues/12747",
     "expression-tests/legacy/interval/tokens-zoom": "https://github.com/mapbox/mapbox-gl-native/issues/12747",
     "expression-tests/resolved-locale/basic": "Even the 'en' locale may not be present on some test systems.",
+    "expression-tests/within/non-supported": "Something is wrong with GeoJSON geometry in test",
```
SiarheiFedartsou, line 19: It throws exception while loading test information (even before we actually run it): 
```
2023-05-05 21:47:43.550 mbgl-expression-test[20437:12390485] [ERROR] {}[General]: Cannot load test data from '/Users/siarheifedartsou/work/maplibre-gl-native/metrics/integration/expression-tests/within/non-supported/test.json': Polygon must be described by 4 or more coordinate points. Improper nesting can also lead to this error. Double check that the coordinates are properly nested and there are 4 or more coordinates.
```

expression-test/expression_test_parser.cpp
```diff
@@ -402,37 +402,42 @@ Ignores parseExpressionIgnores() {
...
-        Log::Error(Event::General, std::string("Can't convert inputs value for '") + path.string() + "'");
+        // Parse inputs
+        if (data.document.HasMember("inputs") && !parseInputs(data.document["inputs"], data)) {
+            Log::Error(Event::General, std::string("Can't convert inputs value for '") + path.string() + "'");
+            return std::nullopt;
+        }
+
+        return {std::move(data)};
+    } catch (const std::exception& ex) {
+        Log::Error(Event::General, std::string("Cannot load test data from '" + path.string() + "': " + ex.what()));
```
SiarheiFedartsou, line 438: If there is a problem during load of test data we just return nullopt.

expression-test/main.cpp
```diff
@@ -47,7 +47,11 @@ int main(int argc, char** argv) try {
         }
 
         if (!testRun) {
-            stats.errored.emplace_back(id);
+            if (shouldIgnore) {
```
SiarheiFedartsou, line 50: If test cannot be loaded for whatever reason we still can ignore it.

SiarheiFedartsou:
I think tests for CI were already red even before this change...

louwers (approved):
Awesome work!

Yes, about the Linux tests see #1070 and #1071.
I will have a look at them this week.

But in this case it is failing because of xvfb. Need to look into that as well...

PR #1110, opened 2023-05-06
Do not run tests for glInvalidateFramebuffer & glInvalidateSubFramebuffer on macOS

SiarheiFedartsou:
They are available only on iOS and iOS Simulator if current platform is Darwin https://github.com/maplibre/maplibre-native/blob/f8c2c6c8b3f1fce6a3529b240d4d3c695a5edfb1/platform/darwin/src/gl_functions.cpp#L268. That's why locally I have such linker errors(for `cmake .. && make -j8 mbgl-test-runner`):
Screenshot 2023-05-06 at 12 47 15

Not sure why it is green on existing macOS CI...

### DIFF
--- a/test/gl/gl_functions.test.cpp
+++ b/test/gl/gl_functions.test.cpp
@@ -251,10 +251,10 @@ TEST(GLFunctions, OpenGLES) {
     EXPECT_NE(glGetProgramBinary, nullptr);
     EXPECT_NE(glProgramBinary, nullptr);
     EXPECT_NE(glProgramParameteri, nullptr);
-    #if !defined(TARGET_OS_MAC)
+#if !defined(TARGET_OS_MAC)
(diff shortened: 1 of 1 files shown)


ntadej:
Darwin is currently not really supported on `main` due to OpenGL ES 3 migration. We do not even run the CI.

SiarheiFedartsou:
> Darwin is currently not really supported on `main` due to OpenGL ES 3 migration. We do not even run the CI.

Hm, indeed, I got confused by presence of https://github.com/maplibre/maplibre-native/blob/f8c2c6c8b3f1fce6a3529b240d4d3c695a5edfb1/.github/workflows/macos-ci.yml#L3. But there are people like me who work on macOS 😄

ntadej:
That's why I maintain `opengl-2` branch.

louwers:
macOS now support Metal (and Vulkan via MoltenVK).

PR #1127, opened 2023-05-10
Uniform buffer objects

alexcristici:
Uniform Buffer Object Implementation.

### DIFF
--- a/include/mbgl/gfx/uniform_block.hpp
+++ b/include/mbgl/gfx/uniform_block.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <memory>
 #include <string>
+#include <memory>
 #include <unordered_map>
 
 namespace mbgl {
@@ -62,23 +62,23 @@ public:
 
     /// Get an uniform block element.
     /// Returns a pointer to the element on success, or null if the uniform block doesn't exists.
-    UniformBlock* get(const std::string& name) const;
+    const std::unique_ptr<UniformBlock>& get(const std::string& name) const;
 
     /// Add a new uniform block element.
     /// Returns a pointer to the new element on success, or null if the uniform block already exists.
-    UniformBlock* add(std::string name, int index, std::size_t size);
+    const std::unique_ptr<UniformBlock>& add(std::string name, int index, std::size_t size);
 
     UniformBlockArray& operator=(UniformBlockArray&&);
     UniformBlockArray& operator=(const UniformBlockArray&);
 
 protected:
-    UniformBlock* add(std::string name, std::unique_ptr<UniformBlock>&& uniformBlock) {
+    const std::unique_ptr<UniformBlock>& add(std::string name, std::unique_ptr<UniformBlock>&& uniformBlock) {
         const auto result = uniformBlockMap.insert(std::make_pair(std::move(name), std::unique_ptr<UniformBlock>()));
         if (result.second) {
             result.first->second = std::move(uniformBlock);
-            return result.first->second.get();
+            return result.first->second;
         } else {
-            return nullptr;
+            return nullref;
         }
     }
 
@@ -87,6 +87,7 @@ protected:
 
 protected:
     UniformBlockMap uniformBlockMap;
+    static std::unique_ptr<UniformBlock> nullref;
 };
 
 } // namespace gfx
--- a/include/mbgl/gfx/uniform_block.hpp
+++ b/include/mbgl/gfx/uniform_block.hpp
@@ -1,0 +1,93 @@
+#pragma once
+
+#include <string>
+
+namespace mbgl {
+namespace gfx {
+
+class UniformBuffer;
+
+class UniformBlock {
+    // Can only be created by UniformBlockArray implementations
+protected:
+    UniformBlock(int index_, std::size_t size_)
+        : index(index_),
+          size(size_) {}
+    UniformBlock(const UniformBlock&) = default;
+    UniformBlock(UniformBlock&& other)
+        : index(other.index),
+          size(other.size) {}
+
+public:
+    virtual ~UniformBlock() = default;
+
+    int getIndex() const { return index; }
+
+    std::size_t getSize() const { return size; }
+
+    virtual void bindBuffer(const UniformBuffer& uniformBuffer) = 0;
+    virtual void unbindBuffer() = 0;
+    
+protected:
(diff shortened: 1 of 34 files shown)


TimSylvester (approved):

ovivoda (approved):

mwilsnd (approved):

include/mbgl/gfx/uniform_block.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    /// Get map of elements.
+    const UniformBlockMap& getMap() const { return uniformBlockMap; }
+
+    /// Number of elements
+    std::size_t size() const { return uniformBlockMap.size(); }
+
+    /// Get an uniform block element.
+    /// Returns a pointer to the element on success, or null if the uniform block doesn't exists.
+    UniformBlock* get(const std::string& name) const;
```
sjg-wdw, line 63: Raw pointers don't bother me all that much, but I can see they bother some.
Should the UniformBlock* be some other C++ structure on return?

TimSylvester: I think this is based on the attribute stuff I did.  It's a `unique_ptr` internally and we don't return ownership, only a temporary reference, but it's optional so it's a pointer.  The nicer semantics would be a `shared_ptr`, but there's no actual need for the refcounting.  Maybe an `optional<reference_wrapper<>>`?

mwilsnd: We could return a `const std::unique_ptr<>&` temporary which would better convey ownership.

sjg-wdw:
This looks good.  I think we can build on this.

sjg-wdw (approved):

include/mbgl/gfx/uniform_block.hpp
sjg-wdw: That feels more C++

alexcristici: I need to return `nullptr`, not sure if this is possible for this return `const std::unique_ptr<>&`.

TimSylvester: works for me

mwilsnd: A more C++-y way, but you could return an optional holding a reference wrapper. Ex: https://godbolt.org/z/3xdrnE3dG

alexcristici: Yes, maybe we should create a global one somewhere.

alexcristici: So we should used `optional<reference_wrapper<>>` as also Tim mentioned?

louwers (approved):

include/mbgl/gfx/uniform_block.hpp
mwilsnd: I'm fine with the pointer for now. The reference wrapper is more verbose but does help convey the returned reference is temporary.

stefankarschti: I'm in favor of `const std::unique_ptr<>&`

include/mbgl/gfx/uniform_block.hpp
```diff
@@ -0,0 +1,91 @@
+#pragma once
+
+#include <string>
```
mwilsnd, line 3: This file doesn't include everything it needs to build the test runner on linux. Include unordered_map and memory here.

include/mbgl/gfx/uniform_buffer.hpp
```diff
@@ -0,0 +1,75 @@
+#pragma once
+
+#include <string>
```
mwilsnd, line 3: ditto

src/mbgl/gl/drawable_gl.cpp
```diff
@@ -67,12 +71,42 @@ void DrawableGL::setVertexArray(gl::VertexArray&& vertexArray_,
...
+
 void DrawableGL::resetColor(const Color& newColor) {
     if (auto* colorAttr = impl->vertexAttributes.get("a_color")) {
         colorAttr->clear();
         colorAttr->set(0, colorAttrValue(newColor));
     }
 }
 
+void DrawableGL::bindUniformBuffers() const {
+    if (const auto& shader = getShader()) {
```
mwilsnd, line 90: `shader` shadows member variable here (-Werror=shadow).

src/mbgl/gl/drawable_gl.cpp
```diff
@@ -67,12 +71,42 @@ void DrawableGL::setVertexArray(gl::VertexArray&& vertexArray_,
...
+            if (!uniformBuffer) {
+                continue;
+            }
+            element.second->bindBuffer(*uniformBuffer);
+        }
+    }
+}
+
+void DrawableGL::unbindUniformBuffers() const {
+    if (const auto& shader = getShader()) {
```
mwilsnd, line 103: ditto

src/mbgl/gl/uniform_block_gl.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <mbgl/gl/defines.hpp>
+#include <mbgl/platform/gl_functions.hpp>
+
+namespace mbgl {
+namespace gl {
+
+using namespace platform;
+
+void UniformBlockGL::bindBuffer(const gfx::UniformBuffer& uniformBuffer) {
+    assert(size == uniformBuffer.getSize());
```
mwilsnd, line 13: `#include <cassert>`

mwilsnd (changes requested):

mwilsnd (changes requested):

src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -235,8 +235,10 @@ void Renderer::Impl::render(const RenderTree& renderTree) {
             }
 
             if (auto& shader = drawable.getShader()) {
```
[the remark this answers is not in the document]
alexcristici: I removed the `if`, not used.

include/mbgl/gfx/uniform_block.hpp
alexcristici: I changed it to `const std::unique_ptr<>&`.

stefankarschti (approved):

mwilsnd (approved):

PR #1132, opened 2023-05-11
Refactor shaders for drawables

mwilsnd:
Brings generated shaders to the drawables branch.

Generate shaders: `node shaders/generate_shader_code.js --i shaders --o shaders/out`
Add new shaders in `renderer_backend.cpp:initShaders` by adding the `BuiltIn` value to the template list in `registerTypes`

Updates `generate_shader_code.js` with a new pathway for manifest parameter `uses_ubos`; Does not emit `uniform` permutations but still references them if the permutation takes the uniform pathway. Uniform names must be preserved for this to work ie `uniform vec4 u_color` becomes a member vec4 of a UBO with the same name.

There is some extra stuff going on with color packing, the library can pack a 4 component color in two floats and interpolate between them. We'll address how to handle that when we run into it.

### DIFF
--- a/include/mbgl/shaders/gl/drawable_fill.hpp
+++ b/include/mbgl/shaders/gl/drawable_fill.hpp
@@ -1,3 +1,6 @@
+// Generated code, do not modify this file!
+// Generated on 2023-05-11T18:11:44.911Z by mwilsnd using shaders/generate_shader_code.js
+
 #pragma once
 #include <mbgl/shaders/shader_source.hpp>
 
@@ -5,12 +8,9 @@ namespace mbgl {
 namespace shaders {
 
 template <>
-struct ShaderSource<BuiltIn::BackgroundProgramUBO, gfx::Backend::Type::OpenGL> {
-    static constexpr const char* vertex = R"(#version 300 es
-precision highp float;
-
-layout (location = 0) in vec2 a_pos;
-
+struct ShaderSource<BuiltIn::FillShader, gfx::Backend::Type::OpenGL> {
+    static constexpr const char* name = "FillShader";
+    static constexpr const char* vertex = R"(layout (location = 0) in vec2 a_pos;
 layout (std140) uniform DrawableUBO {
     mat4 u_matrix;
 };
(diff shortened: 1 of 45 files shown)


include/mbgl/shaders/gl/drawable_fill.hpp
```diff
@@ -19,19 +19,13 @@ void main() {
     gl_Position = u_matrix * vec4(a_pos, 0, 1);
```
alexcristici, line 19: We need to keep the #ifdefs inside the main and the ones that output the varying (in/out). We should remove only the branch that define the uniform, this will go inside the UBO.

alexcristici: Example: 
```
#ifndef HAS_UNIFORM_u_color
uniform lowp float u_color_t;
layout (location = 1) in highp vec4 a_color;
out highp vec4 color;
#else
uniform highp vec4 u_color;
#endif
```
should become

```
#ifndef HAS_UNIFORM_u_color
layout (location = 1) in highp vec4 a_color;
out highp vec4 color;
#endif
```
and 
```
uniform lowp float u_color_t;
uniform highp vec4 u_color;
```
will be included in the UBO.

alexcristici: The one that is in the main will remain the same:
```
    #ifndef HAS_UNIFORM_u_color
color = unpack_mix_color(a_color, u_color_t);
#else
highp vec4 color = u_color;
#endif
```

mwilsnd: This is what the code generator does, this fill placeholder was just copied from the background generic program to satisfy the compiler for now.

In the shader manifest JSON, if you include `uses_ubo` it will drop `uniform` decls but keep all the other pre-processor parameters.

sjg-wdw (approved):

TimSylvester (approved):

alexcristici (approved):

alexcristici (approved):

PR #1139, opened 2023-05-13
Linux tests + coverage report with Bazel

louwers:
Just trying out Bazel and exploring the build setup.

Two C++ unit tests now can be run with:

```
bazel test //test:core
```


louwers:
I wanted to use the new [Platforms](https://bazel.build/extending/platforms) functionality to configure which headers include and which flags to pass etc. but it turns out `rules_apple` does not have support for it yet: https://github.com/bazelbuild/rules_apple/issues/1658

I need to find some workaround.

louwers:
coecov.io needs to be enabled, but I don't have organization access. So the upload will fail.

platform/linux/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+        "-lz",
+        "-lcurl",
+        "-lpng",
+        "-ljpeg",
+    ] + ICU_FLAGS,
+    visibility = ["//visibility:public"],
+    deps = [
+        "//:mbgl-core",
+        "//platform/default:mbgl-default",
+        # "//:headers"
```
louwers, line 31:
```suggestion
```

louwers, line 31:
```suggestion
        # "//:headers"
```

louwers, line 31:
```suggestion
```

ovivoda (approved):

mwilsnd (approved):

PR #1141, opened 2023-05-15
Add push trigger to linux-ci

louwers:


birkskyum (approved):

birkskyum:
So this is after a merge happen, right?

louwers:
@birkskyum In that case it will run for any push, even if just a Markdown file changed. But I can check.

birkskyum:
We know the current code works, so it's fine by me to leave it at that.

Issue #1118, opened 2023-05-08
Line Layer


sjg-wdw:
Drawable/builder implementation of the line layer.

PR #1142, opened 2023-05-15
Line Layer

stefankarschti:
#1118


sjg-wdw (approved):
Looks like a good start!

include/mbgl/shaders/gl/circle.hpp
```diff
@@ -1,5 +1,5 @@
 // Generated code, do not modify this file!
-// Generated on 2023-05-12T17:51:34.322Z by mwilsnd using shaders/generate_shader_code.js
```
alexcristici, line 2: In order to avoid conflicts it will be good to revert these files, that doesn't have any important changes.

TimSylvester: I've been avoiding committing those, but it's a bit tedious.  The same basic info is in git-blame ("authors" pane in Xcode), so I would advocate removing the date and username.

stefankarschti: +1

TimSylvester (approved):

platform/ios/platform/ios/app/MBXViewController.m
```diff
@@ -1457,6 +1462,59 @@ -(void)toggleStyleLabelsLanguage
     [self.mapView.style localizeLabelsIntoLocale:_localizingLabels ? [NSLocale localeWithLocaleIdentifier:@"mul"] : nil];
 }
 
+- (void)styleLineGradient
```
stefankarschti, line 1465: @louwers : this is added to test the line gradient

platform/ios/platform/ios/app/MBXViewController.m
```diff
@@ -1909,6 +1967,9 @@ - (void)setStyles
     [self.styleNames addObject:@"MapLibre Basic"];
     [self.styleURLs addObject:[NSURL URLWithString:@"https://demotiles.maplibre.org/style.json"]];
 
+    [self.styleNames addObject:@"Line Pattern"];
+    [self.styleURLs addObject:[NSURL URLWithString:@"https://api.maptiler.com/maps/46b24e19-6503-4b2a-81a4-ea6d8431fc92/style.json?key=pdr3bbcne6eRBgLXQ1kg"]];
+
```
alexcristici, line 1972: Probably this should be removed.

TimSylvester (approved):

src/mbgl/renderer/layers/render_line_layer.hpp
```diff
@@ -8,18 +8,37 @@
...
+#include <memory>
+
 namespace mbgl {
 
 class LineProgram;
 class LineGradientProgram;
 class LineSDFProgram;
 class LinePatternProgram;
 
+namespace gfx {
```
mwilsnd, line 21: Wrap these in `#if MLN_DRAWABLE_RENDERER`, keeping forward decls out of the legacy path is helpful for catching other places we may forget to exclude code.

mwilsnd (changes requested):

mwilsnd (approved):

PR #1147, opened 2023-05-16
ASAN fixes

mwilsnd:
Fixes use-after-free, missing virtual destructors

### DIFF
--- a/src/mbgl/renderer/render_orchestrator.cpp
+++ b/src/mbgl/renderer/render_orchestrator.cpp
@@ -834,7 +834,8 @@ void RenderOrchestrator::observeLayerGroups(std::function<void(const LayerGroup&
 void RenderOrchestrator::updateLayers(gfx::ShaderRegistry& shaders,
                                       gfx::Context& context,
                                       const TransformState& state,
-                                      const std::shared_ptr<UpdateParameters>& updateParameters) {
+                                      const std::shared_ptr<UpdateParameters>& updateParameters,
+                                      RenderItems& layerRenderItems) {
     const bool isMapModeContinuous = updateParameters->mode == MapMode::Continuous;
     const auto transitionOptions = isMapModeContinuous ? updateParameters->transitionOptions
                                                        : style::TransitionOptions();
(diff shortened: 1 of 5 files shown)


src/mbgl/renderer/render_orchestrator.cpp
```diff
@@ -850,10 +851,11 @@ void RenderOrchestrator::updateLayers(gfx::ShaderRegistry& shaders,
     // Layer index, similar but not identical to `layerRenderItems` index in render_impl
     // int32_t i = static_cast<int32_t>(layerRenderItems.size()) - 1;
     // for (auto it = layerRenderItems.begin(); it != layerRenderItems.end() && i >= 0; ++it, --i) {
-    auto index = static_cast<int32_t>(renderLayers.size()) - 1;
+    auto index = static_cast<int32_t>(renderTree.getLayerRenderItems().size()) - 1;
```
mwilsnd, line 854: @TimSylvester Does this change sound correct to you? Things render and look correct on my end.

TimSylvester: If those are the active layers, yeah.  The potential issue, I think, would be making sure that the layer gets a chance to remove its drawables when the layer is removed or becomes inactive.  That's based on the layer diff in the render tree, but if there's an additional filter step, it should be reflected there as well.

TimSylvester (approved):

stefankarschti (approved):

src/mbgl/renderer/render_orchestrator.cpp
mwilsnd: There is one more layer of filtering, where the orchestrator checks each render layer for for sources and if the zoom fits the layer. Looking further downwind, I see the layer index is used to create tile layer groups and drawables. I don't think either of these uses are entirely correct given it is possible to insert custom layers arbitrarily in the stack (`style->addLayer()`). Doing so means previously rendered layers would possibly have stale layer indices, if we treat the layer index as immutable for the lifetime of a layer.

We should revisit this later and determine the best option, perhaps we need to invalidate render layers when the stack of renderable layers changes (getLayerRenderItems), with render layers updating their respective groups and drawables. Or switch to another means of sorting and tracking layer groups and their renderables.

PR #1148, opened 2023-05-16
apply color attributes

TimSylvester:
I got the colors applied in the fill layer, but only with a terrible downcast on `PaintPropertyBinder` that depends on the style.

I looked at instead adding a virtual method to `PaintPropertyBinder` but I couldn't figure out how to construct the return value type from the type parameter pack.  So I fell back on a `void*`, which is only slightly less bad.

Any ideas on a better way to do this are welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -928,6 +928,7 @@ if(MLN_WITH_OPENGL)
             ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/background.hpp
             ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/drawable_background.hpp
             ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/drawable_fill.hpp
+            ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/drawable_fill_outline.hpp
             ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/background_pattern.hpp
             ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/circle.hpp
             ${PROJECT_SOURCE_DIR}/include/mbgl/shaders/gl/clipping_mask.hpp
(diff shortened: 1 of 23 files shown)


sjg-wdw:
I'd be interested in talking about how to tease all these various malleable properties out of the style sheet.
It seems like this is a key problem in breaking things up.

stefankarschti:
it would be desirable that the DrawableBuilder could "adopt" a vertex buffer with attributes & either an index array or provide an overload addTriangle() method that accepts 3 indexes instead of positions.

sjg-wdw:
> it would be desirable that the DrawableBuilder could "adopt" a vertex buffer with attributes & either an index array or provide an overload addTriangle() method that accepts 3 indexes instead of positions.

Do you mean the vertex buffer would be built elsewhere?

TimSylvester:
It's adopting a set of attributes that will be built into the buffer if they match those on the shader program.

stefankarschti:
An VertexAttributeArray & indexes could be built and supplied, then the builder could build and upload the vertexbuffer.  but that is not where we want to go with the builder concept I think.

The other option is to have specialized builders that can do the buckets' job and remove buckets. I believe we're going into that direction by supplying addTriangle methods. For LineLayer to work, the builder would need to have something similar to LineBucket::addGeometry()

TimSylvester:
The layer is already building vertexes and indexes and supplying them to the builder.  The point of this change is to have the layer also supply attributes in a `VertexAttributeArray`.  The actual build/upload happens later because the attributes can potentially change and the builder is gone by that point.

We probably want to replace at least some of the bucket functionality, but not yet.  If the geometry produced by the bucket doesn't match `addTriangles`, feel free to add another method/override.

Similarly, if it makes more sense to provide the attributes one-by-one along with vertexes, rather than providing an entire `VertexAttributeArray`, that should also be no problem, we would collect them in the `VertexAttributeArray` within the builder.

stefankarschti:
> If the geometry produced by the bucket doesn't match addTriangles, feel free to add another method/override.
Ok

This PR is a step forward, regardless of the void*, I think we can find a better solution for that later.

stefankarschti (approved):

PR #1180, opened 2023-05-26
Remove generated shader code from repo

### LINKED ISSUE #1166: Remove generated shaders from source control
The `shaders/generate_shader_code.js` script generates shaders.

- [ ] Use `add_custom_command` with CMake and a genrule with Bazel to generate the shader code.
- [ ] Install the needed npm dependencies with each respective build tool as well.
- [ ] Remove the generated shaders from source control, add the generated files to `.gitignore`.

Related ticket #310

louwers:
Will close #1166 

I moved `shaders/**` to `shaders/gl` to prepare for the Metal shaders.

CMake is done but I still need to update Bazel (which will be a bit easier I think).

Draft PR for now.


ntadej:
Can we generate shaders separately as an artifact? I do not appreciate that node becomes a dependency of MapLibre native.

louwers:
> Can we generate shaders separately as an artifact? I do not appreciate that node becomes a dependency of MapLibre native.

You don't want to require Node.js for building right?

I guess it's possible, but code generation as part of the build process is probably easier to maintain. We can also use a different language. Maybe C++? (I would not want to implement code generation in CMake...)

PR #1182, opened 2023-05-26
Don't reuse git dir across runs node-ci and ios-ci

louwers:
We recently deleted a submodule, but the self-hosted runners for Node.js and macOS now fail with the message that the deleted submodule cannot be found in `.gitmodules`.

This can be avoided by not re-using the same git directory across runs.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -88,8 +88,11 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          submodules: recursive
+          submodules: true
           path: ${{ github.sha }}
+
+      - run: ls -R
+        if: runner.os == 'Linux'      
 
       - name: Get OS Architecture
         if: runner.os == 'MacOS' || runner.os == 'Linux'
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -88,10 +88,7 @@ jobs:
         with:
           fetch-depth: 0
           submodules: true
-          path: ${{ github.sha }}
-
-      - run: ls -R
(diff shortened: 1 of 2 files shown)


.github/workflows/node-ci.yml
```diff
@@ -80,21 +80,15 @@ jobs:
...
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-
-      - name: Setup submodules
-        shell: bash
-        run: |
-          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
-          git submodule sync --recursive
-          git -c core.longpaths=true -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive || true
```
acalcutt, line 97: just an fyi, there is a reason we did it this way. On the windows platform the normal checkout has issues with long file names, which this got around.

It may be possible that with the recent reorg of files that this path issue was no longer a problem, but we would have to look out for that.

mwilsnd: I remember this causing problems for me cloning on my windows system but I just did a recursive clone and it seems fine now. I checked and I still have long path support disabled at the moment.

If I recall correctly it was test fixtures in a submodule that had extremely long file names which caused problems.

louwers: Thanks! I am indeed running into that problem when I do a recursive merge.

louwers:
@mwilsnd This seems to be the issue https://github.com/actions/checkout/issues/354

The problem is only showing up on self-hosted runners. I created a new issue there and I deleted the `actions-runner/_work` directory on the self-hosted runners. I hope that will solve the problem.

PR #1183, opened 2023-05-26
Add Texture2D and use it in PatternAtlas

mwilsnd:
Adds a new documented Texture2D object, integrates with drawables and builders. Updates the PatternAtlas to use Texture2D as a starting point.


src/mbgl/gl/context.cpp
```diff
@@ -480,6 +481,14 @@ bool Context::setupDraw(const PaintParameters& parameters, const gfx::Drawable&
     setColorMode(parameters.colorModeForRenderPass());
     setCullFaceMode(gfx::CullFaceMode::disabled());
 
+    int32_t nextTextureUnit = 0;
+    for (auto& attachment : drawable.getTextures()) {
+        std::static_pointer_cast<gl::Texture2D>(attachment.texture)->bind(attachment.location, nextTextureUnit++);
```
mwilsnd, line 486: Do we want to bind textures here or in `DrawableGL::draw`? It's actually doing both at the moment because I forgot to pick one.

sjg-wdw (approved):
Looks like a solid design.
I like how we're starting to see patterns in how we store and interact with these assets.

src/mbgl/gl/context.cpp
sjg-wdw: I prefer drawables for this kind of thing, but I defer to the rest of you if you want to keep the Context alive.

src/mbgl/gl/context.hpp
```diff
@@ -124,7 +125,7 @@ class Context final : public gfx::Context {
     State<value::BindFramebuffer> bindFramebuffer;
     State<value::Viewport> viewport;
     State<value::ScissorTest> scissorTest;
-    std::array<State<value::BindTexture>, 2> texture;
+    std::array<State<value::BindTexture>, gfx::MaxActiveTextureUnits> texture;
```
sjg-wdw, line 128: lol

mwilsnd: Yeah that was interesting, I bumped it to 8 which should give plenty of freedom for people implementing custom drawing routines.

TimSylvester (approved):

stefankarschti (approved):

alexcristici (approved):

PR #1187, opened 2023-05-29
Update guide how to run GLFW demo on Ubuntu 22.04

birkskyum:
The instructions how to execute the GLFW demo was missing from the repository, and also the build instructions for linux were obsolete. This updates it

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -37,10 +37,16 @@ Run the following on linux to build the render test runner:
 git submodule update --init --recursive --depth 1
 
 sudo apt update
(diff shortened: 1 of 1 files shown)


louwers (changes requested):
If your system does not support OpenGL ES 3.0, it won't run anymore. ;-)

Can we use the demotiles instead of MapTiler?

birkskyum:
It's not really representative of the features , 3d buildings etc, but yes, I can default the demotiles

louwers (approved):

PR #1188, opened 2023-05-29
Topic/vao per segment

TimSylvester:
Retains segments through to drawing, building separate VAOs referencing the same buffers for each segment.


sjg-wdw (approved):
Good to see the segments addressed.

stefankarschti (approved):

PR #1190, opened 2023-05-30
Consolidate android-ci-pull.yml and android-ci.yml

louwers:
We have to workflows for Android, one for pull requests and one that runs as soon as code is merged into `main`. However there is a lot of duplication, so I removed one and use:

```
        if: github.ref == 'refs/heads/main'
```


ovivoda (approved):
Nice to see this merged!

PR #1194, opened 2023-05-31
Topic/atlas tex 2d

TimSylvester:
Add the background pattern shader and wire it up.  Also fixes a couple of latent bugs that were preventing it from working.

### DIFF
--- a/src/mbgl/gfx/upload_pass.hpp
+++ b/src/mbgl/gfx/upload_pass.hpp
@@ -7,6 +7,7 @@
 #include <mbgl/gfx/index_vector.hpp>
 #include <mbgl/gfx/index_buffer.hpp>
 #include <mbgl/gfx/texture.hpp>
+#include <mbgl/gfx/texture2d.hpp>
 #include <mbgl/util/size.hpp>
 #include <mbgl/util/image.hpp>
 
@@ -20,6 +21,7 @@ class Texture2D;
 class VertexAttributeArray;
 
 using AttributeBindingArray = std::vector<std::optional<gfx::AttributeBinding>>;
+using Texture2DPtr = std::shared_ptr<Texture2D>;
 
 class UploadPass {
 protected:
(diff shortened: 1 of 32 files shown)


src/mbgl/renderer/pattern_atlas.cpp
```diff
@@ -93,7 +93,7 @@ void PatternAtlas::upload(gfx::UploadPass& uploadPass) {
     //     return;
     // }
 
-    if (!atlasTexture2D) {
+    if (!atlasTexture2D || atlasTexture2D->getSize() != atlasImage.size) {
```
mwilsnd, line 96: Looking at this now, if we recreate the atlas after some number of drawables have already been created using it, they're going to have an old atlas texture. When shelfpack reruns, it might cause UV coords to shift enough to not match the older atlas instances in circulation. We should instead move the recreating to inside Texture2D and offer pointer stability here.

src/mbgl/gl/upload_pass.cpp
```diff
@@ -84,6 +84,26 @@ std::unique_ptr<gfx::TextureResource> UploadPass::createTextureResource(const Si
...
 
+gfx::Texture2DPtr UploadPass::createTexture2D(const Size size,
+                                              const void* data,
+                                              gfx::TexturePixelType format,
+                                              gfx::TextureChannelDataType type) {
+    auto tex = std::make_shared<gl::Texture2D>(commandEncoder.context);
+    int textureByteSize = gl::TextureResource::getStorageSize(size, format, type);
+    commandEncoder.context.renderingStats().memTextures += textureByteSize;
+    commandEncoder.context.pixelStoreUnpack = {1};
+    tex->create(data, *this);
```
mwilsnd, line 95:
```
tex->setFormat(format, type);
tex->create(data, *this);
```

mwilsnd (changes requested):

src/mbgl/gl/upload_pass.cpp
TimSylvester: And `setSize`? That's probably why I was seeing zeros.

mwilsnd: Yeah, I've got a second pass in progress stacked on this PR which should make things a bit nicer. Can go ahead and merge this and follow it up on my second pass.

TimSylvester: Feel free to commit directly unless there's something you want to discuss first.
FYI, it there are some `Texture2D` merge conflicts with #1195.

src/mbgl/gfx/upload_pass.hpp
```diff
@@ -124,7 +147,7 @@ class UploadPass {
                                           TexturePixelType,
                                           TextureChannelDataType) = 0;
 
-    virtual std::shared_ptr<gfx::Texture2D> createTexture2D(const PremultipliedImage& image) = 0;
+    virtual gfx::Texture2DPtr createTexture2D() = 0;
```
stefankarschti, line 150: How about moving this method to Context? the merge with Raster will be easier

TimSylvester: on it...

TimSylvester: Not sure that will be better, I'm having to add a `Context` parameter to a bunch of `upload` methods.

stefankarschti: I suppose those methods have uploadPass? Maybe create a getter for Context in the UploadPass?

TimSylvester: oh, I missed that I can get it from the encoder in there, that should be better

PR #1198, opened 2023-06-01
Update existing UBOs instead of replacing

mwilsnd:
Update UBO contents in-place if a suitable buffer is found already present.

### DIFF
--- a/src/mbgl/gl/uniform_buffer_gl.cpp
+++ b/src/mbgl/gl/uniform_buffer_gl.cpp
@@ -1,7 +1,8 @@
-
 #include <mbgl/gl/uniform_buffer_gl.hpp>
 #include <mbgl/gl/defines.hpp>
 #include <mbgl/platform/gl_functions.hpp>
+
+#include <cassert>
 
 namespace mbgl {
 namespace gl {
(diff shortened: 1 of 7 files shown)


sjg-wdw (commented):
I don't know if it belongs in this one, but I think we need some method of queuing up UBO updates.

For Metal, or anywhere we're potentially rendering multiple frames at once, we need a way to queue up the copy into the buffer.  But perhaps we revisit this with Metal.

mwilsnd:
> I think we need some method of queuing up UBO updates.

I can see a Metal version of the buffers holding a block of staging memory, duplicated to whatever maximum frames in flight we want to support. For the most part I think that can be a hidden implementation detail, we'd just include something like a dirty/needsUpload flag for the current frame being recorded and dispatch a deferred transfer from the staging area during the upload pass.

sjg-wdw:
> > I think we need some method of queuing up UBO updates.
> 
> I can see a Metal version of the buffers holding a block of staging memory, duplicated to whatever maximum frames in flight we want to support. For the most part I think that can be a hidden implementation detail, we'd just include something like a dirty/needsUpload flag for the current frame being recorded and dispatch a deferred transfer from the staging area during the upload pass.

The way we did it last time was an explicit copy action on a Metal command buffer.  That way we didn't need to keep an array of buffers around.  But let's revisit when we get to Metal.

src/mbgl/gl/uniform_buffer_gl.cpp
```diff
@@ -23,5 +24,12 @@ UniformBufferGL::~UniformBufferGL() {
     }
 }
 
+void UniformBufferGL::update(const void* data, std::size_t size_) {
+    assert(size == size_);
```
TimSylvester, line 28: Is `size_>size` a memory corruption and potential crash?  Maybe we should return or limit the size in release builds.

TimSylvester (approved):

src/mbgl/gl/uniform_buffer_gl.cpp
stefankarschti: I agree: assert + log error? + return

stefankarschti (approved):

alexcristici (approved):

TimSylvester (approved):

PR #1221, opened 2023-06-07
Handle texture updates better

TimSylvester:
I got fill-patterns working but, since `RenderTiles` isn't available in the tweaker,  it required the `RenderLayer::update` pushing the result of `tile.getIconAtlasTexture` to each tile drawable.

The tile keeps the atlas textures in a more stable `TileAtlasTextures` object with shared ownership, and indirectly passing that through the drawables with a lambda capture also works, and helps make the drawables more independent of `update`.  This might be better done using an abstract interface rather than a lambda.

Thoughts?


stefankarschti (approved):

sjg-wdw (approved):
I see the logic of this, but we may want to come back to it for Metal and see if we can make it a bit more straightforward.

sjg-wdw (approved):
I see the logic of this, but we may want to come back to it for Metal and see if we can make it a bit more straightforward.

Issue #1197, opened 2023-06-01
Unique image support in Raster Layer


sjg-wdw:
In addition to a source with an image per tile, the Raster Layer also supports unique images.

Issue #1168, opened 2023-05-23
Raster Layer


stefankarschti:
Drawable/builder implementation of the raster layer.

PR #1223, opened 2023-06-08
Raster Layer (images)

stefankarschti:
#1168 #1197 
including creation of a new LayerGroup class, that is not tile bound
added "data" to drawables to accommodate the image matrix. The design is flexible to allow for other forms of "data" being attached to drawables.


src/mbgl/renderer/layers/render_raster_layer.cpp
```diff
@@ -255,40 +250,206 @@ void RenderRasterLayer::update(gfx::ShaderRegistry& shaders,
...
-                it == renderTiles->end()) {
-                // remove it
-                drawable.reset();
-                ++stats.tileDrawablesRemoved;
+
+        // textures
+        std::shared_ptr<gfx::Texture2D> tex0 = context.createTexture2D();
+        tex0->setImage(bucket.image);
+        auto location0 = rasterShader->getSamplerLocation("u_image0");
+        if (location0.has_value()) {
```
mwilsnd, line 345: We can check this first and skip allocating a texture instance if the shader is missing a sampler for whatever reason.

mwilsnd (changes requested):

mwilsnd (approved):

PR #1227, opened 2023-06-08
Circle layer final

alexcristici:
Implemented shader groups and shader permutations.
Improved how vertex attributes are filled from data driven paint properties.
Implemented interpolation UBO.


sjg-wdw (approved):

include/mbgl/gfx/vertex_attribute.hpp
```diff
@@ -219,6 +219,32 @@ class VertexAttributeArray {
         return newAttrs;
     }
 
+    template <class... DataDrivenPaintProperty, class Binders, class Evaluated>
+    std::vector<std::string> readDataDrivenPaintProperties(const Binders& binders, const Evaluated& evaluated) {
+        int index = 0;
+        std::vector<std::string> attributesAsUniforms(sizeof...(DataDrivenPaintProperty));
+        (
```
TimSylvester, line 226: Nice work!

TimSylvester (approved):

include/mbgl/shaders/gl/shader_group_gl.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <mbgl/shaders/shader_source.hpp>
+#include <mbgl/programs/program_parameters.hpp>
+
+namespace mbgl {
+namespace gl {
+
+template <shaders::BuiltIn ShaderID>
+class ShaderGroupGL final : public gfx::ShaderGroup {
+public:
+    ShaderGroupGL(const ProgramParameters& programParameters)
```
mwilsnd, line 14: `programParameters` shadows member of same name.

mwilsnd (changes requested):
Changes here have caused MSVC to become very upset :(.
Starting in fill bucket,
```
= mbgl::PaintPropertyBinder<
    mbgl::style::expression::Image,
    std::array<uint16_t,4>,
    mbgl::PossiblyEvaluatedPropertyValue<
        mbgl::Faded<mbgl::style::expression::Image>>,
    mbgl::attributes::pattern_to::Type,mbgl::attributes::pattern_from::Type>
    
= mbgl::CompositeCrossFadedPaintPropertyBinder<T,A1,A2> with T = uint16_t

= std::tuple<
    mbgl::gfx::detail::VertexType<mbgl::gfx::AttributeType<T,8>>,
    mbgl::gfx::detail::VertexType<mbgl::gfx::AttributeType<T,8>>
> mbgl::CompositeCrossFadedPaintPropertyBinder<
        mbgl::style::expression::Image,A1,A2>::getVertexValue(size_t) const'
with T = uint16_t
```
C2027: use of undefined type 'mbgl::gfx::AttributeDataTypeOf<T,8>'

Because: 
```
CompositeFunctionPaintPropertyBinder:
    using AttributeType = ZoomInterpolatedAttributeType<A>;
    
using ZoomInterpolatedAttributeType =
    gfx::AttributeType<typename AttributeType::ElementType, AttributeType::Dimensions * 2>;

with ElementType = uint16_t
```

AttributeDataTypeOf<uint16_t, 8> is not specialized (in `attribute.hpp`).
Not sure why this isn't seen in GCC or clang.

alexcristici:
> 
This is because `CrossFadedDataDrivenPaintProperty` was not supported, now should be ok.

mwilsnd:
> > 
> 
> This is because `CrossFadedDataDrivenPaintProperty` was not supported, now should be ok.


Still trying to create a `mbgl::gfx::AttributeDataTypeOf<uint16_t,8>` which isn't specialized. [compiler output](https://github.com/maplibre/maplibre-native/files/11707982/output.txt).

src/mbgl/renderer/paint_property_binder.hpp
```diff
@@ -222,9 +224,10 @@ class ConstantCrossFadedPaintPropertyBinder final
         return constantPatternPositions;
     }
 
-    std::size_t getVertexCount() const override { return 1; }
-    const /*std::tuple<ExpandToType<A, T>>&*/ void* getVertexValue(std::size_t) const override {
-        return &constantPatternPositions;
+    std::size_t getVertexCount() const override { return 0; }
+
+    std::tuple<ZoomInterpolatedVertexType<As>...> getVertexValue(std::size_t) const override {
+        return {ZoomInterpolatedVertexType<As>{0}...};
```
alexcristici, line 230: @TimSylvester We don't need to send `constantPatternPositions` as vertex, you have it on the `uniformValue` method.  But I think we should decouple `evaluated` from `binders` so I would propose to compute the `constantPatternPositions` from evaluated.

alexcristici: We keep `binders` for now only for vertex attributes and interpolation UBO. For other uniforms `evaluated` and other params, but without `binders` should be enough.

src/mbgl/renderer/layers/render_circle_layer.cpp
```diff
@@ -308,20 +333,58 @@ void RenderCircleLayer::update(gfx::ShaderRegistry& shaders,
...
+                                                                                    CircleRadius,
+                                                                                    CircleBlur,
+                                                                                    CircleOpacity,
+                                                                                    CircleStrokeColor,
+                                                                                    CircleStrokeWidth,
+                                                                                    CircleStrokeOpacity>(
+            paintPropertyBinders, evaluated);
+
+        auto circleShader = circleShaderGroup->getOrCreateShader(context, propertiesAsUniforms);
+        if (!circleShader) {
```
alexcristici, line 385: This is the code where our vertex attributes reads from data driven paint properties and generates the permutation as return to be used to create the shader variant.

src/mbgl/renderer/layers/render_circle_layer.cpp
```diff
@@ -308,20 +333,58 @@ void RenderCircleLayer::update(gfx::ShaderRegistry& shaders,
...
+            /* .stroke_width_t = */
+            std::get<0>(paintPropertyBinders.get<CircleStrokeWidth>()->interpolationFactor(state.getZoom())),
+            /* .stroke_opacity_t = */
+            std::get<0>(paintPropertyBinders.get<CircleStrokeOpacity>()->interpolationFactor(state.getZoom())),
+            /* .padding = */ 0};
+
+        tileLayerGroup->observeDrawables(renderPass, tileID, [&](gfx::Drawable& drawable) {
+            drawable.mutableUniformBuffers().createOrUpdate(CircleInterpolateUBOName, &interpolateUBO, context);
+        });
+
```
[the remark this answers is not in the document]
mwilsnd: Having access to evaluate style expressions in the tweaker sounds sensible, I can see this being part of the refactor of renderable dependencies (RenderTile & atlases).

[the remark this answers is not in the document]
alexcristici: Yes. `expression.interpolationFactor(zoomRange, possiblyRoundedZoom)` - https://github.com/maplibre/maplibre-native/blob/26094930e90c7328ec34c2fb3850a41d4cd609a6/src/mbgl/renderer/paint_property_binder.hpp#L416

mwilsnd (approved):

stefankarschti (approved):

PR #1230, opened 2023-06-09
Heatmap layer

alexcristici:


src/mbgl/renderer/layers/render_heatmap_layer.cpp
```diff
@@ -209,4 +229,261 @@ bool RenderHeatmapLayer::queryIntersectsFeature(const GeometryCoordinates& query
...
+
+    if (!renderTiles || renderTiles->empty()) {
+        removeAllDrawables();
+        return;
+    }
+
+    // Set up a render target
+    if (!renderTarget) {
+        // const auto& viewportSize = parameters.staticData.backendSize;
+        const auto size = Size(500, 500); // Size{viewportSize.width / 4, viewportSize.height / 4};
```
alexcristici, line 289: I still have this know issue, not sure where to set the size of the render target.

sjg-wdw (approved):
From the architecture perspective, I like this quite a lot.  Tim and Marc should weigh in from theirs, though.

TimSylvester (approved):
Looks good to me

src/mbgl/gl/render_target_gl.cpp
```diff
@@ -0,0 +1,85 @@
...
+    glContext.clear(Color{0.0f, 0.0f, 0.0f, 1.0f}, {}, {});
+
+    // Run layer tweakers to update any dynamic elements
+    observeLayerGroups([&](LayerGroupBase& layerGroup) {
+        if (layerGroup.getLayerTweaker()) {
+            layerGroup.getLayerTweaker()->execute(layerGroup, renderTree, parameters);
+        }
+    });
+
+    // draw layer groups, opaque pass
```
mwilsnd, line 61: This does a lot of what `Renderer::Impl::render()` does, we should consider modularization of that logic and break it out of `Impl::render()` so we can also call it here.

src/mbgl/renderer/layers/render_heatmap_layer.cpp
mwilsnd: `state.getSize()` should be the viewport.

src/mbgl/gl/render_target_gl.cpp
sjg-wdw: Yeah, it reminded me a bit of the way we do render targets in WhirlyGlobe.  We eventually mostly consolidated that for the default render target (the screen) and other render targets (other textures we draw to).  But it wasn't trivial.

I think it's worth opening an issue and having someone spend a bit of time on it.  It's something I'd like to see fleshed out in MapLibre.

src/mbgl/renderer/layers/render_heatmap_layer.cpp
sjg-wdw: It's curious to use the viewport for that.  Normally you'd pick a fixed size, like 256 or so, and just have that backed into your tile size for your source.  At least that's how we'd do it.

stefankarschti (approved):

src/mbgl/renderer/layers/render_heatmap_layer.cpp
alexcristici: Thanks @mwilsnd. `getSize()` returns the view size, which ignores the scale factor, so I divided by 2, instead of 4.

src/mbgl/gl/render_target_gl.cpp
alexcristici: Yes, we will need to optimize this.

PR #1235, opened 2023-06-10
[macos] Tried to make it works with OpenGL >= 3.3

vng:
Tried to _alive_ MacOS with new 3.3 shaders.
Now we have [hardcoded shaders "version 300 es"](https://github.com/maplibre/maplibre-native/pull/995) and I suspect that it breaks not only MacOS, but also some desktop Linux that doesn't have ES compatibility support.

I got a working version, but still wrong with the texture's format. Since GL_ALPHA is not supported in >= 3.3 core, I replaced it with GL_RED/GL_R8.

Got this picture:
Screenshot 2023-06-10 at 17 26 34

I'm not an expert in OpenGL, so if somebody has ideas on how to fix it, please welcome :)


louwers:
OpenGL ES 3.0 was a deliberate choice. Please see the discussion here: https://github.com/maplibre/maplibre-native/discussions/923

After the renderer modularization design proposal is implemented, a Metal rendering backend will be implemented, which will bring back macOS support on the main branch. macOS support is currently available on the opengl-2 branch.

Support for other graphics backends are possible at a later point by creating additional alternate rendering backends, such as for Vulkan or other OpenGL variants.

vng:
I read that. My initial goal was not to merge this changes, but ask for help in textures, if any ...

And this is not only about MacOS, current approach is hardcoded for OpenGL ES, but not working with 3.3+ core OpenGL standard  and I think will not work also for Linux desktops.

vng:
So my PR is a playground on how it will be with "other OpenGL variants". Sooner or later we should support OGL 3.3 _classic_ core standard.

vng:
Well, in the Android world many depends on the concrete vendor, but I suppose that ES standard is the most common.

vng:
For those who are interested. My branch works on OpenGL3+ core for desktops.
Screenshot 2023-06-30 at 10 46 44

cjbd:
@vng , i tried manually merge this in latest main branch for linux glfw platform, it doesn't work... my opengl skill is not up to the task ... if possible, can you please fix linux glfw using opengl core profile, thanks!

birkskyum:
@mwilsnd , a lot has happened since the discussion above in 2023 - given our present situation, with metal/vulkan/opengl support, would it make sense to do this shift from .a to .r, for compatibility? When working on Qt recently, i did stumble upon this issue a few times, with texts becoming blocks, which presumably must have been due to a lack of alpha.

PR #1239, opened 2023-06-12
Fix Android code generation templates & regenerate generated sources

louwers:
- I fixed the `.ejs` files I messed up.
- `LightTest.java` is actually generated, but this was one of the tests that @fynngodau fixed. So I edited `light.junit.ejs` to reflect this change (see line 91).


platform/android/DEVELOPING.md
```diff
@@ -58,6 +58,14 @@ final draft of the C++14 standard. More information in [DEVELOPING.md](../../DEV
 
 ## Opening the project
 
+Run
+
+```
+make style-code
+```
+
+To generate the needed generated sources.
```
birkskyum, line 67: If it's possible to make this comment a bit more specific, that would be good, because after reading it I still don't know what make style-code does.

birkskyum:
In Android Studio, I have it setup so that I can run the example app, and it works with this make style-code, but how do I execute the junit test suites?

fynngodau:
I think we would benefit greatly from having this `make style-code` step as a part of the gradle build process. The following snippet in `build.gradle` seems like a good start, but it is not functional yet and does not compile:

```
task generateStyleCode {
    doLast {
        commandLine 'sh', '-c', 'make style-code'
    }
}

preBuild.dependsOn generateStyleCode
```

louwers:
> I think we would benefit greatly from having this `make style-code` step as a part of the gradle build process. The following snippet in `build.gradle` seems like a good start, but it is not functional yet and does not compile:
> 
> ```
> task generateStyleCode {
>     doLast {
>         commandLine 'sh', '-c', 'make style-code'
>     }
> }
> 
> preBuild.dependsOn generateStyleCode
> ```

The `Makefile` is a wrapper around Gradle... We should not let Gradle also use the Makefile. That just does not feel very clean to me.

Ideally the build system should be smart enough to pull in all needed dependencies and also to keep track of when dependencies are changed so that it can re run the code generation (and only then). I could look into https://github.com/node-gradle/gradle-node-plugin

birkskyum (approved):

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/style/layers/PropertyFactory.java
```diff
@@ -1806,7 +1806,7 @@ public static PropertyValue<Expression> symbolSpacing(Expression value) {
   }
 
   /**
-   * If true, the symbols will not cross tile edges to avoid mutual collisions. Recommended in layers that don't have enough padding in the vector tile to prevent collisions, or if it is a point symbol layer placed after a line symbol layer. When using a client that supports global collision detection, like MapLibre GL JS version 0.42.0 or greater, enabling this property is not needed to prevent clipped labels at tile boundaries.
+   * If true, the symbols will not cross tile edges to avoid mutual collisions. Recommended in layers that don't have enough padding in the vector tile to prevent collisions, or if it is a point symbol layer placed after a line symbol layer. When using a client that supports global collision detection, like Mapbox GL JS version 0.42.0 or greater, enabling this property is not needed to prevent clipped labels at tile boundaries.
```
birkskyum, line 1809: Is this correct? going from maplibre to mapbox?

louwers: Technically there never was a MapLibre GL JS 0.42

louwers: But yes this is 'correct' it uses an older `v8.json`. We can update the `v8.json` but that should be done in a separate PR and may break other things.

PR #1247, opened 2023-06-14
Selectable renderer implementation

mwilsnd:
Isolates both legacy rendering and drawable rendering pathways via the preprocessor. Adds CMake options to configure the desired pathway(s).

Configure using the legacy renderer:
`cmake . -B build -DMLN_LEGACY_RENDERER=ON -DMLN_DRAWABLE_RENDERER=OFF`
image

Configure using the drawable pathway (the default configuration):
`cmake . -B build`
image

Configure with both renderers, enabling split view:
`cmake . -B build -DMLN_LEGACY_RENDERER=ON -DMLN_DRAWABLE_RENDERER=ON`
image


sjg-wdw:
This is great!

mwilsnd:
> This is great!

Bazel is up next, this is about 80% of the way there.

sjg-wdw (approved):
There might be some minor changes mixed in?  But seems good in any case.

mwilsnd:
> There might be some minor changes mixed in? But seems good in any case.

This is a near-full isolation of our added objects so some preprocessor branches had to be included which bring back old behavior when using legacy mode.

sjg-wdw:
> > There might be some minor changes mixed in? But seems good in any case.
> 
> This is a near-full isolation of our added objects so some preprocessor branches had to be included which bring back old behavior when using legacy mode.

Gotcha!

src/mbgl/gl/context.cpp
```diff
@@ -3,26 +3,29 @@
...
+#include <mbgl/gl/drawable_gl_builder.hpp>
+#include <mbgl/gl/drawable_gl_tweaker.hpp>
+#include <mbgl/gl/layer_group_gl.hpp>
+#include <mbgl/gl/uniform_buffer_gl.hpp>
+#include <mbgl/gl/texture2d.hpp>
+#include <mbgl/shaders/gl/shader_program_gl.hpp>
+#endif
+
 #include <cstring>
 #include <iterator>
```
stefankarschti, line 31: I am not sure how this is related but I get a crash in `Context::initializeExtensions `at `getProcAddress` line 87.
`probe.second` is _"glPushGroupMarkerEXT"_
Env: Mac, iOS Simulator (iPad), Debug build

TimSylvester (commented):
I tried a few things based on [this](https://cmake.org/pipermail/cmake/2008-February/020010.html) but wasn't able to get CMake to produce consistent project files.  Editing them manually works well, though.

mwilsnd:
I've got a working solution now for Xcode, though I'm not 100% happy with it.
```
> cd platform/ios/platform/ios/scripts
> ./bazel-xcodeproj.sh flavor [legacy|drawable|split] teamid 0000000000
```

The shell script will generate `platform/ios/__generated__/provisioning_profile.bzl` which just puts your team ID in a variable. I wasn't able to find a way to pass it to where I needed it as a custom build setting like renderer flavor works, so this will do for now.

Team ID is required if building for device, otherwise you can put whatever and just do simulator builds. Due to how bazel works with Xcode, trying to change provisioning from inside Xcode under Signing & Capabilities won't apply to your build, bazel overrules it.

This is hard-coded to use your personal/local team, so those without a paid developer account can run on iPhone. Grab your local team ID via the keychain app:
* Open Keychain Access
* Locate "Apple Development: you@email (XXXXXXXXXX)" under the "login" keychain
* Right click it and select "Get Info"
* Your team ID is listed under "Organizational Unit"

With bazel generating the iosapp now, we're probably going to work towards deprecating CMake. At least for Apple platforms first.

tl;dr: **CMake/make iproj are deprecated and non-functional with this PR, use bazel.**

louwers:
I think we need to remove the `profile_name`.

I am getting an error that a profile with this name has not been found.

mwilsnd:
That's very strange. All the renderer variants work when I select them (Tested on a physical device):
`❯ ./bazel-xcodeproj.sh --flavor legacy --teamid`
Screen Shot 2023-06-25 at 10 02 44 AM

`❯ ./bazel-xcodeproj.sh --flavor drawable --teamid`
Screen Shot 2023-06-25 at 10 21 05 AM

`❯ ./bazel-xcodeproj.sh --flavor split --teamid`
Screen Shot 2023-06-25 at 10 35 40 AM

As for provisioning, I wasn't able to make it work without both a team ID and a name. I ended up going to Xcode -> Settings -> Accounts, made sure I was signed in with my Apple ID, selected my personal team in the list and clicked "Download Manual Profiles". Then as you found, located them in `~/Library/MobileDevice/Provisioning\ Profiles` and opened them in a text editor. This snippet shows what is inside one of my local profiles: 
```
<key>ExpirationDate</key>
<date>2023-06-30T17:31:24Z</date>
<key>Name</key>
<string>iOS Team Provisioning Profile: com.maplibre.app</string>
<key>ProvisionedDevices</key>
<array>
	<string> ...-... </string>
</array>
<key>LocalProvision</key>
<true/>
<key>TeamIdentifier</key>
<array>
	<string> ... </string>
```
That is where the name comes from, I'm curious what you find in your profiles. An alternate option we have here is to just give the mobileprovision file to the build instead of letting Xcode try and resolve it. That will involve downloading your profile and copying it to a location the build can reference.

louwers:
@mwilsnd Oops, it works now. I must have accidentally ran `bazel run //platform/ios:xcodeproj`  

I still need to set the `profile_name` manually. Maybe the script should accept that as an argument as well.

We should consider wrapping the script with `sh_binary` and set `local = True` so it doesn't run in the sandbox. Then users don't have to `cd` into the scripts directory but can run it through Bazel.

mwilsnd:
Alright, I've added `--profile--uuid [UUID]` to the bazel build. So to build for device:
* Navigate to `~/Library/MobileDevice/Provisioning\ Profiles`
* Click a profile and press space to quick view it
* Note the UUID and team ID values
* Generate the project: `❯ ./bazel-xcodeproj.sh --flavor drawable --teamid TEAMID --profile-uuid PROFILE-UUID`

louwers (approved):

TimSylvester (approved):

PR #1249, opened 2023-06-15
Topic/drawable symbol

TimSylvester:


TimSylvester:
rebased

mwilsnd:
Something isn't quite right with the layout: 
image

TimSylvester:
Seems to be `text-variable-anchor`...

TimSylvester:
Turns out `SymbolBucket.hasVariablePlacement` changes during the layout phase, which may happen after we create drawables.  I wonder if the render tests are sensitive to that?

Screen Shot 2023-06-22 at 9 36 49 AM

TimSylvester:
I saw the icons rendering incorrectly, like the atlas or offsets were wrong, but Xcode crashed before I could debug it, and I haven't been able to reproduce it.
Otherwise, it seems to be working.

sjg-wdw (approved):
Doesn't look too weird!

src/mbgl/renderer/layers/render_raster_layer.cpp
```diff
@@ -239,7 +242,7 @@ void RenderRasterLayer::update(gfx::ShaderRegistry& shaders,
                                               ? gfx::TextureFilterType::Nearest
                                               : gfx::TextureFilterType::Linear;
 
-    auto createBuilder = [&context, &renderPass, this]() -> std::unique_ptr<gfx::DrawableBuilder> {
+    auto createBuilder = [&context, this]() -> std::unique_ptr<gfx::DrawableBuilder> {
```
mwilsnd, line 245: Capture `renderPass` by value here

mwilsnd (changes requested):

src/mbgl/renderer/layers/render_raster_layer.cpp
TimSylvester: Capturing by value or references produces `Lambda capture 'renderPass' is not required to be captured for this use`, that's why I removed it.

TimSylvester:
I found one problem: when changing between styles that contain the same layer, `RenderLayer::update()` may be called up to several times with the old `bucket` still in place after the tile drawables have been removed.  That caused new drawables to be created based on the old bucket data, which those drawables would later be updated by the tweaker with properties from the new style, leading to various strange results.

I suspect we want to avoid creating drawables until the bucket layout is complete, but I don't see any way to check for that directly.  Checking `needsUpload` did not work.  For now, I'm detecting when an ID value in the bucket is changing, and that seems to suffice.

src/mbgl/renderer/layers/render_raster_layer.cpp
mwilsnd: MSVC wants an explicit capture: https://godbolt.org/z/MxGTEbsdK

TimSylvester: I put it back the way it was, Stefan's still working on this anyway.

mwilsnd (approved):

stefankarschti (approved):

alexcristici (approved):

bazel/core.bzl
```diff
@@ -57,6 +57,7 @@ MLN_CORE_SOURCE = [
     "src/mbgl/gfx/shader_registry.cpp",
     "src/mbgl/gfx/shader_group.cpp",
     "src/mbgl/gfx/stencil_mode.hpp",
+	"src/mbgl/gfx/symbol_drawable_data.hpp",
```
mwilsnd, line 60: You include this file in the core here but in the list of drawables source for CMakeLists. Move this to MLN_DRAWABLES_HEADERS. Everything else looks good.

mwilsnd (changes requested):

mwilsnd (approved):

PR #1250, opened 2023-06-16
Fix Android instrumentation tests

fynngodau:
* Merge faulty tests in `OfflineManagerTest`
   * Tests `a` through `e` merged into one
   * Additional test at the end to ensure that listing regions after deletion yields no regions
* Fix behavior in `FontUtils`


louwers (approved):
Awesome. I think we are really close to having green CI again for Android. 💯

Issue #1248, opened 2023-06-15
Modernize iOS Codebase


louwers:
The iOS SDK has been mostly stale since the fork. During yesterday's TSC Meeting several ideas were proposed to improve it.

- Remove deprecated APIs, replace by modern alternatives.
- Remove 'cruft' that is unlikely to be used anymore.
- Make the library more ergonomic to use with Swift.
- Update and improve (build) tooling and setup (DocC for docs, [Bazel](https://github.com/maplibre/maplibre-native/issues/840)).

Please drop any modernization ideas or more detailed suggestions of the above in this thread. Thanks to the [Metal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2022-11-29-metal-port.md) implementation project that is well underway the low-level graphics rendering of MapLibre Native on iOS will be in a great place, I hope we can give the user-facing SDK the same love together.

Thanks to the [Modernization](https://github.com/maplibre/maplibre/issues/270) bounty direction, there is money available to award those brave [developers](https://github.com/maplibre/maplibre/wiki/Developer-Role) willing to work with Objective-C in 2023. This is an open-ended ticket: suggest a piece of work, we define the scope and agree on [bounty size](https://github.com/maplibre/maplibre/wiki/Bounty-System#bounty-sizes). See also the [Step-by-Step Bounties Guide](https://maplibre.org/roadmap/step-by-step-bounties-guide/).

louwers:
The iOS tests written in Objective-C++ contain a lot of deprecated stuff:

```
platform/ios/platform/darwin/test/MLNCodingTests.mm:27:22: error: 'archiveRootObject:toFile:' is deprecated: first deprecated in iOS 12.0 - Use +archivedDataWithRootObject:requiringSecureCoding:error: and -writeToURL:options:error: instead [-Werror,-Wdeprecated-declarations]
    [NSKeyedArchiver archiveRootObject:annotation toFile:filePath];
                     ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator16.4.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSKeyedArchiver.h:48:1: note: 'archiveRootObject:toFile:' has been explicitly marked deprecated here
[...]

PR #1255, opened 2023-06-18
Android Annotations proposal

### LINKED ISSUE #1189: Plan changes to Android annotation API
When starting work on #1154, I noticed it can't be implemented as easily as we were hoping, for the following reasons:

* The code currently located in the Annotations repo has classes like `Symbol`, `Fill` and so on, that
the older implementation in `maplibre-native` does not have.
* Instead, it has the classes `Marker`, `Polygon` and more that imitate the Google Maps API and call some native code: https://github.com/maplibre/maplibre-native/blob/749458d6c866223c4966b0c6734b8d5cd8ad9744/platform/android/MapboxGLAndroidSDK/src/cpp/native_map_view.cpp#L655-L673
* This native code itself has classes like `FillAnnotation`, `SymbolAnnotation`  and probably generates `Geometry` somewhere, which the annotations code also does, but not in C++ code.
* Because the old API (deprecated and in `maplibre-native`) tries to immitate the Google Maps API whilst the new one (in annotations plugin) does not, there is no obvious path towards moving the code whilst un-deprecating the deprecated methods like we planned, without making the migration from the current API – which I imagine most users are using – unnecessarily complex.

Due to these considerations, we should make a plan for how we want the API to look like in the future. Some points that I think are important for this, regarding what aspects are nice about each of the APIs:

* The old API had methods directly on the map object that allowed to `addMarker`, not bothering users with creating an object like `SymbolAnnotations`. I like that.
* In total the two APIs are quite similar obviously (which is why we considered a merge in the first place); their user-facing renamings are pretty much a matter of taste – but it makes more sense to stick with the new terminology (Symbol, Fill) which is also used internally instead of trying to pretend to be Google Maps.
* In certain places, the newer annotations API has the user wangle more with its internals. Here is one example of this: `SymbolOptions.withIconImage` takes a `String` that references the icon, and it is the user's responsibility to put their resource into the style, whereas `Marker.setIcon` takes an Icon object where it is easy to find out how to construct it. The latter is much preferrable and we should keep such an easy interface for the future.

Goals of this issue:

* Create a concrete design proposal for changes to the API for annotations in Android.
* The API should be a part of `maplibre-native` according to this design proposal.
* It should aim to be harmonic and user-friendly, and create as few friction as necessary for existing users of the annotations plugin during migrations.

fynngodau:
Fixes #1189 by planning a new and improved annotations API for Android.

Latest version for reading: https://github.com/fynngodau/maplibre-native/blob/annotations-proposal/design-proposals/2023-06-17-android-annotations.md

### DIFF
--- a/design-proposals/2023-06-17-android-annotations.md
+++ b/design-proposals/2023-06-17-android-annotations.md
@@ -64,7 +64,7 @@ As stated above, we want to be able to instantiate annotation classes directly. 
 ```
 Symbol(latLng).apply {
     icon = Icon(bitmap)
-    text = "label"
+    text = Text("label", color = Color.RED)
 }.let { map.add(it) }
 ```
 
@@ -133,6 +133,9 @@ The following lists concrete class defintions that should be available to users.
     * `var zLayer: Int` (default `0`)
     * `var draggable: Boolean` (default `false`)
     * `var data: JsonElement?`
+    * `var clickListener: OnAnnotationClickListener`
+    * `var dragListener: OnAnnotationDragListener`
+    * `var longClickListener: OnAnnotationLongClickListener`
 * `Symbol extends Annotation`
     * `var position: LatLng` (non-optional)
     * `var icon: Icon?`
@@ -192,7 +195,7 @@ The following lists concrete class defintions that should be available to users.
     * `val offset: Float`
 * `AbsoluteOffset extends Offset` (inner class of `Text`)
     * `val offset: PointF`
-* `Line`
+* `Line extends Annotation`
     * `var path: List<LatLng>`
     * `var join: Join` (default `MITER`)
     * `var opacity: Float` (default `1f`)
@@ -211,7 +214,7 @@ The following lists concrete class defintions that should be available to users.
     * `val offset: PointF` (maps to `line-translate`)
     * `val anchor: Translate.Anchor`
 * `enum Anchor` (inner class of `Translate`): `MAP`, `VIEWPORT`
-* `Fill`
+* `Fill extends Annotation`
     * `var paths: List<List<LatLng>>`
     * `var opacity: Float` (default `1f`)
     * `var color: @ColorInt Int` (default `Color.BLACK`)
@@ -219,7 +222,7 @@ The following lists concrete class defintions that should be available to users.
     * `var pattern: Bitmap?` (note: if set, `outlineColor` is ignored)
     * `var antialias: Boolean` (default `true`) (NDD)
     * `var translate: Translate?` (NDD)
-* `Circle`
+* `Circle extends Annotation`
     * `var center: LatLng`
     * `var radius: Float` (default `5f`, in pixels)
     * `var color: @ColorInt Int` (default `Color.BLACK`)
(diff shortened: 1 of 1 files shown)


RomanBapst:
@fynngodau Thanks for bringing up this proposal! I went through it today and liked the direction it takes a lot.
Actually it's quite funny because me and @JonasVautherin released [ramani-maps](https://github.com/ramani-maps/ramani-maps) today and many things you have in the proposal are quite similar to what we have done. E.g. managing the layering under the hood and just exposing a zIndex to the user is exactly what we have done. Also the fact that you propose to instantiate the annotation classes directly is pretty much what we do in compose. With compose we have the additional advantage that it's super declarative, e.g. you add a Circle directly to the UI code.

From what I understand the old workflow of adding annotations via the manager would still co-exists, right? That's important for us in our library as we currently make use of it.

What I like a lot are all the options that can be used to fine-tune the appearance of the elements, e.g. Padding, Transform, ...
And of course I am excited to see the as much "Kotlinification" as possible.

fynngodau:
@RomanBapst Your library sounds very exciting!

> From what I understand the old workflow of adding annotations via the manager would still co-exists, right? That's important for us in our library as we currently make use of it.

Yes, this is what I had in mind. It wouls be necessary to create annotations using the changed API, but they could still be added to annotation managers as previously. As explained in the proposal, the two use main cases are

* when you really can't affort the matching algorithm to take O(n log n) (and have additional knowledge about the desired z layers in advance)
* when you want to call methods that are not available on the specific Annotation

systemed:
As a general note, while porting cycle.travel's app from iOS to Android, I've been surprised how different the Android and iOS APIs are.

Obviously each platform has its own design patterns (e.g. iOS delegates, Android builders, etc.) so you wouldn't expect full convergence, but there's a lot that seems needlessly different between the two - annotations being just one example. It means a whole bunch of extra cognitive load when porting or developing in parallel. From looking through issues on the old Mapbox repos it looks like their two teams were working independently.

So perhaps an additional design goal would be appropriate: increased commonality with MapLibre iOS Native. That doesn't necessarily imply changes have to be made here in Androidland - it could equally mean changing the iOS API to match the new, better Android API. But I think it's worth considering as an aim.

louwers:
@systemed Very interesting point! Now that Swift is the language of choice for iOS development, I think it makes sense to offer a Swift-first API in the future. Being a much more modern programming language with quite powerful features (in comparison to Objective-C), I think a lot is possible there. If you have specific ideas in mind or want to contribute, check out the general topic for iOS modernization #1248.

I gave you a follow in LinkedIn, let me know if you ever want to brainstorm about this.

systemed:
Cheers!

Swift and Kotlin are broadly similar languages, each with a good async/await story and heavy use of blocks/closures. To my mind this gives the opportunity to remove some of the more idiomatic design patterns (Java-ish managers and builders, ObjC-ish delegates) and converge on a simpler, more direct API. So @fynngodau's suggestion of de-emphasising manager objects makes a lot of sense in this context.

louwers (changes requested):

design-proposals/2023-06-17-android-annotations.md
```diff
@@ -0,0 +1,281 @@
+# Design Proposal for Android Annotations API
+
+## Motivation
+
+`maplibre-native` contains a deprecated variant of annotation methods that work with objects such as `Marker`, `Polygon` and more that imitate the Google Maps API. On the other hand, `maplibre-plugins-android` repository contains an implementation that works with the objects `Symbol`, `Fill` and more. This situation is confusing to developers and users and should be cleaned up.
```
louwers, line 5:
```suggestion
MapLibre Native contains a deprecated variant of annotation methods that work with objects such as `Marker`, `Polygon` and more that imitate the Google Maps API. On the other hand, `maplibre-plugins-android` repository contains an implementation that works with the objects `Symbol`, `Fill` and more. This situation is confusing to developers and users and should be cleaned up.
```

louwers (approved):
Disregard my earlier comment, I think it is better to use `maplibre-native` since you are referring to the repo, not the project.

LGTM!

PR #1285, opened 2023-06-29
Bazel Updates, Code generation

mwilsnd:
Bazel now runs style & shader codegen tools during build. Early work on Windows platform support (mbgl-core is building on Windows).


BUILD.bazel
```diff
@@ -1,56 +1,118 @@
...
+    srcs = ["shaders/manifest.json"] + glob(["shaders/*.glsl"], allow_empty = False)
 )
 
-config_setting(
-    name = "split_renderer",
-    flag_values = {
-        ":renderer": "split"
-    }
+genrule(
+    name = "generate_shaders",
```
louwers, line 48:
```suggestion
    name = "generated_shaders",
```

bazel/core.bzl
```diff
@@ -1,10 +1,97 @@
-# MLN_RENDERER_DEFINES = select({
-#     ":legacy_renderer": ["MLN_LEGACY_RENDERER=1"],
-#     "//conditions:default": ["MLN_LEGACY_RENDERER=0"]
-# }) + select({
-#     ":drawable_renderer": ["MLN_DRAWABLE_RENDERER=1"],
-#     "//conditions:default": ["MLN_DRAWABLE_RENDERER=0"]
-# })
+MLN_PUBLIC_GENERATED_STYLE_HEADERS = [
```
louwers, line 1: I am still a bit worried about how we can best keep this file up to date.

The generation scripts output a list of files, maybe we can write those out to a `.bzl` file?

shaders/generate_shader_code.js
```diff
@@ -223,13 +220,20 @@ const args = (() => {
 
 
 // Generate shader source headers
+const shaderRoot = "shaders/";
 let generatedHeaders = [];
 let shaderNames = [];
-JSON.parse(fs.readFileSync(path.join(args.input, "manifest.json")))
+
+const targetRoot = path.join(args.root, "__generated__");
```
louwers, line 227: Does these scripts still work for the CMake based in source files?

bazel/core.bzl
mwilsnd: It does, I'm looking at how to control where that gets placed as it doesn't look to be explicitly created and is rather an artifact of the `ejs` module being used.

For context, the [CMake](https://github.com/maplibre/maplibre-native/blob/9374765cb873306606cbd2679958e8d1e2fe9a0f/CMakeLists.txt#L133) had everything mixed together in opaque bricks.

shaders/generate_shader_code.js
mwilsnd: It does, the invocation just changes. Before the script dumped everything into a flat directory and now it emits structured folders with everything where it should be. It will pack that in a `__generated__` folder  wherever you specify the root is. Still a manual copy from inside the `__generated__` folder required for CMake as long as we're checking generated code into the repository, but our goal should probably be to move away from that model.

bazel/core.bzl
louwers: > For context, the [CMake](https://github.com/maplibre/maplibre-native/blob/9374765cb873306606cbd2679958e8d1e2fe9a0f/CMakeLists.txt#L133) had everything mixed together in opaque bricks.

:face_in_clouds:

alexcristici (approved):

TimSylvester (approved):

sjg-wdw (approved):

louwers (approved):

Issue #1286, opened 2023-06-29
Layer Rendering: order / depth buffer


stefankarschti:
![Image](https://user-images.githubusercontent.com/4198736/249857737-8426e5a5-139b-41ba-aff8-48340e561d81.png)

![Image](https://user-images.githubusercontent.com/4198736/249857831-6af35c51-0545-4ab4-a7c9-e8f617965b59.png)

![Image](https://user-images.githubusercontent.com/4198736/249857876-f5f2b7cc-67ae-4dc8-bd1d-af7e17536d0d.png)

PR #1288, opened 2023-06-29
Topic/drawable extrusion

TimSylvester:
Mostly working.  The legacy path drops out sometimes, and it seems to follow tile boundaries so is probably stencil-related, but that might just be a problem with the split view.

Likely related to #1286.


sjg-wdw (approved):

stefankarschti (approved):

mwilsnd:
If I rotate the map, I can see the lighting/shading doesn't update correctly, ex:
image

TimSylvester:
light parameters are updated every frame now

src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
```diff
@@ -241,4 +276,227 @@ bool RenderFillExtrusionLayer::queryIntersectsFeature(const GeometryCoordinates&
...
+
+        if (doDepthPass && !depthBuilder) {
+            if (auto builder = context.createDrawableBuilder(layerPrefix + "depth")) {
+                builder->setShader(shader);
+                builder->setIs3D(true);
+                builder->setEnableStencil(false);
+                builder->setEnableColor(false);
+                builder->setRenderPass(drawPass);
+                builder->setCullFaceMode(gfx::CullFaceMode::backCCW());
+                depthBuilder = std::move(builder);
```
mwilsnd, line 420: I see the reason why we have two builders but the extra copies we're doing here just to have different state isn't great. Can we have a single builder and get shared vertex buffers working? ie Two or more drawables sharing the same VAO but configured with different state.

src/mbgl/gfx/fill_extrusion_drawable_data.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
mwilsnd, line 1: Is this drawable data being used or do you have plans for it to be used?

src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
sjg-wdw: If we are going to do that, can we do it at a separate task?  Be happy to schedule it very soon.

mwilsnd: I'll make an issue for it.

mwilsnd: Tracking this in #1300, OK to leave this for now.

src/mbgl/gfx/fill_extrusion_drawable_data.hpp
TimSylvester: No, it turned out to not be needed, and I forgot about it.

mwilsnd (approved):

src/mbgl/gfx/drawable_builder.cpp
```diff
@@ -34,12 +34,15 @@ void DrawableBuilder::flush() {
...
         draw->setLineWidth(static_cast<int32_t>(lineWidth));
-        draw->setNeedsStencil(needsStencil);
+        draw->setEnableColor(enableColor);
+        draw->setEnableStencil(enableStencil);
         draw->setRenderPass(renderPass);
         draw->setDrawPriority(drawPriority);
         draw->setSubLayerIndex(subLayerIndex);
         draw->setDepthType(depthType);
+        draw->setIs3D(is3D);
         draw->setCullFaceMode(impl->cullFaceMode);
```
alexcristici, line 46: If you merge this after the heatmap layer, you will have to add `setColorMode`.

alexcristici (approved):

PR #1289, opened 2023-06-30
Harfbuzz demo

alanchenboy:
This is a demo of how to use harfbuzz to render complex text.   

Example style:
https://alanchenboy.github.io/harfbuzzResource/hindi.json.  

Screenshots:
simulator_screenshot_AF3734BF-C4E4-40A2-9480-BA9945F19148.  

simulator_screenshot_9C42D2B4-E900-4A7B-9543-1FE47456BC3D

https://github.com/maplibre/maplibre-native/assets/5135389/f8f294ec-1d20-4106-b745-51d5f4a1af1d.  

https://github.com/maplibre/maplibre-native/assets/5135389/7e3857e9-74c4-405e-b12c-e5722cf76919.


birkskyum:
@alanchenboy , the example link doesn't appear to be working.

birkskyum:
Can we potentially replace some in-house logic for text-shaping with Harfbuzz to lower the maintenance burden?

wipfli:
Very interesting!

louwers:
Neat, but it looks like we need some build changes.

FYI @alanchenboy  we are transitioning away from the CMake based setup on iOS.

The `topic/drawable` branch will be merged into main over the coming weeks. https://github.com/maplibre/maplibre-native/blob/topic/drawable/platform/ios/BAZEL.md

wipfli:
> Can we potentially replace some in-house logic for text-shaping with Harfbuzz to lower the maintenance burden?

What part of the codebase are you referring to? MapLibre does not do shaping as far as I know. There is just some line breaking code. But Harfbuzz does explicitly not support line breaking.

birkskyum:
@wipfli , I was just trying to learn more - didn't understand if this was an alternative to something we already did or an addition.

wipfli:
By shaping, people commonly mean this: https://harfbuzz.github.io/what-is-harfbuzz.html#what-is-text-shaping

alanchenboy:
> @alanchenboy , the example style link doesn't appear to be working.

link Corrected,  you must use this branch to verify render effect.

alanchenboy:
> Neat, but it looks like we need some build changes.
> 
> FYI @alanchenboy we are transitioning away from the CMake based setup on iOS.
> 
> The `topic/drawable` branch will be merged into main over the coming weeks. https://github.com/maplibre/maplibre-native/blob/topic/drawable/platform/ios/BAZEL.md

thanks, will look into these changes

src/mbgl/text/harfbuzz.cpp
```diff
@@ -0,0 +1,121 @@
...
+    std::string fontFileName;
+    std::string language;
+    hb_script_t script;
+    hb_direction_t direction;
+};
+
+std::map<GlyphIDType, InternalHBLangInfo> HBShaper::internalInfos = {
+    {GlyphIDType::Khmer, {"khmer.ttf", "kh", HB_SCRIPT_KHMER, HB_DIRECTION_LTR}},
+    {GlyphIDType::Myanmar, {"Myanmar.ttf", "my", HB_SCRIPT_MYANMAR, HB_DIRECTION_LTR}},
+    {GlyphIDType::Devanagari, {"Devanagari.ttf", "hi", HB_SCRIPT_DEVANAGARI, HB_DIRECTION_LTR}},
```
wipfli, line 17: Language: Would it make sense to introduce a `text-language` style spec property? Because for example Devanagari can be used for Hindi and Marathi and the shaping results might be different.

src/mbgl/util/i18n.cpp
```diff
@@ -627,4 +628,17 @@ bool isWhitespace(char16_t chr) {
...
 } // namespace util
+
+GlyphIDType charGlyphIDType(char16_t ch, GlyphIDType lastChType) {
+    if (u'\uFE00' == ch) {
+        return lastChType;
+    }
+
+    if (isInKhmer(ch)) return GlyphIDType::Khmer;
+    if (isInMyanmar(ch) || isInMyanmarExtendedA(ch) || isInMyanmarExtendedB(ch)) return GlyphIDType::Myanmar;
+    if (isInDevanagari(ch) || isInDevanagariExtended(ch)) return GlyphIDType::Devanagari;
```
wipfli, line 639: Here the mapping from unicode range to font file is hard coded. For every new language we would need to add an entry.

Would it make more sense to specify in the font definition what unicode range it covers:

So instead of 

```json
"fonts": "https://alanchenboy.github.io/harfbuzzResource/fonts/{fontstack}/{language}.ttf"
```

we could do this:

```json
"fonts": [
  {
    "name": "devanagari",
    "url": "https://example.com/path-to-font-file.ttf"
    "unicode-range": "U+4??"
  }
]
```

the unicode-range would use the same definition as the CSS specification https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range

wipfli (commented):
This is an amazing achievement, thank you so much for sharing this work, @alanchenboy!

I added some comments.

I think harfbuzz in native is the right approach and for web we could use harfbuzz in webassembly, similar to how the RTL plugin does things.

src/mbgl/text/harfbuzz.cpp
alanchenboy: Good question here. Does you mean when the script is different even the font name is same the render result is different?

src/mbgl/util/i18n.cpp
alanchenboy: This means we leave the code range setting to the style, hmm, does this necessary? cause only a few languages need shaping the entries won't be that many.

alanchenboy:
Hi everyone, this pr is closed because the main branch changed huge, so I have to refactor all the change, but it auto close when I try to force push

PR #1290, opened 2023-06-30
Layer Rendering: order / depth buffer

stefankarschti:
Issue #1286


stefankarschti:
Looks like [this item](https://github.com/orgs/maplibre/projects/8/views/3?pane=issue&itemId=31518986) is also fixed by this PR.

TimSylvester:
So the tile masking and `opaquePassCutoff` check don't need to be done at all?

stefankarschti:
> So the tile masking and `opaquePassCutoff` check don't need to be done at all?

I'm not entirely sure, but apparently the visuals are correct if we're not filtering out drawables. Most likely it has to do with the opaque pass, where depth is written.

stefankarschti:
> > So the tile masking and `opaquePassCutoff` check don't need to be done at all?
> 
> I'm not entirely sure, but apparently the visuals are correct if we're not filtering out drawables. Most likely it has to do with the opaque pass, where depth is written.

tile masking remains for `TileLayerGroupGL`. the `LayerGroupGL` doesn't need it since it's not tile oriented, its drawables _should_ not have tileID set

TimSylvester:
Oh, I see, yeah, I removed the tile masking stuff from `LayerGroupGL` in my PR as well.

stefankarschti:
it would be also interesting to check whether the legacy path draws multiple times. I think it does, and where it doesn't it returns early from the `RenderLayer.render` method. Like it does for line:
```
void RenderLineLayer::render(PaintParameters& parameters) {
  assert(renderTiles);
  if (parameters.pass == RenderPass::Opaque) {
    return;
  }
```
for `RenderFillLayer` it does not early return. and there's some opaquePassCutoff logic in the `Renderer::Impl::render` as well

mwilsnd (approved):

alexcristici (approved):

TimSylvester (approved):

PR #1299, opened 2023-06-30
Topic/tweaker refactor

TimSylvester:
Replace lambdas with drawable tweakers


sjg-wdw (approved):
Much better.  Thanks!

src/mbgl/gfx/drawable_tweaker.cpp
```diff
@@ -0,0 +1 @@
+
```
mwilsnd, line 1: What's this for?

mwilsnd (changes requested):

src/mbgl/gfx/drawable_tweaker.cpp
TimSylvester: Probably not needed, I can remove it.

mwilsnd (approved):

stefankarschti (approved):

alexcristici (approved):

PR #1308, opened 2023-07-04
Make reusable action for downloading workflow run artifact

louwers:
I am going to make more workflows where a PR workflow triggers a priviledged workflow. E.g. I want to upload the render test app for iOS and Android to AWS Device Farm from a PR, not just in main.

This makes the rather verbose JavaScript to download and unzip the artifact into a reusable action.


HarelM (approved):
Seems a bit complicated to me, but if that's what you need.
Also why upload to AWS and not github artifacts?

Issue #1272, opened 2023-06-23
CustomGL Layer Path


sjg-wdw:
We need to accommodate the custom gl layers that developers have written on top of the toolkit.
There are two versions of these, one that's more generic and uses our Drawables/Builders and ones that's just raw GL calls.
This is the second type.

PR #1312, opened 2023-07-04
Custom layer

stefankarschti:
#1272 
First version. Entanglement to GL is quite high, still. Marked TODOs to discuss options.


src/mbgl/gfx/drawable_custom_layer_host_tweaker.cpp
```diff
@@ -0,0 +1,48 @@
...
+    state.getProjMatrix(projMatrix);
+    parameters.projectionMatrix = projMatrix;
+
+    // TODO: check render errors independently of SDK
+    host->render(parameters);
+
+    // Reset the view back to our original one, just in case the CustomLayer
+    // changed the viewport or Framebuffer.
+    // TODO: do this via the context
+    paintParameters.backend.getDefaultRenderable().getResource<gl::RenderableResource>().bind();
```
stefankarschti, line 42: this is still entangled to GL unless we push it to Context.

louwers:
We have an example (legacy) custom layer here: https://github.com/maplibre/maplibre-native/blob/topic/drawable/platform/android/src/example_custom_layer.cpp

Would be good to have something similar for the drawable-based custom layer when the design is finalized.

src/mbgl/gfx/context.hpp
```diff
@@ -137,6 +137,12 @@ class Context {
 
     /// Create a render target
     virtual RenderTargetPtr createRenderTarget(const Size size, const TextureChannelDataType type) = 0;
+
+    /// Resets the context state to defaults
+    virtual void resetState(gfx::DepthMode depthMode, gfx::ColorMode colorMode) = 0;
+
+    /// Sets dirty state
+    virtual void setDirtyState() = 0;
```
stefankarschti, line 145: this was GL specific.

stefankarschti:
> We have an example (legacy) custom layer here: https://github.com/maplibre/maplibre-native/blob/topic/drawable/platform/android/src/example_custom_layer.cpp
> 
> Would be good to have something similar for the drawable-based custom layer when the design is finalized.

At this stage there should be no disruption, so the existing `CustomLayerHost` interface will continue to work.
As this is now internally untangled from GL, the same interface could be used with different backends, once they are implemented.

sjg-wdw:
We'd need to add more data to what's currently passed in to make this a generic custom layer.
I also think we'll want more methods to call in different places.  But let's revisit next phase.

sjg-wdw (approved):
This looks like a good way to do it for now.
For the next version I can see we'll want a bunch of callbacks in various place.
preRender/postRender,update, that sort of thing.

mwilsnd (changes requested):
Looks like there is still more work to get the custom layer wired up, as is  `./build/mbgl-test-runner --gtest_filter=CustomLayer.Basic` fails because `RenderCustomLayer::update` never gets called. The general approach here looks pretty good though.

stefankarschti:
> Looks like there is still more work to get the custom layer wired up, as is `./build/mbgl-test-runner --gtest_filter=CustomLayer.Basic` fails because `RenderCustomLayer::update` never gets called. The general approach here looks pretty good though.

[This condition](https://github.com/maplibre/maplibre-native/blob/c02099e23f325101d1724ec0c89be3bd38de502e/src/mbgl/renderer/renderer.cpp#L40C13-L40C58) is never true during the test run. No layers

mwilsnd:
> > Looks like there is still more work to get the custom layer wired up, as is `./build/mbgl-test-runner --gtest_filter=CustomLayer.Basic` fails because `RenderCustomLayer::update` never gets called. The general approach here looks pretty good though.
> 
> [This condition](https://github.com/maplibre/maplibre-native/blob/c02099e23f325101d1724ec0c89be3bd38de502e/src/mbgl/renderer/renderer.cpp#L40C13-L40C58) is never true during the test run. No layers will be updated with a `HeadlessFrontend` Ideas?

Ah, that would do it! `impl->render` will check for and create `staticData` if it doesn't exist, but we want update to run, so we need to make sure staticData is created before that point.

Why not add a `preRender` method to `Renderer::Impl`, then we can move the staticData creation check from `render` to `preRender` and call `preRender` before that check.

stefankarschti:
> > > Looks like there is still more work to get the custom layer wired up, as is `./build/mbgl-test-runner --gtest_filter=CustomLayer.Basic` fails because `RenderCustomLayer::update` never gets called. The general approach here looks pretty good though.
> > 
> > 
> > [This condition](https://github.com/maplibre/maplibre-native/blob/c02099e23f325101d1724ec0c89be3bd38de502e/src/mbgl/renderer/renderer.cpp#L40C13-L40C58) is never true during the test run. No layers will be updated with a `HeadlessFrontend` Ideas?
> 
> Ah, that would do it! `impl->render` will check for and create `staticData` if it doesn't exist, but we want update to run, so we need to make sure staticData is created before that point.
> 
> Why not add a `preRender` method to `Renderer::Impl`, then we can move the staticData creation check from `render` to `preRender` and call `preRender` before that check.

Shaders get created after (`backend.getDefaultRenderable().wait();`)[https://github.com/maplibre/maplibre-native/blob/c02099e23f325101d1724ec0c89be3bd38de502e/src/mbgl/renderer/renderer_impl.cpp#L75] in `Renderer::Impl::render`, so that needs to tag along to `preRender` I suppose.

mwilsnd (approved):

alexcristici (approved):

PR #1314, opened 2023-07-05
Add WebP decoding support to Linux and Windows

mwilsnd:
I noticed MapTiler is returning DEM data in WebP format when using a terrain source. The iOS platform seems to already support this but the default platform (used by Linux and Windows) do not. I've gone ahead and added it in this PR.


louwers (approved):
Neat! 💪

PR #1321, opened 2023-07-07
Render drawables in order

TimSylvester:
Layer groups store drawables in priority order, and by tile ID in the case of the tile layer group, and ensure that the collections stay synchronized.  Rendering occurs in priority order, currently the order that the drawables are produced because layers are not setting priority values, which replicates the order of the legacy rendering.

Somewhat complicated by the semantics of `std::set`, and I changed the way observe-and-remove works so that the drawable hasn't already been destroyed when the callback returns, which further complicates things.

before/after

### DIFF
--- a/include/mbgl/gfx/drawable.hpp
+++ b/include/mbgl/gfx/drawable.hpp
@@ -208,12 +208,8 @@ struct DrawableLessByPriority {
         }
         return a.getId() < b.getId();
     }
-    bool operator()(const Drawable* left, const Drawable* right) const {
-        return operator()(*left, *right);
-    }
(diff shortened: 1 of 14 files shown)


mwilsnd (approved):
I like the sorted container being used alongside the per-tile map. As long as they don't diverge the non-owning pointer sounds like an effective approach.

TimSylvester:
Yeah, this is error-prone, but as long as it's well-encapsulated it should be okay.  I've used `boost::multi_index` to accomplish the same thing more cleanly, but we're going the other direction so... oh well.  Though it is header-only, IIRC.

PR #1324, opened 2023-07-07
Update vcpkg to ac68eb6

mwilsnd:
Fixes missing msys2 packages

### DIFF
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -34,18 +34,17 @@ target_link_libraries(
 
 if(WIN32)
     find_package(libuv REQUIRED)
-    set(LIBUV_LIB $<IF:$<TARGET_EXISTS::uv_a>,uv_a,uv>)
 
     target_link_libraries(
-        mbgl-cache PRIVATE ${LIBUV_LIB}
+        mbgl-cache PRIVATE $<IF:$<TARGET_EXISTS:libuv::uv_a>,libuv::uv_a,libuv::uv>
     )
 
     target_link_libraries(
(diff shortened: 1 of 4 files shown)


mwilsnd:
Can someone confirm if libuv links correctly for them? `uv_a` no longer links on my machines after updating vcpkg and instead resolves to just `uv`. I updated the CMake following usage output from vcpkg but CI isn't resolving it correctly and I'm not sure why.

louwers:
cc @tdcosta100

tdcosta100:
Sorry for taking so long to answer. I've been busy these days. So @mwilsnd, you are trying to update `vcpkg` to 2023.06.20, but the current reference to `vcpkg` points to a commit in 2023.06.21, that is, you are proposing a downgrade. Maybe when you created this branch, `vcpkg` wasn't updated yet? Anyway, there is a important fix in the 2023.06.21 version, so maybe you could try a newer commit, not necessarily a release tag.

mwilsnd:
I didn't notice the release was detached from a tag - that would explain some things. I'll try updating it to a newer commit that makes some changes to libuv's target name.

tdcosta100:
Okay, but it's strange, I'm not having any trouble with libuv at all. Did you tried to delete the vcpkg database located at ? 
`%LOCALAPPDATA%\vcpkg`?

mwilsnd:
The issue was on CI, which I see now got unintentionally reverted on the drawables development branch - explaining even more things. I'm going to re-target this PR on that branch and keep main as-is for the moment.

TimSylvester (approved):
Still builds fine under Xcode, for whatever that's worth.

PR #1345, opened 2023-07-11
Fix 3D stencil mode incrementing for each drawable

TimSylvester:
Stefan's Pattern Line test style shows a problem with fill extrusions at tile boundaries:

The problem is that `Context::stencilModeFor3D` is not idempotent like `stencilModeForClipping`, and needs to be called only once for all the tiles in the layer.

The translucent 3D extrusions use a not-equal stencil test to prevent the layer from drawing to the same pixel more than once.  So unlike the tile clipping where each tile's drawing is constrained to that tile's bound, the edges of each tile's extrusions may actually be drawn by the overflow from the previous tile.

This is separate from depth testing, which is also important.  When extrusions are translucent, they are first drawn to the depth buffer to accumulate the largest value and then drawn to the color buffer with only that depth value so that only the extruded surfaces nearest to the eye position are drawn, and the back sides don't show through.

### DIFF
--- a/src/mbgl/gl/drawable_gl.cpp
+++ b/src/mbgl/gl/drawable_gl.cpp
@@ -46,7 +46,10 @@ void DrawableGL::draw(PaintParameters& parameters) const {
     // force disable depth test for debugging
     // context.setDepthMode({gfx::DepthFunctionType::Always, gfx::DepthMaskType::ReadOnly, {0,1}});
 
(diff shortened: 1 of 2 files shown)


sjg-wdw (approved):
Huh, sort of makes sense.

mwilsnd (approved):

PR #1346, opened 2023-07-11
Share identical uniform buffers across drawables for background, line.

TimSylvester:
Return item count from tile-matching `observeDrawables` to eliminate a second key search. Eliminate extra blank entries in layer IDs.


sjg-wdw (approved):
Makes sense.  Curious to see how this affects things.

mwilsnd (approved):
Looks to save a bit more, moving in the right direction. We're using around 30% more VRAM over legacy and about 20% more RAM, though the RAM is much more variable.

PR #1348, opened 2023-07-12
[iOS] add-camera-padding

JulianBissekkou:
#### Why is this change needed?
We are maintaining the Flutter plugin for maplibre and we are currently implementing https://github.com/maplibre/flutter-maplibre-gl/issues/258 which brings the padding feature to the sdk.
On android we are already able to create camera updates with a padding value, but this is missing on iOS.

This PR exposes the `padding` value of the `MLNMapCamera` so that we are able to implement this feature.

This introduces the changes from an unmerged PR that we have seen in the old mapbox repository.
(mapbox-gl-native-ios/pull/323/files)


platform/ios/platform/ios/src/MLNMapView.mm
```diff
@@ -4376,6 +4387,9 @@ - (MLNMapCamera *)cameraForCameraOptions:(const mbgl::CameraOptions &)cameraOpti
     {
         options.pitch = camera.pitch;
     }
+    if (!MGLEdgeInsetsEqual(camera.padding, MGLEdgeInsetsZero)) {
+        options.padding = MLNEdgeInsetsFromNSEdgeInsets(camera.padding);
+    }
```
JulianBissekkou, line 4392: Moving the discussion of the old PR over to this PR:

>Not sure if this is correct, as there is edgePadding as argument as well as in padding in MGLMapCamera. The padding in MGLMapCamera overwrites the param here.

louwers:
> This introduces the changes from an unmerged PR that we have seen in the old mapbox repository.
(mapbox-gl-native-ios/pull/323/files)

Which to this date (and at the time the PR was merged) is permissively licensed.

platform/ios/platform/ios/app/MBXViewController.m
```diff
@@ -1672,7 +1677,24 @@ - (void)selectAnOffscreenPointAnnotation {
...
+    // Animate, so that point becomes the the center
+    CLLocationCoordinate2D center = [self.mapView convertPoint:point toCoordinateFromView:self.mapView];
+    MLNMapCamera *camera = self.mapView.camera.copy;
+    camera.centerCoordinate = center;
+    camera.padding = UIEdgeInsetsMake(200, 50, 10, 4);
+    
+    [self.mapView setCamera:camera animated:YES];
+}
+
+- (void)centerSelectedAnnotationWithoutPadding {
```
JulianBissekkou, line 1697: I know that this adjustment on the example app is not perfect.
It's been a while since I wrote some objective-c so that's why I quickly adjusted an existing entry in the feature list instead of creating a new one.

If someone has a better idea and can maybe help me out then let me know 👍🏽

louwers (commented):
Could you add an entry to the iOS changelog?

louwers:
Sorry for the confusion, this is the iOS changelog: https://github.com/maplibre/maplibre-native/blob/main/platform/ios/platform/ios/CHANGELOG.md

platform/ios/platform/ios/src/MLNMapView.mm
JulianBissekkou: Let me do some more testing before we merge this.

JulianBissekkou: I did my testing. However, since this is my first PR in maplibre native I would like to have feedback from someone that knows this part of the code well enough :)

louwers:
These tests fail.

image

E.g.

> testHandleTwoFingerTapGesture: ((padding) not equal to (cameraPadding)) failed: ("{length = 32, bytes = 0x00000000 0000f03f 00000000 0000f03f ... 00000000 0000f03f }") is equal to ("{length = 40, bytes = 0x00000000 0000f03f 00000000 0000f03f ... 01000000 00000000 }")

MLNMapProjectionTests also fails:

> testHandleDoubleTapGestureContentInset: ((padding) not equal to (cameraPadding)) failed: ("{length = 32, bytes = 0x00000000 0000f03f 00000000 0000f03f ... 00000000 0000f03f }") is equal to ("{length = 40, bytes = 0x00000000 0000f03f 00000000 0000f03f ... 01000000 00000000 }")

platform/ios/platform/ios/src/MLNMapView.mm
1ec5: > Not sure if this is correct, as there is edgePadding as argument as well as in padding in MGLMapCamera. The padding in MGLMapCamera overwrites the param here.

Unfortunately, the library currently conflates two different concepts as “padding”:

* `MLNMapView.contentInset` is supposed to be persistent, something the developer sets according to UI that takes up space within the map view, such as a translucent sidebar or a camera “notch” on iOS. The developer should only have to think about this inset when sizing the map view or embedding a subview, not when manipulating or animating the camera. The resulting virtual rectangle is often called a “safe area” on Apple platforms.
* Various camera-manipulating methods like `-[MLNMapView setCamera:withDuration:animationTimingFunction:edgePadding:completionHandler:]` take an `edgePadding` parameter for more precisely targeting the camera transition. In particular, the navigation SDK (and navigation applications in general) need to anchor the user location indicator somewhere other than the center of the safe area, typically in the lower third of the map view.

In the original design in mapbox/mapbox-gl-native#3583, the `mbgl::CameraOptions::padding` was the sum of the (persistent) content insets and the (temporary) edge padding. `mbgl::CameraOptions::padding` affected the frame of reference for `mbgl::CameraOptions::center`, but `mbgl::Map::transform` never persisted the padding because it had no context about how much of that padding was persistent and how much was temporary.

Unfortunately, a change was made to mbgl that conflated these two concepts, creating quite a headache for iOS/macOS developers: mapbox/mapbox-gl-native#15232 mapbox/mapbox-gl-native#15233. I have recommended that `mbgl::Transform` decouple the concepts of “padding” and “insets” and even replace “padding” with a more explicit way to anchor the center coordinate within screen space: https://github.com/mapbox/mapbox-gl-native/issues/15233#issuecomment-621507580.

JulianBissekkou:
Closing this because its not required anymore.

PR #1355, opened 2023-07-12
Drawable PR1 - Foundation

mwilsnd:
Changes to the MapLibre architecture needed by the Modularization effort are being done in several stages to aid in digestion by the community.  This is the first set of changes, associated with the new rendering architecture.

This PR consists of new files that hold the new architectural concepts needed in the modularization.  This PR is not meant to compile and it won’t until all the following PRs are stacked on top.  In this way we can walk the community through the new ideas and then the rendering changes and finally the individual layer changes.

A more detailed discussion of the architecture as implemented can be found in [this document](https://docs.google.com/document/d/1QtsUgDyD-Rt1McvJFz7HbHvEKli3tRVfxdW7fLDF-g8).  It discusses the merge plan as well as the lower level architectural components necessary to understand this first, and subsequent, merges.

Comments and feedback are expected and appreciated.  If we’ve reached out to you directly and indicated we will wait for your review, please let us know if you need anything from us.  If we haven’t reached out, your comments and feedback are just as valuable.

Let the merge begin!

### DIFF
--- a/include/mbgl/gfx/drawable.hpp
+++ b/include/mbgl/gfx/drawable.hpp
@@ -1,0 +1,230 @@
+#pragma once
+
+#include <mbgl/gfx/uniform_buffer.hpp>
+#include <mbgl/tile/tile_id.hpp>
+#include <mbgl/util/color.hpp>
+#include <mbgl/util/identity.hpp>
+#include <mbgl/util/traits.hpp>
+#include <mbgl/gfx/texture2d.hpp>
+#include <mbgl/gfx/drawable_data.hpp>
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <vector>
+
+namespace mbgl {
+
+class Color;
+class PaintParameters;
+enum class RenderPass : uint8_t;
+
+namespace gfx {
+
+class ColorMode;
+class CullFaceMode;
+enum class DepthMaskType : bool;
+class DrawableTweaker;
+class DrawMode;
+class IndexVectorBase;
+class ShaderProgramBase;
+class VertexAttributeArray;
+
+using DrawPriority = int64_t;
+using DrawableTweakerPtr = std::shared_ptr<DrawableTweaker>;
+using IndexVectorBasePtr = std::shared_ptr<IndexVectorBase>;
+using ShaderProgramBasePtr = std::shared_ptr<ShaderProgramBase>;
+using Texture2DPtr = std::shared_ptr<Texture2D>;
+
+class Drawable {
+public:
+    /// @brief Map from sampler location to texture info
+    using Textures = std::unordered_map<int32_t, gfx::Texture2DPtr>;
+
+protected:
+    Drawable(std::string name);
+
+public:
+    virtual ~Drawable();
+
+    struct DrawSegment;
+    using UniqueDrawSegment = std::unique_ptr<DrawSegment>;
+
+    const util::SimpleIdentity& getID() const { return uniqueID; }
+
+    /// Draw the drawable
+    virtual void draw(PaintParameters&) const = 0;
+
+    /// Drawable name is used for debugging and troubleshooting
+    const std::string& getName() const { return name; }
+    void setName(std::string value) { name = std::move(value); }
+
+    /// Which shader to use when rendering this drawable
+    const gfx::ShaderProgramBasePtr& getShader() const { return shader; }
+    void setShader(gfx::ShaderProgramBasePtr value) { shader = std::move(value); }
+
+    /// The pass on which we'll be rendered
+    mbgl::RenderPass getRenderPass() const { return renderPass; }
+    void setRenderPass(mbgl::RenderPass value) { renderPass = value; }
+
+    /// Test whether to draw this drawable in a given render pass.
+    bool hasRenderPass(const mbgl::RenderPass value) const {
+        return (mbgl::underlying_type(renderPass) & mbgl::underlying_type(value)) != 0;
+    }
+
+    /// Test whether to draw this drawable in a given render pass.
+    /// If multiple render pass bits are set, all must be present.
+    bool hasAllRenderPasses(const mbgl::RenderPass value) const {
+        const auto underlying_value = mbgl::underlying_type(value);
+        return (mbgl::underlying_type(renderPass) & underlying_value) == underlying_value;
+    }
+
+    /// Width for lines
+    int32_t getLineWidth() const { return lineWidth; }
+    void setLineWidth(int32_t value) { lineWidth = value; }
+
+    /// @brief Remove an attached texture from this drawable at the given sampler location
+    /// @param location Texture sampler location
+    void removeTexture(int32_t location);
+
+    /// @brief Return the textures attached to this drawable
+    /// @return Texture and sampler location pairs
+    const Textures& getTextures() const { return textures; };
+
+    /// @brief Get the texture at the given sampler location.
+    const gfx::Texture2DPtr& getTexture(int32_t location) const;
+
+    /// @brief Set the collection of textures bound to this drawable
+    /// @param textures_ A Textures collection to set
+    void setTextures(const Textures& textures_) noexcept { textures = textures_; }
+    void setTextures(Textures&& textures_) noexcept { textures = std::move(textures_); }
+
+    /// @brief Attach the given texture to this drawable at the given sampler location.
+    /// @param texture Texture2D instance
+    /// @param location A sampler location in the shader being used with this drawable.
+    void setTexture(gfx::Texture2DPtr texture, int32_t location);
+
+    /// Whether the drawble should be drawn
+    bool getEnabled() const { return enabled; }
+    void setEnabled(bool value) { enabled = value; }
+
+    /// Whether to render to the color target
+    bool getEnableColor() const { return enableColor; }
+    void setEnableColor(bool value) { enableColor = value; }
+
+    /// Whether to do stenciling (based on the Tile ID or 3D)
+    bool getEnableStencil() const { return enableStencil; }
+    void setEnableStencil(bool value) { enableStencil = value; }
+
+    /// not used for anything yet
+    DrawPriority getDrawPriority() const { return drawPriority; }
+    void setDrawPriority(DrawPriority value) { drawPriority = value; }
+
+    /// Determines depth range within the layer for 2D drawables
+    int32_t getSubLayerIndex() const { return subLayerIndex; }
+    void setSubLayerIndex(int32_t value) { subLayerIndex = value; }
+
+    /// Depth writability for 2D drawables
+    DepthMaskType getDepthType() const { return depthType; }
+    void setDepthType(DepthMaskType value) { depthType = value; }
+
+    /// Uses 3D depth mode
+    bool getIs3D() const { return is3D; }
+    void setIs3D(bool value) { is3D = value; }
+
+    /// Is custom
+    bool getIsCustom() const { return isCustom; }
+    void setIsCustom(bool value) { isCustom = value; }
+
+    /// The ID of the tile that this drawable represents, if any
+    const std::optional<OverscaledTileID>& getTileID() const { return tileID; }
+    void setTileID(const OverscaledTileID& value) { tileID = value; }
+
+    const gfx::CullFaceMode& getCullFaceMode() const;
+    void setCullFaceMode(const gfx::CullFaceMode&);
+
+    const gfx::ColorMode& getColorMode() const;
+    void setColorMode(const gfx::ColorMode&);
+
+    /// Get the vertex attributes that override default values in the shader program
+    virtual const gfx::VertexAttributeArray& getVertexAttributes() const = 0;
+    virtual gfx::VertexAttributeArray& mutableVertexAttributes() = 0;
+
+    virtual void setVertexAttributes(const gfx::VertexAttributeArray&) = 0;
+    virtual void setVertexAttributes(gfx::VertexAttributeArray&&) = 0;
+
+    /// Provide raw data for vertices
+    /// Incompatible with adding primitives
+    virtual void setVertices(std::vector<uint8_t>&&, std::size_t, AttributeDataType) = 0;
+
+    /// Provide raw indexes and segments
+    void setIndexData(std::vector<std::uint16_t> indexes, std::vector<UniqueDrawSegment>);
+    virtual void setIndexData(gfx::IndexVectorBasePtr, std::vector<UniqueDrawSegment>) = 0;
+
+    /// Get the tweakers attached to this drawable
+    const std::vector<DrawableTweakerPtr>& getTweakers() const { return tweakers; }
+    void addTweaker(DrawableTweakerPtr value) { tweakers.emplace_back(std::move(value)); }
+    void setTweakers(std::vector<DrawableTweakerPtr> value) { tweakers = std::move(value); }
+    void clearTweakers() { tweakers.clear(); }
+
+    /// Get the uniform buffers attached to this drawable
+    virtual const gfx::UniformBufferArray& getUniformBuffers() const = 0;
+    virtual gfx::UniformBufferArray& mutableUniformBuffers() = 0;
+
+    const UniqueDrawableData& getData() const { return drawableData; }
+    void setData(UniqueDrawableData&& value) { drawableData = std::move(value); }
+
+protected:
+    bool enabled = true;
+    bool enableColor = true;
+    bool enableStencil = false;
+    bool is3D = false;
+    bool isCustom = false;
+    std::string name;
+    const util::SimpleIdentity uniqueID;
+    gfx::ShaderProgramBasePtr shader;
+    mbgl::RenderPass renderPass;
+    std::optional<OverscaledTileID> tileID;
+    DrawPriority drawPriority = 0;
+    int32_t lineWidth = 1;
+    int32_t subLayerIndex = 0;
+    DepthMaskType depthType; // = DepthMaskType::ReadOnly;
+    UniqueDrawableData drawableData{};
+
+    struct Impl;
+    std::unique_ptr<Impl> impl;
+
+    Textures textures;
+    std::vector<DrawableTweakerPtr> tweakers;
+};
+
+using DrawablePtr = std::shared_ptr<Drawable>;
+using UniqueDrawable = std::unique_ptr<Drawable>;
+
+/// Comparator for sorting drawable pointers primarily by draw priority
+struct DrawableLessByPriority {
+    DrawableLessByPriority(bool descending = false)
+        : desc(descending) {}
+    bool operator()(const Drawable& left, const Drawable& right) const {
+        const auto& a = desc ? right : left;
+        const auto& b = desc ? left : right;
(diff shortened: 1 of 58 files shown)


include/mbgl/renderer/render_target.hpp
```diff
@@ -0,0 +1,46 @@
...
+    virtual ~RenderTarget() = default;
+
+    void setTexture(const gfx::Texture2DPtr& texture_) { texture = std::move(texture_); };
+    const gfx::Texture2DPtr& getTexture() const { return texture; };
+
+    bool addLayerGroup(LayerGroupBasePtr, bool replace);
+    bool removeLayerGroup(const int32_t layerIndex);
+    size_t numLayerGroups() const noexcept;
+    const LayerGroupBasePtr& getLayerGroup(const int32_t layerIndex) const;
+    void observeLayerGroups(std::function<void(LayerGroupBase&)>);
```
VladyslavOdobesku-TomTom, line 34: I see lots of functions introduced with the prefix "observe" while effectively what it is doing is visiting? Was such naming agreed upon? If yes I wonder what is the idea behind such naming.

TimSylvester: No specific reasoning, just used the wrong word, I think.  I have no objection to changing it globally.

sjg-wdw: That seems reasonable.

include/mbgl/gfx/drawable.hpp
```diff
@@ -0,0 +1,230 @@
...
+
+    Textures textures;
+    std::vector<DrawableTweakerPtr> tweakers;
+};
+
+using DrawablePtr = std::shared_ptr<Drawable>;
+using UniqueDrawable = std::unique_ptr<Drawable>;
+
+/// Comparator for sorting drawable pointers primarily by draw priority
+struct DrawableLessByPriority {
```
louwers, line 205: Let's use this thread to discuss tests. If there are tests already they should be added to this PR. Without tests I fear a large part of the code base will become read only after the Metal project has concluded, as we will be too afraid to break anything.

At a minimum I think we should create the files for each new concept, so we can easily add more tests later. I think there is a lot of low-hanging fruit like this `struct`. What do you think?

sjg-wdw: What sort of tests do you mean?

sjg-wdw: Ah, class level tests.  It's worth having that discussion, but I've never been much of a fan since they got popular in Java.  It's a different style of working, so if someone wants to push for it, I'll listen.

alanchenboy: There are a lot of tests under the ./test directory. and you can run it via mbgl-test project.

VladyslavOdobesku-TomTom: @alanchenboy in TomTom we have per test case tweakable thresholds and we do use exactly same images for OpenGL and Metal API. The difference indeed usually is on the edges of the geometry where opacity of the pixels don't match sometimes.

There is a possible way of working around it. In OpenCV there are algos for edge detection and then some pixels around the edges can have increased tolerance for pixel-to-pixel comparison.

alanchenboy: This looks like a great solution, cause some shared use line primitive in OpenGL, it's maybe a little bit hard to make it looks same in Metal, ignoring these edges can make unit tests more neat.

louwers: Yes my comment was relating to extending the unit tests (in `./test`) specifically.  What @mwilsnd says is in line with my thinking.

Unit tests are very helpful to keep things functional as other contributors start poking around in the codebase. In addition, they also serve as documentation. For example, a test of VertexAttributeArray::readDataDrivenPaintProperties would be helpful to understand how it is used. Good test coverage also makes my life as maintainer much easier. I don't need to do a deep code dive of very PR to ensure correctness of the requested changes. 

I'm not saying we should add a test for every getter and setter, Mapbox also took unit tests (too?) far, but if you ask me it will help the library a lot if all non-trivial logic has a few tests.

louwers: Yes my comment was relating to extending the unit tests (in ./test) specifically. What @mwilsnd says is in line with my thinking.

Unit tests are very helpful to keep things functional as other contributors start poking around in the codebase. In addition, they also serve as documentation. For example, a test of VertexAttributeArray::readDataDrivenPaintProperties would be helpful to understand how it is used. Good test coverage also makes my life as maintainer much easier. I don't need to do a deep code dive of very PR to ensure correctness of the requested changes.

I'm not saying we should add a test for every getter and setter, Mapbox also took unit tests (too?) far, but if you ask me it will help the library a lot if all non-trivial logic has a few tests.

mwilsnd: Render tests are going to be more complicated but I can see value in at least attempting to reuse the same expected images from the OpenGL metrics. I want to preserve the probe functionality at a minimum and generate new probe metric expectations for the Metal renderer, but I can also see generating a fresh set of expected images for the new pathway as the old ones reach a state of acceptable correctness.

include/mbgl/gfx/drawable.hpp
```diff
@@ -0,0 +1,230 @@
...
+class ShaderProgramBase;
+class VertexAttributeArray;
+
+using DrawPriority = int64_t;
+using DrawableTweakerPtr = std::shared_ptr<DrawableTweaker>;
+using IndexVectorBasePtr = std::shared_ptr<IndexVectorBase>;
+using ShaderProgramBasePtr = std::shared_ptr<ShaderProgramBase>;
+using Texture2DPtr = std::shared_ptr<Texture2D>;
+
+class Drawable {
```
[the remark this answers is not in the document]
sjg-wdw: Some of these are things that can be modified on the fly.  So yes to the tweakers question.
Some is just convenience, I think.  Easier to build the Drawable up in pieces.

include/mbgl/gfx/drawable.hpp
```diff
@@ -0,0 +1,230 @@
...
+
+    /// Whether to render to the color target
+    bool getEnableColor() const { return enableColor; }
+    void setEnableColor(bool value) { enableColor = value; }
+
+    /// Whether to do stenciling (based on the Tile ID or 3D)
+    bool getEnableStencil() const { return enableStencil; }
+    void setEnableStencil(bool value) { enableStencil = value; }
+
+    /// not used for anything yet
```
louwers, line 119: Is this still true?

include/mbgl/gfx/drawable.hpp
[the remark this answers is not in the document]
sjg-wdw: So Drawable Builders are kind of the step by step object you use to build drawables.  At least that's the intent.  The way Buckets worked made that a bit more complicated, but there will be more on that in the Layer PRs.

I can see some of the set methods you would never expect to change after creation (isCustom) and some you would.

Now I don't think we need to get strict with this (ala constructors) but it would be helpful to mark which ones are expected to change and which ones aren't.  @TimSylvester  or @mwilsnd thoughts on this?

include/mbgl/gfx/drawable.hpp
sjg-wdw: For scheduling, let's bring in the render tests after we've got things looking more or less right in Metal.  The probe tests let's schedule near the end of phase 2.

include/mbgl/gfx/drawable.hpp
sjg-wdw: Seems to still be the case.

include/mbgl/gfx/drawable.hpp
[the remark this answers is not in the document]
mwilsnd: The one pattern I find myself gravitating towards in cases like this is a configuration struct given to an object's constructor. Vulkan leverages this pattern extensively, naming such objects `<x>CreateInfo`. In essence, a relevant CreateInfo struct is populated and then passed (or moved) to the constructor of an object, passing all the data which should be immutable (or at least private) for the object's lifetime. Any remaining accessor methods for things not covered by a creation struct are for truly dynamic state.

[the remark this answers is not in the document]
sjg-wdw: That is a nice approach for a rendering SDK.
I don't really want to restructure the Drawable around that idea, though.  It seems like a fair bit of work for a very low level utility object here.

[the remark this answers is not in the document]
sjg-wdw: The Circle Layer is one of the simpler implementations as far as layers go:

https://github.com/maplibre/maplibre-native/blob/topic/drawable/src/mbgl/renderer/layers/render_circle_layer.cpp
Look around 395 in the update call.  You'll see that we don't really build drawables directly, we use the Builders.

alanchenboy (approved):
lgtm

VladyslavOdobesku-TomTom (approved):

include/mbgl/gfx/types.hpp
```diff
@@ -0,0 +1,124 @@
...
+namespace mbgl {
+namespace gfx {
+
+enum class ColorBlendEquationType : uint8_t {
+    Add,
+    Subtract,
+    ReverseSubtract,
+};
+
+enum class ColorBlendFactorType : uint8_t {
```
louwers, line 16: Suggestion: add a triple slash comment to each enum, describing what it does.

It is also possible to add descriptions to enum values.

````
/// This is an enum class
enum class fooenum {
    FOO, ///< this is foo
    BAR, ///< this is bar
};
````

File: include/mbgl/gfx/vertex_attribute.hpp [hunk 6]
```diff
@@ -0,0 +1,359 @@
...
+using mat2 = std::array<double, 2 * 2>;
+using mat3 = std::array<double, 3 * 3>;
+using mat4 = std::array<double, 4 * 4>;
+
+namespace gfx {
+
+class ShaderProgramBase;
+class VertexVectorBase;
+
+class VertexAttribute {
```
louwers, line 27: A documentation comment should be added to this class.

Some (public) methods including the constructors might also benefit from documentation comments, although they are relatively straightforward.

include/mbgl/gfx/vertex_attribute.hpp
```diff
@@ -0,0 +1,359 @@
...
+    VertexAttributeArray& operator=(const VertexAttributeArray&);
+
+    virtual std::unique_ptr<VertexAttributeArray> clone() const {
+        auto newAttrs = std::make_unique<VertexAttributeArray>();
+        newAttrs->copy(*this);
+        return newAttrs;
+    }
+
+    template <class... DataDrivenPaintProperty, class Binders, class Evaluated>
+    std::vector<std::string> readDataDrivenPaintProperties(const Binders& binders, const Evaluated& evaluated) {
```
louwers, line 283: This one definitely needs a documentation comment.

Some comments to help a future programmer make sense of the the internals also would not hurt!

include/mbgl/gfx/vertex_attribute.hpp
```diff
@@ -0,0 +1,359 @@
...
+                            propertiesAsUniforms.emplace_back(attributeName);
+                        }
+                    }
+                }
+            }(DataDrivenPaintProperty::AttributeNames),
+            ...);
+        return propertiesAsUniforms;
+    }
+
+    virtual void copy(const VertexAttributeArray& other) {
```
louwers, line 326: Needs a documentation comment because that includes a description of what happens when a key is already present.

src/mbgl/gl/drawable_gl.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (shader) {
+        const auto& shaderGL = static_cast<const ShaderProgramGL&>(*shader);
+        for (const auto& element : shaderGL.getUniformBlocks().getMap()) {
+            element.second->unbindBuffer();
+        }
+    }
+}
+
+void DrawableGL::upload(gfx::UploadPass& uploadPass) {
+    if (!shader) {
```
louwers, line 140: This seems like a good place to Log.

include/mbgl/renderer/layer_group.hpp
```diff
@@ -0,0 +1,152 @@
...
+    virtual void postRender(RenderOrchestrator&, PaintParameters&) {}
+
+    /// Call the provided function for each drawable in priority order
+    virtual std::size_t observeDrawables(const std::function<void(gfx::Drawable&)>&&) = 0;
+    virtual std::size_t observeDrawables(const std::function<void(const gfx::Drawable&)>&&) const = 0;
+
+    /// Call the provided function for each drawable in undefined order, allowing for removal.
+    /// @param f A function called with each drawable, returning true to keep and false to discard it.
+    /// @return The number of items removed
+    virtual std::size_t observeDrawablesRemove(const std::function<bool(gfx::Drawable&)>&& f) = 0;
```
louwers, line 79: `removeDrawablesIf` sounds like a much better name here.

louwers (changes requested):
General comments:

- We need to be mindful what headers we put in `include` and what headers we put in `src`. The former is intended for public usage. We want to have a stable C++ API so stuff that we might want to refactor later should go in `src`.
- Documentation comments are especially important for headers in `include`, so people can use the library and get help from their IDE, use the library without looking at the implementation (although I think this is more of an ideal for C++ code) and browse the [API docs](https://maplibre.org/maplibre-native/cpp/api/). Each class should have a short description what it does and especially non-trivial methods should have a documentation comment as well. This is already the case for most classes.
- As mentioned in a separate thread, I would like to see some unit tests. They are a good source of documentation and help prevent regressions.
- I pinpointed one possible place to throw in a Log statement. Maybe there are more.

include/mbgl/renderer/layer_group.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+using DrawablePtr = std::shared_ptr<Drawable>;
+using UniqueDrawable = std::unique_ptr<Drawable>;
+} // namespace gfx
+
+class LayerTweaker;
+using LayerTweakerPtr = std::shared_ptr<LayerTweaker>;
+
+/**
+    A layer-like group of drawables, not a group of layers.
```
louwers, line 33: Should it have a different name maybe?

include/mbgl/gl/drawable_gl.hpp
```diff
@@ -0,0 +1,77 @@
...
+using UniqueVertexBufferResource = std::unique_ptr<VertexBufferResource>;
+
+} // namespace gfx
+
+namespace gl {
+
+class Texture2D;
+class VertexArray;
+
+class DrawableGL : public gfx::Drawable {
```
louwers, line 31: Personally I like the GL postfix because can sometimes be omitted and then it is unclear which one it is (base class, Metal Drawable, GL Drawable??). But on Slack people preferred the namespace approach.

Maybe both like we have now?

include/mbgl/renderer/layer_group.hpp
sjg-wdw: We settled on this after evaluating the alternatives.  It's not perfect, but it could be worse.

include/mbgl/gl/drawable_gl.hpp
sjg-wdw: They really like the namespace approach.  If you really want to do it the other way revive the discussion on Slack and see if you can talk them into it.

louwers (approved):
Approved with a caveat that I want to address testing again during the final PR when we merge `topic/drawable-reviewed` into `main`.

include/mbgl/gfx/drawable.hpp
TimSylvester: Yes.  It's part of the comparison, but the only layer where ordering has been observed to matter (circle) is relying on the fallback ordering of assigned unique IDs.  We probably should be using explicit priority instead, as one could argue that we shouldn't assume monotonic increasing IDs.

include/mbgl/mtl/drawable_mtl_builder.hpp
TimSylvester, line 1: I removed some empty files previously, I guess I missed these.

include/mbgl/gfx/vertex_attribute.hpp
VladyslavOdobesku-TomTom: I've also wondered about the comments in the code. What is the policy in this repo about comments?
I see that in some files nearly everything has doxygen comments. In some files only some minor things have comments.
Also in different files different style of doxygen comments is used.

```
class RenderTarget {
public:
    virtual ~RenderTarget() = default;

    void setTexture(const gfx::Texture2DPtr& texture_) { texture = std::move(texture_); };
    const gfx::Texture2DPtr& getTexture() const { return texture; };

    bool addLayerGroup(LayerGroupBasePtr, bool replace);
    bool removeLayerGroup(const int32_t layerIndex);
    size_t numLayerGroups() const noexcept;
    const LayerGroupBasePtr& getLayerGroup(const int32_t layerIndex) const;
    void observeLayerGroups(std::function<void(LayerGroupBase&)>);
    void observeLayerGroups(std::function<void(const LayerGroupBase&)>) const;

    virtual void upload(gfx::UploadPass& uploadPass) = 0;
    virtual void render(RenderOrchestrator&, const RenderTree&, PaintParameters&) = 0;

protected:
    gfx::Texture2DPtr texture;
    using LayerGroupMap = std::map<int32_t, LayerGroupBasePtr>;
    LayerGroupMap layerGroupsByLayerIndex;
};
```

then this

```
/// The RenderFrontend is the bridge between the Map and
/// platform used to update and observer the Renderer
///
/// It hides any threading specifics and always replies on
/// the original thread.
class RendererFrontend {
public:
    virtual ~RendererFrontend() = default;

    /// Must synchronously clean up the Renderer if set
    virtual void reset() = 0;

    /// Implementer must bind the renderer observer to the renderer in a
    /// appropriate manner so that the callbacks occur on the main thread
    virtual void setObserver(RendererObserver&) = 0;

    /// Coalescing updates is up to the implementer
    virtual void update(std::shared_ptr<UpdateParameters>) = 0;
};
```

then this

```
    // Debug
    void dumpDebugLogs();

    /**
     * @brief In Tile map mode, enables or disables collecting of the placed
     * symbols data, which can be obtained with `getPlacedSymbolsData()`.
     *
     * The placed symbols data collecting is disabled by default.
     */
    void collectPlacedSymbolData(bool enable);

    /**
     * @brief If collecting of the placed symbols data is enabled, returns the
     * reference to the `PlacedSymbolData` vector holding the collected data.
     *
     * Note: the returned vector gets re-populated at every `render()` call.
     *
     * @return collected placed symbols data
     */
    const std::vector<PlacedSymbolData>& getPlacedSymbolsData() const;
```

Are there some strict rules about doxygen comments or they are relaxed (if doxygen works without errors then it's okay). And also what are the rules (if there are any) about what should be commented/explained and what should not?

louwers: @VladyslavOdobesku-TomTom Any non-trivial methods and classes part of the public API should have some documentation comments.

I think the general pattern is that triple slash comments is used when something only has a description whereas `/**` is used when other things are included such as documentation for parameters or the return value.

PR #1356, opened 2023-07-12
Avoid running privileged workflows on forks

louwers:
These workflows run on forks, where the needed secrets are not available.

This PR prevents that.

### DIFF
--- a/.github/workflows/ios-render-test.yml
+++ b/.github/workflows/ios-render-test.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   ios-render-test:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


HarelM (approved):

.github/workflows/ios-render-test.yml
```diff
@@ -9,6 +9,7 @@ on:
 jobs:
   ios-render-test:
     runs-on: ubuntu-22.04
+    if: secrets.MAPLIBRE_NATIVE_BOT_APP_ID
```
wipfli, line 12: Do you keep track of what secrets are use somewhere?

PR #1357, opened 2023-07-12
Platform flattening overhaul

mwilsnd:
This PR merges the nested platforms inside `platform/ios`, re-wires the bazel to support the new hierarchy and deletes a bunch of the old CMake build system from the darwin, macos and ios platforms.

Stacked on topic/drawable to try and minimize the number of merge conflicts.


louwers:
Happy to review this, but do you want to set topic/drawable as target branch?

mwilsnd:
Sure, I had it based on main just to trigger CI last night. We should go ahead and enable it for topic/drawable at this point.

louwers:
Can we apply this reorg to both main and topic/drawable?

Otherwise the merge might be painful later.

mwilsnd:
A change like this is going to be painful whichever side of the branch we target. I'll see if I can duplicate this for `main`. If the conflicts between the two aren't too bad then we can certainly go that route.

PR #1358, opened 2023-07-12
Share bucket indices

TimSylvester:


sjg-wdw (approved):

mwilsnd (approved):
I think we've about reached the limit of what we can win while still retaining the raw data CPU-side in the buckets. Our VRAM consumption is good, but we're paying the cost twice in instances where we keep VertexVectors alive in the buckets.

The symbol bucket is a good example of this, we keep a copy of both static and dynamic vertices around all the time when it would probably be better to free them once uploaded to our GPU buffers.

sjg-wdw:
We've got a couple directions we could go here.
1) We could clear out the CPU side buffers once we're done with them, but I'm not clear on whether you still need them later.  Do they modify them in place.

2) We could move to SDK specific storage for the buckets.  I believe both OpenGL and Metal will let you use shared memory (at least on mobile) directly and just manipulate it.

PR #1359, opened 2023-07-13
Bug/fill attributes

TimSylvester:
When the style changes, requests are initiated but not yet complete.  The buckets are re-created only when those requests complete, or after a layout pass, if they require layout.  If we drop the existing drawables immediately, they will just be re-created based on the old data on the next frame render.

So instead, we give each bucket instance a unique ID, and track that against the tile IDs.  This resolves the problem where things don't change when switching between styles with the same layer names, like the [FB light/dark example](https://osmus.slack.com/archives/C04SSR0S78B/p1689089320597589?thread_ts=1689089261.684019&cid=C04SSR0S78B).

This can be further improved by adding a mechanism to avoid updating the UBOs on the old drawables with the new style constants during this interval, which produces an intermediate between the old and new styles that isn't really valid.

### DIFF
--- a/include/mbgl/gfx/drawable.hpp
+++ b/include/mbgl/gfx/drawable.hpp
@@ -48,7 +48,7 @@ public:
     struct DrawSegment;
     using UniqueDrawSegment = std::unique_ptr<DrawSegment>;
 
-    const util::SimpleIdentity& getId() const { return uniqueID; }
+    const util::SimpleIdentity& getID() const { return uniqueID; }
 
     /// Draw the drawable
     virtual void draw(PaintParameters&) const = 0;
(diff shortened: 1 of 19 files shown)


sjg-wdw (approved):
Seems like a fairly elegant solution.

mwilsnd:
Symbols aren't working right, here I switch from light to dark:
image

Interaction with the map to trigger extra frame renders after reaching this state does eventually result in fully switching,
image

Sometimes, it switches fully without having to render extra frames.

TimSylvester:
Sounds like it depends on the timing of loads?  I assume there's a mechanism for forcing a render after things load, I wonder if we need to actively participate in that.

stefankarschti (approved):

mwilsnd:
I think this may be related to the flickering we can see when drawables load in, especially noticeable in split view. We seem to be running at least a frame late somewhere.

mwilsnd (approved):

TimSylvester:
Looks like the key issue is that the layer tweaker runs before the drawable tweaker, so the atlas textures aren't set up yet, and the texture sizes come out zero.  Things not showing up right away seems to happen without this, and I think will be fixed by #1370.

PR #1360, opened 2023-07-13
Generate plist files during CI

mwilsnd:
Needed by the iOS app which is used as a host for the UI test.

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -139,6 +139,10 @@ jobs:
 
       - run: cp bazel/example_config.bzl bazel/config.bzl
 
+      - name: Generate app plists
+        working-directory: ./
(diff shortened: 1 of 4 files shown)


alexcristici (approved):

louwers (commented):
I'm not a big fan of this script expanding over time, I'd personally prefer to just have to deal with Bazel.

For the version we should use https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-versioning.md for the version

mwilsnd:
I had changed the `infoplists` field of `ios_application` to reference a generated `"build/Info-app.plist"` to support MapTiler API key insertion. The old Xcode project, part of the CMake build system, would copy these plist files and run `insert_access_token.sh` during an Xcode build.

TimSylvester (approved):
I was able to run the tests from Xcode without having to set up signing or anything.

stefankarschti (approved):

PR #1381, opened 2023-07-14
iOS render test on AWS Device Farm

### LINKED ISSUE #942: Run iOS render tests on CI
In this moment render tests are running in Xcode on devices.
Trying to run the render test on simulator will show this error: 
`Building for iOS Simulator, but linking in dylib built for iOS, file '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/Frameworks/XCTest.framework/XCTest' for architecture arm64`

louwers:
Closes #942

When the `ios-ci` workflow ran it triggers `ios-render-test` with a `workflow_run`.

Then this workflow fetches a token from the MapLibre Native Bot so it can start a check on the PR.

It is in a separate workflow because we need to have access to the GitHub secrets to upload to AWS Device Farm.

I hope the SHA of the PR commit is available in `${{ github.event.workflow_run.sha }}`. I am printing `${{ github.event.workflow_run }}` for debugging purposes in case it is not. It's not very well documented it seems.

### DIFF
--- a/.github/workflows/ios-render-test.yml
+++ b/.github/workflows/ios-render-test.yml
@@ -60,7 +60,7 @@ jobs:
           name: RenderTest.xctest.zip
           type: XCTEST_TEST_PACKAGE
 
-      - name: Upload Android UI test
+      - name: Upload iOS Render test
         run: |
           curl -T RenderTestApp.ipa '${{ steps.upload-ios-app.outputs.url }}'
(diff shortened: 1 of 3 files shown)


louwers:
One test is failing right now that should be fixed before merging:

image

mwilsnd (approved):

PR #1383, opened 2023-07-16
Windows: update OSMesa HeadlessBackend to OpenGL ES 3.0

tdcosta100:
Update the OSMesa HeadlessBackend version (`MLN_WITH_OSMESA=ON`) in Windows to support OpenGL ES 3.0. Also, some initialization adjustements were made, making it more reliable avoiding race conditions.

### DIFF
--- a/platform/windows/src/headless_backend_osmesa.cpp
+++ b/platform/windows/src/headless_backend_osmesa.cpp
@@ -14,8 +14,9 @@ namespace gl {
 
 class OSMesaBackendImpl final : public HeadlessBackend::Impl {
 public:
-    OSMesaBackendImpl() {
-        context = OSMesaCreateContextAttribs(std::initializer_list<int>({OSMESA_FORMAT,
+    OSMesaBackendImpl()
+        : buffer(std::make_unique<uint8_t[]>(2048 * 2048 * 4)),
+          context(OSMesaCreateContextAttribs(std::initializer_list<int>({OSMESA_FORMAT,
                                                                          OSMESA_RGBA,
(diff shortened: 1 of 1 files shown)


platform/windows/src/headless_backend_osmesa.cpp
```diff
@@ -14,19 +14,23 @@ namespace gl {
...
                                                                          OSMESA_COMPAT_PROFILE,
+                                                                         OSMESA_CONTEXT_MAJOR_VERSION,
+                                                                         3,
+                                                                         OSMESA_CONTEXT_MINOR_VERSION,
+                                                                         0,
                                                                          NULL})
                                                  .begin(),
-                                             NULL);
-    }
+                                             NULL)) {}
```
louwers, line 33:
```suggestion
                                             nullptr)) {}
```

Might as well update this as well.

tdcosta100: I'm calling a C function. Should I use `nullptr` instead `NULL` anyway?

louwers: Yes I don't think there's a good reason to use `NULL` in C++ code, even when calling C libraries.

mwilsnd: They ultimately mean the same thing, and at least in MSVC/Windows land, NULL is still used in a lot of places. I do agree though a chance to modernize while we're in here is good.

mwilsnd (approved):

tdcosta100:
Hey guys, is there anything missing to merge this PR?

PR #1384, opened 2023-07-16
Fix Linux render test CI failure showing up as pass

louwers:
- Don't show a failure as passed.
- Only create artifacts on failure & upload Linux render test results when the render test has failed.


HarelM (approved):

louwers:
Needed to use 

```
if: always() && steps.render_test.outcome != 'failure'
```

https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f#comment112735158_58859404

Issue #1259, opened 2023-06-19
Android instrumented render tests


louwers:

louwers:
Started looking into this.

https://developer.android.com/ndk/samples/sample_na

https://bazel.build/docs/android-ndk

https://github.com/maplibre/maplibre-native/blob/main/render-test/android/app/src/main/AndroidManifest.xml#L18

PR #1387, opened 2023-07-17
Android Render Test CI

louwers:
When running the render test .apk it fails to write the result.

The reason is that we target Android 11+ so writing to `/sdcard` is not allowed anymore, we have to use Scoped Storage.

```
07-17 14:35:47.072  3677  9800 E libc++abi: terminating with uncaught exception of type std::runtime_error: Failed to open file /sdcard/android-render-test-runner-metrics.html
```

But it builds again... Related #1259

---

Edit: 

- I created an action that can be used to run both iOS and Android tests on AWS Device Farm `.github/actions/aws-device-farm-run/action.yml`. Will refactor the duplicated YAML in `android-ci.yml` and `ios-ci.yml` after I confirmed it works.
- I had to ignore some additional tests because I want CI to be green.


louwers:
@artakka Do you have a moment to take a look with me?

1. When I open `render-test/android` with Android Studio, the app can be built but the test related Java files are not showing up. Maybe the Gradle files need to be updated?
2. Apps can no longer write to `/sdcard`. Do you know which directory an app *can* write to?

mwilsnd (approved):

PR #1392, opened 2023-07-18
Render Tests: fix icon texture filter type for symbol layer

stefankarschti:
Issue #1374

### DIFF
--- a/include/mbgl/gfx/drawable_atlases_tweaker.hpp
+++ b/include/mbgl/gfx/drawable_atlases_tweaker.hpp
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <mbgl/gfx/drawable_tweaker.hpp>
+#include <mbgl/style/types.hpp>
+#include <mbgl/renderer/layers/render_symbol_layer.hpp>
 
 #include <memory>
 #include <string>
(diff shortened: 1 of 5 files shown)


sjg-wdw (approved):

mwilsnd (approved):

TimSylvester:
There may be a case where the atlases are shared between symbols that end up having different filtering results.  If so, this will make it work most of the time, but leave an edge case that we might not test for.

mwilsnd:
Can different symbols from the same bucket evaluate such that both linear and nearest filtering is needed? The atlases are owned by the geometry tile render data so I don't _think_ there can be sharing outside of that scenario.

TimSylvester:
It looks like all the variables either come from the bucket or apply to everything, so as long as buckets never share it should be fine.

stefankarschti:
the assumption is that the drawables that need to use the atlases (especially icon, synce glyph is always set to linear by the legacy path) would set the sampler configuration in their drawable/layer tweakers before drawing. 
If the usage is serialized then it works, the issues arise with multi threaded rendering (which we don't have yet)...

TimSylvester (approved):

PR #1393, opened 2023-07-18
Platform flattening overhaul

mwilsnd:
This PR merges the nested platforms inside platform/ios, re-wires the bazel to support the new hierarchy and deletes a bunch of the old CMake build system from the darwin, macos and ios platforms.

This version is drafted against main, and does much the same as  #1357 drafted against topic/drawable.

Deferred to a later PR:
* ios-sanitize[-address]
* ios-static-analyzer


louwers:
@mwilsnd Great stuff!

> 2023-07-26 13:33:26.415917+0200 App[568:14667] Could not load the "settings.png" image referenced from a nib in the bundle with identifier "com.firstnamelastname.maplibre.app"

I am getting this with the sample app. Also the location icon in the top right does not show. I remember this being a problem before. Not even sure were it is getting this file from.

louwers (approved):

PR #1399, opened 2023-07-20
Windows: protect WGLDisplayConfig from being created multiple times by different threads

tdcosta100:
Usually when using a single map instance, `WGLDisplayConfig` singleton is created only once when the rendering thread tries to access it. But when multiple rendering threads (each one with its own instance of map) try to access the singleton, it may be created multiple times. So this PR protects the creation of `WGLDisplayConfig` with a mutex, so it gets created only once, like it should.

### DIFF
--- a/platform/windows/src/headless_backend_wgl.cpp
+++ b/platform/windows/src/headless_backend_wgl.cpp
@@ -196,15 +196,10 @@ public:
 
     ~WGLDisplayConfig() { UnregisterClassA(renderingWindowClass.lpszClassName, renderingWindowClass.hInstance); }
 
-    static std::shared_ptr<WGLDisplayConfig> create() {
-        static std::weak_ptr<WGLDisplayConfig> instance;
-        auto shared = instance.lock();
+    static WGLDisplayConfig& create() {
+        static WGLDisplayConfig instance = WGLDisplayConfig{Key{}};
 
-        if (!shared) {
-            instance = shared = std::make_shared<WGLDisplayConfig>(Key{});
-        }
-
-        return shared;
+        return instance;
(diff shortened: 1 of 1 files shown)


platform/windows/src/headless_backend_wgl.cpp
```diff
@@ -197,13 +197,19 @@ class WGLDisplayConfig {
     ~WGLDisplayConfig() { UnregisterClassA(renderingWindowClass.lpszClassName, renderingWindowClass.hInstance); }
 
     static std::shared_ptr<WGLDisplayConfig> create() {
```
mwilsnd, line 199: Could we not refactor this whole method like so:
```cpp
static WGLDisplayConfig& create() {
    static auto instance = WGLDisplayConfig{Key{}};
    return instance;
}
```

C++ 11 added the whole thread-safe static trick so this should be functionally the same.

tdcosta100: Much nicer solution, thanks!

tdcosta100: @mwilsnd @louwers Could someone please re-run the failed jobs? The temporary node files download problem strikes again.

mwilsnd (approved):

alanchenboy (approved):
lgtm

tdcosta100:
Hey guys, is there anything missing to merge this PR?

louwers:
Missed this one. Feel free to ping me next time if a PR is approved and still needs to be merged.

tdcosta100:
Nice, thank you! I will ping you if it happens again =).

PR #1401, opened 2023-07-21
Remove ccache ios-ci

louwers:


ntadej:
Is there a reason why we do not want cache?

louwers:
@ntadej I mostly created this PR to check if the iOS instrumented render test is running. (It does, but there is one failure.)

To answer your question: Bazel has a built in cache so we do not need ccache on iOS anymore.

PR #1402, opened 2023-07-21
Render Tests: fix hill shade failures

stefankarschti:
Issue #1373 
The render tests pass, but I had to revert to a non-shared approach to passing vertices, indices and segments.
The buffers usually contain 1 quad per tile.

### DIFF
--- a/src/mbgl/renderer/layers/render_hillshade_layer.cpp
+++ b/src/mbgl/renderer/layers/render_hillshade_layer.cpp
@@ -327,11 +327,106 @@ void RenderHillshadeLayer::update(gfx::ShaderRegistry& shaders,
     stats.drawablesRemoved += tileLayerGroup->observeDrawablesRemove(
         [&](gfx::Drawable& drawable) { return (!drawable.getTileID() || hasRenderTile(*drawable.getTileID())); });
 
-    if (!staticDataSharedVertices) {
-        staticDataSharedVertices = std::make_shared<HillshadeVertexVector>(RenderStaticData::rasterVertices());
-    }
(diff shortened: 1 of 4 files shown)


src/mbgl/renderer/layers/render_hillshade_layer.cpp
```diff
@@ -423,78 +495,20 @@ void RenderHillshadeLayer::update(gfx::ShaderRegistry& shaders,
...
-                drawable.setIndexData(indices->vector(), std::move(drawSegments));
-
-                auto imageLocation = hillshadeShader->getSamplerLocation("u_image");
-                if (imageLocation.has_value()) {
-                    drawable.setTexture(bucket.renderTarget->getTexture(), imageLocation.value());
-                }
-            });
-            continue;
+            // TODO: check whether to remove tile
+            removeTile(renderPass, tileID);
```
alexcristici, line 500: You are recreating the drawable every frame. Instead of `continue` you remove the tile. You can try this fix on the shared vertices version and probably the render tests will pass. I also curios if this is actually the fix.

stefankarschti:
After testing the code from [Alex's branch](https://github.com/alexcristici/maplibre-native/tree/hillshade-shared) it looks like the only variant that passes tests is using static (non-shared) vertex buffers.
[These defines make it easy to switch and test](https://github.com/maplibre/maplibre-native/blob/ec2064be1761a84eed596a45ad4b150c93857198/src/mbgl/renderer/layers/render_hillshade_layer.cpp#L287C1-L288C48)
@TimSylvester - any ideas where to look further?

alexcristici (approved):

Issue #1405, opened 2023-07-21
Fix iOS query test: location_indicator fails


louwers:
Follow-up on #942. The render tests are passing but there is one failing test case.

There is one failing test case.

![image](https://github.com/maplibre/maplibre-native/assets/649392/f780f1d8-907b-464c-9ffa-391665528678)

PR #1406, opened 2023-07-24
Disable location-indicator query-test on iOS

louwers:
The instrumented render test has one failure. https://github.com/maplibre/maplibre-native/issues/1405

Adding to `metrics/ignores/platform-ios.json` until we get around to fixing it.


louwers:
> Should I be approving this? Not sure I fully understand this part...

@HarelM That's alright. I don't either. 😄 

Nothing special here except disabling a test.

PR #1417, opened 2023-07-25
Metal textures

alexcristici:


src/mbgl/mtl/texture2d.cpp
```diff
@@ -75,126 +70,104 @@ size_t Texture2D::numChannels() const noexcept {
...
+    samplerDescriptor->setMagFilter(samplerState.filter == gfx::TextureFilterType::Nearest
+                                        ? MTL::SamplerMinMagFilterNearest
+                                        : MTL::SamplerMinMagFilterLinear);
+    samplerDescriptor->setSAddressMode(samplerState.wrapU == gfx::TextureWrapType::Clamp
+                                           ? MTL::SamplerAddressModeClampToEdge
+                                           : MTL::SamplerAddressModeRepeat);
+    samplerDescriptor->setTAddressMode(samplerState.wrapV == gfx::TextureWrapType::Clamp
+                                           ? MTL::SamplerAddressModeClampToEdge
+                                           : MTL::SamplerAddressModeRepeat);
+    metalSamplerState = context.createMetalSamplerState(samplerDescriptor);
```
alexcristici, line 139: @sjg-wdw For each metal texture I create a metal sampler state and then I pass it to fragment shader (bind function).

sjg-wdw: That meshes nicely with the sampler state idea added earlier.  Seems logic to me!

sjg-wdw (approved):
Seems pretty logical.
For the Raster Layer I like that most of the changes are just the new shader.  That suggests we got the architecture pretty right.

platform/ios/app/MBXViewController.m
```diff
@@ -1974,7 +1974,7 @@ - (void)setStyles
     
     /// Style that does not require an `apiKey` nor any further configuration
     [self.styleNames addObject:@"MapLibre Basic"];
-    [self.styleURLs addObject:[NSURL URLWithString:@"https://demotiles.maplibre.org/style.json"]];
+    [self.styleURLs addObject:[NSURL URLWithString:@"https://api.maptiler.com/maps/ac5e9411-61c8-4e5d-a4b2-2dc4c792b858/style.json?key=G4MQXsYbLiUxOu3SV4lh"]];
```
mwilsnd, line 1977: Need to revert this

src/mbgl/mtl/context.cpp
```diff
@@ -123,11 +123,18 @@ UniqueShaderProgram Context::createProgram(std::string name,
         std::move(name), backend, std::move(vertexFunction), std::move(fragmentFunction));
 }
 
+MTLTexturePtr Context::createMetalTexture(MTLTextureDescriptorPtr textureDescriptor) const {
+    return NS::TransferPtr(backend.getDevice()->newTexture(textureDescriptor.get()));
```
mwilsnd, line 127: We should be collecting metrics in the metal context like the GL context does, it'll be helpful for developing new Metal render tests and doing general profiling.

src/mbgl/mtl/texture2d.cpp
```diff
@@ -75,126 +70,103 @@ size_t Texture2D::numChannels() const noexcept {
...
+                    return MTL::PixelFormat::PixelFormatRGBA16Float;
+                case gfx::TexturePixelType::Alpha:
+                    return MTL::PixelFormat::PixelFormatR16Float;
+                default:
+                    assert(false);
+            }
+    }
+}
+
 void Texture2D::createObject() noexcept {
```
mwilsnd, line 96: Do we still need a split concept of 'object' and 'storage' in our Metal implementation?

src/mbgl/mtl/texture2d.cpp
```diff
@@ -75,126 +70,103 @@ size_t Texture2D::numChannels() const noexcept {
...
-                                  size.height,
-                                  0,
-                                  Enum<gfx::TexturePixelType>::to(pixelFormat),
-                                  Enum<gfx::TextureChannelDataType>::to(channelType),
-                                  data));
+    assert(metalTexture.get());
+
+    MTL::Region region = MTL::Region::Make2D(0, 0, size.width, size.height);
+    NS::UInteger bytesPerRow = size.width * getPixelStride();
+    metalTexture->replaceRegion(region, 0, data, bytesPerRow);
```
mwilsnd, line 111: Since this is pretty much identical to `uploadSubRegion`, I think we can drop `createStorage` here.

src/mbgl/mtl/upload_pass.cpp
```diff
@@ -82,35 +82,27 @@ std::unique_ptr<gfx::TextureResource> UploadPass::createTextureResource(const Si
...
-    //    updateTextureResource(*resource, size, data, format, type);
-    //    // We are using clamp to edge here since OpenGL ES doesn't allow GL_REPEAT
-    //    // on NPOT textures. We use those when the pixelRatio isn't a power of two,
-    //    // e.g. on iPhone 6 Plus.
-    //    MBGL_CHECK_ERROR(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE));
-    //    MBGL_CHECK_ERROR(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE));
-    //    MBGL_CHECK_ERROR(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST));
-    //    MBGL_CHECK_ERROR(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST));
-    return std::make_unique<TextureResource>();
+    assert(false);
```
mwilsnd, line 85: We should probably throw since assert will get dropped in release builds. Calling stubs should always terminate.

mwilsnd (changes requested):
Happy to see things moving away from the legacy texture resource.

src/mbgl/mtl/texture2d.cpp
sjg-wdw: Seems harmless enough.

src/mbgl/mtl/context.cpp
sjg-wdw: That's a good idea.  Would you open an Issue and let's figure out what metrics we want.
What are we already collecting, like frame rate?

mwilsnd: The GL context tracks the number of allocated textures and the size of requested texture memory. [Here](https://github.com/maplibre/maplibre-native/blob/b792309d6adc0ec1d3aad6f292f1e08df96061e2/include/mbgl/gfx/rendering_stats.hpp#L10-L18) is everything the renderer tracks.

src/mbgl/mtl/texture2d.cpp
alexcristici: I keep similar with GL to catch errors faster. Now that it works I will refactor it a little bit. 👍

src/mbgl/shaders/mtl/shader_program.cpp
```diff
@@ -37,7 +37,7 @@ MTLRenderPipelineStatePtr ShaderProgram::getRenderPipelineState(const gfx::Rende
     const auto& renderable = renderPassDescriptor.renderable;
     const auto& renderableResource = renderable.getResource<RenderableResource>();
 
-    auto colorFormat = MTL::PixelFormat::PixelFormatBGRA8Unorm_sRGB;
+    auto colorFormat = MTL::PixelFormat::PixelFormatBGRA8Unorm;
```
alexcristici, line 40: I set it to linear color space.

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -163,7 +163,7 @@ void swap() override {
     resource.mtlView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
     resource.mtlView.contentScaleFactor = contentScaleFactor();
     resource.mtlView.contentMode = UIViewContentModeCenter;
-    resource.mtlView.colorPixelFormat = MTLPixelFormatBGRA8Unorm_sRGB;
+    resource.mtlView.colorPixelFormat = MTLPixelFormatBGRA8Unorm;
```
alexcristici, line 166: I set it to linear color space.

src/mbgl/mtl/texture2d.cpp
```diff
@@ -75,171 +71,117 @@ size_t Texture2D::numChannels() const noexcept {
...
+                case gfx::TexturePixelType::RGBA:
+                    return MTL::PixelFormat::PixelFormatRGBA16Float;
+                case gfx::TexturePixelType::Alpha:
+                    return MTL::PixelFormat::PixelFormatR16Float;
+                default:
+                    assert(false);
+                    throw std::runtime_error("Pixel format not supported on Metal!");
+            }
+    }
+}
```
alexcristici, line 97: Same here, I used linear color space.

stefankarschti (approved):

src/mbgl/mtl/texture2d.cpp
TimSylvester: I think Apple recommends using sRGB where possible, but this could lead to testing variance even if the result is subjectively better.  Perhaps in a later version this could be an API option.

TimSylvester (approved):

mwilsnd (approved):

src/mbgl/mtl/texture2d.cpp
alexcristici: Yes, we can adjust later on, when reaching the render test phase.

PR #1424, opened 2023-07-26
Render tests: Heatmap rendering

stefankarschti:
Issue #1372 partially. The heatmap is getting rendered now.

Remaining issues:
- Combinations with background and fill. To be investigated in those layers
image

- Pixel differences versus expected results:
 
image
image
image


sjg-wdw (approved):
Interesting.  Good fix.

mwilsnd (approved):

alexcristici (approved):

Issue #1435, opened 2023-07-27
Debug render: Collision circles


stefankarschti:
Enable drawing of collision circles

Issue #1434, opened 2023-07-27
Debug render: Collision boxes


stefankarschti:
Enable drawing of collision boxes

PR #1438, opened 2023-07-27
Collision boxes and circles

stefankarschti:
Issues #1375 #1434 #1435 
Implementation of collision boxes and circles


stefankarschti:
the static implementation does work. the shared one produces scrambled outputs, it may be that the buffers are changed after creation by the placement algorithm.

sjg-wdw (approved):
Looks pretty logical.
For the static vs shared issue since this is debugging output is this just a performance issue?

TimSylvester:
I assume you checked on the raw types for the shared buffers?  I see `int2` vs `UByte2`, but that isn't necessarily a problem.

TimSylvester (approved):

stefankarschti:
> Looks pretty logical.
> For the static vs shared issue since this is debugging output is this just a performance issue?

Yes, it is a performance issue. Memory gets copied if run in "static" mode.

sjg-wdw:
For a debugging mode this doesn't worry me particularly.

stefankarschti:
> I assume you checked on the raw types for the shared buffers? I see `int2` vs `UByte2`, but that isn't necessarily a problem.

Yes, I checked `int2` as well. It didn't make a difference. The `UByte2` came from `MBGL_DEFINE_ATTRIBUTE(uint8_t, 2, placed);` from [attribute.hpp](https://github.com/maplibre/maplibre-native/blob/6733dce7650a988962cac859f16c8a9595cf052a/src/mbgl/programs/attributes.hpp#L21C1-L21C43)

platform/ios/platform/ios/scripts/bazel-generate-plists.sh
mwilsnd, line 1: This empty file should be safe to delete

mwilsnd (approved):

platform/ios/platform/ios/scripts/bazel-generate-plists.sh
stefankarschti: on `topic/drawable-fixes` this is[ still not empty](https://github.com/maplibre/maplibre-native/blob/topic/drawable-fixes/platform/ios/platform/ios/scripts/bazel-generate-plists.sh) in use and referenced from
```
./platform/ios/platform/ios/scripts/bazel-xcodeproj.sh
./platform/ios/platform/ios/scripts/bazel-package.sh
./.github/workflows/ios-ci.yml
```

mwilsnd: Right, I was just curious seeing it empty in this PR. Removing it here wouldn't have any effect on drawable-fixes when we merge it, but neither would just leaving it here as is.

stefankarschti: I think I added an executable flag to it, since it was called from another script.

PR #1439, opened 2023-07-28
Harffbuzz freetype Demo

alanchenboy:
refactor for #1289

This is a demo of how to use harfbuzz to render complex text.   

Example style:
https://alanchenboy.github.io/harfbuzzResource/hindi.json.  

Screenshots:
simulator_screenshot_AF3734BF-C4E4-40A2-9480-BA9945F19148.  

simulator_screenshot_9C42D2B4-E900-4A7B-9543-1FE47456BC3D

https://github.com/maplibre/maplibre-native/assets/5135389/f8f294ec-1d20-4106-b745-51d5f4a1af1d.  

https://github.com/maplibre/maplibre-native/assets/5135389/7e3857e9-74c4-405e-b12c-e5722cf76919.


vendor/BUILD.bazel
```diff
@@ -234,3 +235,28 @@ cc_library(
...
     ],
 )
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
+    visibility = ["//visibility:public"],
+)
+
+cmake(
```
alanchenboy, line 245: Let's discuss here, about user cc_library to replace foreign_cc.cmake.
Using foreign_cc.cmake can let us directly use the libraries' shipped compiling setting, but it will cost too much time on compiling. I didn't do the final test, but I do run raw cmake build, the cost is under 100s with raw cmake on my end but will cost 400s+ when using foreign_cc.cmake.

The con for cc_library is we must compile the library on our own, but I don't think it's an obstacle.

alanchenboy:
Swipe out macos end is such a pain for me, I have to use my personal computer to fix unit tests

vendor/harfbuzz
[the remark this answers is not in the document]
louwers: Which of the Qt workflows is failing?

cc @ntadej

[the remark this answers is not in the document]
ntadej: I'll update the CMake version to 3.19 in the next few days.

ntadej:
I wonder if we want to make this an optional feature? This would allow us to fix Qt later for example.

Also what is the binary size increase? (@louwers, didn't we have a CI test for that in the past? I guess the rest of the actions need to pass.)

alanchenboy:
https://github.com/maplibre/maplibre-native/actions/runs/5747452827/job/15578480347 does anybody known what is the reason for this failure?

alanchenboy:
I didn't have linux end, will fix linux CI in next weeks.

vendor/harfbuzz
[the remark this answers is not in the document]
alanchenboy: https://www.google.com.hk/search?q=fatal+error+C1128%3A+number+of+sections+exceeded+object+file+format+limit%3A+compile+with+%2Fbigobj&rlz=1C5GCEM_en&oq=fatal+error+C1128%3A+number+of+sections+exceeded+object+file+format+limit%3A+compile+with+%2Fbigobj&aqs=chrome..69i57.987j0j7&sourceid=chrome&ie=UTF-8

looks like harfbuzz 5.1.0 is not able to be compiled with msvc 2019. update cmake to 3.19 may resolve this issue. the latest harbuzz depends on 3.12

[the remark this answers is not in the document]
ntadej: Hmm, but Windows should use latest CMake, is that not the case?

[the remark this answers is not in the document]
alanchenboy: I didn't test if the latest harbuzz(8.1.0) can be compiled by msvc 2019, cause the runner's CMake is 3.10, I have to use the old version of harfbuzz (5.1.0), let's update CMake first, then I could update harfbuzz to 8.10 and test if the issue still there.

louwers:
@alanchenboy That means the iOS render tests are failing. Did you run them locally?

alanchenboy:
This looks quite hard to pass all unit tests. I will investigate why these changes impact some render tests like pure fill render results.

.gitmodules
```diff
@@ -49,3 +49,9 @@
 [submodule "platform/windows/vendor/vcpkg"]
 	path = platform/windows/vendor/vcpkg
 	url = https://github.com/microsoft/vcpkg.git
+[submodule "vendor/harfbuzzFreetype/freetype"]
+	path = vendor/harfbuzzFreetype/freetype
+url=https://github.com/freetype/freetype.git
+[submodule "vendor/harfbuzzFreetype/harfbuzz"]
+	path = vendor/harfbuzzFreetype/harfbuzz
+url=https://github.com/alanchenboy/harfbuzz
```
louwers, line 57: Nice that CI (almost) passes!

Let me know when you feel this is ready for review.

Can you get your changes upstreamed?

alanchenboy: Rollback to https://github.com/harfbuzz/harfbuzz 
Need to update qt runner's cmake to 3.12

alanchenboy: And needn't review it now, I think I will operate with @wipfli on how to change style things, once the style define is done, and then I believe this PR will be ready for review.

louwers:
@alanchenboy CMake has been updated.

alanchenboy:
No cmake in this job?
https://github.com/maplibre/maplibre-native/actions/runs/6464095713/job/17551288992?pr=1439

louwers (changes requested):
Hi @alanchenboy thank you for working on this.

I want to get this merged, but due to the +26% binary size increase I think it makes sense to make using Harfbuzz opt-in (or at least offer an easy opt-out). The easy way to make it opt-in would be to sprinkle `#ifdef`s throughout all the touched files, but this would not make the code very readable. Especially since more features that will be opt-in are expected in the future.

So to merge this we need to find a good pattern to make functionality opt-in. Perhaps it is possible to make a few overloads where only at instantiation the right overload is instantiated (with an `#ifdef`).

alanchenboy:
I think opt-out/in is feasible. let me  take a try

alanchenboy:
@louwers  add a flag to turn off harfbuzz freetype default, can use bazel to enable it in the iOS platform example:

`bazel run //platform/ios:xcodeproj --@rules_xcodeproj//xcodeproj:extra_common_flags="--//:renderer=legacy --//:maplibre_platform=ios --//:shaping=legacy"`

wipfli:
I would love to try out this pull request with the linux renderer. @louwers is this possible and are there some compilation guidelines?

louwers:
@wipfli Yes, follow the Linux link in the main README.

Check out this branch and don't forget to 

```
git submodule update --init
```

To pull in Harfbuzz as a submodule.

alanchenboy:
@wipfli Linux to turn on harfbuzz can add cmake parameter: -DMLN_TEXT_SHAPING_HARFBUZZ=ON

thanks for test and review.

wipfli:
Looking at your code and the example style sheet, the only change to the style specification is the new `font` property, right?

```json
"fonts": "https://alanchenboy.github.io/harfbuzzResource/fonts/{fontstack}/{language}.ttf",
```

wipfli:
I think it would be best if the user could indicate what unicode range a font supports in the style sheet. Something similar to how browsers do it in css `@font-face` would be ideal I think.

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range

alanchenboy:
@wipfli  Sound reasonable, Let me redesign the style when I am free. Thanks for advice.

wipfli:
We can use the geojson from https://github.com/wipfli/i18n-testsuite-maplibre to test if this pull request works with most relevant scripts...

wipfli:
Should the `-DMLN_TEXT_SHAPING_HARFBUZZ=ON` flag be in the github workflow?

https://github.com/maplibre/maplibre-native/blob/9e7f1cc3a579e096fb2f750c14078c823b2fe545/.github/workflows/linux-ci.yml#L96 ?

wipfli:
When I compile without the ` -DMLN_TEXT_SHAPING_HARFBUZZ=ON` and run `./build/bin/mbgl-render --style https://alanchenboy.github.io/harfbuzzResource/hindi.json --output out.png` I get this image:

**without the flag**:
out-no-flag

And when I compile it with the flag I get:
**with the flag**:
out

In the first image, the Hindi labels are not rendered correct which is what one expects without harfbuzz. But actually I thought that MapLibre Native just shows nothing if it encounters Hindi by default, no?

louwers:
> Should the `-DMLN_TEXT_SHAPING_HARFBUZZ=ON` flag be in the github workflow?
> 
> https://github.com/maplibre/maplibre-native/blob/9e7f1cc3a579e096fb2f750c14078c823b2fe545/.github/workflows/linux-ci.yml#L96
> ?

This brings up an interesting point. In the current state, if we decide to compile with Harfbuzz support on CI, builds without Harfbuzz will be untested. In fact, it not even sure they will even compile. 

Ideally we don't rely on `#ifdef` to choose a code path. We should have a specializations of `SymbolLayout`, `GlyphRange`, `GlyphRequestor` etc. that use HarfBuzz and specializations that don't. It should be possible to compile both versions at the same time (so we can at least compile both on CI without a proliferation of build config permutations).

Our binary distribution for Android and iOS should probably contain the 'kitchen sink' of all features available. Binary size is probably only important to big players who will make their own builds anyway (feel free to correct me here). We could consider introducing a plugin system, but I doubt that will be worth the effort.

vendor/harfbuzzFreetype/harfbuzz_wrap.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    glyphIDs.reserve(glyphCount);
+    adjusts.reserve(glyphCount);
+
+    for (uint32_t i = 0; i < glyphCount; ++i) {
+        glyphIDs.emplace_back(glyphInfo[i].codepoint);
+
+        float x_advance = static_cast<float>(glyphPos[i].x_advance / 64.0f);
+        float x_offset = static_cast<float>(glyphPos[i].x_offset / 64.0f);
+        float y_offset = static_cast<float>(-glyphPos[i].y_offset / 64.0f);
```
biodranik, line 102: Why the offset is negated here compared to L67 above?

alanchenboy: Thanks for correcting this, L67 should be negated, will correct it in next commit.

alanchenboy:
Keeping both versions for render tests is challenging, some render tests will get different download results between PBF file and the font file. This project is not like metal, Basically, it aims for the same result, but changing the shaping will make many tests turn out differently, such as the render test containing the Khmer text.

alanchenboy:
Hindi could be draw incorrectly with PBF, like your first screen shot @wipfli

alanchenboy:
use font face to to define font files, example: https://alanchenboy.github.io/harfbuzzResource/hindi.json

```
    "font-faces": [
        {
            "text-font": ["Noto Sans Regular"],
            "font-files": [
                {
                    "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
                    "unicode-range": [
                        "U+1780-17FF"
                    ]
                },
                {
                    "url":"https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
                    "unicode-range": [
                        "U+1000-109F", "U+A9E0-0xA9FF", "U+AA60-0xAA7F"
                    ]
                },
                {
                    "url":"https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
                    "unicode-range": [
                        "U+0900-097F", "U+A8E0-0xA8FF"
                    ]
                }
                
            ]
        }
    ],
```

vendor/harfbuzzFreetype/harfbuzz_wrap.cpp
alanchenboy: origin L67 removed

louwers:
@keith Do you know if rules_foreign_cc can be used with bzlmod? I created an issue here

https://github.com/bazelbuild/rules_foreign_cc/issues/1130

louwers:
!benchmark android

keith:
phew! I think we might still have to have the empty WORKSPACE file tho

louwers:
@keith Seems to work without.

@alanchenboy I am enabling Harfbuzz by default to run some benchmarks.

wipfli:
I think this is coming together nicely. Thanks for keeping working on this pull request @alanchenboy.

We want to be able to compile without HarfBuzz because of binary size. To reduce the number of ifdefs we could just have the ifdef only inside the function which calls HarfBuzz.

If you compile with HarfBuzz, this function will return the glyph vector and metrics vector filled with text shaping results. If you compile without HarfBuzz, this function will return empty vectors.

What do you think?

alanchenboy:
Sorry for the late response, I was a little bit occupied these days.
If without harfbuzz, we can simply ignore the try shaping intent at the layout thread, 

I may spare some time this week, let me figure this out .

alanchenboy:
Make harfbuzz as the default shaping engine now。

Need time to pass all the unit  @tests.

I will try to make it through, but I will be occupied by some work at my company, so maybe till next month.

BTW, I the render test of the main branch always fails on my M1 Mac do you experience the same issues?

abebeos:
@wipfli , this looks like a godd bounty-candidate (e.g. Bounty X to fix n tests). Not for me, but could be attractive to other devs.

ramSeraph:
> Hi @alanchenboy thank you for working on this.
> 
> I want to get this merged, but due to the +26% binary size increase I think it makes sense to make using Harfbuzz opt-in (or at least offer an easy opt-out). The easy way to make it opt-in would be to sprinkle `#ifdef`s throughout all the touched files, but this would not make the code very readable. Especially since more features that will be opt-in are expected in the future.
> 
> So to merge this we need to find a good pattern to make functionality opt-in. Perhaps it is possible to make a few derived classes where only at instantiation the right class is instantiated (with an `#ifdef`).

Hey @louwers What is the acceptable limit for increase in binary size? Would pulling in data during runtime be an acceptible solution here?

louwers:
@ramSeraph We don't really have a policy for it at the moment, but if we don't make a feature that has a large binary size increase opt-in or opt-out then we risk losing contributions from large corporations for whom binary size is important. They might decide to maintain an internal fork.

Each binary size increate should always be looked at in the context of the value the contribution brings.

Personally I think we should release the 'kitchen sink' with all features enabled, users that care about binary size need to make their own builds. Although we can also consider making minimal builds in adition to full builds.

> Would pulling in data during runtime be an acceptable solution if it helps reduce binary size?

No that is not acceptable, at least not for iOS because Apple forbids it.

ramSeraph:
> @ramSeraph We don't really have a policy for it at the moment, but if we don't make a feature that has a large binary size increase opt-in or opt-out then we risk losing contributions from large corporations for whom binary size is important. They might decide to maintain an internal fork.
> 
> Each binary size increate should always be looked at in the context of the value the contribution brings.

The problem with this argument is that you are indeed encouraging the people who need this feature to fork maplibre( as Grab has probably already done ).

Anyway, I do understand where you are coming from. 


> Personally I think we should release the 'kitchen sink' with all features enabled, users that care about binary size need to make their own builds. Although we can also consider making minimal builds in addition to full builds.

I am thinking a modular architecture with feature flags would be more appropriate, but in the end that will cause a test matrix overload.


> No that is not acceptable, at least not for iOS because Apple forbids it.

This is surprising.. I do mean pulling in data, not code. I was asking this because a lot of harfbuzz binary size might be coming from unicode data it loads, which can be pulled in lazily.

ramSeraph:
> I am thinking a modular architecture with feature flags would be more appropriate, but in the end that will cause a test matrix overload.

Doesn't maplibre-native already have a plugin architecture? Can this  particular feature be moved to plugin?

alanchenboy:
Pipelines fixed， lucky me. let me add some complex render tests to finish this PR.

.gitmodules
alanchenboy: I think we can review now.

louwers:
@ramSeraph 

> This is surprising.. I do mean pulling in data, not code. I was asking this because a lot of harfbuzz binary size might be coming from unicode data it loads, which can be pulled in lazily.

That is an interesting idea.

> Doesn't maplibre-native already have a plugin architecture? Can this particular feature be moved to a plugin?

Not for the C++ Core at the moment.

vendor/harfbuzzFreetype/freetype_wrap.cpp
```diff
@@ -0,0 +1,105 @@
...
+    FT_Load_Glyph(face,
+                  (FT_UInt)glyph, // the glyph_index in the font file
+                  flags);
+
+    FT_GlyphSlot slot = face->glyph;
+    FT_Render_Glyph(slot, FT_RENDER_MODE_NORMAL);
+
+    const FT_Bitmap &ftBitmap = slot->bitmap;
+
+//    fixedMetrics.id = glyphID;
```
louwers, line 74: There is some commented out code here.

vendor/BUILD.bazel
```diff
@@ -229,6 +230,71 @@ cc_library(
     ],
 )
 
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
```
louwers, line 236: This includes all source code from all libraries in `vendor`.

alanchenboy: There is a problem here: Harfbuzz's header files can't pass the tidy-clang check and will raise warnings, cause we treat warnings as errors in the core, so I have to create a wrap header file to hide the warning on the vendor side. So I Can put freetype and harfbuzz under vendor and must create a warp directory to store my wrap files, is that acceptable?

wipfli:
If this is ready for review then I would propose to have a discussion in the style spec repo about the changes needed to the style spec.

vendor/BUILD.bazel
alanchenboy: resettle the vendors.

vendor/harfbuzzFreetype/freetype_wrap.cpp
alanchenboy: deleted

alanchenboy:
> @alanchenboy Thank you so much for the effort you are putting in. I was hoping to try it out, but whenever I am trying to build from the code for Android, it's failing. Can you please help me build from the code?

Hi @robinhood245 the pipeline is passed so I think the CMake change is correct, but the new vendor requires higher CMake version, so if you installed cmake 3.18 in your Android studio,  it will use that and cause the build failure. I update cmakeVersion in `dependencies.gradle` to from `3.18.1+` to `3.19.1+`. And the app can build by android studio.
So maybe we should also update this file, or we build our own cmake file to build freetype and harfbuzz, cause most vendors are using Maplibe cmake instead of vendors' cmake file.

alanchenboy:
https://app.codecov.io/gh/maplibre/maplibre-native/commit/2f76cee32dff34d10f43f8abc4985f77a9578680 How can I fix this pipeline, is it a requirement for approve?

louwers:
It means there are some code paths that are not reached during testing.

alanchenboy:
@maxammann 
I think this approach needs to edit the style specification to support downloading font files via style config.  Before this done we should merge this PR. 
I think @wipfli once helped me edit the specification, any update?

1ec5:
> edit the style specification to support downloading font files via style config

See also the discussion in https://github.com/maplibre/maplibre-gl-js/pull/4564#discussion_r1730061958. Maybe we should coordinate on a strategy for making font selection more flexible across platforms.

louwers:
Ideally we should create an issue (and then a PR) in the [MapLibre Style Spec](https://github.com/maplibre/maplibre-style-spec/issues) repo to achieve consensus on the style spec changes and have it officially changed before merging.

louwers:
Closing in favor of https://github.com/maplibre/maplibre-native/pull/3611

PR #1454, opened 2023-08-03
add-fly-to-with-padding

JulianBissekkou:


platform/ios/src/MLNMapView.mm
```diff
@@ -4241,15 +4249,6 @@ - (void)_flyToCamera:(MLNMapCamera *)camera edgePadding:(UIEdgeInsets)insets wit
...
     }
-    
-    if ([self.camera isEqualToMapCamera:camera] && UIEdgeInsetsEqualToEdgeInsets(_contentInset, insets))
-    {
-        if (pendingCompletion)
-        {
-            [self animateWithDelay:duration animations:pendingCompletion];
-        }
-        return;
-    }
```
JulianBissekkou, line 4252: This was removed otherwise we wouldn't move the camera if someone calls the new `flyToCamera` method with `UIEdgeInsetsZero` after calling `flyToCamera` with the same camera, but with not zero `UIEdgeInsets`.

This is exactly the case that the example shows.

louwers (commented):
I'm not sure if this makes sense since `MLNMapCamera` can have a `altitude`/`viewingDistance`.

@1ec5 Do you know why something like this was never added?

platform/ios/src/MLNMapView.mm
louwers: How does this affect current users of this method?

platform/ios/src/MLNMapView.h
```diff
@@ -1253,6 +1253,25 @@ MLN_EXPORT
...
+ @param camera The new viewpoint.
+ @param duration The amount of time, measured in seconds, that the transition
+    animation should take. Specify `0` to jump to the new viewpoint
+    instantaneously. Specify a negative value to use the default duration, which
+    is based on the length of the flight path.
+ @param edgePadding The minimum padding (in screen points) that would be visible
+ around the returned camera object if it were set as the receiver’s camera.
+ @param completion The block to execute after the animation finishes.
+ */
+- (void)flyToCamera:(MLNMapCamera *)camera edgePadding:(UIEdgeInsets)insets withDuration:(NSTimeInterval)duration completionHandler:(nullable void (^)(void))completion;
```
1ec5, line 1273: > @1ec5 Do you know why something like this was never added?

I think by the time we added the flight methods, we were already reeling from how many different versions of `-setCamera:…` we needed to support. (Oh if only we could have written this library in pure Swift to begin with!) We tried to keep the number of variations of `-flyToCamera:…` to a minimum. At the time, we had a practical use case for publicly setting edge padding with `-setCamera:…`, for keeping something anchored on screen while zooming, but not for publicly setting edge padding with `-flyToCamera:…`. But internally, we did call essentially this method when jumping to the user’s location from afar:

https://github.com/maplibre/maplibre-native/blob/0af3f21367a4cb044fdd55fcb629dc7ee29fa765/platform/ios/src/MLNMapView.mm#L6229-L6233

It wasn’t until later that developers started asking for this particular method: https://github.com/mapbox/mapbox-gl-native/issues/8625#issuecomment-451557016.

platform/ios/src/MLNMapView.h
```diff
@@ -1253,6 +1253,25 @@ MLN_EXPORT
  */
 - (void)flyToCamera:(MLNMapCamera *)camera withDuration:(NSTimeInterval)duration peakAltitude:(CLLocationDistance)peakAltitude completionHandler:(nullable void (^)(void))completion;
 
+
+/**
+ Moves the viewpoint to a different location using a transition animation that
+ evokes powered flight and an optional transition duration and peak altitude.
```
1ec5, line 1259: This method does take a duration (and edge padding), but unlike the method above, it doesn’t take a peak altitude.

> I'm not sure if this makes sense since `MLNMapCamera` can have a `altitude`/`viewingDistance`.

The peak altitude is the altitude you want the camera to have if you get the map’s camera at the moment when the map is zoomed out the most.

platform/ios/src/MLNMapView.mm
JulianBissekkou: Here are 3 proposals how we can proceed with this:

#### Don't do an early return aka keep the lines deleted
This would be consistent with the android sdk since they are not handling this case. I think its totally fine for the consumers of this library to make sure that there is no ongoing camera animation when they start a new one.
We can keep and open issue and come up with a solution before releasing the next major version bump.

#### Add the old check again and document the behavior. 
We can add the behavior into the docs and write `This method will not do anything if the camera didn't change but only the padding changed`. This would be my personal preference from all the options since we have to backport the changes to 5.13 in order to fix another bug in the flutter sdk.

#### Hold the latest `mbgl::CameraOptions` as a member var and don't do anything if the options are the same
This might be the "best" solution but I don't think this should be part of the java/iOS sdks. This can be implemented in C and all platforms can benefit from it. If we do this later on we will fix android automatically as well.

We would like to have this method asap deployed since this is currently blocking us. We can continue our discussion in the slack channel which might be a faster way of communicating

platform/ios/src/MLNMapView.h
JulianBissekkou: Yeah, swift would be nice. I created a discussion for that if you are interested: https://github.com/maplibre/maplibre-native/discussions/1460

Agree, there are too many different methods to move a camera to position X with parameters Y.
This is also a problem of the C library and I have the feeling that the native android and iOS sdks just expose whatever people needed. This is no expection because we also just exposed this method because we need it.

The question is how to avoid that? Having a smaller API surface? But how can we make sure that all use cases are covered? 🤔

louwers:
Thanks for sharing your thoughts @1ec5!

JulianBissekkou:
This tries to improve https://github.com/maplibre/maplibre-native/issues/1462 by adding the same method to iOS that is already available on android

platform/ios/src/MLNMapView.mm
JulianBissekkou: Option 1 would be more consistent to android. Of course its debatable if its correct!

1ec5: Option 1 (eliminating the special case) seems risky to me because it’s inconsistent with past behavior and also inconsistent with the other camera-related methods.

I think it’s possible to fix the problem you ran into without introducing other weird behavior. Ignoring the implementation for a moment, the developer would want the method to be a no-op if the camera effectively keeps the viewport in the same position.

Obviously you’d have to compare the current camera to the passed-in camera, but I don’t think you’d want to factor in the content inset, because any change to the content inset has already kicked off a camera change automatically. So the only other comparison you’d make is whether the passed-in edge insets match the excess padding in `mbgl::CameraOptions::padding` caused by mapbox/mapbox-gl-native#15233. You can get the current camera options by calling `getCameraOptions()` on `mbglMap`. Then it’s a matter of subtracting `_contentInset` from `padding` to find that excess padding left over from the previous camera change.

If this doesn’t work, I think your next best bet would be option 1. As long as the passed-in `duration` is negative, this code would ensure that the map only jumps to the target camera (imperceptibly to the user rather than bouncing up and down:

https://github.com/maplibre/maplibre-native/blob/3c57d37ef26ca84d11e376a08a22f00e60c237ce/src/mbgl/map/transform.cpp#L276-L291

1ec5 (commented):
Hope this helps. Note that I’m just gesticulating wildly based on code inspection. I haven’t tried my suggestions out in the current codebase, and it’s been years since I had my hands on this code.

platform/ios/src/MLNMapView.mm
JulianBissekkou: I reverted the change.

platform/ios/src/MLNMapView.h
JulianBissekkou: I updated the documentation as required.

louwers (approved):
Thanks, let's revisit this API when we introduce a Swift API for MapLibre Native.

PR #1455, opened 2023-08-03
ios-setLatLngBounds

JulianBissekkou:


louwers (commented):
I think this can already be achieved by implementing the `shouldChangeFromCamera` method of a `MLNMapViewDelegate`.

JulianBissekkou:
@louwers 
The reason we introduced this was to have some more consistency between the android and the iOS framework. #1454 is trying to do the same.

This is something that might be overlooked, but we are currently struggeling a lot in getting all platforms of our flutter maplibre package to behave the same.
If the native (C/C++) implementation will be adjusted, we would expect all platforms to work the same and, if needed, we would expect to do similar changes to all platforms.

We worked with other map frameworks in the past and ArcGIS for example is handling this very nicely. Implementing a flutter plugin was quite easy.
We can also open an issue for platform consistency if that helps.

louwers:
> We can also open an issue for platform consistency if that helps.

Ah, right. I was missing that piece of context here. That would definitely help.

This is an aside, but have you looked into if there is any way we can add tests for these new methods? I know the instrumentation tests on Android are quite extensive... On iOS, not so much.

JulianBissekkou:
> This is an aside, but have you looked into if there is any way we can add tests for these new methods? I know the instrumentation tests on Android are quite extensive... On iOS, not so much.

Tbh I expected tests on the C side and since I am only calling a single method I skipped the tests for this.
How do you decide where and what to test?

JulianBissekkou:
This tries to improve #1462 by adding the same method to iOS that is already available on android

louwers (approved):

platform/ios/src/MLNMapView.h
```diff
@@ -1364,6 +1364,18 @@ MLN_EXPORT
  */
 - (CGPoint)anchorPointForGesture:(UIGestureRecognizer *)gesture;
 
+
+/**
+ * Sets a LatLngBounds that constraints map transformations to this bounds.
+ * @param latLngBounds the bounds to constrain the map with
+ */
+- (void)setLatLngBounds:(MLNCoordinateBounds)latLngBounds;
```
1ec5, line 1372: Can it be called something other than `setLatLngBounds`? The iOS/macOS SDK now has a method named after the Android SDK’s `LatLngBounds` class, which makes little sense. That class is analogous to the `MGLCoordinateBounds` struct on iOS/macOS, so it should be called something like `setCoordinateBounds`. Better yet, it should be modeled as a property called something more descriptive like `viewableCoordinateBounds`. As it is, this is a write-only property, nearly unprecedented in any Objective-C framework.

1ec5 (commented):
Please try to conform to Cocoa API naming conventions rather than Android naming conventions…

1ec5:
> This is an aside, but have you looked into if there is any way we can add tests for these new methods? I know the instrumentation tests on Android are quite extensive... On iOS, not so much.

It’s definitely possible to test these additions using the existing test harness.

platform/ios/src/MLNMapView.h
```diff
@@ -1364,6 +1364,18 @@ MLN_EXPORT
...
+
+/**
+ * Clears the bounds that were set via `setLatLngBounds`
+ */
+- (void)clearLatLnBounds;
```
plainheart, line 1377: Hi, thanks for your work! I'd like to mention that there may be a typo in the `clearLatLnBounds` API. Should it be `clearLatLngBounds`? The changelog and GitHub release notes have the same typo.

PR #1458, opened 2023-08-03
Removed public texture resource from texture 2d class

alexcristici:
I made this change on the metal branch, but I was thinking to add it now instead keeping it as a metal change.

Note: This change will break `flavor split`.


mwilsnd (approved):
Nice! Breaking split view at this point is probably fine. We won't be able to use it for comparing different rendering APIs anyways.

We should plan for when we want to pull out the rest of the split view code if we're going to go ahead and break it.

sjg-wdw (approved):

stefankarschti (approved):

PR #1464, opened 2023-08-04
Add expression test runner to Bazel

louwers:
- Replaces ghc::filesystem with std::filesystem for the expression test runner.
- Adds Bazel target for the expression test runner and includes expression test in coverage calculation.

### DIFF
--- a/expression-test/expression_test_parser.cpp
+++ b/expression-test/expression_test_parser.cpp
@@ -53,7 +53,7 @@ void writeJSON(rapidjson::PrettyWriter<rapidjson::StringBuffer>& writer, const V
 
 using ErrorMessage = std::string;
 using JSONReply = variant<JSDocument, ErrorMessage>;
-JSONReply readJson(const filesystem::path& jsonPath) {
+JSONReply readJson(const std::filesystem::path& jsonPath) {
     auto maybeJSON = util::readFile(jsonPath);
     if (!maybeJSON) {
         return {"Unable to open file "s + jsonPath.string()};
@@ -309,8 +309,9 @@ bool parseInputs(const JSValue& inputsValue, TestData& data) {
 
 } // namespace
 
-std::tuple<filesystem::path, std::vector<filesystem::path>, bool, uint32_t> parseArguments(int argc, char** argv) {
-    args::ArgumentParser argumentParser("Mapbox GL Expression Test Runner");
+std::tuple<std::filesystem::path, std::vector<std::filesystem::path>, bool, uint32_t> parseArguments(int argc,
+                                                                                                     char** argv) {
+    args::ArgumentParser argumentParser("MapLibre Native Expression Test Runner");
 
     args::HelpFlag helpFlag(argumentParser, "help", "Display this help menu", {'h', "help"});
     args::Flag shuffleFlag(argumentParser, "shuffle", "Toggle shuffling the tests order", {'s', "shuffle"});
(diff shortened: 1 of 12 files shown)


expression-test/expression_test_parser.cpp
```diff
@@ -53,7 +53,7 @@ void writeJSON(rapidjson::PrettyWriter<rapidjson::StringBuffer>& writer, const V
 
 using ErrorMessage = std::string;
 using JSONReply = variant<JSDocument, ErrorMessage>;
-JSONReply readJson(const filesystem::path& jsonPath) {
+JSONReply readJson(const std::filesystem::path& jsonPath) {
     auto maybeJSON = util::readFile(jsonPath);
```
mwilsnd, line 57: You need to call `jsonPath.string()` here for `readFile`

expression-test/expression_test_parser.cpp
```diff
@@ -339,13 +340,13 @@ std::tuple<filesystem::path, std::vector<filesystem::path>, bool, uint32_t> pars
         exit(2);
     }
 
-    filesystem::path rootPath{std::string(TEST_RUNNER_ROOT_PATH).append("/metrics/integration/expression-tests")};
-    if (!filesystem::exists(rootPath)) {
+    const auto rootPath = std::filesystem::current_path().append("metrics/integration/expression-tests");
+    if (!std::filesystem::exists(rootPath)) {
```
mwilsnd, line 344: ditto

mwilsnd (changes requested):

expression-test/expression_test_parser.cpp
louwers: > Paths are implicitly convertible to and from [std::basic_string](https://en.cppreference.com/w/cpp/string/basic_string)s, which makes it possible to use them with other file APIs.

https://en.cppreference.com/w/cpp/filesystem/path

louwers: OK just noticed it does not compile for Windows. 💢

mwilsnd: It's a bit weird but what I think is going on: The implicit conversion is to the system native type, string for POSIX and wstring for Win32. So an implicit conversion to `std::wstring` would compile on Windows.

louwers: Yes makes sense.

platform/linux/src/headless_backend_glx.cpp
```diff
@@ -39,7 +39,7 @@ class GLXDisplayConfig {
             if (xDisplay != nullptr) {
                 break;
             }
-            Log::Debug(Event::OpenGL, "[CI] Failed to open X display, retrying...");
+            Log::Error(Event::OpenGL, "[CI] Failed to open X display, retrying...");
```
mwilsnd, line 42: This was changed from Error to Debug intentionally, some of the tests use a [log observer](https://github.com/maplibre/maplibre-native/blob/88917b18065f6c7ee5d11ab2e109e0fab7af6edf/src/mbgl/util/logging.cpp#L53-L56) and only [Debug is omitted](https://github.com/maplibre/maplibre-native/blob/88917b18065f6c7ee5d11ab2e109e0fab7af6edf/src/mbgl/util/logging.cpp#L79). If XOpenDisplay fails during a test using the [FixtureLogObserver](https://github.com/maplibre/maplibre-native/blob/main/test/src/mbgl/test/fixture_log_observer.hpp), the test will fail because of the error level used in this retry message.

mwilsnd (changes requested):

platform/linux/src/headless_backend_glx.cpp
louwers: Thanks for clarifying.

I didn't see any retry messages in the logs, so I didn't know if it was actually retrying.

mwilsnd (approved):

PR #1459, opened 2023-08-03
Drawable PR2 - Renderer Integration

mwilsnd:
Extending on the new concepts introduced in #1355, this PR:
- Integrates with the existing rendering infrastructure
  - Render orchestration, ordering, updating, traversal and rendering
  - Atlases (Not all, remaining will accompany relevant layer PRs)
  - Paint properties, data driven attributes
  - Build system: CMake, Bazel, Gradle
- Adds pre-processor checks and build system flags to build with or without the new rendering pathway
  - Adds the temporary debugging split view mode for viewing both pathways concurrently
- A clang-tidy pass was run at one point, locating some missing/transient includes

Some notes when reviewing:
- References to files added in future layer PRs haven't been removed from the build system(s)
- References to shaders added in future layer PRs haven't been removed
- As mentioned in the last PR, refer to the [design results](https://docs.google.com/document/d/1QtsUgDyD-Rt1McvJFz7HbHvEKli3tRVfxdW7fLDF-g8) document for some extra guiding information during review

PR #1501, opened 2023-08-09
Drawable PR 3 - Layers

mwilsnd:
In this last PR we introduce alternative ways of rendering the various layer types using drawables. Within each `RenderLayer` instance drawables are constructed via a new `update` method. Legacy rendering via the `render` method is compiled out in drawable mode. Texture atlases and custom layers are also addressed when running with drawables. New drawable variants of the shaders have also been added, prefixed with `drawable.` Refer to the proceeding PRs #1459 and #1355 for extra context when reviewing.

### DIFF
--- a/src/mbgl/renderer/layers/hillshade_layer_tweaker.cpp
+++ b/src/mbgl/renderer/layers/hillshade_layer_tweaker.cpp
@@ -1,0 +1,85 @@
+#include <mbgl/renderer/layers/hillshade_layer_tweaker.hpp>
+
+#include <mbgl/gfx/context.hpp>
+#include <mbgl/gfx/drawable.hpp>
+#include <mbgl/renderer/layer_group.hpp>
+#include <mbgl/renderer/paint_parameters.hpp>
+#include <mbgl/renderer/render_tree.hpp>
+#include <mbgl/style/layers/hillshade_layer_properties.hpp>
+#include <mbgl/util/convert.hpp>
+
+namespace mbgl {
+
+using namespace style;
+
+struct alignas(16) HillshadeDrawableUBO {
+    std::array<float, 4 * 4> matrix;
+    std::array<float, 2> latrange;
+    std::array<float, 2> light;
+};
+static_assert(sizeof(HillshadeDrawableUBO) % 16 == 0);
+
+struct alignas(16) HillshadeEvaluatedPropsUBO {
+    Color highlight;
+    Color shadow;
+    Color accent;
+};
+static_assert(sizeof(HillshadeEvaluatedPropsUBO) % 16 == 0);
+
+static constexpr std::string_view HillshadeDrawableUBOName = "HillshadeDrawableUBO";
+static constexpr std::string_view HillshadeEvaluatedPropsUBOName = "HillshadeEvaluatedPropsUBO";
+
+std::array<float, 2> getLatRange(const UnwrappedTileID& id) {
+    const LatLng latlng0 = LatLng(id);
+    const LatLng latlng1 = LatLng(UnwrappedTileID(id.canonical.z, id.canonical.x, id.canonical.y + 1));
+    return {{static_cast<float>(latlng0.latitude()), static_cast<float>(latlng1.latitude())}};
+}
+
+std::array<float, 2> getLight(const PaintParameters& parameters,
+                              const HillshadePaintProperties::PossiblyEvaluated& evaluated) {
+    float azimuthal = util::deg2radf(evaluated.get<HillshadeIlluminationDirection>());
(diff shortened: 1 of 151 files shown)


sjg-wdw (approved):
That's a big 'un all right.

src/mbgl/renderer/layers/hillshade_layer_tweaker.cpp
```diff
@@ -0,0 +1,85 @@
...
+                                                        /* .shadow = */ evaluated.get<HillshadeShadowColor>(),
+                                                        /* .accent = */ evaluated.get<HillshadeAccentColor>()};
+        evaluatedPropsUniformBuffer = parameters.context.createUniformBuffer(&evaluatedPropsUBO,
+                                                                             sizeof(evaluatedPropsUBO));
+    }
+
+    layerGroup.visitDrawables([&](gfx::Drawable& drawable) {
+        drawable.mutableUniformBuffers().addOrReplace(HillshadeEvaluatedPropsUBOName, evaluatedPropsUniformBuffer);
+
+        if (!drawable.getTileID()) {
```
VladyslavOdobesku-TomTom, line 71: I see 10 layer tweakers that do this check but also have some code before this check.

Why `drawable.mutableUniformBuffers().addOrReplace(HillshadeEvaluatedPropsUBOName, evaluatedPropsUniformBuffer);` is done before the check and not after?

VladyslavOdobesku-TomTom (approved):

src/mbgl/renderer/layers/render_circle_layer.cpp
```diff
@@ -236,4 +262,162 @@ bool RenderCircleLayer::queryIntersectsFeature(const GeometryCoordinates& queryG
...
+        if (!renderData || !renderData->bucket || !renderData->bucket->hasData()) {
+            removeTile(renderPass, tileID);
+            continue;
+        }
+
+        const auto& bucket = static_cast<const CircleBucket&>(*renderData->bucket);
+        const auto vertexCount = bucket.vertices.elements();
+        const auto& paintPropertyBinders = bucket.paintPropertyBinders.at(getID());
+
+        const auto prevBucketID = getRenderTileBucketID(tileID);
```
louwers, line 340: During the meeting it was mentioned this could have a little more explanation (that is, why this is needed).

src/mbgl/renderer/layers/hillshade_prepare_layer_tweaker.cpp
```diff
@@ -0,0 +1,73 @@
...
+    // https://aws.amazon.com/public-datasets/terrain/
+    static const std::array<float, 4> unpackTerrarium = {{256.0f, 1.0f, 1.0f / 256.0f, 32768.0f}};
+
+    return encoding == Tileset::DEMEncoding::Terrarium ? unpackTerrarium : unpackMapbox;
+}
+
+void HillshadePrepareLayerTweaker::execute(LayerGroupBase& layerGroup,
+                                           [[maybe_unused]] const RenderTree& renderTree,
+                                           const PaintParameters& parameters) {
+    // const auto& evaluated = static_cast<const HillshadeLayerProperties&>(*evaluatedProperties).evaluated;
```
louwers, line 39: Stray comment

src/mbgl/renderer/layers/render_circle_layer.cpp
TimSylvester: This code, and the terse comment below, exists in each layer, so an explanatory comment probably belongs somewhere more central.  I'll add something to `getRenderTileBucketID` separately.

louwers (approved):
Does it make sense to have a short description per layer (above `update()`) and tweaker (above` execute()`) that gives a high-level overview description what the layer/tweaker does and lists some of the particularities of that layer/tweaker ?

E.g. for the heat map layer it can be as simple as "creates drawables for off screen target per tile, then creates drawable to blend texture on the screen". I think this might help a future reader understand a layer/tweaker without getting lost in the details.

sjg-wdw:
@alanchenboy did you want to review this?
I realize you're using it in the fill layer implementation, so you're already familiar with what we did.

src/mbgl/renderer/layers/hillshade_layer_tweaker.cpp
TimSylvester: The tile ID is optional, so should always be checked before it's used.  For most layers, it's expected to always be set.

Attaching the shared UBO to the drawable doesn't depend on the tile ID.  The order is likely different because the layers were built in parallel.  It would be nice to go through and make them look more similar, just for clarity, though.

TimSylvester (approved):

alanchenboy (approved):
lgtm

PR #1507, opened 2023-08-11
Ios 5 hot fix padding and bbox

stefanschaller:


louwers (approved):
Note that this older branch has an outdated CI config which does not run on the runner anymore. 

https://github.com/maplibre/maplibre-native/actions/runs/5830210125

I still feel confident releasing this as 5.14.0 since the changes are only adding two new methods without changing existing code.

I hope the release workflow runs through.

PR #1534, opened 2023-08-17
iOS C++ Unit Tests Bazel target & running on CI

### LINKED ISSUE #1001: Run iOS C++ unit tests on CI
Run iOS C++ unit tests via the [iOS UnitTestsRunner App](https://github.com/maplibre/maplibre-gl-native/tree/fc0873a682363ab1b095341b9a406785ccfc7620/test/ios).

This is especially important because we will disable the macOS CI while the Metal backend implementation is underway due to macOS' lacking OpenGL 3.0 support.

The test runner is being run here:

https://github.com/maplibre/maplibre-gl-native/blob/fc0873a682363ab1b095341b9a406785ccfc7620/test/ios/ios_test_runner.cpp#L14

louwers:
Closes #1001 

Still TODO:

- [x] Use the C++ test server with the `TEST_HAS_SERVER` macro
- [x] Run on CI ~in simulator~ on device
- [x] Make sure tests are picked up...

About that last point. When I add a test to `test/src/mbgl/test/test.cpp` it is picked up. But the existing tests `test/**/*.test.cpp` are not right now. I tried various things but I can't seem to get it to work. The Google Test docs state

> TEST() and TEST_F() implicitly register their tests with GoogleTest. So, unlike with many other C++ testing frameworks, you don't have to re-list all your defined tests in order to run them.

> After defining your tests, you can run them with RUN_ALL_TESTS(), which returns 0 if all the tests are successful, or 1 otherwise. Note that RUN_ALL_TESTS() **runs all tests in your link unit**--they can be from different test suites, or even different source files.

As far as I can tell the tests are in the same 'link unit' as `test/src/mbgl/test/test.cpp`. @mwilsnd do you have any ideas?

Can be run with

```
bazel run //platform/ios/test/core:CppUnitTests
```

---

Edit: This PR is now read for review.

- I made the `ios-render-test` workflow generic (renamed to `ios-device-test`) so it can run both the render tests and the C++ unit tests on AWS Device Farm.
- Had to add an epsilon instead of `0.0` in `bounding_volumes.cpp` to make the `BoundingVolumes.AabbFullyInsideFrustum` and `BoundingVolumes.AabbIntersectsFrustum` tests pass.
- Moved the relevant iOS specific files from `test/ios` to `platform/ios/test/core`.


louwers:
OK I figured it out thanks to someone on the Bazel Slack, need to use the `alwayslink` option https://bazel.build/reference/be/c-cpp#cc_import_args:~:text=in%20dependent%20rules.-,alwayslink,-Boolean%3B%20optional%3B%20default

louwers:
Yay!

```
[==========] 904 tests from 116 test suites ran. (121066 ms total)
[  PASSED  ] 901 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] Expression.IsExpression
[  FAILED  ] BoundingVolumes.AabbFullyInsideFrustum
[  FAILED  ] BoundingVolumes.AabbIntersectsFrustum
```

Just needed to run on a real device instead of a simulator for the HTTP lib to work.

So will be integrating with AWS Device Farm again to get it running on CI.

src/mbgl/util/bounding_volumes.cpp
```diff
@@ -213,13 +213,15 @@ IntersectionResult Frustum::intersects(const AABB& aabb) const {
 
     bool fullyInside = true;
 
+    const auto epsilon = 1e-10;
```
louwers, line 216: `std::numeric_limits<double>::epsilon()` was too small.

Maybe we should define this globally somewhere.

louwers:
Something to keep an eye on is rules_apple support for XCTest bundles https://github.com/bazelbuild/rules_apple/pull/1994

src/mbgl/util/bounding_volumes.cpp
mwilsnd: I think it's fine for now, if we start having to do this in other places we should add some test constants though.

platform/ios/BUILD.bazel
```diff
@@ -213,7 +213,7 @@ genrule(
 xcodeproj(
     name = "xcodeproj",
     bazel_env = {"PATH": "/bin:/usr/bin:/usr/local/bin:/opt/homebrew/bin"},
-    build_mode = "xcode",
+    build_mode = "bazel",
```
mwilsnd, line 216: A couple things:
- bwb is really slow, it isn't building in parallel. There might be a way to fix that though.
- bwb fails to locate a provisioning profile which is why we switched to bwx. I tested building the demo app and it still fails here in bwb mode.

mwilsnd (changes requested):

platform/ios/BUILD.bazel
louwers: It does find the provisioning profile for me now. Could we make a bug report over at https://github.com/MobileNativeFoundation/rules_xcodeproj ?

BwX does not have first-class support from rules_xcodeproj and support for it might be removed altogether in the future. I will add your comment to the discussion [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/discussions/2391).

louwers: I made the `BUILD_MODE` configurable. You can set it to your liking in your `config.bzl`.

mwilsnd (approved):
I get a warning about an incorrect bundle ID but it doesn't stop the build or test runner from working - `MapLibre.xcodeproj: warning: User-supplied CFBundleIdentifier value 'org.maplibre.TestData' in the Info.plist must be the same as the PRODUCT_BUNDLE_IDENTIFIER build setting value ''.` I'm still not able to get BwB to locate a provisioning profile but everything works fine in BwX mode.

PR #1541, opened 2023-08-17
Set uniform values based on paint property binders to match `computeAllUniformValues`

TimSylvester:
The behavior of `render()` is to compute the uniform values using the paint property binders, which are not available in the tweaker.  To match that, we need to set up and update the property UBO in the render layer.

### DIFF
--- a/src/mbgl/renderer/layers/render_heatmap_layer.cpp
+++ b/src/mbgl/renderer/layers/render_heatmap_layer.cpp
@@ -278,10 +278,26 @@ struct alignas(16) HeatmapInterpolateUBO {
 };
 static_assert(sizeof(HeatmapInterpolateUBO) % 16 == 0);
 
+struct alignas(16) HeatmapEvaluatedPropsUBO {
+    float weight;
+    float radius;
+    float intensity;
+    float padding;
+};
+static_assert(sizeof(HeatmapEvaluatedPropsUBO) % 16 == 0);
+
 constexpr auto HeatmapShaderGroupName = "HeatmapShader";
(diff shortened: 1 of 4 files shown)


sjg-wdw (approved):
Kind of makes sense.

src/mbgl/renderer/layers/render_heatmap_layer.cpp
```diff
@@ -368,9 +384,27 @@ void RenderHeatmapLayer::update(gfx::ShaderRegistry& shaders,
...
+        const HeatmapEvaluatedPropsUBO evaluatedPropsUBO = {
+            /* .weight = */ uniformValue<style::HeatmapWeight>(paintPropertyBinders, evaluated),
+            /* .radius = */ uniformValue<style::HeatmapRadius>(paintPropertyBinders, evaluated),
+            /* .intensity = */ evaluated.get<HeatmapIntensity>(),
+            /* .padding = */ 0};
+        if (evaluatedPropsUniformBuffer) {
+            evaluatedPropsUniformBuffer->update(&evaluatedPropsUBO, sizeof(evaluatedPropsUBO));
+        } else {
+            evaluatedPropsUniformBuffer = context.createUniformBuffer(&evaluatedPropsUBO, sizeof(evaluatedPropsUBO));
+        }
```
alexcristici, line 402: From what I know the evaluated props are uniform values and we don't need property binders to extract them. In case the values is a vertex attribute the uniform value it will be unused so it's not important what contains.

stefankarschti:
It's ok, but the tweaker code executes before drawables get drawn anyway, If I'm not missing something.

Issue #1548, opened 2023-08-21
Frame rendering time


sjg-wdw:
Look into what stats we've got for frame rendering times.
Min/max, median, etc.

stefankarschti:
Running the iOS Benchmark App on an iPad 5th generation outputs:

Legacy:
```
2023-08-22 15:22:55.185335+0300 Bench GL[15910:8270163] Result:
2023-08-22 15:22:55.185400+0300 Bench GL[15910:8270163] | paris      | 59.9 fps |
2023-08-22 15:22:55.185452+0300 Bench GL[15910:8270163] | paris2     | 59.7 fps |
2023-08-22 15:22:55.185501+0300 Bench GL[15910:8270163] | alps       | 59.8 fps |
2023-08-22 15:22:55.185548+0300 Bench GL[15910:8270163] | us east    | 59.8 fps |
2023-08-22 15:22:55.185594+0300 Bench GL[15910:8270163] | greater la | 59.9 fps |
2023-08-22 15:22:55.185641+0300 Bench GL[15910:8270163] | sf         | 41.3 fps |
2023-08-22 15:22:55.185686+0300 Bench GL[15910:8270163] | oakland    | 59.8 fps |
2023-08-22 15:22:55.185731+0300 Bench GL[15910:8270163] | germany    | 59.7 fps |
2023-08-22 15:22:55.185768+0300 Bench GL[15910:8270163] Total FPS: 460.1
2023-08-22 15:22:55.186270+0300 Bench GL[15910:8270163] Average FPS: 57.5
```

Drawable:
```
2023-08-22 15:30:17.050719+0300 Bench GL[15917:8272762] Result:
2023-08-22 15:30:17.051069+0300 Bench GL[15917:8272762] | paris      | 59.9 fps |
2023-08-22 15:30:17.051178+0300 Bench GL[15917:8272762] | paris2     | 60.0 fps |
2023-08-22 15:30:17.051258+0300 Bench GL[15917:8272762] | alps       | 59.8 fps |
2023-08-22 15:30:17.051336+0300 Bench GL[15917:8272762] | us east    | 59.9 fps |
2023-08-22 15:30:17.051412+0300 Bench GL[15917:8272762] | greater la | 59.9 fps |
2023-08-22 15:30:17.051488+0300 Bench GL[15917:8272762] | sf         | 44.7 fps |
[...]

PR #1561, opened 2023-08-23
VAO creation

stefankarschti:
Improvement for #1548


sjg-wdw (approved):
Oh!  Were we creating the VAOs every frame?
That would be expensive.

mwilsnd (approved):

PR #1562, opened 2023-08-24
Topic/metal symbol

TimSylvester:
Text/icon symbols basically working.
I noticed in the process that the SDF text and SDF icon paths were identical, so I combined them to reduce size.
GL variants don't build in the base branch, and so don't build here either.


sjg-wdw (approved):
The implementation looks pretty logical.

stefankarschti (approved):

PR #1592, opened 2023-08-31
Drawable performance fixes

stefankarschti:
_(retargeting #1588 at `main`)_

Improves #1548 
- Reuse Line Pattern UBO
- Avoid unnecessary deletion of Fill drawables
- CRC32 UBO data to elide upload
- split out high frequency `u_fade_change` uniform in its own UBO in Symbol shaders
- add a generic `type` to drawable 
- Use drawable `type` in the Line layer + tweaker to determine drawable line type instead of UBO query by name

### DIFF
--- a/platform/ios/scripts/bazel-generate-plists.sh
+++ b/platform/ios/scripts/bazel-generate-plists.sh
@@ -82,5 +82,4 @@ cat > platform/ios/bazel/config.bzl <<EOF
 APPLE_MOBILE_PROVISIONING_PROFILE_TEAM_ID = "$teamid"
 APPLE_MOBILE_PROVISIONING_PROFILE_NAME = "iOS Team Provisioning Profile: *"
 BUNDLE_ID_PREFIX = "$bundleidprefix"
-BUILD_MODE = "xcode"
 EOF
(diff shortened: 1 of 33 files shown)


louwers (approved):

TimSylvester (approved):

platform/ios/scripts/bazel-generate-plists.sh
```diff
@@ -82,4 +82,5 @@ cat > platform/ios/bazel/config.bzl <<EOF
 APPLE_MOBILE_PROVISIONING_PROFILE_TEAM_ID = "$teamid"
 APPLE_MOBILE_PROVISIONING_PROFILE_NAME = "iOS Team Provisioning Profile: *"
 BUNDLE_ID_PREFIX = "$bundleidprefix"
+BUILD_MODE = "xcode"
```
louwers, line 85: Should be good to leave this out now.

louwers (changes requested):
Android CI complains about an unused variable.

```
      bool updateLineProps = propertiesUpdated;
```

platform/ios/scripts/bazel-generate-plists.sh
stefankarschti: ok, I'll revert. but this way the script will produce a `config.bzl` file that won't work because of the missing `BUILD_MODE`.

louwers (approved):

PR #1610, opened 2023-09-05
Drawable performance fixes (2)

stefankarschti:
Continue performance fixes for more demanding map styles used by the community (FB, Americana).
Improves #1548


src/mbgl/shaders/gl/shader_program_gl.cpp
```diff
@@ -89,21 +89,27 @@ ShaderProgramGL::ShaderProgramGL(UniqueProgram&& glProgram_)
...
       glProgram(std::move(other.glProgram)),
       uniformBlocks(std::move(other.uniformBlocks)),
-      vertexAttributes(std::move(other.vertexAttributes)) {}
+      vertexAttributes(std::move(other.vertexAttributes)),
+      samplerLocations(std::move(other.samplerLocations)) {}
 
-std::optional<uint32_t> ShaderProgramGL::getSamplerLocation(const std::string_view name) const {
-    GLint sampler_location = MBGL_CHECK_ERROR(glGetUniformLocation(glProgram, name.data()));
-    return (sampler_location == -1) ? std::optional<uint32_t>{} : sampler_location;
+std::optional<uint32_t> ShaderProgramGL::getSamplerLocation(const std::string& name) {
```
louwers, line 107: Method looks like it could still be const.

sjg-wdw (approved):

TimSylvester (approved):

alexcristici (approved):

include/mbgl/util/string_indexer.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <cstddef>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace mbgl {
+
+using StringIdentity = size_t;
+
+class StringIndexer {
```
louwers, line 13: Would you mind adding a few simple tests for this class in `test/util/string_indexer.test.cpp`?

stefankarschti: sure. good idea.

stefankarschti:
Render tests failures are the known ones.
```
1199 passed (91.4%)
11 passed but were ignored (0.8%)
89 ignored (6.8%)
13 failed (1.0%)
```

src/mbgl/gl/vertex_attribute_gl.cpp
mwilsnd, line 1: I'm failing an assert here in `::getRaw()` with type GL_FLOAT_VEC4 on the Meta style. Looks to be caused by a fill-pattern drawable.

mwilsnd (changes requested):

src/mbgl/gl/vertex_attribute_gl.cpp
TimSylvester: Is that on the spreadsheet?  I'm not seeing in the FB light/dark/3D styles.  I don't think `getRaw` is even supposed to be called any more, everything should be shared data.

stefankarschti: what is your test scenario and platform?

TimSylvester:
Looks like there's an issue with depth between extrusion and symbols here.

src/mbgl/gl/vertex_attribute_gl.cpp
TimSylvester: I'm just browsing areas where I know there are fill patterns on my iPad.

mwilsnd: I can trigger the assert on the Windows GLFW platform. FB Light style and navigating to the financial district in NYC is my repro method.
image

mwilsnd: Reproduced on an iPhone 12 at the same location:
Screenshot 2023-09-11 at 4 16 32 PM

platform/ios/scripts/bazel-generate-plists.sh
```diff
@@ -82,4 +82,5 @@ cat > platform/ios/bazel/config.bzl <<EOF
 APPLE_MOBILE_PROVISIONING_PROFILE_TEAM_ID = "$teamid"
 APPLE_MOBILE_PROVISIONING_PROFILE_NAME = "iOS Team Provisioning Profile: *"
 BUNDLE_ID_PREFIX = "$bundleidprefix"
+BUILD_MODE = "xcode"
```
louwers, line 85:
```suggestion
BUILD_MODE = "bazel"
```

Still having problems?

louwers (approved):

platform/ios/scripts/bazel-generate-plists.sh
stefankarschti: Both work for me, I prefer "xcode" because of the information level during the build, that's all. 
It should not have been committed.

platform/ios/app/MBXViewController.m
```diff
@@ -2030,6 +2029,21 @@ - (void)setStyles
...
+    [self.styleURLs addObject:[NSURL URLWithString:@"https://zelonewolf.github.io/openstreetmap-americana/style.json"]];
+
+    [self.styleNames addObject:@"Facebook Light"];
+    [self.styleURLs addObject:[NSURL URLWithString:@"https://external.xx.fbcdn.net/maps/vt/style/canterbury_1_0/?locale=en_US"]];
+
+    [self.styleNames addObject:@"Facebook Dark"];
+    [self.styleURLs addObject:[NSURL URLWithString:@"https://external.xx.fbcdn.net/maps/vt/style/dark/?locale=en_US"]];
+
+    [self.styleNames addObject:@"Facebook 3D"];
+    [self.styleURLs addObject:[NSURL URLWithString:@"https://external.xx.fbcdn.net/maps/vt/style/default_3d/?locale=en_US"]];
```
mwilsnd, line 2045: We shouldn't be adding these endpoints here (licensing complexities) - we can probably provide an offline package for the MapLibre organization to self-host however.

mwilsnd (changes requested):

mwilsnd (approved):

src/mbgl/util/string_indexer.cpp
```diff
@@ -9,8 +9,10 @@ const std::string empty;
 }
 
 StringIdentity StringIndexer::get(const std::string& string) {
-    MapType& stringToIdentity = getMap();
-    VectorType& identityToString = getVector();
+    std::unique_lock<std::shared_mutex> writerLock(instance().sharedMutex);
```
mwilsnd, line 12: We can just take a read lock here. If we don't find our string in the cache, we release the read lock and take a write lock to do our insertion.

stefankarschti: Right, I thought about it but then assumed that write contention is quite low and happening mostly in initialization phase. 
I'll change it the way you suggest.
If the string access is proving to be an issue we could switch to a lock-free solution using atomics and spins, so we avoid potential thread priority inversions as well.

Other aspect to consider is the indexer's use of two containers, which is not ideal memory wise - both in terms of size and fragmentation. Considering we're only doing additions, a single vector and linear search could be more L1 cache friendly on a certain scale.

PR #1611, opened 2023-09-05
Merge latest from main

TimSylvester:
Building and running under Xcode in legacy, drawable, and metal variants.


sjg-wdw:
Nice!

TimSylvester:
After merging the Metal fill work, I'm seeing a bug in pattern fills which applies to both OpenGL and Metal.  The tiles have different icon atlases which all contain the pattern, but some tiles appear to be using the wrong texture coordinates within those atlases.  Looking into it now...

Screenshot 2023-09-06 at 9 14 55 AM

Screenshot 2023-09-06 at 9 13 49 AM

stefankarschti (approved):

src/mbgl/renderer/layers/fill_layer_tweaker.cpp
```diff
@@ -52,16 +55,19 @@ void FillLayerTweaker::execute(LayerGroupBase& layerGroup,
     auto UpdateFillUniformBuffers = [&]() {
         if (fillUniformBufferUpdated) return;
 
+#if MLN_RENDER_BACKEND_METAL
```
sjg-wdw, line 58: Not worth doing on this one, but might be nice to make a more generic hasPermutation check.

src/mbgl/renderer/layers/heatmap_layer_tweaker.hpp
```diff
@@ -9,8 +9,8 @@ namespace mbgl {
  */
 class HeatmapLayerTweaker : public LayerTweaker {
 public:
-    HeatmapLayerTweaker(std::string id, Immutable<style::LayerProperties> properties)
-        : LayerTweaker(std::move(id), properties) {}
+    HeatmapLayerTweaker(std::string id_, Immutable<style::LayerProperties> properties)
```
sjg-wdw, line 12: I feel like id should be a reserved keyword in every language just because.

src/mbgl/renderer/layers/fill_layer_tweaker.cpp
TimSylvester: There's more pending, but tied up with render targets.

https://github.com/maplibre/maplibre-native/blob/6f6970ebf48532f526a12b62a27e3cd202dba844/include/mbgl/renderer/layer_tweaker.hpp#L42-L46

sjg-wdw (approved):
That was a lot.

src/mbgl/renderer/layers/fill_layer_tweaker.cpp
sjg-wdw: Gotcha.

PR #1616, opened 2023-09-07
Metal setup merge

TimSylvester:
Merging back the result of merging main into metal-setup.


include/mbgl/shaders/mtl/common.hpp
```diff
@@ -0,0 +1,202 @@
+#pragma once
+
+#define MLN_MTL_UNIFORM_BLOCK(idx, vert, frag, struc) \
```
louwers, line 3: In my opinion it is not worth a saving few extra keystrokes for introducing a macro here.

louwers (changes requested):
Suggestion: let's add a line to `ios-ci.yml` that compiles the toolkit with the `//:metal_renderer` flag, so we ensure that it at least continues to compile without errors.

include/mbgl/shaders/mtl/common.hpp
TimSylvester: I added this to avoid needing to have the structure name listed twice, once directly and once as a string, and ensure they stay identical.
I was trying to figure out how to make `MLN_STRINGIZE` ensure that the given symbol is valid, but couldn't make it work.

louwers:
🙏

stefankarschti (approved):

include/mbgl/mtl/renderer_backend.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+using ProcAddress = void (*)();
+using FramebufferID = uint32_t;
+
+class RendererBackend : public gfx::RendererBackend {
+public:
+    RendererBackend(gfx::ContextMode);
+    ~RendererBackend() override;
+
+    /// Called prior to rendering to update the internally assumed OpenGL state.
```
louwers, line 27: Should be 'Metal state'
```suggestion
    /// Called prior to rendering to update the internally assumed Metal state.
```

include/mbgl/mtl/renderer_backend.hpp
```diff
@@ -0,0 +1,67 @@
...
+protected:
+    std::unique_ptr<gfx::Context> createContext() override;
+
+    /// Reads the color pixel data from the currently bound framebuffer.
+    PremultipliedImage readFramebuffer(const Size&);
+
+    /// A constant to signal that a framebuffer is bound, but with an unknown ID.
+    static constexpr const FramebufferID ImplicitFramebufferBinding = std::numeric_limits<FramebufferID>::max();
+
+    /// Tells the renderer that OpenGL state has already been set by the windowing toolkit.
```
louwers, line 45: Here too?
```suggestion
    /// Tells the renderer that Metal state has already been set by the windowing toolkit.
```

include/mbgl/mtl/renderer_backend.hpp
```diff
@@ -0,0 +1,67 @@
...
+    /// It sets the internal assumed state to the supplied values.
+    void assumeFramebufferBinding(FramebufferID fbo);
+    void assumeViewport(int32_t x, int32_t y, const Size&);
+    void assumeScissorTest(bool);
+
+    /// Returns true when assumed framebuffer binding hasn't changed from the implicit binding.
+    bool implicitFramebufferBound();
+
+public:
+    /// Triggers an OpenGL state update if the internal assumed state doesn't
```
louwers, line 55:
```suggestion
    /// Triggers an Metal state update if the internal assumed state doesn't
```

src/mbgl/mtl/buffer_resource.cpp
```diff
@@ -0,0 +1,41 @@
...
+BufferResource::BufferResource(BufferResource&& other)
+    : device(std::move(other.device)),
+      buffer(std::move(other.buffer)),
+      usage(other.usage) {}
+
+void BufferResource::update(const void* data, std::size_t size, std::size_t offset) {
+    if (buffer && data) {
+        if (void* content = buffer->contents()) {
+            std::memcpy(static_cast<uint8_t*>(content) + offset, data, size);
+            // if (mode == MTL::StorageModeManaged) buffer->didModifyRange(NS::Range::Make(0, size));
```
louwers, line 35: Stray comment.

include/mbgl/shaders/mtl/common.hpp
louwers: How about something like this: https://wandbox.org/permlink/GUq6XSaVVzETU4A4

include/mbgl/mtl/buffer_resource.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <Metal/MTLBuffer.hpp>
+#include <Metal/MTLDevice.hpp>
+
+namespace mbgl {
+namespace mtl {
+
+class BufferResource {
+public:
+    BufferResource() = default;
+    BufferResource(MTLDevicePtr, const void*, std::size_t size, NS::UInteger usage);
```
louwers, line 16: Is the `usage` parameter obvious to someone familiar with the Metal API?

Otherwise, consider documenting this parameter.

include/mbgl/mtl/context.hpp
```diff
@@ -0,0 +1,157 @@
...
+                                      std::string_view source,
+                                      std::string_view vertexName,
+                                      std::string_view fragmentName,
+                                      const ProgramParameters& programParameters,
+                                      const std::unordered_map<std::string, std::string>& additionalDefines);
+
+    MTLTexturePtr createMetalTexture(MTLTextureDescriptorPtr textureDescriptor) const;
+    MTLSamplerStatePtr createMetalSamplerState(MTLSamplerDescriptorPtr samplerDescriptor) const;
+
+    /*
```
louwers, line 52: I am assuming this comment block will be uncommented in a future PR.

include/mbgl/mtl/context.hpp
```diff
@@ -0,0 +1,157 @@
...
+    void setColorMode(const gfx::ColorMode&);
+    void setCullFaceMode(const gfx::CullFaceMode&);
+
+    void draw(const gfx::DrawMode&, std::size_t indexOffset, std::size_t indexLength);
+
+    void finish();
+*/
+
+    // Actually remove the objects we marked as abandoned with the above methods.
+    // Only call this while the OpenGL context is exclusive to this thread.
```
louwers, line 86:
```suggestion
    // Only call this while the Metal context is exclusive to this thread.
```

include/mbgl/mtl/texture2d.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+class Context;
+class RenderPass;
+
+class Texture2D : public gfx::Texture2D {
+public:
+    Texture2D(Context& context_);
+    ~Texture2D() override;
+
+public:
```
louwers, line 23: The `public` keyword is used three times in a row in this class, and `private` twice in a row.

include/mbgl/shaders/mtl/common.hpp
TimSylvester: Clever, but that's going to create a lot of changes.  I think it should wait until #1618 is merged, and Stefan's work with string indexing should make some of those obsolete, anyway.

include/mbgl/shaders/layer_ubo.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include <mbgl/util/color.hpp>
+
+#include <array>
+#include <cstdint>
+
+#define MLN_STRINGIZE(X) #X
```
louwers, line 8: I suggest we get rid of this if we decide we don't need the other macro.

include/mbgl/mtl/texture2d.hpp
TimSylvester: I believe @alexcristici's render target work makes major changes to `Texture2D`.

include/mbgl/mtl/context.hpp
TimSylvester: Yes, the target/framebuffer stuff should be implemented shortly.  I've removed some of that which definitely won't be replaced.

include/mbgl/mtl/buffer_resource.hpp
TimSylvester: Good catch, that's the wrong type anyway.

platform/ios/scripts/bazel-generate-plists.sh
```diff
@@ -82,4 +82,5 @@ cat > platform/ios/bazel/config.bzl <<EOF
 APPLE_MOBILE_PROVISIONING_PROFILE_TEAM_ID = "$teamid"
 APPLE_MOBILE_PROVISIONING_PROFILE_NAME = "iOS Team Provisioning Profile: *"
 BUNDLE_ID_PREFIX = "$bundleidprefix"
+BUILD_MODE = "xcode"
```
louwers, line 85:
```suggestion
BUILD_MODE = "bazel"
```

include/mbgl/shaders/mtl/common.hpp
louwers: OK sure!

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -0,0 +1,227 @@
...
+MLNMapViewMetalImpl::MLNMapViewMetalImpl(MLNMapView* nativeView_)
+    : MLNMapViewImpl(nativeView_),
+      mbgl::mtl::RendererBackend(mbgl::gfx::ContextMode::Unique),
+      mbgl::gfx::Renderable({ 0, 0 }, std::make_unique<MLNMapViewMetalRenderableResource>(*this)) {
+}
+
+MLNMapViewMetalImpl::~MLNMapViewMetalImpl() {
+    auto& resource = getResource<MLNMapViewMetalRenderableResource>();
+    /*if (resource.context && [[EAGLContext currentContext] isEqual:resource.context]) {
+        [EAGLContext setCurrentContext:nil];
```
louwers, line 117: Is this intentional?

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -0,0 +1,227 @@
...
+    MLNMapViewMetalImpl& backend;
+    mbgl::mtl::MTLCommandBufferPtr commandBufferPtr;
+
+public:
+    MLNMapViewImplDelegate* delegate = nil;
+    MTKView *mtlView = nil;
+    id<CAMetalDrawable> currentDrawable;
+    id <MTLCommandBuffer> commandBuffer;
+    id <MTLCommandQueue> commandQueue;
+    const bool atLeastiOS_12_2_0;
```
louwers, line 101: Is this even still used somewhere?

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -0,0 +1,227 @@
...
+        metalLayer.presentsWithTransaction = value;
+    }
+}
+
+void MLNMapViewMetalImpl::display() {
+    auto& resource = getResource<MLNMapViewMetalRenderableResource>();
+
+    // Calling `display` here directly causes the stuttering bug (if
+    // `presentsWithTransaction` is `YES` - see above)
+    // as reported in https://github.com/mapbox/mapbox-gl-native-ios/issues/350
```
louwers, line 140: This comment should be removed, because I doubt it applies for Metal.

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -0,0 +1,227 @@
...
+}
+
+void MLNMapViewMetalImpl::createView() {
+    auto& resource = getResource<MLNMapViewMetalRenderableResource>();
+    if (resource.mtlView) {
+        return;
+    }
+
+    /*if (!resource.context) {
+        resource.context = [[EAGLContext alloc] initWithAPI:kEAGLRenderingAPIOpenGLES3];
```
louwers, line 155: Can be removed?

src/mbgl/mtl/context.cpp
```diff
@@ -0,0 +1,206 @@
...
+namespace mtl {
+
+Context::Context(RendererBackend& backend_)
+    : gfx::Context(16), // TODO
+      backend(backend_),
+      stats() {}
+
+Context::~Context() noexcept {
+    /*
+        if (cleanupOnDestruction) {
```
louwers, line 36: Probably TODO?

src/mbgl/renderer/layers/circle_layer_tweaker.cpp
```diff
@@ -93,10 +113,11 @@ void CircleLayerTweaker::execute(LayerGroupBase& layerGroup,
 
     layerGroup.visitDrawables([&](gfx::Drawable& drawable) {
         auto& uniforms = drawable.mutableUniformBuffers();
-        uniforms.addOrReplace(CirclePaintParamsUBOName, paintParamsUniformBuffer);
-        uniforms.addOrReplace(CircleEvaluatedPropsUBOName, evaluatedPropsUniformBuffer);
+        uniforms.addOrReplace(MLN_STRINGIZE(CirclePaintParamsUBO), paintParamsUniformBuffer);
```
louwers, line 116: I am not sure what purpose this usage of `MLN_STRINGIZE` serves here. Can we get rid of it?

louwers (approved):

platform/ios/src/MLNMapView+Metal.mm
TimSylvester: I don't see any references to it, even in legacy/OpenGL

platform/ios/src/MLNMapView+Metal.mm
TimSylvester: Doesn't look familiar, @alexcristici?  I think it's just left over from basing the Metal impl on OpenGL, and can be dropped, as can every trace of global state management other than the Context.

alexcristici: Yes, we should remove it. 👍

src/mbgl/mtl/context.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    auto options = MTL::CompileOptions::alloc()->init();
+    options->setPreprocessorMacros(nsDefines);
+    options->setFastMathEnabled(true);
+    options->setLanguageVersion(MTL::LanguageVersion3_0);
+    options->setLibraryType(MTL::LibraryTypeExecutable);
+    options->setPreserveInvariance(true);
+    options->setOptimizationLevel(MTL::LibraryOptimizationLevelDefault);
+    options->setCompileSymbolVisibility(MTL::CompileSymbolVisibilityDefault);
+    options->setAllowReferencingUndefinedSymbols(false);
```
alexcristici, line 84: These 3 functions are available starting with iOS 16, so we should remove them. Or better we should remove the entire `options` block, I don't see where it is used.

alexcristici (approved):

src/mbgl/mtl/context.cpp
TimSylvester: Good catch on it not being used, that's just a bug.  We may want to make the optimization level available to improve initial load time, we just need to make it conditional on the target.

include/mbgl/gfx/vertex_attribute.hpp
```diff
@@ -104,6 +111,7 @@ class VertexAttribute {
 
     /// @brief Get the stride of the vertex attribute
     std::size_t getStride() const { return stride; }
+    void setStride(std::size_t value) { stride = value; }
```
mwilsnd, line 114: Need a static cast here, narrowing `value` from `size_t` to `int`

TimSylvester:
Looks like there's an issue with depth between extrusion and symbols here.

src/mbgl/mtl/layer_group.cpp
```diff
@@ -0,0 +1,60 @@
...
+        return;
+    }
+
+#if !defined(NDEBUG)
+    const auto debugGroup = uploadPass.createDebugGroup(getName() + "-upload");
+#endif
+
+    visitDrawables([&](gfx::Drawable& drawable) {
+        if (drawable.getEnabled()) {
+            auto& drawableGL = static_cast<Drawable&>(drawable);
```
mwilsnd, line 31: `drawableGL` should probably get renamed here to avoid confusion

mwilsnd (changes requested):

src/mbgl/mtl/texture2d.cpp
```diff
@@ -0,0 +1,188 @@
...
+    switch (channelType) {
+        case gfx::TextureChannelDataType::UnsignedByte:
+            switch (pixelFormat) {
+                case gfx::TexturePixelType::RGBA:
+                    return MTL::PixelFormat::PixelFormatRGBA8Unorm;
+                case gfx::TexturePixelType::Alpha:
+                    return MTL::PixelFormat::PixelFormatA8Unorm;
+                default:
+                    assert(false);
+                    throw std::runtime_error("Pixel format not supported on Metal!");
```
stefankarschti, line 84: `noexcept` method throws

stefankarschti (changes requested):

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -0,0 +1,210 @@
...
+
+    id<MTLDevice> device = (__bridge id<MTLDevice>)resource.getBackend().getDevice().get();
+
+    resource.mtlView = [[MTKView alloc] initWithFrame:mapView.bounds device:device];
+    resource.mtlView.delegate = resource.delegate;
+    resource.mtlView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    resource.mtlView.contentScaleFactor = contentScaleFactor();
+    resource.mtlView.contentMode = UIViewContentModeCenter;
+    resource.mtlView.colorPixelFormat = MTLPixelFormatBGRA8Unorm;
+    resource.mtlView.depthStencilPixelFormat = MTLPixelFormatDepth32Float_Stencil8;
```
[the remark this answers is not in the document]
TimSylvester: The only other combined option available for iOS doesn't have depth.

```
    MTLPixelFormatDepth24Unorm_Stencil8  API_AVAILABLE(macos(10.11), macCatalyst(13.0)) API_UNAVAILABLE(ios) = 255,
    MTLPixelFormatDepth32Float_Stencil8  API_AVAILABLE(macos(10.11), ios(9.0)) = 260,
    MTLPixelFormatX32_Stencil8  API_AVAILABLE(macos(10.12), ios(10.0)) = 261,
    MTLPixelFormatX24_Stencil8  API_AVAILABLE(macos(10.12), macCatalyst(13.0)) API_UNAVAILABLE(ios) = 262,

```

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -0,0 +1,210 @@
...
+    resource.mtlView.opaque = mapView.opaque;
+    resource.mtlView.layer.opaque = mapView.opaque;
+    resource.mtlView.enableSetNeedsDisplay = YES;
+    //resource.mtlView.clearColor = MTLClearColorMake(1,0,0,1);
+    
+    if (@available(iOS 13.0, *)) {
+        CAMetalLayer* metalLayer = MLN_OBJC_DYNAMIC_CAST(resource.mtlView.layer, CAMetalLayer);
+        metalLayer.presentsWithTransaction = NO;
```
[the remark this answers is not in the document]
TimSylvester: It looks like `setPresentsWithTransaction` is expected to only be called after `createView`, but it's easy enough to change that.

src/mbgl/mtl/context.cpp
```diff
@@ -0,0 +1,209 @@
...
+        &rawDefines[numDefines], rawDefines.data(), static_cast<NS::UInteger>(numDefines));
+    rawDefines.clear();
+
+    auto options = NS::TransferPtr(MTL::CompileOptions::alloc()->init());
+    options->setPreprocessorMacros(nsDefines);
+    options->setFastMathEnabled(true);
+
+    // TODO: We should specify the language version explicitly, but which one?
+    // v3.0 requires iOS 16
+    // options->setLanguageVersion(MTL::LanguageVersion3_0);
```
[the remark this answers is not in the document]
TimSylvester: Not sure what you mean by the shader header.  I don't see a benefit to putting it in `mbgl/gfx/shader.hpp`, `mbgl/shaders/mtl/shader_program.hpp`, etc..  I've just set it to `...2_1` inline for now, the latest version available on iOS 12.0, and made a few minor changes to make our shaders compatible with that version.

stefankarschti (approved):

src/mbgl/mtl/context.cpp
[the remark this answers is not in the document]
mwilsnd: I was thinking have it in the generated headers for each shader as we've already got other metadata stored there. Having it in a single place and forcing all shaders to conform is fine too.

mwilsnd (approved):

PR #1618, opened 2023-09-07
Rebase fill-extrusion work to eliminate dependence on render target branch

TimSylvester:


sjg-wdw (approved):

src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
```diff
@@ -293,10 +293,10 @@ void RenderFillExtrusionLayer::update(gfx::ShaderRegistry& shaders,
     }
 
     if (!fillExtrusionGroup) {
-        fillExtrusionGroup = shaders.getShaderGroup(std::string(FillExtrusionShaderName));
+        fillExtrusionGroup = shaders.getShaderGroup(MLN_STRINGIZE(FillExtrusionShader));
```
louwers, line 296: Why not

```suggestion
        fillExtrusionGroup = shaders.getShaderGroup("FillExtrusionShader"));
```

?

And in a few other places as well.

TimSylvester: My hope was to get compile-time validation, but haven't been able to achieve it.

src/mbgl/gfx/vertex_attribute.cpp
```diff
@@ -92,6 +92,13 @@ void VertexAttributeArray::resolve(const VertexAttributeArray& overrides, Resolv
     for (auto& kv : attrs) {
         delegate(kv.first, *kv.second, overrides.get(kv.first));
     }
+    // #if !defined(NDEBUG)
+    //     // Every override should match a defined attribute.
+    //     for (const auto& kv : overrides.attrs) {
+    //         const auto hit = attrs.find(kv.first);
+    //         assert(hit != attrs.end());
+    //     }
+    // #endif
```
stefankarschti, line 101: do we want to re-enable this at some point?

TimSylvester: Yeah, IIRC it fails on the missing pattern stuff.  It should really be `#if`d rather than commented, actually.

TimSylvester:
@stefankarschti Note changes to string indexer.  I added some overloads and comments so that string copies can be handled better later without changing the calls.  And `unordered_map` performs much better when initialized with capacity (avoiding numerous rebuilds as it's set up).  Probably an even larger number would be good.

include/mbgl/shaders/mtl/common.hpp
```diff
@@ -10,10 +10,20 @@ constexpr auto prelude = R"(
...
+
+#define SDF_PX 8.0
+
+// OpenGL `mod` is `x-y*floor(x/y)` where `floor` rounds down.
+// Metal `fmod` is `x-y*trunc(x/y)` where `trunc` rounds toward zero.
 template <typename T1, typename T2>
-inline auto mod(T1 x, T2 y) -> decltype(x - y * floor(x/y)) {
-    return x - y * metal::floor(x/y);
-}
+inline auto glMod(T1 x, T2 y) -> auto { return x - y * metal::floor(x/y); }
```
alexcristici, line 26: Changing the name to `glMod` will break new PRs that uses `mod` and not yet merged. I am ok with `mod` name, considering that is not a conflict with metal name, which is `fmod`. `mod` is consistent between Metal and GL.

alexcristici (approved):
One small comment.

include/mbgl/shaders/mtl/common.hpp
TimSylvester: IMO `mod` and `fmod` having different behavior is counterintuitive and should be avoided.

alexcristici: Yes, but `mod` and `fmod` have different names, so different behaviour.

alexcristici: Anyway, not so important, the PR is approved from my side.

TimSylvester: @louwers @mwilsnd Do you have an opinion?

mwilsnd: It looks like the difference is between the behavior of `floor()` and `trunc()`? I think being clear is a pro here, but perhaps being more explicit, with a comment attached.

Reading the metal specs, it's the rounding mode that is different - "negative infinity rounding mode" for `floor()` and "round toward zero rounding mode" for Metal. I'd include that info in a comment and make it clear that GL uses the `floor()` variant. The name doesn't matter too much, so long as it's clear what's happening and why.

TimSylvester: There's more context in the code.

https://github.com/maplibre/maplibre-native/pull/1618/files/8c46ea2141c81c0c3e5e8662f51a5e3efc8b73bc#diff-eb8e1270fc899b64368fc68f795b7bf297ef32892f1535fbd5ac8ce69320716c

mwilsnd: Yeah, then I'm fine with it being `glMod`. Merges for in-flight work should be trivial for this.

louwers (approved):

TimSylvester:
The [constructor](https://en.cppreference.com/w/cpp/container/unordered_map/unordered_map) provides "minimal number of buckets to use on initialization" while [reserve](https://en.cppreference.com/w/cpp/container/unordered_map/reserve) says "reserves at least the specified number of buckets" so, by my reading, they're setting the same thing.  Regardless, I have previously observed significant performance improvement by passing even relatively small values to the constructor, but going slightly over the final size would be best.

stefankarschti:
> The [constructor](https://en.cppreference.com/w/cpp/container/unordered_map/unordered_map) provides "minimal number of buckets to use on initialization" while [reserve](https://en.cppreference.com/w/cpp/container/unordered_map/reserve) says "reserves at least the specified number of buckets" so, by my reading, they're setting the same thing. Regardless, I have previously observed significant performance improvement by passing even relatively small values to the constructor, but going slightly over the final size would be best.

`void reserve( size_type count );`
"Sets the number of buckets to the number needed to accommodate at least _count elements_ ". I interpret that count is the number of elements, the resulting bucket_count is something else, depending on internal implementation.

Don't get me wrong, I agree on the general idea, I just think the means to achieve it through constructor's `bucket_count` may create an unreasonably large number of buckets - depending on implementation.

PR #1639, opened 2023-09-14
Simplify bazel xcode scripts

mwilsnd:
Simplify `bazel-xcodeproj.sh` and `bazel-package.sh` by moving plist generation entirely within bazel. Removes API key insertion and provisioning profile options from the scripts, moving all configuration to `example_config.bzl` instead.


TimSylvester:
So `config.bzl` won't be generated each time any more, we'll just set it up once per workspace?

TimSylvester (approved):

mwilsnd:
> So `config.bzl` won't be generated each time any more, we'll just set it up once per workspace?

Correct

alexcristici (approved):

louwers (approved):

PR #1638, opened 2023-09-14
Add checks for zero indexes on fill patterns to match fills.

TimSylvester:
Convert assertion on zero indexes to if-with-assert. Apply raster layer tweaker to main and image layer groups.

PR #1642, opened 2023-09-14
Don't error on missing artifact

louwers:
The iOS job sometimes doesn't run. Which means no artifact will be available.

The ios-device-test workflow uses the download artifact action, which errors when no artifact is available. This PR should fix that.

~Note that this PR will error because the ios-device-test workflow only runs the version already part of the main branch.~ I made a change to ios-ci to kick off the workflow.


TimSylvester (approved):

TimSylvester:
I think bringing in the `apt update` change would make this pass, but it shouldn't be needed any more.  I'd expect the changes to have gone to zero now that the same thing was merged in #1638.

louwers:
Seems to work... Let's close this and keep an eye on it.

PR #1643, opened 2023-09-17
Update minimum CMake version to 3.19

ntadej:
Update minimum CMake version to 3.19. Will need a separate PR to update to a container that is containing this CMake version but the one we have now should do it by then.

/cc @alanchenboy

### DIFF
--- a/.github/actions/qt5-build/Dockerfile
+++ b/.github/actions/qt5-build/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/maplibre/linux-builder:centos7-cmake3.10
+FROM ghcr.io/maplibre/linux-builder:centos7-cmake3.19
 
 # Copy and set the entry point
 COPY entrypoint.sh /entrypoint.sh
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -6,11 +6,11 @@ on:
     branches:
       - main
     tags:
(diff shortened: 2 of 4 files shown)


louwers (changes requested):
Looks like the Node.js workflow also needs to be updated.

ntadej:
> Looks like the Node.js workflow also needs to be updated.

Added a setup-cmake acction, but not sure if it will work on a self-hosted runner. Let's see.

louwers:
@ntadej Looks like it is not platform aware (self-hosted runner uses ARM64).

.github/actions/qt5-build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/maplibre/linux-builder:centos7-cmake3.10
+FROM ghcr.io/maplibre/linux-builder:centos7-latest
```
ntadej, line 1:
```suggestion
FROM ghcr.io/maplibre/linux-builder:centos7-cmake3.19
```

alanchenboy:
Thanks for this updating

ntadej:
@louwers, I think we need to just figure out how to update CMake in the Linux ARM runner? Can you just do it system-wide maybe?

louwers:
@ntadej Thanks for the heads up. Yes, I can do that.

louwers (approved):

PR #1644, opened 2023-09-18
String Indexer refactor

stefankarschti:
String indexer to use string_view map

### DIFF
--- a/src/mbgl/util/string_indexer.cpp
+++ b/src/mbgl/util/string_indexer.cpp
@@ -9,23 +9,12 @@ const std::string empty;
 constexpr std::size_t initialCapacity = 100;
 } // namespace
 
-StringIndexer::StringIndexer()
-    : stringToIdentity(initialCapacity) {
+StringIndexer::StringIndexer() {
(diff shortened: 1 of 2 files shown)


louwers (approved):

src/mbgl/util/string_indexer.cpp
```diff
@@ -49,7 +38,10 @@ StringIdentity StringIndexer::get(const std::string& string) {
         auto result = stringToIdentity.insert({string, id});
         if (result.second) {
             // this writer made the insert
-            identityToString.push_back(string);
+            identityToString.push_back(std::string(string));
+            stringToIdentity.erase(result.first);
+            result = stringToIdentity.insert({std::string_view(identityToString.back().data()), id});
```
TimSylvester, line 43: I could be missing something, but it seems like `find` would be better than `insert` and `erase`.

And really minor, but `string_view(const char*)` does a `strlen`, using the overload that takes the already-known length would be ever so slightly faster.

mwilsnd (approved):

alexcristici (approved):

Issue #1657, opened 2023-09-19
Metal Performance


stefankarschti:
Measure and improve metal render performance.
Initial performance is about 3.5 fps on iPad 5th gen.

stefankarschti:
Initial instrumentation shows significant time being spent in `mtl::ShaderProgram::getRenderPipelineState`

![Image](https://github.com/maplibre/maplibre-native/assets/4198736/4eeb66bd-d849-45d9-b3c5-5b4f698e1b01)

stefankarschti:
After caching the pipeline state:
```
Benchmark completed.
Result:
| boston     | 46.0 ms | 21.7 fps |
| paris      | 69.8 ms | 14.3 fps |
| paris2     | 52.0 ms | 19.2 fps |
| alps       | 39.0 ms | 25.7 fps |
| us east    | 37.9 ms | 26.4 fps |
| greater la | 49.2 ms | 20.3 fps |
| sf         | 83.8 ms | 11.9 fps |
| oakland    | 59.0 ms | 16.9 fps |
| germany    | 53.9 ms | 18.5 fps |
Average frame time: 54.5 ms
Average FPS: 18.3
```

PR #1658, opened 2023-09-19
Cache metal pipeline state

stefankarschti:
Improves #1657


src/mbgl/mtl/drawable.cpp
```diff
@@ -189,8 +189,11 @@ void Drawable::draw(PaintParameters& parameters) const {
     encoder->setCullMode(cullMode.enabled ? mapCullMode(cullMode.side) : MTL::CullModeNone);
     encoder->setFrontFacingWinding(mapWindingMode(cullMode.winding));
 
-    if (auto state = shaderMTL.getRenderPipelineState(renderPassDescriptor, impl->vertexDesc, getColorMode())) {
-        encoder->setRenderPipelineState(state.get());
+    if (!pipelineState) {
+        pipelineState = shaderMTL.getRenderPipelineState(renderPassDescriptor, impl->vertexDesc, getColorMode());
+    }
```
alexcristici, line 194: We need to recreate the pipelineState also if the `renderPassDescriptor`, `vertexDesc` or `getColorMode()` got changed.

alexcristici: I think instead caching the RenderPipelineState on drawable, a better place will be caching in ShaderProgram.

sjg-wdw: I'd agree with Tim here that Drawable or somewhere near that class is probably the right place.  Too many things to check in the program.  We can also just clear it out in Drawable if we change something.

I'll also add that we put it in Drawable in WhirlyGlobe and it never came up as an issue after that.

TimSylvester: We will still have to account for the render pass descriptor for clear colors, and if a drawable is allowed to be rendered to multiple targets.

alexcristici: Ok, we should cache it on the drawable. The remaining issue is that the vertex descriptor is changing to often because of the vertex attributes being dirty. Not sure if this is a bug, because it force us to recreate the render pipeline state to often.

sjg-wdw: In terms of what they intend for Metal, we're way off, but that's something we'll need to address incrementally by moving more work into the shaders and such.  In the short term, maybe we can make this work better.

Technically, we should only need to change the vertex descriptor object if the actual structure of the vertices changes.  That is, if we switch from a uniform to a vertex array, or something like that.  If we just put new data into one of the buffers for a vertex array, we don't need to update the vertex descriptor.

alexcristici: This is will a fix, to check if the structure changes. But the underlaying issue remains, is it normal that the vertex attributes change so often? Maybe this can be another issue that decrease the performance.

sjg-wdw: Right, that is the central issue.  The vertex attributes shouldn't change quite so often for our use case.  We can move most of that into the shaders.

But!  There are cases where the vertex attributes change constantly and we can make this path faster in Metal.  We just need to hold on to the vertex descriptors and update the buffers, rather than changing them each time.

sjg-wdw: They talk about triple buffering in the Metal docs.  I'm not a fan of that, personally.  I prefer to use a blit encoder instead.

You set up your new data for the existing buffers, by creating new buffers.  Then you copy from the new buffers to the old ones.  If you combine that with a MTLHeap it's cheap enough.

TimSylvester: Yeah, we need to differentiate between the structure and the content being different.  Adding an attribute would make the structure dirty, while the bucket changing the shared data just requires re-uploading that buffer.

That latter flag is maintained in `VertexBufferBase`/`IndexBufferBase`, but the `VertexAttribute`/`VertexAttributeArray` level doesn't track the former.

I think we can assume that the structure of the attribute blocks won't change for a drawable, the layer should produce a new one in that case, but we should at least check that with an assertion.

alexcristici (approved):

stefankarschti:
```
Result:
| boston     | 45.0 ms | 22.2 fps |
| paris      | 68.1 ms | 14.7 fps |
| paris2     | 50.0 ms | 20.0 fps |
| alps       | 37.6 ms | 26.6 fps |
| us east    | 36.3 ms | 27.5 fps |
| greater la | 47.1 ms | 21.2 fps |
| sf         | 80.9 ms | 12.4 fps |
| oakland    | 57.3 ms | 17.5 fps |
| germany    | 51.5 ms | 19.4 fps |
Average frame time: 52.6 ms
Average FPS: 19.0
```

sjg-wdw (approved):

PR #1660, opened 2023-09-19
Update Boost to 1.83, replace Boost Locale with deprecated codecvt STL

louwers:
I got rid of the submodule.


mwilsnd:
Boost locale has different platform implementations, including a WinAPI provider for Windows. It looks like these platform implementations are recent-ish (Our submodule version was 4 years out of date). These platform implementations aren't header-only, you need the src implementations to go with them.

I'd investigate using something like [boost-cmake](https://github.com/Orphis/boost-cmake) which lets you cherry-pick the various boost modules we require. It has an automated downloader for boost. We can also use [rules_boost](https://github.com/nelhage/rules_boost) for bazel.

louwers:
@mwilsnd I used the `scripts/update-boost.sh` script here.

Will updating to a slightly less recent Boost resolve the issue you think?

louwers:
```
        using Md = describe::describe_members<T, describe::mod_any_access>;

        mp11::mp_for_each<Md>([&](auto D){

            boost::hash_combine( r, v.*D.pointer );

        });
```

Results in:

```
/Users/ec2-user/actions-runner/_work/maplibre-native/maplibre-native/vendor/boost/include/boost/container_hash/hash.hpp:431:27: error: Use of undeclared identifier 'Md'
        mp11::mp_for_each<Md>([&](auto D){

```

ntadej:
Just saw this on the phone. ICU update may be tricky as Qt uses an older one by default.

Does the code still compile without any API changes? Then I guess we're fine (as we use system ICU).

louwers:
@ntadej I think ICU is stable as a rock. We'll see :)

PR #1662, opened 2023-09-19
[Windows] Update libwebp

mwilsnd:
Updates vcpkg to include `libwebp-1.3.2` which includes the OOB write fix.


louwers (approved):

etnav:
@louwers Can we do a new Node release? The last release was back in March for [v5.2.0](https://github.com/maplibre/maplibre-native/releases/tag/node-v5.2.0)

ntadej:
The problem is if we make it, it will not work on macOS.

louwers:
@etnav I am not a Node.js maintainer. Maybe @acalcutt wants to look into it, but it would be a (breaking) pre-release first I think.

etnav:
> @etnav I am not a Node.js maintainer. Maybe @acalcutt wants to look into it, but it would be a (breaking) pre-release first I think (without macOS support, unless we can get Metal to work already).

Understood, wasn't aware about the macOS part being "broken"!

PR #1666, opened 2023-09-19
Fix string indexer reallocation bug

stefankarschti:
corrupt strings result in case the string list grows over the initial capacity.
Refactor to use a compact buffer.


TimSylvester (approved):

src/mbgl/util/string_indexer.cpp
```diff
@@ -36,13 +39,25 @@ StringIdentity StringIndexer::get(std::string_view string) {
...
-            assert(result.second);
+            auto& buffer = instance().buffer;
+            const auto previousCapacity = buffer.capacity();
 
+            const StringIdentity id = identityToString.size();
+            identityToString.push_back(buffer.size());
+            buffer.insert(buffer.end(), string.begin(), string.end());
+            buffer.push_back(0);
+            if (buffer.capacity() != previousCapacity) {
+                // reallocation happened
```
mwilsnd, line 50: How often do we actually see reallocation happening here?

src/mbgl/util/string_indexer.cpp
```diff
@@ -51,20 +66,21 @@ StringIdentity StringIndexer::get(std::string_view string) {
...
 
-const std::string& StringIndexer::get(const StringIdentity id) {
+std::string_view StringIndexer::get(const StringIdentity id) {
     std::shared_lock<std::shared_mutex> readerLock(instance().sharedMutex);
 
     const auto& identityToString = instance().identityToString;
     assert(id < identityToString.size());
 
-    return id < identityToString.size() ? identityToString[id] : empty;
+    return id < identityToString.size() ? (instance().buffer.data() + identityToString[id]) : empty;
```
mwilsnd, line 75: Have you considered what might happen if the caller holds on to the returned `string_view` while another thread (or even the same one later on) triggers a reallocation of `buffer`?

It might be prudent to couple the read lock with the returned string_view to at least reduce such a possibility.

src/mbgl/util/string_indexer.cpp
stefankarschti: not that often, as `std::vector`'s default strategy afaik is exponential growth.
we could tune the `initialCapacity` so that it never happens, depending on the amount and length of strings to be indexed.

src/mbgl/util/string_indexer.cpp
stefankarschti: I see the issue, but I don't quite follow the solution. It would be prudent to return a copy?

TimSylvester: That's a good point, and is another argument for something else we talking about, just using fixed-sized, pre-allocated blocks that never change, at a small cost of unused memory.

There aren't many uses for getting strings by ID outside of debugging, though, so we could just return `std::string` with minimal performance effect.

stefankarschti: > using fixed-sized, pre-allocated blocks that never change, at a small cost of unused memory.

yeah, turning `buffer` into a vector of buffers would be a cool solution. if the vector of buffers uses move semantics to grow, that is :)

mwilsnd: Yes, a copy would be the safest option if you feel that is acceptable. I would consider the frequency at which strings are fetched from the buffer to determine if constructing `std::string`s is an acceptable performance trade-off.

mwilsnd:
Does get() making a copy impact the effectiveness of the original optimization intent?

alexcristici (approved):

stefankarschti:
> Does get() making a copy impact the effectiveness of the original optimization intent?

The previous implementation returned `const std::string&` and had the potential for a dangling reference when `identityToString` grew while the caller hung on to the string reference. No copy was made.

In MLN code, the cases when we need `get(id)` are the error cases, when a certain _uniform_ or _attribute_ name is needed to be logged, like so:
```c++
... 
else {
            using namespace std::string_literals;
            Log::Warning(Event::General,
                         "Vertex attribute type mismatch: "s + name + " / " +
                             StringIndexer::get(impl->idVertexAttrName));
            assert(false);
        }
```
 These are rare cases by definition and I think making a copy of the string is an acceptable compromise.

mwilsnd (approved):

TimSylvester:
Any places aside from error logging that do get strings by ID, like `RenderSymbolLayer::update`, can and should cache that value, or use the same string that was used to set up the ID value.  As a rule, results of using `StringIndexer` should be static or cached; if we're looking things up each time it will more than negate the benefits of faster key searches.

PR #1667, opened 2023-09-19
Try to improve ccache usage Android

louwers:
I think ccache does not work well the way we are using it now. The render test app takes over an hour to build.

I hope this improves it.

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -86,29 +86,9 @@ jobs:
         run: npm install --ignore-scripts
         working-directory: platform/android
 
-      - name: Install ccache
-        run: sudo apt-get install -y ccache
-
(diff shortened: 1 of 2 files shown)


ntadej:
I see no explicit reason why this would help. I suspect the main build is just not using `ccache` correctly. Although now I see that some cache was built. Maybe we retry the pipeline and see if it's much faster now.

louwers:
> Maybe we retry the pipeline and see if it's much faster now.

That's the plan :)

birkskyum (approved):

louwers:
It's 15 minutes quicker the second, but could be a fluke.

mwilsnd (approved):

louwers:
2h 25m 34s vs 75 minutes.

So I think it does something.

PR #1668, opened 2023-09-19
Issue/1648

### LINKED ISSUE #1648: Assert while cycling through styles
[This line asserts](https://github.com/maplibre/maplibre-native/blob/fb496dcc7c0f1ccc7fb07c8e5adefc004e9aeb5b/src/mbgl/renderer/render_orchestrator.cpp#L827C1-L827C36) while cycling through the Maptiler styles using the iOS App

TimSylvester:
Cleans up some stuff left over from when layer indexes were unique, and fixes the assertion by removing it, because that's not actually a bug.

Also fixes the build of the unit tests under drawable mode, though the test still fails.  (#1670)

Resolves #1648

### DIFF
--- a/test/util/offscreen_texture.test.cpp
+++ b/test/util/offscreen_texture.test.cpp
@@ -75,7 +75,11 @@ struct Buffer {
     GLuint buffer = 0;
 };
 
+#if MLN_LEGACY_RENDERER
 TEST(OffscreenTexture, RenderToTexture) {
+#else
+TEST(OffscreenTexture, DISABLED_RenderToTexture) {
+#endif
     if (gfx::Backend::GetType() != gfx::Backend::Type::OpenGL) {
         return;
     }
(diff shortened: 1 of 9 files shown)


test/util/offscreen_texture.test.cpp
```diff
@@ -167,7 +167,11 @@ void main() {
     test::checkImage("test/fixtures/offscreen_texture/render-to-fbo", image, 0, 0);
 
     // Now, composite the Framebuffer texture we've rendered to onto the main FBO.
+#if MLN_LEGACY_RENDERER
     gl::bindTexture(context, 0, {texture.getTexture().getResource(), gfx::TextureFilterType::Linear});
+#else
+    // TODO: Implement based on gfx::Texture2D
```
mwilsnd, line 173: Could you disable this test? Letting it pass without an implementation under test can end up being forgotten about.

louwers (approved):

mwilsnd (changes requested):

test/util/offscreen_texture.test.cpp
TimSylvester: That makes the test fail, so it shouldn't fall through the cracks.

stefankarschti (approved):

PR #1676, opened 2023-09-21
[MSVC] Use string view literals with StringIndexer

mwilsnd:
When building with MSVC, a linker issue can be observed with `StringIndexer::get`. This is apparently caused by string literals above the string SSO size failing to implicitly convert to `std::string_view`s. This PR brings `std::string_view_literals` in with `string_indexer.hpp` and calls to `get` with string literals are explicitly provided as string view literals (`"string"sv).

### DIFF
--- a/include/mbgl/util/string_indexer.hpp
+++ b/include/mbgl/util/string_indexer.hpp
@@ -9,6 +9,8 @@
 #include <mutex>
 #include <shared_mutex>
 
+using namespace std::string_view_literals;
+
 namespace mbgl {
 
 using StringIdentity = size_t;
--- a/src/mbgl/gfx/drawable_builder.cpp
+++ b/src/mbgl/gfx/drawable_builder.cpp
@@ -12,7 +12,7 @@ namespace gfx {
 
 DrawableBuilder::DrawableBuilder(std::string name_)
     : name(std::move(name_)),
-      vertexAttrNameId(StringIndexer::get("a_pos")),
+      vertexAttrNameId(StringIndexer::get("a_pos"sv)),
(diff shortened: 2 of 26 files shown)


sjg-wdw (approved):
I'm glad you're doing real time integration on Windows.  This would be an incredible headache later.

TimSylvester:
It seems likely that new code might fail to include `sv`.  Would the implicit conversion work better if we removed the bare `char*` overload?  Or at least produce a compile issue, thus forcing it on other platforms?

mwilsnd:
Is there a `char*` overload? I don't remember seeing one. I do agree it's likely the explicit `sv` will be forgotten sooner or later.

TimSylvester:
There was, but it looks like it already got removed.  So... maybe add it back to make the conversion explicit?  Or add `explicit` to the argument?

https://github.com/maplibre/maplibre-native/pull/1644/files

include/mbgl/util/string_indexer.hpp
```diff
@@ -9,6 +9,8 @@
 #include <mutex>
 #include <shared_mutex>
 
+using namespace std::string_view_literals;
```
TimSylvester, line 12: Better not to have this in the global namespace, but at least within `mbgl` or not in the header at all.

TimSylvester (approved):

mwilsnd:
Well today I can't reproduce the linker issue, closing this for now.

stefankarschti:
we could move from `string_view` to `const char*` parameter type and a few overloads to accept `std::string` / `string_view` maybe? I think this would work out of the box...

PR #1678, opened 2023-09-21
Bug/style-change

TimSylvester:
When the style is changed:

- Each layer is added as the style source is parsed
- Any new layers since the last frame are sorted into added/deleted/changed in the next frames `createRenderTree`
- Changed layers are updated, retaining the `RenderLayer` instance, and thus the existing drawables.
- The results of evaluated properties were then applied to drawables created previously, leading to incorrect results in some cases

With this PR:

- Layer tweakers are owned by render layers and individual drawables, and held weakly by layer groups
- When the style changes (`RenderLayer::layerChanged`), the render layer drops the reference to the tweaker, leaving it owned only by the existing drawables.
- For layers where the layer tweaker does not yet support property updates, the tweaker is re-created on calls to `evaluate`, but the reference on the corresponding drawables is updated.  This will become unnecessary as more tweakers support updating properties without re-creating the corresponding UBOs.
- Frame updates that occur while the tile is loading skip updating the old tiles with the new properties by comparing the tweaker reference on the drawable to that of the render layer.
- Once the async tile load is complete, a new bucket ID will be seen on the next frame, and the old drawables for that tile discarded, releasing the old layer tweaker.

Also fixes some related bugs:
- Builder flush didn't clear textures, so they could be carried over to the next tile
- Symbol UBOs weren't fully rebuilt on property change, only either text or icon were rebuilt, but both can be needed.

[drawable main](https://wetdogpublic.s3.us-west-2.amazonaws.com/share/drawable+main+RPReplay_Final1695404601.gif)
[drawable PR](https://wetdogpublic.s3.us-west-2.amazonaws.com/share/drawable+RPReplay_Final1695402429.gif)
[legacy PR](https://wetdogpublic.s3.us-west-2.amazonaws.com/share/legacy+RPReplay_Final1695403157.gif)
[metal PR](https://wetdogpublic.s3.us-west-2.amazonaws.com/share/metal+RPReplay_Final1695404111.gif)


sjg-wdw (approved):
Nice!

include/mbgl/renderer/layer_tweaker.hpp
```diff
@@ -78,14 +86,15 @@ class LayerTweaker {
 protected:
     std::string id;
     Immutable<style::LayerProperties> evaluatedProperties;
-    bool propertiesUpdated = true;
 
 #if MLN_RENDER_BACKEND_METAL
     // For Metal, whether a property is provided through attribtues or uniforms is specified in
     // a uniform buffer rather than by a shader compiled with different preprocessor definitions.
     std::vector<std::string> propertiesAsUniforms;
```
stefankarschti, line 93: This could benefit from `StringIndexer`, and could become a `std::unordered_set<StringIdentity>` and that could help with performance in `LayerTweaker::hasPropertyAsUniform` which is called per frame to build the `permutationUBO`

TimSylvester: This PR doesn't add `propertiesAsUniforms`, so I'd suggest that should be in a separate one.  
That optimization should apply all the way through, including the return of `readDataDrivenPaintProperties` and the arguments of `getOrCreateShader`

stefankarschti: Correct. I created an issue for that. #1687

stefankarschti (approved):

mwilsnd:
I see the following when flipping between two styles (FB Dark, 3D), looks like fill layer doesn't update correctly:
[mbgl-glfw_dBb0c7LWXV.webm](https://github.com/maplibre/maplibre-native/assets/53413200/6b30e466-6558-4712-8f16-132a7798f3b2)

In the clip, I can get the style to apply correctly by re-loading it. I move between both dark and light styles this way to better show the problem.

TimSylvester:
@mwilsnd I'm getting what seem like the same visual result during the transition, but yours never seems to complete and load some new layers.  Do you see anything being logged?

Could it be that the iOS app is rendering more frames on style switch, and that last frame is needed here?  If you pan or anything does it clear up?



If I switch rapidly now I've been getting an assertion in building the icon atlas that doesn't seem like anything I changed.

Screenshot 2023-09-25 at 4 07 19 PM
Screenshot 2023-09-25 at 4 07 41 PM

mwilsnd:
The issue is present in your image too - the fill pattern for the grass / park land use disappears after the first style change and never returns. The dark mode map is also missing the fill pattern. Look near the end of the video clip I posted where I switch from dark to light, then re-load the light style a second time. Calling `Style::loadURL` with the same style twice after having loaded a different style (dark in my example) correctly loads everything.

I did try panning around suspecting it might be work deferred to a pending frame but no such luck.

Here on the left is the style correctly loaded, on the right is after cycling to the dark style and then back again once. In this state re-loading the same style should correctly load the missing features.
image image

TimSylvester:
There seems to be a deeper problem here, in that `RenderOrchestrator::createRenderTree` removes the images before doing anything with the layers.

Issue #1677, opened 2023-09-21
Elide unnecessary raster layer buffer upload


stefankarschti:
The raster layer should update drawable's vertex vector in place, and only if it changed in the `bucket`.

stefankarschti:
Also bind the same texture in both locations.

PR #1680, opened 2023-09-22
Raster Layer fixes 

stefankarschti:
Addresses #1677 
Use just 1 texture, bind in both locations, update drawable's vertex vector in place.


TimSylvester:
I'll see about pulling this into #1678, it's all the same code changing there.

The sampler locations can be stored with the shader reference as well.

TimSylvester (approved):

TimSylvester:
This should be subsumed by #1678 now, including sharing the textures between bindings, caching the indexes, and only rebuilding drawables when the bucket or vertices change.

stefankarschti:
> This should be subsumed by #1678 now, including sharing the textures between bindings, caching the indexes, and only rebuilding drawables when the bucket or vertices change.

Closing this PR as it's included into #1678.

Issue #687, opened 2023-01-12
Replace xcpretty by xcbeautify


louwers:
[xcpretty](https://github.com/xcpretty/xcpretty) has not been updated for over 4 years and should probably be replaced at some point.

Active successor seems to be [xcbeautify](https://github.com/tuist/xcbeautify).

Note to self: M1 build runners need to be updated as well. #485

PR #1683, opened 2023-09-23
Use xbeautify instead of xcpretty

SiarheiFedartsou:
closes https://github.com/maplibre/maplibre-native/issues/687


louwers (commented):
All the scripts you changed in this PR could actually be removed, because we are not using them anymore.

The pre-release workflow is also on the way out.

PR #1686, opened 2023-09-24
Re-enable cmake-format, format Bazel files

louwers:
Uses a previous [.cmake-format](https://github.com/mapbox/mapbox-gl-native/blob/master/.cmake-format) but with `algorithm_order` and `max_subargs_per_line` options removed (as they have been removed from `cmake-format`).

See [0.6.0 release notes](https://cmake-format.readthedocs.io/en/latest/release_notes.html).

> Additionally, cmake-format will now tend to prefer a normal “horizontal” wrap for relatively long lists of positional arguments (e.g. source files in add_library) whereas it would previously prefer a vertical layout (one-entry per line). This is a consequence of an ambiguity between which positional arguments should be vertical versus which should be wrapped.

I want to clean up the CMake build config a bit soon, so this seemed like a good first step.

Also, since I ran `pre-commit run --all-files`, it turns out that some Bazel files slipped by the eye of pre-commit. So those are now formatted too.


platform/glfw/CMakeLists.txt
```diff
@@ -16,12 +16,9 @@ endif()
 
 add_executable(
     mbgl-glfw
-    ${PROJECT_SOURCE_DIR}/platform/glfw/main.cpp
-    ${PROJECT_SOURCE_DIR}/platform/glfw/glfw_view.cpp
-    ${PROJECT_SOURCE_DIR}/platform/glfw/glfw_renderer_frontend.cpp
-    ${PROJECT_SOURCE_DIR}/platform/glfw/settings_json.cpp
-    ${PROJECT_SOURCE_DIR}/platform/glfw/test_writer.cpp
-    ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/map/map_snapshotter.cpp
+    ${PROJECT_SOURCE_DIR}/platform/glfw/main.cpp ${PROJECT_SOURCE_DIR}/platform/glfw/glfw_view.cpp
```
louwers, line 19: This does not look right to me.

louwers:
Looks like the `always_wrap` option is bugged https://github.com/cheshirekow/cmake_format/issues/180

Also, I'm not sure if cmake-format is maintained anymore...

There is an alternative, but it does not seem to do sorting: https://github.com/BlankSpruce/gersemi

PR #1690, opened 2023-09-26
Metal headless backend

alexcristici:
Implemented headless backend on Metal.
Implemented collision shaders.


louwers (approved):

stefankarschti (approved):

platform/default/include/mbgl/mtl/headless_backend.hpp
```diff
@@ -0,0 +1,34 @@
...
+    void updateAssumedState() override;
+    gfx::Renderable& getDefaultRenderable() override;
+    PremultipliedImage readStillImage() override;
+    RendererBackend* getRendererBackend() override;
+    SwapBehaviour getSwapBehaviour();
+
+private:
+    void activate() override;
+    void deactivate() override;
+    ;
```
mwilsnd, line 26: nit: Extra semicolon

src/mbgl/mtl/offscreen_texture.cpp
```diff
@@ -41,8 +41,13 @@ class OffscreenTextureResource final : public RenderableResource {
     }
 
     PremultipliedImage readStillImage() {
-        assert(false);
-        return PremultipliedImage();
+        auto data = std::make_unique<uint8_t[]>(texture->getDataSize());
+        MTL::Region region = MTL::Region::Make2D(0, 0, size.width, size.height);
+        NS::UInteger bytesPerRow = size.width * texture->getPixelStride();
+
+        static_cast<Texture2D*>(texture.get())->getMetalTexture()->getBytes(data.get(), bytesPerRow, region, 0);
```
mwilsnd, line 48: Does this offer implicit synchronization (like GL) or do we need a read barrier?

mwilsnd (approved):

src/mbgl/mtl/offscreen_texture.cpp
sjg-wdw: In general Metal doesn't do much of anything implicitly so it probably depends on what we're using this for.
If we're using it to read back an offscreen render buffer, then we want to (and are probably already) waiting for the frame to finish so this would be fine.  In that use case.

include/mbgl/shaders/collision_layer_ubo.hpp
```diff
@@ -14,8 +14,8 @@ struct alignas(16) CollisionUBO {
 static_assert(sizeof(CollisionUBO) % 16 == 0);
 static_assert(sizeof(CollisionUBO) == 80);
 
-typedef CollisionUBO CollisionBoxUBO;
-typedef CollisionUBO CollisionCircleUBO;
+using CollisionUBO CollisionBoxUBO;
```
mwilsnd, line 17: syntax: `using CollisionBoxUBO = CollisionUBO;`

alexcristici: Yes, sure. I was focused on something else when I corrected this. Now should be fine. 😄

TimSylvester (approved):

mwilsnd (approved):

PR #1695, opened 2023-09-27
Merge branch 'topic/metal-sep21' into main

mwilsnd:
Merging the Sep. 21 feature branch back into main. `render_hillshade_layer.cpp/h` had conflicts, ensure this has been merged correctly before approving.

### DIFF
--- a/include/mbgl/gfx/drawable.hpp
+++ b/include/mbgl/gfx/drawable.hpp
@@ -18,7 +18,10 @@ namespace mbgl {
 
 class Color;
 class PaintParameters;
+class LayerTweaker;
 enum class RenderPass : uint8_t;
+
+using LayerTweakerPtr = std::shared_ptr<LayerTweaker>;
 
 namespace gfx {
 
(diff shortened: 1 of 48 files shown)


louwers (approved):

stefankarschti (commented):
I get an [assertion here](https://github.com/maplibre/maplibre-native/blob/9ae0d8f0e41c6fc1800c32020e629a633b976cca/src/mbgl/mtl/drawable.cpp#L524C9-L524C39) after panning the map a few times with `Satellite Hybrid` style over Boston (_Metal_ flavor). It takes several style changes before reaching `Satellite Hybrid`.

On _OpenGL/drawable_ it doesn't happen.

TimSylvester:
We ran into that one with fill, I think, when the bucket produced empty output.  But that wasn't specific to Metal.  Probably it has to do with not keeping the data around after a buffer is created for it.  I'll take a look...

TimSylvester:
Ah, `RasterBucket` is changing the vertices and indices in the shared data, and sometimes updating them to empty, but the raster layer isn't considering that.

TimSylvester:
Okay, I think I have that figured out.  It doesn't show up on GL because it's just not being checked, but it's still a potential problem.

It's enough that I created #1700 for it.  Unfortunately, I seem to have introduced a symbol UBO problem in the process.

stefankarschti:
In _legacy_ code the missing indexes on raster `bucket` signals to use the raster quad (`staticDataVertices`, `staticDataIndices` & `staticDataSegments`), so a `drawable` without indexes shouldn't be created in any case.

alexcristici (approved):

TimSylvester (approved):

stefankarschti (approved):

Issue #1692, opened 2023-09-26
Builder thick line implementation


stefankarschti:
Enable the Builder to create thick lines (drawables) by using the current implementation from LineBucket.

stefankarschti:
![image](https://github.com/maplibre/maplibre-native/assets/4198736/21654f86-a37a-439a-be90-f9f7e9303553)

stefankarschti:
the _user_ code fragment:
https://github.com/maplibre/maplibre-native/blob/b92982296aa11fbd77d46f1259d8f9d5970ef0ae/src/mbgl/renderer/renderer_impl.cpp#L397C1-L501C1

stefankarschti:
Now available from the `App` iOS application via the "Add Custom Drawable Layer" menu item in the "Runtime Styling" category.

![Image](https://github.com/maplibre/maplibre-native/assets/4198736/90460f12-b173-4b7a-a846-3f2381cf23b7)

PR #1699, opened 2023-09-27
Custom Drawable Layer v1 + DrawableBuilder thick lines

stefankarschti:
#1692 Use `LineBucket` geometry generator implementation into `DrawableBuilder::Impl`. Introduce a method to produce `Drawable`s capable of drawing thick lines.

Implements the first version of the Custom Drawable Layer as described in #1735.

Use the Builder / Drawable thick line functionality externally.


platform/default/src/mbgl/layermanager/layer_manager.cpp
```diff
@@ -69,6 +70,9 @@ LayerManagerDefault::LayerManagerDefault() {
     addLayerType(std::make_unique<LocationIndicatorLayerFactory>());
 #endif
 #endif
+#if !defined(MBGL_LAYER_CUSTOM_DRAWABLE_DISABLE_ALL)
```
louwers, line 73: Let's use `MLN_` as a prefix for new stuff.

louwers (changes requested):
The duplicated code between `drawable_builder_impl.cpp` and `line_bucket.cpp` should be refactored.

stefankarschti:
> The duplicated code between `drawable_builder_impl.cpp` and `line_bucket.cpp` should be refactored.

good point. I'll extract it in a common place for both.

src/mbgl/gfx/drawable_builder.cpp
```diff
@@ -252,7 +270,27 @@ void DrawableBuilder::addTriangles(const std::vector<uint16_t>& indexes,
...
+void DrawableBuilder::addPolyline(const GeometryCoordinates& coordinates,
+                                  const util::PolylineGeneratorOptions& options) {
+    // mark the current mode
+    if (!checkAndSetMode(Mode::Polylines)) return;
+
+    // append polyline
+    impl->addPolyline(*this, coordinates, options);
+}
+
+bool DrawableBuilder::checkAndSetMode(Mode target) {
```
louwers, line 285: Can the mode be an implementation detail? I.e. set it in `addPolyline`?

CMakeLists.txt
```diff
@@ -943,6 +945,7 @@ list(APPEND SRC_FILES
     ${PROJECT_SOURCE_DIR}/src/mbgl/util/mat4.cpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/util/mat4.hpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/util/math.hpp
+    ${PROJECT_SOURCE_DIR}/src/mbgl/util/polyline_generator.cpp
```
louwers, line 948: Thanks for factoring out this code!

Question: we have quite a bit of different modules in `util` already. Would adding it to another existing module or creating a new module/directory be a better fit?

include/mbgl/util/polyline_generator.hpp
```diff
@@ -0,0 +1,103 @@
...
+public:
+    PolylineGenerator(Vertices& polylineVertices,
+                      LayoutVertexFunc layoutVertexFunc,
+                      Segments& polylineSegments,
+                      CreateSegmentFunc createSegmentFunc,
+                      GetSegmentFunc getSegmentFunc,
+                      Indexes& polylineIndexes);
+    ~PolylineGenerator() = default;
+
+    void generate(const GeometryCoordinates& coordinates, const PolylineGeneratorOptions& options);
```
louwers, line 67: Is there a way we can make the API more functional?

src/mbgl/renderer/layers/render_custom_drawable_layer.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+private:
+    void transition(const TransitionParameters&) override {}
+    void evaluate(const PropertyEvaluationParameters&) override;
+    bool hasTransition() const override;
+    bool hasCrossfade() const override;
+    void markContextDestroyed() override;
+    void prepare(const LayerPrepareParameters&) override;
+
+#if MLN_LEGACY_RENDERER
```
louwers, line 31: Is this layer used or included when compiling the legacy renderer?

include/mbgl/renderer/layer_tweaker.hpp
```diff
@@ -71,20 +71,20 @@ class LayerTweaker {
 
     void enableOverdrawInspector(bool);
 
-    virtual void execute(LayerGroupBase&, const RenderTree&, const PaintParameters&) = 0;
+    virtual void execute(LayerGroupBase&, const PaintParameters&) = 0;
 
     void updateProperties(Immutable<style::LayerProperties>);
 
-protected:
+    // protected:
```
louwers, line 78: Line can be removed.

src/mbgl/gfx/drawable_builder.cpp
```diff
@@ -101,6 +110,9 @@ void DrawableBuilder::setTexture(const std::shared_ptr<gfx::Texture2D>& texture,
 }
 
 void DrawableBuilder::addTriangle(int16_t x0, int16_t y0, int16_t x1, int16_t y1, int16_t x2, int16_t y2) {
+    // check and set the current mode
+    if (!checkAndSetMode(Mode::Primitives)) return;
```
louwers, line 114: Should some logging be added in these cases?

platform/default/src/mbgl/layermanager/layer_manager.cpp
```diff
@@ -69,6 +73,12 @@ LayerManagerDefault::LayerManagerDefault() {
     addLayerType(std::make_unique<LocationIndicatorLayerFactory>());
 #endif
 #endif
+
+#if MLN_DRAWABLE_RENDERER
+#if !defined(MLN_LAYER_CUSTOM_DRAWABLE_DISABLE_ALL)
```
louwers, line 78: This is the only place where this define is used.

Should we include it by default for now, and do a pass later to allow people to not include this functionality?

include/mbgl/util/polyline_generator.hpp
```diff
@@ -0,0 +1,103 @@
...
+    uint32_t overscaling{1}; // TODO: what is this???
+    std::optional<PolylineGeneratorDistances> lineDistances;
+};
+
+template <class PolylineLayoutVertex, class PolylineSegment>
+class PolylineGenerator {
+public:
+    using Vertices = gfx::VertexVector<PolylineLayoutVertex>;
+    using Segments = std::vector<PolylineSegment>;
+    using LayoutVertexFunc = std::function<PolylineLayoutVertex(
```
louwers, line 52: The allocation overhead for std::function can be avoided by adding these as template arguments to `generate`.

include/mbgl/util/polyline_generator.hpp
louwers: I.e. replace the whole class with a function.

CMakeLists.txt
stefankarschti: yeah, I thought about it, maybe `include/mbgl/gfx` and `src/mbgl/gfx` would be better? Since it's platform agnostic graphics code...

include/mbgl/util/polyline_generator.hpp
stefankarschti: we could, but that would be a function with a lot of parameters - `template` + `constructor` parameters + `generate` parameters.

CMakeLists.txt
louwers: Yes sounds good!

src/mbgl/renderer/layers/render_custom_drawable_layer.hpp
stefankarschti: no, it should not. there's no dependency to it from _legacy_. I'm rechecking build files.

platform/default/src/mbgl/layermanager/layer_manager.cpp
stefankarschti: I followed the pattern used for the other layer types. it can be done either way, let's discuss about it.

src/mbgl/gfx/drawable_builder.cpp
stefankarschti: I moved the mode checking into `DrawableBuilder::Impl` and logged an error there if the expected `mode` doesn't match the `target`.

platform/default/src/mbgl/layermanager/layer_manager.cpp
louwers: Discussed!

include/mbgl/util/polyline_generator.hpp
stefankarschti: discussed

sjg-wdw (approved):
A big step in the right direction for reusability.

louwers (approved):
Thanks for addressing the duplication!

For this to be widely used I think we will need some more documentation comments and ideally examples. But I think we want this v1 to be merged first and see how much more complexity can be hidden.

platform/darwin/src/MLNCustomDrawableStyleLayer.mm
```diff
@@ -0,0 +1,279 @@
...
+- (void)setNeedsDisplay {
+    [self.mapView setNeedsRerender];
+}
+
+@end
+
+using namespace mbgl;
+using namespace mbgl::style;
+
+class TestDrawableTweaker : public gfx::DrawableTweaker {
```
mwilsnd, line 84: This appears to be duplicated in `test/api/custom_drawable_layer.test.cpp`. If this is intended to be used in a test, it should be relocated to `platform/darwin/test`.

platform/darwin/src/MLNCustomDrawableStyleLayer.mm
```diff
@@ -0,0 +1,279 @@
...
+    void initialize() {
+        if (layerRef == nil) return;
+        else if (layer == nil) layer = layerRef;
+        
+        if (layer.mapView) {
+            [layer didMoveToMapView:layer.mapView];
+        }
+    }
+    
+    void update(mbgl::RenderLayer& proxyLayer,
```
mwilsnd, line 178: More test code, see above

platform/windows/vendor/vcpkg
mwilsnd, line 1: This is reverting the version of `libwebp` used by Windows.

mwilsnd (changes requested):

platform/windows/vendor/vcpkg
stefankarschti: this is unintentional. I'm reverting.

mwilsnd (approved):

platform/darwin/src/MLNCustomDrawableStyleLayer.mm
stefankarschti: yes, this is similar to `test/api/custom_layer.test.cpp` versus `platform/darwin/src/MLNOpenGLStyleLayer.mm` + `platform/darwin/app/LimeGreenStyleLayer.m". 
As far as I can see, the former is meant to run as a platform independent test, while the latter has dependencies to the `darwin` platform and is meant to support the _demo_ in the iOS App. 
I'm refactoring to remove code duplication, but as I'm devising a friendlier API to be used in subsequent PRs and the whole _user code_ areas will change.

stefankarschti: I am relocating `platform/darwin/src/MLNCustomDrawableStyleLayer.mm` to `platform/darwin/app` as it is essentially example code, to be demonstrated by the `App`

TimSylvester (approved):

alexcristici (approved):

platform/darwin/src/MLNCustomDrawableStyleLayer.mm
stefankarschti: this looks to be much more work than I anticipated, a whole new objective C interface is needed. I'm cutting this PR here with the note that platforms are not yet supported.

stefankarschti: I'm removing the code from this branch and we'll expose the functionality at platform level when it's ready. 
the API test remains, since that uses the core directly through C++.

PR #1700, opened 2023-09-27
Fix raster geometry updates

TimSylvester:
Handle the case where the raster bucket changes the geometry after a drawable is created from it.

Allow index vector to be empty if there's a buffer Skip drawable create in some cases where there's no data to draw. Reset buffers as well as values when clearing index/vertex containers.  Add more validation. Fixed a bug in layer tweaker update that prevented it from working.

### DIFF
--- a/include/mbgl/renderer/layer_tweaker.hpp
+++ b/include/mbgl/renderer/layer_tweaker.hpp
@@ -93,9 +93,12 @@ protected:
     std::vector<std::string> propertiesAsUniforms;
 #endif // MLN_RENDER_BACKEND_METAL
 
+    // Indicates that the evaluated properties have changed
     bool propertiesUpdated = true;
 
-    bool propertiesChanged = true;
+    // Indicates that the properties-as-uniforms has changed
+    bool permutationUpdated = true;
+
     bool overdrawInspector = false;
 };
 
--- a/src/mbgl/renderer/layers/symbol_layer_tweaker.cpp
+++ b/src/mbgl/renderer/layers/symbol_layer_tweaker.cpp
@@ -90,7 +90,7 @@ void SymbolLayerTweaker::execute(LayerGroupBase& layerGroup,
     const auto zoom = parameters.state.getZoom();
 
 #if MLN_RENDER_BACKEND_METAL
-    if (propertiesChanged) {
+    if (permutationUpdated) {
         const SymbolPermutationUBO permutationUBO = {
             /* .fill_color = */ {/*.source=*/getAttributeSource("a_fill_color"), /*.expression=*/{}},
             /* .halo_color = */ {/*.source=*/getAttributeSource("a_halo_color"), /*.expression=*/{}},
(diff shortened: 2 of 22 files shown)


stefankarschti:
In _legacy_ code the missing indexes on raster `bucket` signals to use the raster quad (`staticDataVertices`, `staticDataIndices` & `staticDataSegments`), so a `drawable` without indexes shouldn't be created in any case.

This condition can vary back & forth during the program execution, the bucket generates buffers only when it decides to clip visible tiles, otherwise leaves them empty so the standard quad is used. [Legacy code here.](https://github.com/maplibre/maplibre-native/blob/9ae0d8f0e41c6fc1800c32020e629a633b976cca/src/mbgl/renderer/layers/render_raster_layer.cpp#L194)

TimSylvester:
The problem is that a drawable is created with masked vertices in the shared data, but the bucket later changes that, or clears the shared vertices.  The segments aren't shared, though, so that won't work.  Or the drawable is set up with the static vertices and the bucket later updates to masked shared vertices which has no effect on the drawable at all.

I considered emplacing the static values into the shared buffers and updating the segments instead, but changing the geometry in a drawable will lead to problems later on.

As an improvement, we should find a way to transfer the textures over, but that's not easy any more.

TimSylvester:
Turns out the buffers can't be reset when `VertexVector`/`IndexVector` is cleared, or it breaks OpenGL.  I believe this indicates a potential bug, in that `SymbolBucket` is changing the values dynamically (`updateBucketOpacities`), but the resulting dirty vectors are not being re-applied in the GL implementation.  I'll create an issue so it doesn't get lost.

src/mbgl/renderer/layers/symbol_layer_tweaker.cpp
```diff
@@ -131,7 +131,7 @@ void SymbolLayerTweaker::execute(LayerGroupBase& layerGroup,
         const auto& symbolData = static_cast<gfx::SymbolDrawableData&>(*drawable.getData());
         const auto isText = (symbolData.symbolType == SymbolType::Text);
 
-        if (isText && (textPaintBuffer || textPropertiesUpdated)) {
+        if (isText && (!textPaintBuffer || textPropertiesUpdated)) {
```
TimSylvester, line 134: Note: This was causing a UBO to be updated every time, and should improve performance.

include/mbgl/renderer/layer_tweaker.hpp
```diff
@@ -93,9 +93,12 @@ class LayerTweaker {
     std::vector<std::string> propertiesAsUniforms;
 #endif // MLN_RENDER_BACKEND_METAL
 
+    // Indicates that the evaluated properties have changed
     bool propertiesUpdated = true;
 
-    bool propertiesChanged = true;
```
stefankarschti, line 98: `propertiesChanged` is still referenced. I get a compile error in `LineLayerTweaker` / metal.

TimSylvester: Forgot to try Metal, sorry.

stefankarschti (approved):
works nicely!

mwilsnd (approved):

PR #1702, opened 2023-09-29
Implement metal line sdf shader.

alanchenboy:
Implement metal line sdf shader.

some effects:
Simulator Screen Shot - iPhone 14 Plus - 2023-09-29 at 17 28 57

Simulator Screen Shot - iPhone 14 Plus - 2023-09-29 at 16 57 53


TimSylvester:
You should rebase this on `topic/metal-sep25`, there are some significant changes to the shader template definition.  The string IDs are looked up at init now, so the definitions aren't `constexpr` any more and can't be inline in the header.  I can make the changes if you'd like.

TimSylvester:
See #1730

PR #1713, opened 2023-09-30
[ci] Disable broken tests in OpenGL 2 branch

ntadej:
Disable broken macOS tests in the OpenGL 2 branch.


ntadej:
This makes the macOS test basically useless, but I'll keep it running so we are aware it is in principle supposed to run.

ntadej:
With help of @louwers the macOS tests are now as fixed as it makes sense.

PR #1720, opened 2023-10-03
Fix Drawable / Legacy renderer matrix Linux CI

louwers:
@stefankarschti 

Getting this assertion error now that the Linux tests run for the Drawable path. 

> mbgl-test-runner: /home/runner/work/maplibre-native/maplibre-native/src/mbgl/gl/drawable_gl.cpp:130: void mbgl::gl::DrawableGL::bindUniformBuffers() const: Assertion `false' failed.


src/mbgl/renderer/render_layer.cpp
```diff
@@ -186,7 +186,8 @@ std::size_t RenderLayer::removeAllDrawables() {
 }
 
 void RenderLayer::updateRenderTileIDs() {
-    const auto oldMap = std::move(renderTileIDs);
+    const auto oldMap = renderTileIDs;
```
stefankarschti, line 189: this copies the map instead of moving. there's a performance penalty.

stefankarschti:
> @stefankarschti
> 
> Getting this assertion error now that the Linux tests run for the Drawable path.
> 
> > mbgl-test-runner: /home/runner/work/maplibre-native/maplibre-native/src/mbgl/gl/drawable_gl.cpp:130: void mbgl::gl::DrawableGL::bindUniformBuffers() const: Assertion `false' failed.

The `StringIndexer` tests corrupt the global string list, because the `StringIndexer` is a singleton. Then the core is incorrectly using indexes that have been replaced and attempts to use them for UBO retrieval. 

To move ahead please disable `StringIndexer.*` tests temporarily, while I work on a proper fix.

src/mbgl/renderer/render_layer.cpp
louwers: I can change it, but then I have to re-initialize the data member, otherwise it is used after being moved.

stefankarschti (approved):

alexcristici (approved):

Issue #1634, opened 2023-09-13
Android Annotations: add click and drag listeners


fynngodau:
[Android annotations API proposal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md) implementation issue as a part of https://github.com/maplibre/maplibre-native/issues/1491

---

Add click, drag, and long-click listeners.

* Kotlinify `DraggableAnnotationController`.
* Items should be draggable if `draggable` is set to `true`.
* The click and long-click listeners addable to the individual markers should work.
* Tests for `DraggableAnnotationListener` should pass.

louwers:
This ticket is in scope of the Modernization bounty direction. https://github.com/maplibre/maplibre/issues/270

louwers:
This bounty can be paid out.

fynngodau:
Included in https://opencollective.com/maplibre/expenses/193947.

Issue #1494, opened 2023-08-08
Android Annotations: automatic grouping


fynngodau:
[Android annotations API proposal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md) implementation issue as a part of https://github.com/maplibre/maplibre-native/issues/1491

---

Add automatic grouping by non data-driven properties and by Z layer. For this, the non data-driven properties themselves need to be added in the first place.

* Add non data-driven properties
* Add automatic grouping into manager objects by Z layer and by non data-driven properties as specified in the proposal
* Grouping should run in O(n log n) with n as the amount of annotations
* Add support for cleaning up no longer needed bitmaps

louwers:
This ticket is in scope of the Modernization bounty direction. https://github.com/maplibre/maplibre/issues/270

fynngodau:
On second thought, adding the grouping algorithm that I had in mind adds a large amount of code for possibly little benefit, as the grouping algorithm in Kotlin's `groupBy` method has a bad runtime only if there _lots_ of different groups are spawned. I think we should probably delay asymptotically more efficient grouping until all things are in place.

This would that assigning a different z layer value to each symbol is not a supported use case at first. What do you think @louwers?

louwers:
Sure. We can always swap out the algorithm later.

louwers:
This bounty can be paid out.

fynngodau:
Included in https://opencollective.com/maplibre/expenses/193947.

Issue #1493, opened 2023-08-08
Android Annotations: add new classes (1)


fynngodau:
[Android annotations API proposal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md) implementation issue as a part of https://github.com/maplibre/maplibre-native/issues/1491

---

Implement the new classes (and required additional classes) specified in the proposal for `Symbol`.

* These classes should be implemented and should be usable.
* At this point, only one manager per type needs to be created; automatic grouping will be implemented in a later issue (#1494).
* For the same reason, only data driven properties should be added to the classes. Non data-driven properties will be added later on.
* `CollisionGroup` and `ClusterGroup` are added in later issues (#1495, #1496).
* Dragging and click listeners will be added in a later issue (#1634).

louwers:
This ticket is in scope of the Modernization bounty direction. https://github.com/maplibre/maplibre/issues/270

fynngodau:
@louwers After the first issue is solved, I'll gladly start working on this one.

fynngodau:
@louwers This issue has turned out to be more work than anticipated. For this reason, I propose to move the following sub-tasks to new issues with the following bounty sizes:

* Click, drag, long-click listeners (S)
* `Line`, `Circle`, `Fill` (M)

The following sub-tasks should be moved to existing issue:

* Cleaning up no longer needed bitmaps → https://github.com/maplibre/maplibre-native/issues/1494; the trivial implementation would cause bugs as soon as multiple managers are required.

Passing tests based on `SymbolManagerTest` will still part of the scope of this issue.

louwers:
Thanks for making a suggestion for this @fynngodau.

Please create / update the issues like you suggest and update #1491, I can assign the bounties.

louwers:
This bounty can be paid out.

fynngodau:
Is included in this invoice: https://opencollective.com/maplibre/expenses/162741

Issue #1491, opened 2023-08-08
Android Annotations


fynngodau:
In https://github.com/maplibre/maplibre-native/pull/1255, [a proposal for the future of the annotations API on Android][0] has been merged. This issue tracks the overall implementation of the proposal.

To make the task more manageable, the work can be split into the following issues – roughly in this order:

* (the issues will be created and linked here shortly, please stand by)

[0]: https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md

PR #1724, opened 2023-10-03
Android annotations: implement `Circle`, `Line`, `Fill`

### LINKED ISSUE #1635: Android Annotations: add new classes (2)

---

Implement the new classes (and required additional classes) specified in the proposal for `Fill`, `Circle`, `Line`.

* These classes should be implemented and should be usable.
* At this point, only one manager per type needs to be created; automatic grouping will be implemented in a later issue (#1494).
* For the same reason, only data driven properties should be added to the classes. Non data-driven properties will be added later on.

fynngodau:
[Android annotations API proposal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md) implementation PR as a part of https://github.com/maplibre/maplibre-native/issues/1491. Closes #1635.

---

* `Circle`, `Line` and `Fill` classes are now usable.
* Updates should propagate correctly.
* Bitmaps may not work correctly if used with an annotation that is directly added to a manager. Per https://github.com/maplibre/maplibre-native/issues/1493#issuecomment-1718094327, better bitmap handling and cleanup magic is in scope of #1494.
* Existing tests for `CircleManager`, `LineManager` and `FillManager` have been updated, and pass.
* The following activities in the test app are modified to allow manual testing:
   * Add Markers in Bulk
   * Polygon
   * Polyline
* There is no existing activity for testing Circles; in the future, the demo activities should definitely be reworked. For now, it is easy enough to modify "Add Markers in Bulk" to use circles instead of markers for manual testing.
* To avoid conflicts, the following classes still have temporary names that differ from their intended names:
   * `Annotation` → `KAnnotation`
   * `AnnotationContainer` → `KAnnotationContainer`
* `Symbol` no longer has the temporary name `KSymbol`.
* Commented-out classes temporarily added in #1637 are removed or no longer commented out. The exception is `DraggableAnnotationControllerTest`, which will be re-added in #1634.
* Unresolved TODOs are in the scope of future issues.
* For some more detailed changes, please refer to the commit log.


platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/annotations/AnnotationContainer.kt
```diff
@@ -82,13 +94,21 @@ class KAnnotationContainer(
...
     private fun MutableMap<Key, AnnotationManager<*, *>>.getOrCreate(key: Key): AnnotationManager<*, *>? =
         get(key) ?: style?.let {
-            SymbolManager(mapView, mapLibreMap, it)
+            when (key.type) {
+                Symbol::class -> SymbolManager(mapView, mapLibreMap, it)
+                Circle::class -> CircleManager(mapView, mapLibreMap, it)
+                Line::class -> LineManager(mapView, mapLibreMap, it)
+                Fill::class -> FillManager(mapView, mapLibreMap, it)
+                else -> throw IllegalArgumentException(
+                    "Impossible key! This should never occur because KAnnotation is a sealed class."
```
louwers, line 109: Is this needed when it is a sealed class?

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -996,9 +999,35 @@ void notifyStyleLoaded() {
...
   //
 
-  public void addSymbol(@NonNull KSymbol symbol) {
-    symbol.attach(this, nextId++);
-    annotationContainer.add(symbol);
+  /* To make the types of annotations more discoverable, we have implemented the same method
+   * four times for all of the subtypes of Annotation.
+   */
+
+  public void addAnnotation(@NonNull Symbol annotation) {
```
louwers, line 1006: Nice!

Could you add some doc comments to the new classes here, it will look pretty in the documentation.

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/annotations/AnnotationContainer.kt
fynngodau: @louwers From what I've tried, yes. `key.type` is not of type `KAnnotation` (unlike the other `when` statements in this file), and it seems like the compiler can't derive from the fact that `key.type` is of type `KClass<KAnnotation<*>>` that there's only a limited number of (equivalence groups of) instances that could be given as inputs.

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/MapLibreMap.java
fynngodau: @louwers Documentation comments are now added.

I also un-deprecated the `removeAnnotations()` method, while rewriting the `clear()` doc comment to refer to `removeAnnotations()`. This decision was made because `clear()` could be understood to also remove other things, like map listeners, and hence `removeAnnotations()` is more explicit about what it actually does.

louwers: Great.

PR #1737, opened 2023-10-05
Android annotations: click, long-click and dragging

### LINKED ISSUE #1634: Android Annotations: add click and drag listeners

fynngodau:
[Android annotations API proposal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md) implementation PR as a part of https://github.com/maplibre/maplibre-native/issues/1491. Closes #1634. To be merged after #1724.

---

* Click listeners, long-click listeners and dragging, as well as drag listeners are usable.
* Existing tests for `DraggableAnnotationController` have been updated, and pass.
* The "Add Markers in Bulk" activity in the test app has been modified to allow manual testing: the symbol number 0 toasts when you click on it, and symbol number 1 is draggable.
* To avoid conflicts, the following classes still have temporary names that differ from their intended names:
   * `Annotation` → `KAnnotation`
   * `AnnotationContainer` → `KAnnotationContainer`
* It is not yet certain whether any lag while dragging comes primarily from debug builds, or from any new annotations code. I have not been able to find problems with annotations code but would welcome any clues.


platform/android/MapboxGLAndroidSDKTestApp/src/main/java/org/maplibre/android/testapp/activity/annotation/BulkMarkerActivity.kt
```diff
@@ -102,12 +101,25 @@ class BulkMarkerActivity : AppCompatActivity(), OnItemSelectedListener {
...
-                KSymbol(
+            maplibreMap.addAnnotation(
+                Symbol(
                     position = latLng,
                     text = Text(i.toString(), color = Color.WHITE),
                     icon = icon
-                )
+                ).apply {
+                    if (i == 0) {
+                        clickListener = object : OnSymbolClickListener {
```
louwers, line 111: Could this be Kotlin-ified further by just using a lambda? Or do we want to keep it this way for limited backwards compatibility?

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/org/maplibre/android/testapp/activity/annotation/PolygonActivity.kt
```diff
@@ -56,18 +55,16 @@ class PolygonActivity : AppCompatActivity(), OnMapReadyCallback {
     override fun onMapReady(map: MapLibreMap) {
         maplibreMap = map
         map.setStyle(Style.getPredefinedStyle("Streets"))
-        map.setOnPolygonClickListener { polygon: Polygon ->
+        /*map.setOnPolygonClickListener { polygon: Fill ->
             Toast.makeText(
                 this@PolygonActivity,
                 "You clicked on polygon with id = " + polygon.id,
```
louwers, line 61: Can this example be revived?

louwers:
I created a trace dragging the annotation around on the Bulk Marker Activity...

image

louwers:
The dragging issue seems to get progressively worse with more markers.

Are you iterating over all the markers to handle the drag event?

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/org/maplibre/android/testapp/activity/annotation/BulkMarkerActivity.kt
fynngodau: Sure can! This is a great idea because it will clean up Kotlin code but not have negative impacts on Java code (provided that lambdas are used). Somehow it will make our code a bit less pretty but that is well worth it.

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/org/maplibre/android/testapp/activity/annotation/PolygonActivity.kt
fynngodau: @louwers

> `map.setOnPolygonClickListener` still works right?

Actually, no, this is not covered by the spec. Probably it should still be added – I think it would be best to do this at the same time that we add functionality for users being able to add managers to `MapLibreMap` (which, now to notice, is not actually covered by any of the existing issues. I'd suggest to do it after automatic grouping is implemented since we wanted a method that adds a new manager above a specific z layer).

I reinstated the click listener by adding it to the individual polygon.

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/annotations/AnnotationContainer.kt
```diff
@@ -41,17 +44,23 @@ class KAnnotationContainer(
         }
     }
 
+    @UiThread
     fun update(annotation: KAnnotation<*>) {
         managers[annotation.key()]?.updateSource()
-        if (annotation is KSymbol) annotation.icon?.let { style?.addImage(it.image.toString(), it.image) }
+        if (annotation is Symbol) annotation.icon?.let { style?.addImage(it.image.toString(), it.image) }
+        if (annotation is Line) annotation.pattern?.let { style?.addImage(it.toString(), it) }
+        if (annotation is Fill) annotation.pattern?.let { style?.addImage(it.toString(), it) }
```
fynngodau, line 52: > Are you iterating over all the markers to handle the drag event?

@louwers No, like previously, we are only calling `AnnotationManager.queryMapForFeatures` one time to get the annotation to be dragged when dragging begins. What I considered could be a bottleneck is that we are re-adding the icon image to the map every update. However, commenting out the lines marked above does not noticeably improve the situation.

louwers: Was the performance this bad before too?

Let's have a chat on Slack when you have time.

louwers (approved):
There are some performance issues, but it's unclear if they are regressions or not.

I spoke with @fynngodau and we will merge this for now and when the project is complete we can do another pass for performance QA.

fynngodau:
(Tests fail because we didn't have the render tests at the time. I'll try merging main into `android-annotations` next.)

CodeK1988:
@fynngodau @louwers hi mate, When creating a symbol using SymbolManager, I noticed that fast dragging is unresponsive and has a high latency. Is there a solution?  What I want is to edit lines by dragging and dropping Markers (symbols); the line dragging function in the map plugin does not meet my needs.

PR #1740, opened 2023-10-08
[qt] Only build core platform as part of main maplibre-native repo

ntadej:
This makes only core part of the Qt platform build as part of this repository.

Unit tests are also enabled on Linux (with other platforms to follow later). I only needed to disable WebP test as Qt is not built with it by default.

I could not enable render tests as they do not compile so this will come later.


ntadej:
@louwers, from my point of view this is ready for review. I decided to keep some files in the core (the ones recently touched due to the Metal effort) and are not really public-facing.

PR #1743, opened 2023-10-09
[cmake] Make most of vendor libraries OBJECT libraries

ntadej:
Now that we have new enough CMake we can make most of vendor libraries OBJECT libraries which helps deploying MapLibre Native as a static library in only one file.


ntadej:
I will abandon this for now as I do not understand the issue.

PR #1750, opened 2023-10-10
Further stencil optimizations

TimSylvester:
Avoid re-building the tile stencil mask when the current tile set is a subset of the already-rendered tiles, which would previously be considered not-equal and rebuild it.  This would sometimes happen when rendering while not all the tiles have loaded, so probably won't affect the benchmark where the tiles sets are effectively static.

Eliminate an extra template expansion of `renderTileClippingMasks` on `RenderTiles` which is only used in legacy builds.  This can be further simplified once the legacy path is removed entirely.

### DIFF
--- a/src/mbgl/renderer/paint_parameters.cpp
+++ b/src/mbgl/renderer/paint_parameters.cpp
@@ -8,21 +8,12 @@
 #include <mbgl/renderer/render_source.hpp>
 #include <mbgl/renderer/render_tile.hpp>
 #include <mbgl/renderer/update_parameters.hpp>
-
-#if MLN_RENDER_BACKEND_METAL
-#include <mbgl/mtl/context.hpp>
-#include <mbgl/mtl/render_pass.hpp>
-#include <mbgl/mtl/renderer_backend.hpp>
-#include <mbgl/shaders/layer_ubo.hpp>
-#include <mbgl/shaders/mtl/clipping_mask.hpp>
-#include <mbgl/shaders/mtl/shader_program.hpp>
-#include <mbgl/shaders/shader_program_base.hpp>
-#include <mbgl/shaders/shader_source.hpp>
 #include <mbgl/util/convert.hpp>
 #include <mbgl/util/logging.hpp>
 
-#include <Foundation/Foundation.hpp>
-#include <Metal/Metal.hpp>
+#if MLN_RENDER_BACKEND_METAL
+#include <mbgl/mtl/context.hpp>
+#include <mbgl/shaders/mtl/clipping_mask.hpp>
 #endif // MLN_RENDER_BACKEND_METAL
 
 namespace mbgl {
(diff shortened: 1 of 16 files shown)


sjg-wdw (approved):

alexcristici (approved):

louwers (commented):
Any chance the Metal-specific code in `PaintParameters` can be moved to the Metal namespace (and similarly for the OpenGL specific code)? I think it is getting a bit cluttered.

src/mbgl/renderer/paint_parameters.cpp
```diff
@@ -103,53 +103,71 @@ gfx::DepthMode PaintParameters::depthModeFor3D() const {
...
 namespace {
 
 template <typename TIter>
 using GetTileIDFunc = std::function<const UnwrappedTileID&(const typename TIter::value_type&)>;
 using TileMaskIDMap = std::map<UnwrappedTileID, int32_t>;
 
-// Detects a difference in keys of renderTiles and tileClippingMaskIDs
+// `std::includes` adapted to unwrap arbitrary collection iterators
+template <class I1, class I2>
+bool includes(I1 beg1, const I1 end1, I2 beg2, const I2 end2, const GetTileIDFunc<I2>& f) {
```
louwers, line 114: It's not really arbitrary because of the `GetTileIDFunc` type. Maybe make that parameter a template (too) and move it to its own file?

src/mbgl/renderer/paint_parameters.cpp
```diff
@@ -103,53 +103,71 @@ gfx::DepthMode PaintParameters::depthModeFor3D() const {
...
-    renderTileClippingMasks(ids.cbegin(), ids.cend(), std::move(f), /*clear=*/true);
-#else
-    context.clearStencilBuffer(0b00000000);
-#endif
-}
-
 namespace {
 
 template <typename TIter>
 using GetTileIDFunc = std::function<const UnwrappedTileID&(const typename TIter::value_type&)>;
```
louwers, line 109: Not related to this PR but I think this a function pointer would work as well here and would be more performant.

src/mbgl/renderer/paint_parameters.cpp
TimSylvester: It works on any iterable container of tile IDs, but not arbitrary containers of anything, which is why it's there instead of in `util` somewhere.  It could be made more general, but the un-wrapper function would get more complex.  All of this can be deleted entirely when the legacy code is removed, it's just to avoid duplicating the code for iterating `RenderTiles`.

louwers (approved):
Thanks for addressing my suggestions.

TimSylvester:
@sjg-wdw I converted the stencil rendering to a single indexed draw call, please take another look.   ([ace165b](https://github.com/maplibre/maplibre-native/pull/1750/commits/ace165b16f61471e1034d022b1e1511ba5c63efd) in particular)

sjg-wdw (approved):

sjg-wdw:
> @sjg-wdw I converted the stencil rendering to a single indexed draw call, please take another look. ([ace165b](https://github.com/maplibre/maplibre-native/pull/1750/commits/ace165b16f61471e1034d022b1e1511ba5c63efd) in particular)

Well that looks pretty clever.  Is this our first instancing?

TimSylvester:
Yeah, I think so.

PR #1751, opened 2023-10-11
[ci] Fix Gradle cache in Android workflows

ntadej:
It seems the expressions were not done properly, probably due to some find-replace.

Also remove trailing whitespace.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -14,21 +14,22 @@ jobs:
       BUILDTYPE: Release
       IS_LOCAL_DEVELOPMENT: false
     steps:
-
       - uses: actions/checkout@v4
         with:
           submodules: recursive
           fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


.github/workflows/android-release.yml
```diff
@@ -14,21 +14,27 @@ jobs:
...
-
       - uses: actions/checkout@v4
         with:
           submodules: recursive
           fetch-depth: 0
 
+      - name: Setup cmake
+        uses: jwlawson/actions-setup-cmake@v1.14
+        with:
+          cmake-version: "3.19.x"
```
louwers, line 25: CMake 3.27.6 is included by default.

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#tools

ntadej: This just unifies it with the CI. I can revert it from both.

louwers: Ah! Yes please.

ntadej: Done.

louwers (approved):

PR #1758, opened 2023-10-14
[ci] Cleanup caches in the target repo

ntadej:
We need to use `pull_request_target` so that caches from forks are deleted. We only request relevant permissions so this should be safe enough.

Also quote variables, just in case.

Why delete caches? We cache a lot of data and the quota we have is low. This prevents relevant caches to get evicted too soon.

### DIFF
--- a/.github/workflows/cache-cleanup.yml
+++ b/.github/workflows/cache-cleanup.yml
@@ -1,6 +1,6 @@
 name: cleanup-branch-caches
 on:
-  pull_request:
+  pull_request_target:
     types:
       - closed
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


ntadej:
We may want to clear all caches once this is merged. I can do that if fine with you @louwers.

louwers (changes requested):
The Qt workflows are using up 4260MB of cache storage for a single workflow run, almost half of the 10GB cache limit. So even if this is merged relevant caches probably evicted too soon.

image

Let's solve that problem first... One idea is to use an S3 Cache:

https://github.com/marketplace/actions/s3-cache-for-github-actions

We would save only on main (because a secret is needed) but retrieve with a public read-only key.

I'm not sure if I like removing caches from PRs because when workflows get skipped on `main` (ideally almost always when a workflow already ran for a PR with the same code) we'd potentially end up with a stale cache.

ntadej:
Yes, I will disable the debug symbols.

Still now all PR caches stay there. I see no reason why this would not get merged. It will never affect the caches of the branch merged into.

ntadej:
To add, we already evict caches for PRs open from branches directly in this repo. It's just the forks that do not work.

louwers:
>  It will never affect the caches of the branch merged into.

If a PR adds a node module (as an example), it will create a cache for it, then when it is merged the workflow is skipped on main and the cache is deleted. A subsequent PR won't have a cache of the node modules.

We can still merge this but we should not check out any code in a `pull_request_target` workflow.

ntadej:
PRs only use caches from `main` or their own caches. `main` will always recreate caches (I guess to avoid getting some malicious data from PRs). At least this is my experience.

AFAIK we need the clone for GitHub CLI to work, but I can try to workaround it. Also I see no security issue in cloning what was just merged in `main` because this runs after merge so as long nothing malicious is merged, we are fine.

Regarding `ccache` size, I think we accumulate historic data each time the code changes. We need to set some sensible limit e.g. 200 MB (or maybe even lower). I will look into it.

ntadej:
Sure, and then we should also delete the caches as they are irrelevant. As far as I understand this will still take the action from the `main` branch so we should not have a problem (but we can test it).

louwers (approved):

ntadej:
Hmm, nothing happened. I guess this is good from security point of view, although not solving the caches issues 😊

Let's merge this for now and monitor the behaviour.

PR #1765, opened 2023-10-17
CI fixes and improvements

louwers:
- Checkout the repo in the new `pr-bloaty` workflow (otherwise the reusable action cannot be found).
- Simplify `ios-device-test` workflow by using reusable action.
- Don't fail `android-device-test` workflow when artifact cannot be found.


louwers:
@HarelM Thanks :)

PR #1767, opened 2023-10-17
Compile bloaty in action

louwers:
Google has not made a release for over 3 years, so we have to build it ourselves...


HarelM (approved):
I would try and remove this dependency then...

PR #1769, opened 2023-10-17
Upload coverage in separate workflow

louwers:
We need to upload the coverage in a separate workflow, becuase it needs access to the codecov secret.


ntadej (changes requested):
Why do we need a secret? I don't need it in any of my repos?

louwers:
@ntadej Because Codecov gets rate limited by GitHub otherwise. This is causing the workflow to fail periodically. https://community.codecov.com/t/github-api-forbidden/1295/4

ntadej (approved):

louwers:
Temporarily disabling code coverage check to merge.

louwers:
Enabled again.

PR #1770, opened 2023-10-17
Don't use Ninja for bloaty build

louwers:
It's not available on the default Ubuntu image...

### DIFF
--- a/.github/workflows/pr-bloaty.yml
+++ b/.github/workflows/pr-bloaty.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Compile Bloaty
         run: |
           git clone https://github.com/google/bloaty.git
-          cmake -B bloaty/build -G Ninja -S bloaty
+          cmake -B bloaty/build -S bloaty
           cmake --build bloaty/build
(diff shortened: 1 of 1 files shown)


louwers:
@ntadej Sure I can also do that.

.github/workflows/pr-bloaty.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Compile Bloaty
         run: |
           git clone https://github.com/google/bloaty.git
-          cmake -B bloaty/build -G Ninja -S bloaty
+          cmake -B bloaty/build -S bloaty
```
ntadej, line 39: This will AFAIK build serially which we probably do not want.

ntadej (changes requested):

.github/workflows/pr-bloaty.yml
louwers: Fixed.

ntadej (approved):
We should probably unify actions between workflows, but fine with me for now.

PR #1775, opened 2023-10-18
cd in bloaty compile step

louwers:


ntadej (approved):
I recommend next time you make a temporary trigger for PR to test it 😊

louwers:
@ntadej It's not possible because it only uses the version in `main`. :(

ntadej:
What I meant is add the PR trigger, test it as much as possible (i.e. build) and then remove the trigger.

PR #1776, opened 2023-10-18
Benchmark app frame time improvements

alexcristici:
Refactored benchmark app to use headless frontend for an accurate measurement of the frame time.

Make sure to set the Benchmark app on `Release`, without `Debug executable` set and disable `Metal API validation`:
Screenshot 2023-11-02 at 20 47 16
Screenshot 2023-11-02 at 20 47 06


sjg-wdw (approved):
Looks reasonable, but someone with more familiarity with the MLN View should weigh in.

louwers (commented):
Could you share some more details about why these changes were needed here, as well as your test results?

platform/default/src/mbgl/gfx/headless_frontend.cpp
```diff
@@ -169,6 +157,24 @@ void HeadlessFrontend::renderOnce(Map&) {
     util::RunLoop::Get()->runOnce();
 }
 
+void HeadlessFrontend::renderFrame() {
```
louwers, line 160: Should this code maybe moved to `MLNRenderFrontend`?

I know it was here before but I feel like it doesn't belong here because it is specific to `MLNRenderFrontend`.

alexcristici: This code is specific to headless_frontend, which is used in render tests and now in Benchmark app. 
`MLNRenderFrontend` is not headless frontend, it is used in the MapView to render directly to the screen.

louwers: Ah I was confusing `RendererFrontend` and `HeadlessFrontend`. Thanks.

mwilsnd (approved):

TimSylvester (approved):
On my iPad mini 5:

```
| boston | 3.2210 ms | 4.7509 ms |
| paris | 2.5287 ms | 12.5014 ms |
| paris2 | 2.1421 ms | 12.5297 ms |
| alps | 2.9439 ms | 21.9420 ms |
| us east | 2.0176 ms | 14.9323 ms |
| greater la | 2.2015 ms | 14.6801 ms |
| sf | 4.0980 ms | 18.4790 ms |
| oakland | 2.5990 ms | 8.9986 ms |
| germany | 3.4378 ms | 22.3297 ms |
Average frame encoding time: 2.7988 ms
Average frame rendering time: 14.5715 ms
```

alexcristici:
> Could you share some more details about why these changes were needed here, as well as your test results?

Old implementation was rendering directly to the `MTKView` or on screen. Because of that, frame measurement was affected by waiting for new drawable to be available in the `MTKView`. Because the display is set to 60 fps, new drawable is available one time at 16.666 ms. Even though the encoding of the gpu commands would have taken less than 16.666 ms, the frame time would measure 16.66 ms because it was waiting for the new drawable. 

In order have an accurate measurement and not being affected by this waiting time (when new drawable is available) I had to render the map to an offscreen texture, using the headless frontend. In this way I can render frame after frame without being limited by vsync of 60 fps and measure correctly the encoding time of the gpu command.

Latest results using refactored Benchmark app:
Screenshot 2023-11-02 at 11 48 26
s.

PR #1780, opened 2023-10-18
Triangulated fill outlines

stefankarschti:
feat: Fill layer to generate outlines composed of triangle primitives
- Created a new addition to the fill generator functions to generate this new fill+outline variant
- The Fill bucket generates fill outlines with triangle primitives, compatible with the Line shader
- The Fill bucket uses `gfx::generateFillAndOutineBuffers` to generate its buffers
- Added an option to enable/disable this feature at compile time with the `MLN_TRIANGULATE_FILL_OUTLINES` #define in `fill_bucket.hpp`

Refs: #1768


stefankarschti:
WIP
image

stefankarschti:
Comparison of fill outlines

_Legacy renderer:_
IMG_0032

_Drawable renderer:_
IMG_0031

stefankarschti:
_Fill pattern outline drawn with the legacy renderer:_
IMG_0033-legacy

_And with the Metal renderer (no outline):_
IMG_0034-metal

stefankarschti:
performance impact versus [main@a73e88fee28593a0e66be1e67d37ae444b0bfc6c ](https://github.com/maplibre/maplibre-native/commit/a73e88fee28593a0e66be1e67d37ae444b0bfc6c) 
Screenshot 2023-11-06 at 22 13 30

platform/ios/src/MLNMapView+Metal.mm
```diff
@@ -82,7 +82,7 @@ void swap() override {
 
         // Un-comment for synchronous, which can help troubleshoot rendering problems,
         // particularly those related to resource tracking and multiple queued buffers.
-        //[commandBuffer waitUntilCompleted];
+        [commandBuffer waitUntilCompleted];
```
sjg-wdw, line 85: Probably want to turn this back off.
It might be changing the frame rendering time too.

stefankarschti: oops. correct. remeasuring.

stefankarschti:
new measurements after commenting out `[commandBuffer waitUntilCompleted];`
image

stefankarschti:
New timing measured by averaging 3 runs on _iPad 5th gen_.
In _LineBasicShader v1_ the `interpolation` UBO is deleted.
In _LineBasicShader v2_ the `blur`, `gap width`, `offset` UBO fields are removed; `permutation` and `expression` UBOs are also removed on Metal.

image

sjg-wdw:
Looks like reasonable numbers.  
One more useful measurement might be lines with the single pixel Metal implementation (e.g. what Metal provides).  Just to see what lines are costing us.

stefankarschti:
> Looks like reasonable numbers. One more useful measurement might be lines with the single pixel Metal implementation (e.g. what Metal provides). Just to see what lines are costing us.

This is what metal on `main` has, 2 pixel wide lines (which render as 1 pixel)...

sjg-wdw:
Hmmm, still some cost to the simpler version.

sjg-wdw:
Staring at the numbers a bit more.... I assume they'll be slightly different on more recent hardware, but they're unlikely to break the pattern we've seen before.

I think we want to go with the v2 shader and we may also want to provide an option to turn that off and use Metal lines.
But after we've got the v2 shader merged in, we should see how it fares in a Metal frame analysis.  Maybe those tools can see something we can't.

louwers (commented):
Why is the `MLN_TRIANGULATE_FILL_OUTLINES` flag needed? Is it only because the legacy renderer does not support it? In this case, please add a comment to the issue to remove the legacy renderer to remove this build flag as well.

In general, I think we should avoid build time configuration as much as possible because only the default is tested / compiled on CI.

stefankarschti:
@louwers : I added that flag following Steve's recommendation to "provide an option to turn that off and use Metal lines"...
CC @sjg-wdw

sjg-wdw:
> @louwers : I added that flag following Steve's recommendation to "provide an option to turn that off and use Metal lines"... CC @sjg-wdw

Nice!

alexcristici:
With `#define MLN_TRIANGULATE_FILL_OUTLINES 1` render test `render-tests/fill-outline-color/property-function` is drawing white background, before it was drawing ok but failing because of line width.

Before:
Before

After:
After

louwers (changes requested):
Should we make the more correct (less performant) version the default?

@sjg-wdw @stefankarschti Can one of you create an issue that links to this PR to re-address this after variants have been introduced?

stefankarschti:
While investigating the failed render tests I realized that we also need shader variations for the triangulated lines in case of outline-color and opacity per-vertex, which is the case with some expressions. 
`render-tests/fill-outline-color/zoom-and-property-function`
`render-tests/fill-outline-color/property-function`
ref #1855

PR #1790, opened 2023-10-20
Build node with GCC 12

mwilsnd:


acalcutt:
I personally wouldn't mind moving to 22.04, since it could be supported longer.  It would mean some work on the tileserver-gl side to update the base docker images, but I have already had requests to update the base 20.04 os, which isn't possible right now.

ntadej:
> I personally wouldn't mind moving to 22.04, since it could be supported longer. It would mean some work on the tileserver-gl side to update the base docker images, but I have already had requests to update the base 20.04 os, which isn't possible right now.

But would that affect people who install the package directly and still want to use 20.04?

acalcutt:
Pretty sure it would, yes. for the same reason the current version doesn't work on 22.04. the built statically links to specific versions of things like libicu

ntadej:
> Pretty sure it would, yes. for the same reason the current version doesn't work on 22.04. the built statically links to specific versions of things like libicu

OK, then fine with me. I was just asking as Qt we build on the lowest supported version (and I'll investigate moving to static ICU).

acalcutt:
You can see this long thread on the tileserver-gl repo, trying to get the current not release to work on 22.04
https://github.com/maptiler/tileserver-gl/issues/643

In the end we got stuck at libicui18n.so.66 since ubuntu 22.04 uses libicu70 and had libicui18n.so.70
https://github.com/maptiler/tileserver-gl/issues/643#issuecomment-1317116437

PR #1792, opened 2023-10-20
Use `TinyUnorderedMap` for render tile sets

TimSylvester:
Seems to have a small but measurable effect on `updateRenderTileIDs`, which is called every frame for every layer.

Before:
```
336.30 ms   36.80 ms  mbgl::RenderLayer::prepare(mbgl::LayerPrepareParameters const&)
170.90 ms    5.80 ms  mbgl::GeometryTileRenderData::getLayerRenderData(mbgl::style::Layer::Impl const&) const
112.20 ms   22.50 ms   mbgl::RenderLayer::updateRenderTileIDs()
 45.00 ms   45.00 ms    mbgl::(anonymous namespace)::findTile(std::__1::vector<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>, std::__1::allocator<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>>>&, mbgl::OverscaledTileID const&, bool)
 42.30 ms   42.30 ms    std::__1::__wrap_iter<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>*> std::__1::vector<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>, std::__1::allocator<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>>>::emplace<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>>(std::__1::__wrap_iter<std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity> const*>, std::__1::pair<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity>&&)
```
After:
```
238.70 ms   30.20 ms  mbgl::RenderLayer::prepare(mbgl::LayerPrepareParameters const&)
124.90 ms    4.70 ms   mbgl::GeometryTileRenderData::getLayerRenderData(mbgl::style::Layer::Impl const&) const
 73.70 ms   48.00 ms   mbgl::RenderLayer::updateRenderTileIDs()
 24.70 ms   23.10 ms    mbgl::util::TinyUnorderedMap<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity, 12ul, std::__1::hash<mbgl::OverscaledTileID>, std::__1::equal_to<mbgl::OverscaledTileID>, std::__1::allocator<std::__1::pair<mbgl::OverscaledTileID const, mbgl::util::SimpleIdentity>>>::swap(mbgl::util::TinyUnorderedMap<mbgl::OverscaledTileID, mbgl::util::SimpleIdentity, 12ul, std::__1::hash<mbgl::OverscaledTileID>, std::__1::equal_to<mbgl::OverscaledTileID>, std::__1::allocator<std::__1::pair<mbgl::OverscaledTileID const, mbgl::util::SimpleIdentity>>>&)
```


sjg-wdw:
Messing with the tileIDs was pretty heavy in the last few sets of traces I looked at.  But I may be out of date.

mwilsnd:
For a linear search, we'd want it to fit in L1 data nicely so the CPU streaming prefetch works in our favor.

OverscaledTileID = 12B (might be a bit more due to compiler-inserted padding) + SimpleIdentity = 8B. Assuming a 64B cache line and ignoring alignment we prefetch 3 entries in a cache line.

If we have a larger number of IDs to search we might split the pair into two vectors, then the prefetch is only grabbing the tile IDs and not wasting cache line on the simple identity.

louwers:
There is a build issue with TinyUnorderedMap on Windows.

TimSylvester:
Yeah, I saw that, but it doesn't make sense to me.  I can try setting up a Windows environment.

louwers:
@tdcosta100 Do you mind having a quick look?

mwilsnd:
Is that Qt build using C++14? It's a really old compiler version - GNU 8.1.0 (2018). This builds correctly for me (minus unrelated narrowing warnings from the custom drawable layer).

I found this which might explain the problem https://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p1286r2.html, which looks like it wasn't implemented until GCC 10: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90537.

ntadej:
In principle I'd be fine to drop MinGW support for Qt 5 on Windows if that helps.

TimSylvester:
@stefankarschti The Metal uniform binding process relies on iterating over the map.  I ended up being dropping the iterator wrappers that make it possible to iterate regardless of which internal storage is in use.

The obvious workaround of iterating over the uniforms listed in the shader metadata would mean doing lookups of each one, which I think would end up being wasteful.

mwilsnd (approved):

PR #1794, opened 2023-10-22
update for node 21

acalcutt:
Fix broken node workflow

1.) Update NAN from 2.17.0 to 2.18.0
2.) Update syntax of node_request.cpp to work with node 21
3.) Update Ubuntu version to 22.04
4.) Update build node version to 20
5.) Add node 20 to the published binaries. Remove node 14
6.) Workaround cmake issue caused by recent update to Server-2022 image ( https://github.com/actions/runner-images/issues/8598 )


acalcutt:
I also added code in https://github.com/maplibre/maplibre-native/pull/1793 that sets a maximum node version to download.

I was thinking increasing the max node version manually might be preferable to the workflow just randomly breaking when a new version gets released. 

let me know if you want me to bring that code over to this PR.

louwers (approved):
Thank you! 💯

Issue #1808, opened 2023-10-26
Benchmarks that use the HeadlessFrontend / Map are hanging


louwers:
The benchmarks in `benchmark/api/query.benchmark.cpp` and `benchmark/api/render.benchmark.cpp` are hanging in this loop:

https://github.com/maplibre/maplibre-native/blob/619408d09d107b92630a4e04848de1178a2f9e81/platform/default/src/mbgl/gfx/headless_frontend.cpp#L157-L159

I'm enabling the other benchmarks and plan to look into this issue.

ntadej:
Which platform?

Note that in this specific case I suspect that the image never gets valid. Can you confirm that?

louwers:
@ntadej I tried both Android and Linux. Possibly the sprites are never loaded.

Yes the image never gets valid.

PR #1809, opened 2023-10-26
Linux benchmark on CI

louwers:
The `API_*` benchmarks are disabled until I figure out why they hang: https://github.com/maplibre/maplibre-native/issues/1808

Similar to the Bloaty setup, this creates a (release) benchmark binary in the `linux-ci.yml` workflow. Then depending on in which context the workflow is ran...

- For a PR: create artifact to be picked up by the `pr-tests.yml` workflow.
- On main: upload binary to S3.

In `pr-tests.yml`:

1. The artifact is downloaded (PR binary) the main version is pulled from S3.
2. The test is run.
3. The result is commented on the PR (with a link to the full report).

I also updated Google Benchmark.


.github/workflows/linux-ci.yml
```diff
@@ -96,21 +96,29 @@ jobs:
...
-      # mbgl-render (used for size test)
+      # mbgl-render (used for size test) & mbgl-benchmark-runner
 
-      - name: Build mbgl-render
+      - name: Build mbgl-render & mbgl-benchmark-runner
         if: matrix.renderer == 'drawable'
         run: |
-          cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -B build-size -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DMLN_DRAWABLE_RENDERER=ON
-          cmake --build build-size --target mbgl-render
+          cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -B build-release -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DMLN_DRAWABLE_RENDERER=ON
```
ntadej, line 104: Why do we need `Debug` build above to run tests? Can we just build everything with `RelWithDebInfo`?

ntadej (changes requested):
In interest of speeding-up the CI, I wonder why we need to build the code twice.

.github/workflows/linux-ci.yml
ntadej: I guess because you don't have optimisations, but `RelWithDebInfo` should have all the debug symbols. As Linux runners are not so critical, I guess it's fine for now, but just a tip how to potentially speed things up a bit.

ntadej (approved):

louwers:
@tdcosta100 Do you happen to know why there are linker errors on Windows?

ntadej:
It seems that symbols are not properly exported. Depending who has time first, I can also try to fix it next week.

tdcosta100:
Hi guys, sorry for getting late in the discussion. It seems benchmark library needs a static build under Windows, because the Windows library is static. Adding the `BENCHMARK_STATIC_DEFINE` define to the `mbgl-vendor-benchmark` target seems to fix the issue, at least for me. This is the modification I've made to `vendor/benchmark.cmake`:

```diff --git a/vendor/benchmark.cmake b/vendor/benchmark.cmake
index b3927868f..87d80c458 100644
--- a/vendor/benchmark.cmake
+++ b/vendor/benchmark.cmake
@@ -30,6 +30,13 @@ target_compile_definitions(
     PRIVATE HAVE_STEADY_CLOCK
 )

+if(WIN32)
+    target_compile_definitions(
+        mbgl-vendor-benchmark
+        PUBLIC BENCHMARK_STATIC_DEFINE
+    )
+endif()
+
 target_include_directories(
     mbgl-vendor-benchmark SYSTEM
     PUBLIC ${CMAKE_CURRENT_LIST_DIR}/benchmark/include
```

PR #1813, opened 2023-10-29
Ignore Qt installation when finding ICU

ntadej:
This is a bit hacky but I can't find another solution. We want to ignore Qt-provided ICU as it has no headers provided so it's impossible to use it out-of-the-box.

Also move Linux-specific commands before Qt setup.


louwers (approved):

ntadej:
I made the feature opt-out if needed.

PR #1827, opened 2023-10-30
Rename `FileSource` to `ResourceLoader`

louwers:


ntadej:
What is the motivation for this? I guess also other references should be renamed if we want to proceed.

louwers:
@ntadej Yes I'll rename the other references (and filenames) as well. There's an open TODO above `FileSource` which I agree with.

> // TODO: Rename to ResourceProvider to avoid confusion with
> // GeoJSONSource, RasterSource, VectorSource, CustomGeometrySource and other *Sources.

It's also generally badly named. `FileSource` instances are called `resourceProvider` in the Android codebase, and I think this name is much more descriptive.

PR #1831, opened 2023-10-31
Refactor bindings to allow for eliminating duplicates...

TimSylvester:
Reviewing the profiling results, I noticed we're spending as much time binding as in draw calls.

```
4.54 s     14.7% 735.00 ms  mbgl::mtl::Drawable::draw(mbgl::PaintParameters&) const
1.02 s      3.3%  43.00 ms    mbgl::mtl::Drawable::bindUniformBuffers(mbgl::mtl::RenderPass const&) const
964.00 ms   3.1%  16.00 ms    -[AGXA12FamilyRenderContext drawIndexedPrimitives:indexCount:indexType:indexBuffer:indexBufferOffset:instanceCount:baseVertex:baseInstance:]
```

The frame capture results also include a large number of "redundant binding" warnings, so I thought I look into reworking that.

I made `RenderPass` (which owns the encoder) keep track of what buffer was last bound in each index and skip the binding API call if the same buffer is provided for the same index.

Based on [this](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/BufferBindings.html) I also made the buffer resource only create a `MTLBuffer` for large buffers.  If this turns out to be useful, we can look at not making a copy of the data where possible.

This also removes the explicit "un-binding" of buffers by replacing them with null, since this does take additional time and provides no benefit except when troubleshooting incorrect bindings.

The effect on performance is relatively small, at least as measured by the fixed-view benchmarks on my device.

```
Main

default 12:43:38.018286-0700  BenchmarkApp  Average frame time: 15.3 ms
default 12:45:21.956852-0700  BenchmarkApp  Average frame time: 14.7 ms
default 12:46:14.420546-0700  BenchmarkApp  Average frame time: 14.7 ms

Threshold=0

default 12:35:53.251260-0700  BenchmarkApp  Average frame time: 14.4 ms
default 12:39:27.748389-0700  BenchmarkApp  Average frame time: 15.1 ms
default 12:40:22.097796-0700  BenchmarkApp  Average frame time: 14.6 ms

Threshold = 4096

default 14:51:13.121840-0700  BenchmarkApp  Average frame time: 13.2 ms
default 14:56:15.659756-0700  BenchmarkApp  Average frame time: 12.5 ms
default 14:59:06.623332-0700  BenchmarkApp  Average frame time: 14.7 ms
default 15:00:41.122817-0700  BenchmarkApp  Average frame time: 14.9 ms
default 15:01:46.593355-0700  BenchmarkApp  Average frame time: 14.5 ms
default 15:02:52.889321-0700  BenchmarkApp  Average frame time: 13.8 ms
```


sjg-wdw (approved):

src/mbgl/mtl/buffer_resource.cpp
```diff
@@ -4,62 +4,158 @@
...
     : context(context_),
       size(static_cast<NS::UInteger>(size_)),
-      usage(usage_) {
-    auto& device = context.getBackend().getDevice();
-    buffer = NS::TransferPtr((data && size) ? device->newBuffer(data, size, usage) : device->newBuffer(size, usage));
-    if (buffer) {
+      usage(usage_),
+      isIndexBuffer(isIndexBuffer_),
+      persistent(persistent_) {
+    // If the buffer is small, not for indexes, and not explicitly persustent, skip the overhead of an `NSBuffer`
```
louwers, line 33:
```suggestion
    // If the buffer is small, not for indexes, and not explicitly persistent, skip the overhead of an `NSBuffer`
```

src/mbgl/mtl/context.cpp
```diff
@@ -212,20 +217,28 @@ const BufferResource& Context::getTileVertexBuffer() {
     if (!tileVertexBuffer) {
         const auto vertices = RenderStaticData::tileVertices();
         constexpr auto vertexSize = sizeof(decltype(vertices)::Vertex::a1);
-        tileVertexBuffer.emplace(createBuffer(vertices.data(), vertices.bytes(), gfx::BufferUsageType::StaticDraw));
+        tileVertexBuffer.emplace(createBuffer(vertices.data(),
+                                              vertices.bytes(),
+                                              gfx::BufferUsageType::StaticDraw,
+                                              /*isIndexBuffer=*/true,
+                                              /*persustent=*/true));
```
louwers, line 224:
```suggestion
                                              /*persistent=*/true));
```

src/mbgl/mtl/drawable.cpp
```diff
@@ -364,103 +344,85 @@ void Drawable::setVertexAttrNameId(const StringIdentity value) {
...
-                const auto tileID = getTileID() ? util::toString(*getTileID()) : "<no tile>";
-                const auto tileLabel = util::toString(getID()) + "/" + getName() + "/" + tileID;
-                Log::Error(Event::General,
-                           "bindUniformBuffers: UBO "s + stringIndexer().get(element.first) + " not found on " +
-                               tileLabel + ". skipping.");
+                try { // discard logging exceptions to conform to `noexcept`
+                    using namespace std::string_literals;
+                    const auto tileID = getTileID() ? util::toString(*getTileID()) : "<no tile>";
+                    const auto tileLabel = util::toString(getID()) + "/" + getName() + "/" + tileID;
+                    Log::Error(Event::General,
```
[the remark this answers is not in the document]
TimSylvester: It would be a bug, but it might be in a custom layer implementation.  It would be nice to get useful feedback without a debug build, but maybe not worth it since we need to reduce code size.

mwilsnd (approved):

alexcristici (approved):

Issue #1815, opened 2023-10-30
First map rendering takes a long time when you have a lot of offline data stored


geolives:
**Describe the bug**
I found a nasty issue on MapLibre native when you have a lot of data downloaded. The time for the map first rendering increases significantly. With 1,5GB of offline data, the first map rendering lasts 5 seconds instead of 1 second. 

**To Reproduce**
Steps to reproduce the behavior with the test app:
1. Change the default style for the test app with this style: https://carto.geolives.com/vector/geolivestopoosm_vector_grey.json
2. Go to folder directory to download offline areas.
3. Download an amount of 1,5GB+ of offline data.
4. Exit the app and re-run it.
5. The first map rendering will take more time to render completely.

The time increases as the local database also increases.
The referenced style includes multiple sources that emphasizes the issue, but you can use any style that includes sources with a lot of data.

**Expected behavior**
A first map rendering that takes the same time or near the same time as when no offline data are stored.

**Screenshots**
https://github.com/maplibre/maplibre-native/assets/13694294/a7c2bf01-54ff-43ed-86b1-675efb79e00a

**Platform information:**
- OS: iOS 14. I suspect Android to have the same issue, but less noticable.
 - branch main, version 6.0.0.

**Additional context**
Historically, this issue was found in Mapbox in 2020 : https://github.com/mapbox/mapbox-gl-native/issues/16519
I searched in the code and I agree with the initial issue : the initialization of the ambient cache size (`OfflineDatabase::initAmbientCacheSize`) takes a long time when the database gets a lot of stored tiles.

I modified the code by ignoring the request and returning directly 0 for the ambient cache size, and the first map rendering was instantly done.

geolives:
2 possible alternatives to solve this issue:
- Store the current ambient cache size inside the database, and update it each time his cache size is modified. This way, the initialization simply request the computed cache size from the last execution.
- Store a flag inside the tiles and resources table that tells if the resource/tile is ambient or not. That way we simplify the request by deleting the need to make a left join. With the cost that for each download/cleaning, the flag must be re-set correctly.

louwers:
Thanks for this super detailed bug report.

Which of the two solutions you suggest would you prefer?

geolives:
With more study on this issue, i could see a third option that is better than the two previous ones.

[...]

PR #1839, opened 2023-11-06
Added design proposal for ambient cache initialization speed improvement

geolives:
Added design proposal for ambient cache initialization speed improvement to fix issue https://github.com/maplibre/maplibre-native/issues/1815


louwers (approved):
We're giving other community members some time to respond, but looks good to me. Thanks for writing it up.

JesseCrocker:
Will the planned implementation of this support copying resources/tiles that exist in the cache tables to the download tables when a new download is created that uses them? A nice side effect of the current structure is that it avoids redownloading in that case. And if not will resources be deleted from the cache table if they are downloaded?

Did you consider putting the cache tables in a separate database, instead of just in new tables? In my experience a frequent cause of slowness when there are large downloads is db VACUUMing, and vacuuming tends to get triggered when there is lots of turnover in the cache table but the speed of vacuum depends on the overall size of the db.  It would also be nice to make clearing cache just be deleting a file instead of an sql `DELETE`.

geolives:
Yes, when a new download is created that uses tiles / resources that are present in the ambient cache, we will move the data from the ambient cache tables to the offline data tables in order to avoid downloading it again.

For your other proposal of using two separate databases, it may be interesting to do that with a 2nd database attached to the main database connection via ATTACH DATABASE feature of SQLite.
The main drawback of using a second database is the performance to obtain data. 
We would have to query the two databases for each resource / tile, with a SELECT UNION query between the two tables in two different databases or with a SELECT in ambient cache table + a second SELECT in the offline table (in the other database) when the data is not present in the ambient cache.
What do you think about that?

It would be interesting to benchmark the performance in the two cases (all the data in the same database VS ambient cache in a second database).

We would suggest implementing it in two steps :

1st step : all the data in the same database

2nd step : ambient cache in a second database

Thank you.

louwers (changes requested):
Please do the implementation to a different PR.

PR #1842, opened 2023-11-06
Replace rather than update Metal buffers

TimSylvester:
With #1831, small non-index buffers (<= 4KiB) are encoded directly into the render command encoder rather than referenced as buffers.  This should eliminate sequencing problems where the buffer is updated while the previous frame rendering is still using them.

In order to avoid the same problem with larger buffers (e.g., vertex attributes), this change makes buffer updates allocate a new Metal buffer instead of modifying an existing one.  Index buffers were already being replaced rather than updating.

The effect on performance in the benchmarks seems minimal, but cases where the vertex buffers are large and changing regularly (e.g., symbols with layout in dense areas) there may be a more significant effect.

The previous buffers will remain allocated until they are no longer referenced, so memory utilization will increase somewhat.

```
Before:
Average frame encoding time: 2.6382 ms
Average frame rendering time: 14.5671 ms

Average frame encoding time: 2.6830 ms
Average frame rendering time: 14.5455 ms

Average frame encoding time: 2.5971 ms
Average frame rendering time: 14.5503 ms

---

After:
Average frame encoding time: 2.6585 ms
Average frame rendering time: 14.5019 ms

Average frame encoding time: 2.6152 ms
Average frame rendering time: 14.5189 ms

Average frame encoding time: 2.6410 ms
Average frame rendering time: 14.5200 ms
```


sjg-wdw (approved):
What a nice, small change.

louwers (approved):

stefankarschti:
After some panning & zooming the street name labels are drawn in the wrong place. Maybe some UBO buffer is not getting updated and the matrix is wrong? 

https://github.com/maplibre/maplibre-native/assets/4198736/1c070374-e1e4-4948-a638-df5ce2c4dd7f

src/mbgl/mtl/buffer_resource.cpp
```diff
@@ -92,7 +92,12 @@ void BufferResource::update(const void* data, std::size_t updateSize, std::size_
     if (updateSize > 0) {
         if (buffer) {
             if (auto* content = static_cast<uint8_t*>(buffer->contents())) {
-                std::memcpy(content + offset, data, updateSize);
+                // Until we can be sure that the buffer is not still in use to render the
+                // previous frame, replace it with a new buffer instead of updating it.
+                auto& device = context.getBackend().getDevice();
+                if (auto newBuffer = NS::TransferPtr(device->newBuffer(content, size, usage))) {
+                    buffer = std::move(newBuffer);
```
stefankarschti, line 99: If I get this right we're getting a new `buffer` with the old `content` here, but the `data` patch is not being applied?

stefankarschti (approved):

alexcristici (approved):

TimSylvester:
About 1/3 of buffers above the 4k threshold in the benchmark, but only 11K of 1.8M = 0.6% of updates are to large buffers.

```
totalBuffers = 48322
totalBufferObjs = 15239
bufferUpdates = 1769897
bufferObjUpdates = 10878
bufferUpdateBytes = 296601088
uniformUpdateBytes = 107493168
```

src/mbgl/mtl/buffer_resource.cpp
```diff
@@ -90,13 +95,42 @@ void BufferResource::update(const void* data, std::size_t updateSize, std::size_
...
-            if (auto* content = static_cast<uint8_t*>(buffer->contents())) {
-                std::memcpy(content + offset, data, updateSize);
+            if (auto* const content = static_cast<uint8_t*>(buffer->contents())) {
+                // Until we can be sure that the buffer is not still in use to render the
+                // previous frame, replace it with a new buffer instead of updating it.
+                auto& device = context.getBackend().getDevice();
+
+                const bool updateIsEntireBuffer = (offset == 0 && updateSize == size);
+                const auto copySource = updateIsEntireBuffer ? data : content;
+                auto newBuffer = NS::TransferPtr(device->newBuffer(copySource, size, usage));
```
mwilsnd, line 107: Would it be any more efficient to keep a set number of buffers alive in a ring buffer, instead of allocating and releasing buffers all the time?

src/mbgl/mtl/buffer_resource.cpp
```diff
@@ -90,13 +95,42 @@ void BufferResource::update(const void* data, std::size_t updateSize, std::size_
     updateSize = std::min(updateSize, size - offset);
 
     if (updateSize > 0) {
+        auto& stats = context.renderingStats();
         if (buffer) {
-            if (auto* content = static_cast<uint8_t*>(buffer->contents())) {
-                std::memcpy(content + offset, data, updateSize);
+            if (auto* const content = static_cast<uint8_t*>(buffer->contents())) {
```
mwilsnd, line 100: I'd be curious if calling `buffer->contents()` is doing us any favors here. Mapping a buffer to host memory is usually a heavier operation.

The [docs](https://developer.apple.com/documentation/metal/mtlbuffer/1515716-contents) don't make it immediately obvious if this is a full map operation, but I would assume it is.

TimSylvester: Because we're forced to use `MTLResourceStorageModeShared`, I think this is just CPU memory that's mapped when the command buffer is committed, but it could be doing some mapping here.  We'll just have to measure it.

mwilsnd: I don't think there's any way for this API to understand the context of a call to `contents()`. It could be for a read only purpose or the caller could try and write to the memory it points at. In the latter case a transfer has to happen between host and GPU. And then thinking to the case of a compute shader that writes to a buffer, the API would also possibly need to transfer the other way when called as well. This is essentially a full buffer map then, if we cover those situations. Definitely measure it, I'm interested to see how responsive it is on the happy path where it doesn't need to do anything.

stefankarschti: we could skip calling `buffer->contents()` if we know we're going to replace it entirely, if calling it is expensive.

TimSylvester: I re-did the buffer update logic, please have another look.

PR #1848, opened 2023-11-07
Shader Tooling Overhaul

mwilsnd:
Expands the shader code generation used in the GL backend to Metal. The revised tooling supports emitting shaders for both backends with a number of size reduction strategies applied. The most aggressive of these is the compressed amalgamation, which concatenates all shaders together into one buffer and then compresses the whole thing together to hit a high compression ratio.

* New to the Metal backend, shaders have been pulled out into `.metal` files and a new manifest has been created to describe them, along with the relevant reflection data we need.
* New to shader generation is generation of runtime reflection info. To enable de-templating of `ShaderGroup`, the generator now emits functions to enable access to reflection data and shader source at runtime given a shader type enum.
* On the Metal backend, shader source has been moved into `ProgramParameters`, a required step to support shader source modification at runtime by users.
* This PR also adds additional steps to CI to invoke this tool with the amalgamation strategy.

TODO:
- [ ] Compile the Prelude (referred to here as a "link") as a Metal shader library and wire it together with all shaders that specify relevant link dependencies in their reflection data.
- [ ] Investigate an observed issue with `rules_xcodeproj` not compiling generated code from this tool correctly, when that code is emitted as an output of a genrule.

Note: This disables bazel invocation of shader generation during a build and moves back to manual invocation of the tool, pending resolution of the bazel problem.

Invoke the tool as follows to update generated files for a PR: `node scripts/shaders/generate_shader_code.js --root ./ --opengl --metal`

To enable size reduction strategies, `node scripts/shaders/generate_shader_code.js --root ./ --opengl --metal [--strip|--compress|--amalgamate]`
If `--compress` is provided, individual source files are compressed separately. If `--amalgamate` is *also* specified, all source is combined into a single compressed buffer for maximum compression efficiency.


.github/workflows/qt-ci.yml
```diff
@@ -233,6 +233,14 @@ jobs:
           key: Qt_${{ matrix.name }}_${{ matrix.qt_version }}
           max-size: 200M
 
+      - name: npm install
+        run: npm install --ignore-scripts
+        working-directory: ./
+
+      - name: Generate shaders
+        working-directory: ./
+        run: node shaders/generate_shader_code.js --root ./ --opengl --metal --strip --compress --amalgamate
```
ntadej, line 242: I'm on holidays so just seeing a Qt comment on my phone. Is this for testing or is it a build dependency? I would really like to avoid npm dependence for Qt.

ntadej (changes requested):

.github/workflows/qt-ci.yml
mwilsnd: This is just for CI to try and validate shaders are kept up to date with PRs that change them. When changes to shaders are made, this tool is run to re-generate them and changes are (supposed to be) submitted with the PR. During normal builds, you shouldn't need to run this tool.

This is also run on CI for the Linux platform so Bloaty can see the compressed version of shaders, while we only check in the human readable versions.

ntadej: Ok, thanks for clarification.

ntadej (approved):
Approving not to avoid delays. Trusting what is done is ok 😊

shaders/generate_shader_code.js
```diff
@@ -315,8 +124,188 @@ struct ShaderSource<BuiltIn::None, gfx::Backend::Type::OpenGL> {
...
+        parser.add_argument("--verbose", "--d", {
+            help: "Perform all steps but don't write anything to disk.",
+            required: false,
+            action: "store_true"
+        });
+        parser.add_argument("--dryrun", "--v", {
+            help: "Print extra process information.",
+            required: false,
+            action: "store_true"
+        });
```
stefankarschti, line 287: `verbose` and `dryrun` arguments are mixed.

louwers:
Looks unlikely to be merged at this point.

Hopefully someone wants to pick this up again sometime.

PR #1849, opened 2023-11-08
Fix tile collection update

TimSylvester:
I would have been certain that this was being cleared...  

In any case, drawables are sticking around longer than they should, causing various problems.


sjg-wdw (approved):
Nice find!

I am beginning to very much dislike how MapLibre handles loading.

PR #1850, opened 2023-11-08
Benchmark activity Android

louwers:
Add Benchmark activity for Android. I used the existing Java / Kotlin API for FPS measurement and a [TextureView](https://developer.android.com/reference/android/view/TextureView).

The World Tour Benchmark activity is available as a menu item in the test app.

You can add styles to `developer-config.xml` like so:

```xml
 <array name="benchmark_style_names">
   <item>Americana</item>
 </array>
 <array name="benchmark_style_urls">
   <item>https://zelonewolf.github.io/openstreetmap-americana/style.json</item>
</array>
```

The time it takes to call

https://github.com/maplibre/maplibre-native/blob/c8010dba7d8de0f759dc33c243a40091814848cc/platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/renderer/MapRenderer.java#L135

is measured. Which is this method:

https://github.com/maplibre/maplibre-native/blob/c8010dba7d8de0f759dc33c243a40091814848cc/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp#L155

Here is where the timing happens.

https://github.com/maplibre/maplibre-native/blob/c8010dba7d8de0f759dc33c243a40091814848cc/platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/renderer/MapRenderer.java#L85-L103

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
@@ -14,6 +14,7 @@
         android:roundIcon="@drawable/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
+
         <activity
             android:name=".activity.FeatureOverviewActivity"
             android:exported="true"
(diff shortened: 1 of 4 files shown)


louwers:
@mwilsnd OK, let's merge then I will make a follow-up PR.

mwilsnd (approved):

PR #1851, opened 2023-11-09
Complex, Animatable, Interactive Annotations Proposal

theocalmes:


wipfli:
Thanks for putting this proposal together. Sounds like a good idea to me in general. 

Could this system be used to put GIFs on a map with collisions enabled?

Could this system be used to show real-time moving markers for example of busses and trains like [here](https://mobility.portal.geops.io/world.geops.transit?layers=paerke,strassennamen,haltekanten,haltestellen,pois,world.geops.traviclive&x=951046.09&y=6001466.43&z=13.56)?

Are collisions part of this proposal or more like a stretch goal?

theocalmes:
> Thanks for putting this proposal together. Sounds like a good idea to me in general.
> 
> Could this system be used to put GIFs on a map with collisions enabled?
> 
> Could this system be used to show real-time moving markers for example of busses and trains like [here](https://mobility.portal.geops.io/world.geops.transit?layers=paerke,strassennamen,haltekanten,haltestellen,pois,world.geops.traviclive&x=951046.09&y=6001466.43&z=13.56)?
> 
> Are collisions part of this proposal or more like a stretch goal?

1) Gifs and spritesheets should definitely be supported, in fact if you look at apple maps' pin selection animation you will see it is just an animated sequence of images
2) I'd love to enable these kind of animations. In theory, by separating out the annotation layout and render path from the style system, we would have control over positioning like in the example you linked. And since annotations would be aware of the layer system, we could even do things like have a train be drawn under a road when it is below ground and over a road when it is above ground.
3) It looks like I forgot to add a bullet point for before the [stretch], but yes I would consider collisions to be a requirement and clustering to be a stretch.

sjg-wdw:
This would be a fantastic use case for the Custom Layer support we've (half) added in the Modularization & Metal development.  It's exactly the kind of thing you'd want to support as a new layer and as a first class citizen telling the rest of the toolkit what to do.

On iOS there are some really interesting possibilities for cheaply connecting textures back and forth.  I could see using those to great effect.  For Android.... it's a lot harder.  So developing those in conjunction is a good idea so you don't do something you can't do on Android.

Overall, seems like an excellent plan.

louwers:
Thanks for the proposal. I can give this a detailed read early next week.

theocalmes:
> This would be a fantastic use case for the Custom Layer support we've (half) added in the Modularization & Metal development. It's exactly the kind of thing you'd want to support as a new layer and as a first class citizen telling the rest of the toolkit what to do.
> 
> On iOS there are some really interesting possibilities for cheaply connecting textures back and forth. I could see using those to great effect. For Android.... it's a lot harder. So developing those in conjunction is a good idea so you don't do something you can't do on Android.
> 
> Overall, seems like an excellent plan.

Custom layers sound perfect for this!

Native views as annotations would be the holy grail for maximum flexibility, but also potentially quite challenging. I think even with just bitmaps we would get a big win.

There is also a world where this can be extended, to expose some of the rendering context so that users could provide 3d models to be drawn all the while having access to the various camera matrices etc. I'd love to enable, for example, drawing a 3d model of the eiffel tower much like what apple and google are doing. But this can be a separate pursuit.

sjg-wdw:
> Native views as annotations would be the holy grail for maximum flexibility, but also potentially quite challenging. I think even with just bitmaps we would get a big win.

It feels plausible on iOS.  Whole lotta details though.

wipfli:
I was about to ask if the content of the annotations could be 3d models. Can you clarify if this is part or not part of this design proposal?

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+By introducing a specialized rendering path for annotations, we bypass some of the limitations of the current layer-based system. This leads to better performance, especially in scenarios with many dynamic elements, where full layout re-calculations would otherwise cause noticeable lag and degrade the user experience.
+
+The customization options that these annotations will provide, together with the existing MapLibre stack, aim to offer a system that is more adaptable and robust than those from Google or Apple.
+
+## Proposed Change
+
+### Phase 1
+For this phase we would like to just build our support for bitmaps. The idea being that all native views eventually become bitmaps and supporting them robustly will eventually allow the renderer to draw native views correctly.
+
+* Introduce “Annotation” as a first class citizen of maplibre-gl core c++ layer. This feature would only be available to native clients and platforms and not maplibre-js.
```
louwers, line 22:
```suggestion
* Introduce “Annotation” as a first class citizen of the MapLibre Native C++ Core.
```

Since this design proposal is for the MapLibre Native repo, it is understood that it does not apply to MapLibre GL JS.

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+
+These new annotations will provide a more flexible tool for drawing user content on the map that is more inline with what developers expect from the built-in mobile map toolkits. The focus on animations and interactivity allows developers to create differentiated map experiences that feel made for mobile.
+
+By introducing a specialized rendering path for annotations, we bypass some of the limitations of the current layer-based system. This leads to better performance, especially in scenarios with many dynamic elements, where full layout re-calculations would otherwise cause noticeable lag and degrade the user experience.
+
+The customization options that these annotations will provide, together with the existing MapLibre stack, aim to offer a system that is more adaptable and robust than those from Google or Apple.
+
+## Proposed Change
+
+### Phase 1
```
louwers, line 19:
```suggestion
### Phase 1

```

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+* Introduce “Annotation” as a first class citizen of maplibre-gl core c++ layer. This feature would only be available to native clients and platforms and not maplibre-js.
+* Annotations rely on their own bitmap handling and not on the symbol layer texture atlas. Bitmaps should be able to be transformed and quickly swappable to support animations.
+* Annotations can define a label (reuse symbol layer logic)
+* Annotations provide callbacks and handling / can trigger full map re-renders. These callbacks need to be in sync with the native render loop so that animations can be synchronized properly.
+* Annotations participate in label and symbol layer collisions.
+[Stretch] Annotations support clustering, like symbol layers do.
+* Platform layers allow annotations to participate in hit tests and “onTap” events.
+* Layout, styling and positioning of Annotations can happen outside of the style’s layout and paint loops. However they should still participate in the rendering process.
+
+### Phase 2
```
louwers, line 31:
```suggestion
### Phase 2

```

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
+# Complex, Animatable, Interactive Annotations Proposal
+
+An annotation as described in this proposal is a piece of UI that is drawn on the map at a single coordinate. In Apple MapKit this would be an MKAnnotationView, and in Google world this would be similar to a Marker. Annotations are expected to be user interactable, animatable and allow for complex ui to be represented.
+
+## Motivation
+
+Most mobile map experiences display some sort of custom content on the map, often in the form of annotations. MapLibre’s current annotation system is difficult to use, performs poorly, and lacks many of the features developers expect from modern mapping libraries.
+
+This proposal introduces the concept annotations that are complex, animatable and interactive to maplibre’s core. The first phase allows bitmap backed annotations to be rendered. The second phase will enable native platform views (ios, android) to be rendered directly by maplibre-gl.
```
louwers, line 9:
```suggestion
This proposal introduces the concept of annotations that are complex, animatable and interactive to MapLibre’s core. The first phase allows bitmap backed annotations to be rendered. The second phase will enable native platform views (iOS, Android) to be rendered directly by MapLibre Native.
```

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+
+These new annotations will provide a more flexible tool for drawing user content on the map that is more inline with what developers expect from the built-in mobile map toolkits. The focus on animations and interactivity allows developers to create differentiated map experiences that feel made for mobile.
```
louwers, line 11:
```suggestion
These new annotations will provide a more flexible tool for drawing user content on the map that is more in line with what developers expect from the built-in mobile map toolkits. The focus on animations and interactivity allows developers to create differentiated map experiences that feel made for mobile.
```

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Proposed Change
+
+### Phase 1
+For this phase we would like to just build our support for bitmaps. The idea being that all native views eventually become bitmaps and supporting them robustly will eventually allow the renderer to draw native views correctly.
+
+* Introduce “Annotation” as a first class citizen of maplibre-gl core c++ layer. This feature would only be available to native clients and platforms and not maplibre-js.
+* Annotations rely on their own bitmap handling and not on the symbol layer texture atlas. Bitmaps should be able to be transformed and quickly swappable to support animations.
+* Annotations can define a label (reuse symbol layer logic)
+* Annotations provide callbacks and handling / can trigger full map re-renders. These callbacks need to be in sync with the native render loop so that animations can be synchronized properly.
```
louwers, line 25:
```suggestion
* Annotations provide callbacks and can trigger full map re-renders. These callbacks need to be in sync with the native render loop so that animations can be synchronized properly.
```

What do you mean that annotations can provide handling?

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+* Annotations participate in label and symbol layer collisions.
+[Stretch] Annotations support clustering, like symbol layers do.
```
louwers, line 27:
```suggestion
* [Stretch] Annotations support clustering, like symbol layers do.
```

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+* Platform layers allow annotations to participate in hit tests and “onTap” events.
+* Layout, styling and positioning of Annotations can happen outside of the style’s layout and paint loops. However they should still participate in the rendering process.
```
louwers, line 29: This seems to conflict with the earlier point that annotations participate in label and symbol layer collisions.

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+The customization options that these annotations will provide, together with the existing MapLibre stack, aim to offer a system that is more adaptable and robust than those from Google or Apple.
+
+## Proposed Change
+
+### Phase 1
+For this phase we would like to just build our support for bitmaps. The idea being that all native views eventually become bitmaps and supporting them robustly will eventually allow the renderer to draw native views correctly.
+
+* Introduce “Annotation” as a first class citizen of maplibre-gl core c++ layer. This feature would only be available to native clients and platforms and not maplibre-js.
+* Annotations rely on their own bitmap handling and not on the symbol layer texture atlas. Bitmaps should be able to be transformed and quickly swappable to support animations.
+* Annotations can define a label (reuse symbol layer logic)
```
louwers, line 24:
```suggestion
* Annotations can define a label (reuse symbol layer logic).
```

What does this mean exactly? Do you mean that Annotations can *have* a label?

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
```diff
@@ -0,0 +1,55 @@
...
+[Stretch] Annotations support clustering, like symbol layers do.
+* Platform layers allow annotations to participate in hit tests and “onTap” events.
+* Layout, styling and positioning of Annotations can happen outside of the style’s layout and paint loops. However they should still participate in the rendering process.
+
+### Phase 2
+Once we have a system in place that can handle bitmaps we would like to take it one step further and support native platform views. The changes described here would likely be done inside the individual platforms. 
+
+For inspiration we can look at Flutter and their rendering engine (see https://github.com/flutter/flutter/wiki/Texture-Layer-Hybrid-Composition and https://github.com/flutter/flutter/wiki/Hybrid-Composition-iOS).
+
+The big challenges here would be:
```
louwers, line 36: Please expand a bit what you mean by each of these points and why they are a challenge.

louwers (approved):
Annotations are in my eyes an essential piece of functionality of a map toolkit and making them part of the core for better performance, expanded functionality and an easier to use API sounds great. It sounds like an ambitious proposal, but if you can pull it off it would be a great improvement.

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
theocalmes: Right, and this is one of tricky parts. It makes sense to separate out the rendering of the label and the rendering of the annotation content. Labels need to know about the rest of the map to best position themselves (or in some cases hide themselves). Most of the logic exists in some form already inside of SymbolLayer, so it should be possible to optionally allow Annotations to opt into displaying a label.

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
theocalmes: Lets take it out, the word is confusing.

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
theocalmes: There is a balancing act here where we don't want every frame of an Annotation animation to trigger a global map re-layout and re-styling. But once animations "settle" (or through some other criteria, maybe a debounce) the annotation should once again participate in collisions etc.

For example, if we have an animation that moves from one point of the map to another. The movement animation needs to be smooth (maybe even at the expense of any layout which may happen underneath it). If while it move it collides with a map label (and we determine the label should be hidden), then we want to gracefully fade out that label on the next layout pass.

design-proposals/2023-11-08-complex-animatable-interactive-annotations.md
theocalmes: I can offer an iOS centric explanation, but I think there are similar challenges on Android:

1) UIViews animate using a framework called CoreAnimation, much of which actually happens on a inaccessible rendering backend. However, with Metal it is now possible to synchronize with Core Animation via https://developer.apple.com/documentation/quartzcore/cametallayer/1478157-presentswithtransaction. Core Animation will animate your view either through transforms or any number of techniques. We need to capture each frame of the animation and the state of the bitmap so that maplibre-gl can render it.
2) Touch events are notoriously tricky to get right (even the Flutter team points it out as a difficult problem). We need to come up with a way to correctly proxy the touch events from the MapView to the UIView backed annotations (which may have complex gesture recognizers attached to them). We may want to settle for limited touch support (select/deselect via single tap), but full touch will be complex.
3) The view content itself may not represent what is actually being drawn, for example imagine an annotation backed by a UI view that has parts of the view hidden but can be shown through some touch event. Which parts of this view should participate? Should the whole view (based on the frame) or the visible elements. And lastly, back to (1) how do these get sync'd with the map in case of animations. We can start with the simpler case of just looking at bounding boxes as well as checking for visibility of the superview.

theocalmes:
> I was about to ask if the content of the annotations could be 3d models. Can you clarify if this is part or not part of this design proposal?

Right I should clarify. This proposal does not include rendering of 3D models on Annotations. But I would suggest that it is something we can build the architecture around, so that it can be extended to support 3D models in the future.

I.e
Custom Layer -> AnnotationBaseLayer -> (2DAnnotation, 3DAnnotation)

PR #1853, opened 2023-11-09
Fixed metal symbol render tests

alexcristici:
Implemented the missing `SymbolDynamicUBO` in the metal  symbol shaders.

Before:
Before

After:
After


include/mbgl/shaders/symbol_layer_ubo.hpp
```diff
@@ -46,20 +46,12 @@ struct alignas(16) SymbolDrawableUBO {
...
     /* 228 */ float aspect_ratio;
-    /* 232 */ std::array<float, 2> pad;
+    /* 232 */ float fade_change;
+    /* 236 */ float pad;
     /* 240 */
 };
 static_assert(sizeof(SymbolDrawableUBO) == 15 * 16);
 
-/// Dynamic UBO
-struct alignas(16) SymbolDynamicUBO {
```
TimSylvester, line 55: This was broken out into its own UBO for performance, IIRC, is it no longer better/relevant?

alexcristici: SymbolDynamicUBO is not implemented in the shader code, that's why the symbol render tests were failing.

alexcristici: My fix was to add the `fade_change` from `SymbolDynamicUBO` to `SymbolDrawableUBO`. But I can implement `SymbolDynamicUBO` in the shader code instead.

alexcristici: What is the performance improvement here? Both UBOs are updated in the same place.

TimSylvester: I believe that value changes every frame, and the benefit was that it could be updated once and shared with all the drawables rather than updated separately for each drawable's UBO.

I guess that benefit would obviated (at least on Metal) by having the UBO bind as bytes rather than an explicit buffer.

alexcristici: I see that was implemented in GL, so I will implement in Metal too. Still not clear what is the improvement now, not in the future, the split in this moment seems to decrease performance from my point of view, because is updating the same as a single UBO and we have 2, instead of 1, also the size is bigger, `fade_change` could occupy the pad space.

src/mbgl/renderer/layers/symbol_layer_tweaker.cpp
```diff
@@ -188,16 +187,12 @@ void SymbolLayerTweaker::execute(LayerGroupBase& layerGroup, const PaintParamete
...
+            /*.pad=*/0,
         };
 
-        const SymbolDynamicUBO dynamicUBO = {/*.fade_change=*/parameters.symbolFadeChange,
-                                             /*.pad1=*/0,
-                                             /*.pad2=*/{0, 0}};
-
         auto& uniforms = drawable.mutableUniformBuffers();
         uniforms.createOrUpdate(idSymbolDrawableUBOName, &drawableUBO, context);
-        uniforms.createOrUpdate(idSymbolDynamicUBOName, &dynamicUBO, context);
```
alexcristici, line 200: @TimSylvester This is place where this UBO is update, there is no improvement now.

include/mbgl/shaders/symbol_layer_ubo.hpp
alexcristici: My point is maybe we should split this UBO when we do also the improvement too, like update less often.

alexcristici: Reverted.

src/mbgl/renderer/layers/symbol_layer_tweaker.cpp
TimSylvester: Yeah, looks like it's not even shared.  Seems like perhaps the second half of that effort got lost along the way.  @stefankarschti Are we missing something here?

I'm fine with re-combining them, or moving `DynamicUBO` to a shared UBO and only updating it once per tweaker run.

On Metal, if I can resolve the problems I'm seeing, it won't make much difference.  On GL, a shared approach should lead to fewer `glBufferData` calls, as the dynamic UBO can be updated only once per frame and that keeps the drawable UBO from needing to be updated even when the matrix, etc., hasn't changed.  But that probably benefits the benchmark more than it does for more realistic scenarios.

alexcristici: I implemented the `SymbolDynamicUBO` on Metal too, and reverted my old approach. In case we want to merge them we can do them in a separate PR.

TimSylvester (approved):

stefankarschti (approved):

mwilsnd (approved):

Issue #641, opened 2022-12-27
Feature parity with the GL JS implementation (muli-sprites)


smellyshovel:
This PR https://github.com/maplibre/maplibre-gl-js/pull/1805 was recenly merged. In short, it adds support for multiple sprite declaraions per one style file. All the details are inside. The spec was slightly updated (but in a backward-compatible manner), so the changes should also be addressed here to maintain the feature parity between the implementations.

PR #1858, opened 2023-11-13
Implementation of multi-sprites in mbgl-core

geolives:
In order to solve https://github.com/maplibre/maplibre-native/issues/641.
Implemented in the same way as for Maplibre GL JS.

### In detail
In Style JSON, `sprite` property can now be a string or an array. If this is an array that must contains object(s) with id and urls.
`SpriteLoader` can now manage multiple sprites in parallel.
`Style::Impl` now stores the information about how much sprites are loaded. Included a method `areSpritesLoaded()` to be adequate with the old boolean property that serves the same goal.

When using a string as sprite, that creates internally a Sprite object with id default and the url.
When using an array, there is a parsing to create adequate Sprite objects.

### Usage
In images identifiers you need to prepend the sprite id to the image. Example :
```
"icon-image": "hiking:myimage",
```
That leads to the image identifier 'myimage' for the sprite object with id 'hiking'.
If id of your sprite is 'default', you don't need to prepend it :
```
"icon-image": "myimage",
```

**Tip:** If your icon-image is property driven, you can use this instead :
```
"icon-image": [
          "concat",
          "hiking:",
          [
            "get",
            "myproperty"
          ]
        ],
```


louwers (changes requested):
Thank you for your contribution.

Please don't use raw pointers. Use references (preferred) or unique/shared pointes instead.

Please add tests to `style_parser.test.cpp` to validate both the old (single string URL) and new (multiple sprites in array) are parsed correctly. Does an empty array get parsed correctly? What is the behavior when specifying the same sprite id twice? Some inspiration can be taken from https://github.com/maplibre/maplibre-gl-js/pull/1805/files#diff-3f25c81a5d56e4b9c90f8e63d1bdc10eeaf8e0de78818ff567d87c5ff5950710

We need to copy the render tests from MapLibre GL JS.

src/mbgl/style/parser.hpp
```diff
@@ -27,7 +27,7 @@ class Parser {
 
     StyleParseResult parse(const std::string&);
 
-    std::string spriteURL;
+    std::vector<std::unique_ptr<Sprite>> sprites;
```
louwers, line 30: I think using unique pointers but then using `.get()` to get a raw pointer is dangerous because now someone might hold onto a dangling pointer.

I would try to avoid using raw pointers as much as possible, in this case since `Sprite` is such a lightweight object copying is probably even OK.

src/mbgl/style/style_impl.hpp
```diff
@@ -86,10 +86,12 @@ class Style::Impl : public SpriteLoaderObserver,
     Immutable<std::vector<Immutable<Layer::Impl>>> getLayerImpls() const;
 
     void dumpDebugLogs() const;
+    bool areSpritesLoaded() const;
 
     bool mutated = false;
     bool loaded = false;
-    bool spriteLoaded = false;
+    bool countOfSprites = 0;
```
louwers, line 93: Use `= false`.

louwers, line 93: Probably a `size_t` as intended here.

src/mbgl/style/parser.hpp
geolives: I rewrited `Sprite` to be used as single objets and not references. Sprites are no longer stored into `std::unique_ptr` and are passed by copy. I think also for 2 strings it's not worth using `std::unique_ptr` and potentially struggle with references.
By copying, we don't use raw pointers from `.get()` anymore.

src/mbgl/style/conversion/sprite.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <mbgl/style/conversion/constant.hpp>
+#include <mbgl/style/conversion_impl.hpp>
+
+#include <array>
+
+namespace mbgl {
+namespace style {
+namespace conversion {
+
+std::optional<std::unique_ptr<Sprite>> Converter<std::unique_ptr<Sprite>>::operator()(const Convertible& value,
```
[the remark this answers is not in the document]
geolives: As Sprite are no longer stored in `std::unique_ptr` I consider the converter method is simplified.

src/mbgl/style/style_impl.cpp
```diff
@@ -255,12 +264,16 @@ Source* Style::Impl::getSource(const std::string& id) const {
     return sources.get(id);
 }
 
+bool Style::Impl::areSpritesLoaded() const {
+    return (countOfSpritesLoaded >= countOfSprites);
```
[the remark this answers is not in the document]
geolives: I added an assert about this to be sure it cannot trigger issues. And it does!
It seems, when a sprite is expired, the `onSpriteLoaded()` method can be called mutliple times. Seen with the cpp unit test `Map.ExpiredSpriteSheet`. So it will be an issue.

I updated the code to store a summary of sprite object into a map in the form `std::unordered_map<std::string,bool>` in order to solve the issue.
Sadly, I still have some memory issues I need to check before the final commit.

src/mbgl/style/style_impl.hpp
geolives: Loading of sprites are now managed with a map of ids/booleans to dodge possible issues with multiple calls to onSpriteLoaded.

geolives:
I implemented checks about id uniqueness.
I added also 2 style parsing tests `sprites-missing-fields` and `sprites-not-same-ids` within the `style_parser` test directory to check attributes of sprites and id uniqueness.
4 additional tests (`SpriteAsArrayEmpty`,`SpriteAsArraySingle`, `SpriteAsArrayMultiple` and `SpriteAsString`) added to test compatibility with array of sprite or single sprite url inside `style_parser.test.cpp`.

Seems to be good for me. Tell me if something is wrong.

scripts/generate-style-code.list
```diff
@@ -0,0 +1,42 @@
+/private/var/tmp/_bazel_cbrasseur/4fc5a82355f9b4b5805311889652eaea/rules_xcodeproj.noindex/build_output_base/execroot/Maplibre/bazel-out/ios-x86_64-min12.0-applebin_ios-ios_x86_64-dbg-ST-1a083bdc8518/bin/src/mbgl/style/layers/fill_layer_properties.hpp
```
louwers, line 1: I think this file was comitted by accident. We should probably `.gitignore` it.

louwers:
@geolives Will do a manual test today!

wipfli:
Since this feature already exists in MapLibre GL JS if I remember correctly, can we copy-paste the render test from there?

louwers:
@wipfli Yes I think so. @geolives Let me know if you want me to take a look at this, because I have been meaning to do that anyway.

geolives:
It's probably better if you can do it, as my understandings about the render tests are minor.
Thanks :)

louwers (approved):
Thanks for your patience with this one. The new render tests passed locally. Thanks for this awesome contribution!

louwers:
iOS render test passes locally.

On CI it complains about missing metrics.

metrics/linux-gcc8-release/render-tests/sprites/array-default-only/metrics.json
```diff
@@ -0,0 +1,35 @@
+{
```
alexcristici, line 1: This file should be also in `ios-render-test-runner` in order to be found on iOS platform.

metrics/linux-gcc8-release/render-tests/sprites/array-multiple/metrics.json
```diff
@@ -0,0 +1,35 @@
+{
```
alexcristici, line 1: This file should be also in `ios-render-test-runner` in order to be found on iOS platform.

metrics/linux-gcc8-release/render-tests/sprites/array-default-only/metrics.json
alexcristici: And in `android-render-test-runner` for Android platform.

metrics/linux-gcc8-release/render-tests/sprites/array-multiple/metrics.json
alexcristici: And in `android-render-test-runner` for Android platform.

Issue #1672, opened 2023-09-20
Metal Render Tests


stefankarschti:
Run the render tests for Metal and start fixing things.

PR #1859, opened 2023-11-14
Collision drawables shared buffers

stefankarschti:
Switch collision boxes and circles to shared vertex buffers and fix a few render tests failures. 

Ref #1672 
Related to #1855 

Before:
Before

After:
After


TimSylvester (approved):

alexcristici (approved):

sjg-wdw (approved):

src/mbgl/mtl/vertex_attribute.cpp
```diff
@@ -4,78 +4,303 @@
...
+                break;
+            case gfx::AttributeDataType::UInt:
+                get<uint32_t>(value);
+                break;
+            case gfx::AttributeDataType::Float:
+                get<float>(value);
+                break;
+            default:
+                assert(false);
+                buffer += stride;
```
louwers, line 49: Could you explain these?

src/mbgl/mtl/vertex_attribute.cpp
```diff
@@ -4,78 +4,303 @@
...
+        buffer += stride;
+    }
+
+private:
+    gfx::AttributeDataType dataType;
+    std::uint8_t* buffer{nullptr};
+    std::size_t stride{0};
+
+    template <typename To, typename From, typename = std::enable_if_t<std::is_assignable_v<To&, From>>>
+    void get(const From value) {
```
louwers, line 283: `get` is not a good name for what this does.

louwers: `set` seems more applicable?

src/mbgl/mtl/vertex_attribute.cpp
```diff
@@ -4,78 +4,303 @@
...
+        assert(stride == sizeof(To));
+        *reinterpret_cast<To*>(buffer) = static_cast<To>(value);
+        buffer += stride;
+    }
+
+    template <typename To,
+              typename From,
+              std::size_t Size,
+              typename = std::enable_if_t<std::is_assignable_v<To&, From>>>
+    void get(const std::array<From, Size>& value) {
```
louwers, line 293: This one does not seem to be used. YANGI?

src/mbgl/mtl/vertex_attribute.cpp
```diff
@@ -4,78 +4,303 @@
...
+
+private:
+    gfx::AttributeDataType dataType;
+    std::uint8_t* buffer{nullptr};
+    std::size_t stride{0};
+
+    template <typename To, typename From, typename = std::enable_if_t<std::is_assignable_v<To&, From>>>
+    void get(const From value) {
+        assert(stride == sizeof(To));
+        *reinterpret_cast<To*>(buffer) = static_cast<To>(value);
```
louwers, line 285: I know this is just an internal helper class, but this seems asking for trouble. It takes one off by one error or one call to many to operator() to introduce an out of bounds write.

Adding this functionality to `mtl::VertexAttribute` may be a more memory safe alternative to consider.

src/mbgl/mtl/vertex_attribute.cpp
stefankarschti: sure, I'll make a few renames to make it clearer.

stefankarschti:
I think we don't need static vertex attribute buffer building anymore. I removed it to save on size.

PR #1881, opened 2023-11-21
Add ankerl::unordered_dense containers

mwilsnd:
Add `ankerl::unordered_dense` containers as a default build option, replacing std containers in various hot-paths. This yields a notable performance improvement on Android and Windows platforms.


mwilsnd:
@ntadej @louwers I think its time to retire GCC 8. Ran into what appears to be a defect in `std::pmr` with GCC 8 alone - trying to use `dynamic_cast` when RTTI is off. See [this run](https://github.com/maplibre/maplibre-native/actions/runs/6948394481/job/18904296023).

No other compilers tested (MSVC 17 2022, Clang 11, 14, GCC 10, 12) show the same defect. I've temporarily forked this library to not try and include PMR if RTTI is off but before this is merged I want my temporary fork reverted.

ntadej:
Which platform is that?

mwilsnd:
Looks like Ubuntu.

ntadej:
Can you just revert the fork then? I'll have a look and probably then we can either try to update the compiler or disable the test.

mwilsnd:
I've reverted the fork modification that addressed GCC 8.

ntadej:
Can you please try updating `.github/actions/qt5-build/Dockerfile` and set the image to `FROM ghcr.io/maplibre/linux-builder:centos8-latest`? I think this should fix it. I can later prepare an image with older CMake version so we have sufficient test coverage.

mwilsnd:
Changed the image but no dice. I think as long as GCC 8 is trying to build this the workflow will fail.

ntadej:
Hmm, I thought this image has a newer compiler. Let me update it tomorrow.

ntadej:
You can rebase now.

mwilsnd:
Awesome! Thanks for taking a look.

sjg-wdw (approved):

TimSylvester (approved):

alexcristici (approved):

stefankarschti (approved):

PR #1882, opened 2023-11-21
Mark layer renderability inline, don't copy render items

mwilsnd:
Copying render items from the orchestrator out of the layer set instead of iterating directly is only faster on legacy because of the number of times the list is traversed. On the new renderer we don't iterate this as much so we're wasting time doing this copy.

Layer renderability was left with a TODO that wasn't addressed, I've addressed this as well in this PR.


src/mbgl/renderer/render_orchestrator.cpp
```diff
@@ -334,6 +335,8 @@ std::unique_ptr<RenderTree> RenderOrchestrator::createRenderTree(
     }
 
     // Update all sources and initialize renderItems.
+    std::vector<bool> updateList(orderedLayers.size());
```
louwers, line 338: I think you should move the comment to above the for-loop, and maybe add a comment to `updateList`?

src/mbgl/renderer/render_layer.hpp
```diff
@@ -189,6 +189,9 @@ class RenderLayer {
     /// @param willRender Indicates if this layer should render or not
     /// @param changes The collection of current pending change requests
     virtual void markLayerRenderable(bool willRender, UniqueChangeRequestVec& changes);
+
+    /// Returns the current renderability mode if the layer
```
louwers, line 193:
```suggestion
    /// Returns the current renderability mode of the layer
```

louwers (approved):

stefankarschti (approved):

louwers (approved):

alexcristici (approved):

PR #1883, opened 2023-11-21
Don't use std::function in the renderer hot paths

mwilsnd:
Replaces use of `std::function` in hot paths with template functions to support naked lambdas. Switching to template functions meant lifting some hidden implementation into headers, which I've tried to limit somewhat.


louwers:
Nice. I wonder how this will affect binary size.

include/mbgl/renderer/layer_group.hpp
```diff
@@ -154,15 +203,78 @@ class LayerGroup : public LayerGroupBase {
...
 protected:
-    struct Impl;
-    std::unique_ptr<Impl> impl;
+    using DrawableCollection = std::set<gfx::UniqueDrawable, gfx::DrawableLessByPriority>;
+    DrawableCollection drawables;
 };
 
+template <typename Func /* void(gfx::Drawable&) */>
+void visitLayerGroupDrawables(mbgl::LayerGroupBase& layerGroup, Func dg) {
+    switch (layerGroup.getType()) {
```
mwilsnd, line 242: I'm not super happy with this, but I can't specialize a template member function that also accepts a lambda without using `std::function` to erase the lambda type.

stefankarschti: maybe something like this would work?

```
        if constexpr (!std::is_invocable<Func, gfx::Drawable&>::value) {
               static_assert(...)
        }
```

mwilsnd: The problem stems from the polymorphism of LayerGroupBase - a template virtual function isn't possible, so we need to know what kind of layer group we're dealing with to then invoke the right instance's visit template.

louwers: Are there more lightweight `std::function` alternatives?

louwers:
Did you benchmark the difference?

mwilsnd:
> Did you benchmark the difference?

I just did a fresh run on my Android device,

`no-std-func`
Americana: average=475.570162304881, low1p=40.14525684862589
FB Lite: 393.6385052027523, low1p=48.10081148400923

`main (5dbe89dea1954cf9a6bf4f2a94373c9b20136479)`
Americana: average=379.6550905650921, low1p=42.84222499315657
FB Lite: average=360.5070261859236, low1p=42.77173683287311

The device was allowed to cool off between runs, so these should be fairly comparable as far as peak SoC performance. I did a second run of the latter test (on `main`) by first putting the phone in the freezer for a few minutes. Chilled phone, no observable difference, so these numbers appear genuine.

mwilsnd:
And another run off this branch,

 Americana: average=386.8102953349323, low1p=43.50911167783764
 FB Lite: average=374.67087503739106, low1p=43.52770034993408
 
 Probably had the build from unordered_dense stuck on the phone in the initial run.

include/mbgl/renderer/layer_group.hpp
mwilsnd: I don't think so. There are two aspects to overhead with `std::function`, allocation and indirection. We probably weren't incurring allocation overhead in any of these changes, but we were paying the indirection cost.

`std::function`'s type erasure works by using virtual dispatch. The compiler can't inline these virtual function calls as the functions provided could vary and *possibly* carry an allocated closure.

stefankarschti (approved):

include/mbgl/renderer/layer_group.hpp
louwers: This solution is pretty easy to understand. There may be a fancier solution but looks OK to me.

PR #1885, opened 2023-11-21
Fill: Don't query UBOs to determine shader variant

mwilsnd:
Profiling directed my attention to these map lookups. Allocating drawable data for this profiled a bit faster.

Given the tiny amount of data we need here, another option might be to tag the drawable data pointer which should let us avoid allocating at all. (Use the 3 LSB positions of the 64 bit pointer to indicate 4 distinct values - the fill variant).


stefankarschti:
For lines I introduced a simple `std::size_t` field called `type` in `gfx::Drawable`, complete with accessors. It's less type safe but it's an option to use instead of a DrawableData descendant...

Ref: https://github.com/maplibre/maplibre-native/blob/3aacc0aae8a537c92a128a45f949dbc223d0d096/src/mbgl/renderer/layers/line_layer_tweaker.cpp#L135C1-L138C1

src/mbgl/renderer/layers/render_fill_layer.cpp
```diff
@@ -521,21 +522,36 @@ void RenderFillLayer::update(gfx::ShaderRegistry& shaders,
...
+                case gfx::FillVariant::FillOutlinePattern: {
+                    uniforms.createOrUpdate(
+                        idFillOutlinePatternInterpolateUBOName, &getFillOutlinePatternInterpolateUBO(), context);
+                    uniforms.createOrUpdate(
+                        idFillOutlinePatternTilePropsUBOName, &getFillOutlinePatternTilePropsUBO(), context);
+                    break;
+                }
+                default: {
+                    assert(false);
+                    std::terminate();
```
louwers, line 553: I don't think we do this anywhere else in the codebase. Not sure what the program behavior is if this is ever hit, but it can't be any worse than an apparant crash.

louwers (changes requested):
See comment.

src/mbgl/renderer/layers/render_fill_layer.cpp
mwilsnd: Oops, addressed with a debug log.

stefankarschti (approved):

louwers (approved):

cgalvan (approved):

alexcristici (approved):

PR #1888, opened 2023-11-22
Metal debug lines

stefankarschti:
Use the polyline generator to create triangulated lines for tile borders in Metal.

image

Ref: #1855


sjg-wdw:
Rather use the GL lines on GL if we can.

alexcristici (approved):
Nice! 👍 

Before:
Before


After:
After

mwilsnd (approved):

cgalvan (approved):

PR #1897, opened 2023-11-28
Polyline generator refactor: use templates instead of std::function

stefankarschti:
Use template arguments instead of `std::function` in `PolylineGenerator` for performace reasons.


include/mbgl/gfx/polyline_generator.hpp
```diff
@@ -7,97 +7,823 @@
...
-    GetSegmentFunc getSegment;
-    Indexes& indexes;
+template <class Vertices,
+          class LayoutVertexFunc,
+          class Segments,
+          class CreateSegmentFunc,
+          class GetSegmentFunc,
+          class Indexes>
+void PolylineGenerator::generate(Vertices& vertices,
+                                 LayoutVertexFunc layoutVertex,
```
mwilsnd, line 144: I think in this case (and for `addCurrentVertex`), it's probably preferable to stick with `std::function`. What I expect to happen here is each unique lambda instantiates a new instance of this template. Since there is a lot of code in this template function, each new lambda given duplicates all this code. That quickly sends the trade-off in the other direction towards `std::function` the moment more than one lambda is passed to this template. Since each lambda is unique and the template approach is an almost certain inline, that prevents the compiler from de-duping the 95+% of the function body that is otherwise identical.

Issue #1899, opened 2023-11-28
Re-enable ignored metal render tests


alexcristici:
4 ignored render tests on Metal:

- 2 will be fixed after triangulated lines will support data driven properties:
render-tests/fill-outline-color/zoom-and-property-function
render-tests/fill-outline-color/property-function

- 2 because of the layer z-fighting https://github.com/maplibre/maplibre-native/issues/1847:
render-tests/text-pitch-alignment/map-text-depthtest
render-tests/fill-extrusion-color/function

PR #1900, opened 2023-11-28
Metal render tests ignore

alexcristici:
Ignored the 4 failing render tests on Metal:
Screenshot 2023-11-28 at 21 51 19

We should re-enable those 4 failing render tests when they will be fixed: https://github.com/maplibre/maplibre-native/issues/1899

Also `ignoreProbing` is now enabled to ignore the probe tests.
Maybe we should consider to re-baseline probes at some point.


louwers (approved):
Thanks! Looking forward to flip the switch.

PR #1903, opened 2023-11-30
Metal permutations

TimSylvester:
Compile a different variation of the shader for each combination of values passed through attributes vs. uniforms, allowing the fragment shader to use the uniforms directly instead of passing everything through from the vertex shader.  Having a large number of fragment staging elements was causing performance degradation.

This only applies a few types with the greatest performance issues observed; symbols, lines, and fills (+circles).  We could wring out some additional performance by applying the same changes to other shaders.

Also eliminates vertex attribute declarations and bindings when uniforms are being used, converts some values from `float` to `half`, and moves some code out of the shader templates into non-templated classes to reduce code size.

```
| boston | 2.2309 ms | 4.4627 ms |
| paris | 1.9586 ms | 12.6372 ms |
| paris2 | 1.9285 ms | 14.9594 ms |
| alps | 1.9657 ms | 25.5018 ms |
| us east | 1.7381 ms | 19.4209 ms |
| greater la | 1.7491 ms | 15.9696 ms |
| sf | 3.8067 ms | 17.9621 ms |
| oakland | 1.8359 ms | 9.1588 ms |
| germany | 2.5683 ms | 26.3344 ms |
Average frame encoding time: 2.1980 ms
Average frame rendering time: 16.2674 ms
```

```
| boston | 2.3524 ms | 3.8479 ms |
| paris | 1.9614 ms | 8.8417 ms |
| paris2 | 1.5754 ms | 10.1237 ms |
| alps | 1.8416 ms | 19.1445 ms |
| us east | 1.3510 ms | 14.5941 ms |
| greater la | 1.6276 ms | 10.5458 ms |
| sf | 3.2322 ms | 16.8448 ms |
| oakland | 1.8104 ms | 6.6524 ms |
| germany | 2.5182 ms | 20.0616 ms |
Average frame encoding time: 2.0300 ms
Average frame rendering time: 12.2952 ms
```

Screenshot 2023-12-04 at 10 47 52 AM

### DIFF
--- a/include/mbgl/shaders/mtl/shader_group.hpp
+++ b/include/mbgl/shaders/mtl/shader_group.hpp
@@ -50,16 +50,17 @@ public:
         constexpr auto& source = ShaderSource::source;
         constexpr auto& vertMain = ShaderSource::vertexMainFunction;
         constexpr auto& fragMain = ShaderSource::fragmentMainFunction;
+        constexpr auto permutations = usePermutations();
 
-        const size_t key = usePermutations()
+        const size_t key = permutations
                                ? util::order_independent_hash(propertiesAsUniforms.begin(), propertiesAsUniforms.end())
                                : 0;
-        const std::string shaderName = usePermutations() ? getShaderName(name, key) : name;
+        const std::string shaderName = permutations ? getShaderName(name, key) : name;
 
         auto shader = get<mtl::ShaderProgram>(shaderName);
         if (!shader) {
             DefinesMap additionalDefines;
-            if (usePermutations()) {
+            if (permutations) {
                 addAdditionalDefines(propertiesAsUniforms, additionalDefines);
             }
 
@@ -73,9 +74,13 @@ public:
                 throw std::runtime_error("Failed to register " + shaderName + " with shader group!");
             }
 
+            shader->setBindMissingAttributes(!permutations);
+
             using ShaderClass = shaders::ShaderSource<ShaderID, gfx::Backend::Type::Metal>;
             for (const auto& attrib : ShaderClass::attributes) {
-                shader->initAttribute(attrib);
+                if (!permutations || !propertiesAsUniforms.count(attrib.nameID)) {
+                    shader->initAttribute(attrib);
+                }
             }
             for (const auto& uniform : ShaderClass::uniforms) {
                 shader->initUniformBlock(uniform);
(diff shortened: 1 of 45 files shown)


alexcristici (approved):

sjg-wdw:
That's good news!

stefankarschti (approved):

include/mbgl/shaders/mtl/shader_group.hpp
```diff
@@ -57,6 +92,15 @@ class ShaderGroup final : public gfx::ShaderGroup {
         return shader;
     }
 
+protected:
+    static constexpr bool usePermutations() noexcept {
+#if NO_METAL_PERMUTATIONS
```
louwers, line 97: I assume you used this for benchmarking. Does it make sense to keep it in?

test/util/hash.test.cpp
```diff
@@ -9,46 +9,169 @@
...
+#include <mbgl/shaders/mtl/raster.hpp>
+#include <mbgl/shaders/mtl/symbol_icon.hpp>
+#include <mbgl/shaders/mtl/symbol_sdf.hpp>
+#include <mbgl/shaders/mtl/symbol_text_and_icon.hpp>
+#endif
+
 using namespace mbgl;
 
+// Enable for troubleshooting
+#define LOG_STEPS 0
```
louwers, line 39: Do you want to keep this in? Please make it a `bool`.

test/gl/gl_functions.test.cpp
```diff
@@ -1,3 +1,5 @@
+#if MLN_RENDER_BACKEND_OPENGL
```
louwers, line 1: Maybe it's better to handle these at the build config level? Not sure.

louwers (changes requested):
Please provide a quick summary in the PR description what shader permutations are (for the uninitiated), why they are being introduced, which shaders you added a permutations for and generally what permutations exist of these shaders.

This will be helpful for future reference. Thanks!

include/mbgl/shaders/mtl/shader_group.hpp
TimSylvester: My only concern is if there are cases where shader compilation time ends up a problem.  But it would be easy enough to put back in

test/gl/gl_functions.test.cpp
TimSylvester: I tried that, but the `select` didn't seem to know about `//:metal_renderer` there.

louwers (approved):

louwers:
Much appreciated!

test/gl/gl_functions.test.cpp
louwers: OK, I'll have a look at it sometime.

PR #1911, opened 2023-12-04
OpenGL: Monotonic UBO allocator

mwilsnd:
Addresses a major performance regression when using UBOs: each buffer instance allocating its own buffer from the API.

This PR creates a new buffer sub-allocation scheme and a means to pool and recycle larger pages of buffer memory. This notably accelerates UBO performance by streaming unsynchronized writes to memory guaranteed to be unused by the GPU, eliminating the possibility of a pipeline stall during buffer updates.

How it works:
* A UBO is created or attempts to update its contents
* The allocator locates the next free block of memory, maps that buffer region in unsynchronized  fashion and writes
* That memory region is marked as occupied and a strong reference linking the memory region and UBO together is created
* Once a buffer page fills up, it gets moved to an 'in-flight' list where it waits to either become free, or -
* De-fragmentation happens on a set interval to reclaim buffers with poor occupancy. This has a pretty important implication: buffers must be relocatable and *must* be bound every frame.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1067,6 +1067,8 @@ if(MLN_WITH_OPENGL)
             ${PROJECT_SOURCE_DIR}/src/mbgl/gl/command_encoder.hpp
             ${PROJECT_SOURCE_DIR}/src/mbgl/gl/context.cpp
             ${PROJECT_SOURCE_DIR}/src/mbgl/gl/context.hpp
+            ${PROJECT_SOURCE_DIR}/src/mbgl/gl/fence.cpp
+            ${PROJECT_SOURCE_DIR}/src/mbgl/gl/fence.hpp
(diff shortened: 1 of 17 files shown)


TimSylvester:
Would it be practical to have `RelocatableBuffer` be a member rather than a base of `UniformBufferGL`, maybe via a mandatory `getRelocatableBuffer` on `OwnerClass` types?

mwilsnd:
> Would it be practical to have `RelocatableBuffer` be a member rather than a base of `UniformBufferGL`, maybe via a mandatory `getRelocatableBuffer` on `OwnerClass` types?

Sure, we can do that. I made it a base mostly out of convenience.

louwers:
!benchmark android

stefankarschti (approved):

alexcristici (approved):

PR #1923, opened 2023-12-06
Metal permutations 2

TimSylvester:
Apply preprocessor permutations to the rest of the Metal shaders, and remove support code for dynamic permutations.

Optimizes fragment stage items by reducing precision and removing where possible.

Relatively minor performance effect with the static benchmark, as expected since the previous work covered the bottlenecks, but should increase the number of layers/elements that can be displayed before running into staging/interpolator limitations.

main:
```
Average frame encoding time: 1.9609 ms
Average frame rendering time: 12.4482 ms
```

metal-permutations-2:
```
Average frame encoding time: 1.9174 ms
Average frame rendering time: 12.2038 ms
```

For an arbitrary frame, the fragment interpolator peaks at around 80%.

image

image

### DIFF
--- a/include/mbgl/renderer/layer_tweaker.hpp
+++ b/include/mbgl/renderer/layer_tweaker.hpp
@@ -17,11 +17,6 @@ class UniformBuffer;
 using UniformBufferPtr = std::shared_ptr<UniformBuffer>;
 } // namespace gfx
 
-namespace shaders {
-enum class AttributeSource : int32_t;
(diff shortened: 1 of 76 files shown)


sjg-wdw (approved):
Makes sense.

alexcristici (approved):

stefankarschti (approved):

louwers (approved):

PR #1934, opened 2023-12-08
Add explicit frame capture

TimSylvester:
In profiling the benchmark app, I saw some results indicating particular frames, but wasn't able to tie that to what's going on in those frames:

Screenshot 2023-12-07 at 10 18 57 AM

This code, when enabled, captures specific frames without involving the Xcode capture UI.  It also allows for capturing more than the arbitrary limit of 5 frames imposed by that UI.

image


louwers (approved):
Neat!

sjg-wdw (approved):
Looks good.

mwilsnd (approved):

stefankarschti (approved):

PR #1940, opened 2023-12-09
[node][opengl-2] Update vcpkg so windows builds in workflow can succeed

acalcutt:
This PR updates vcpkg in the 'opengl-2' to match the main branch. This is needed for the windows node builds to succeed

This is a cherry pick of two commits which touch vcpkg 
* Update vcpkg to e65af7b ([#1433](https://github.com/maplibre/maplibre-native/issues/1433))
* [Windows] Update libwebp ([#1662](https://github.com/maplibre/maplibre-native/issues/1662))

there are a few other commits that update vcpkg in main, but they get superseded by the two PRs above, so they don't need to be included.

### DIFF
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -36,15 +36,15 @@ if(WIN32)
     find_package(libuv REQUIRED)
 
     target_link_libraries(
-        mbgl-cache PRIVATE uv_a
+        mbgl-cache PRIVATE $<IF:$<TARGET_EXISTS:libuv::uv_a>,libuv::uv_a,libuv::uv>
     )
 
     target_link_libraries(
-        mbgl-offline PRIVATE uv_a
+        mbgl-offline PRIVATE $<IF:$<TARGET_EXISTS:libuv::uv_a>,libuv::uv_a,libuv::uv>
     )
 
     target_link_libraries(
-        mbgl-render PRIVATE uv_a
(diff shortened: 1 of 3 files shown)


ntadej:
Hmm, I guess I could merge this as is, but is the CI expected to use Node 21 already?

I guess you can't enable the Node CI for all platforms first as this change is not really tested by anything now.

acalcutt:
This PR doesn't matter the version of node, the vcpkg update is just so the the windows build can work again. right now it fails in this branch, but the vcpkg update fixes that.

For the node 21 question, we have only ever published even stable releases, so before we published 14, 16, 18. In https://github.com/maplibre/maplibre-native/pull/1941 it has been updated to node 16, 18, 20 . I also considered keeping this on Ubuntu 20.04 , but the self hosted arm64 runner has already been update and we no longer have a 20.04 arm64 to build on.

acalcutt:
Without the vcpkg update it gets this error
https://github.com/WifiDB/maplibre-native/actions/runs/7148148808/job/19468776378

acalcutt:
I think the CI that is failing is actually 'node-ci-mac.yml' , which also has the name 'node-ci' . that name being the same lead to some confusion in testing. Looking at what is failing, I think this will work again after https://github.com/maplibre/maplibre-native/pull/1941 , which has the updates that allow node 20/21 to build.

ntadej:
I can't seem to rebase the MR like I usually can on `main`. Can you please do it manually?

acalcutt:
I think this should be re-based now, and all checks but the macos arm64 one we are aware of have passed

PR #1941, opened 2023-12-09
[node][opengl-2] Update workflow for node 21 + allow publish from opengl-2 branch

acalcutt:
This PR allows a node release to be published from the 'opengl-2' branch.  This would be run by selecting the 'opengl-2' branch when running then workflow
image

Changes
- added back macos building
- allow publish steps to work outside main
- put 'fallback-to-build' back false. I had changed this to true in another PR long ago, but it should be false since the gyp build isn't set up to work.
- I realized for node 20 support that node-pre-gyp needed the abi list updated. since we are still using my version of that package, I went and updated it to add support for the latest node versions.

Cherry Picks
- update for node 21 ([#1794](https://github.com/maplibre/maplibre-native/issues/1794))
- Fix flaky Node.js CI with cache ([#1443](https://github.com/maplibre/maplibre-native/issues/1443))


ntadej:
I think we should just merge the two Node workflows. Here the the frequency of PRs is so low that we can afford also running Linux and Windows.

ntadej (approved):
macOS ARM fails due to missing font NotoCJK (we need @louwers to fix it, I think).

Windows build fails due to outdated VCPKG which will come in a separate PR.

louwers:
@ntadej I installed it.

acalcutt:
I tried to re-run that macos arm64 job that failed to see if it would pick up the update, but it still failed. I'm not sure it would pick up the new changes yet though.

ntadej:
I'll merge this. We can fix the missing fonts in parallel as I think this has already been failing earlier.

PR #1946, opened 2023-12-11
Line layer UBO improvements

alexcristici:
Extracted `units_to_pixels` field from line drawable UBOs in a separate UBO which is updated per layer.


sjg-wdw (approved):
I'm interested to see how this is affecting those update numbers.

stefankarschti (approved):
Nice work! Render times look to be around legacy levels now!

mwilsnd (approved):

PR #1947, opened 2023-12-11
[bazel] Switch to bzlmod

keith:
This is the new and improved way to manage dependencies in bazel. It is the default in bazel 7.x but works fine with 6.x until we want to update to that.


louwers:
Neat! Thank you. Let me check if Dependabot works with Bzlmod already, then it looks good to go if CI passes.

louwers (approved):

keith:
Yep should make keeping these updated much easier!

keith:
Thanks!

Issue #1910, opened 2023-12-04
Remove dependency on globally installed Node to build iOS


bsudekum:
Currently, to build iOS, there is a requirement that the caller has node installed globally:

https://github.com/maplibre/maplibre-native/blob/454efb64ac0aa549883d8401858687ef419b8d54/BUILD.bazel#L37

Ideally, we could remove this dependency or at least update the workspace to include it.

louwers:
Yes, we could use https://github.com/bazelbuild/rules_nodejs instead.

PR #1948, opened 2023-12-12
[bazel] Run nodejs inside of bazel for genrules

keith:
This fixes https://github.com/maplibre/maplibre-native/issues/1910 by using rules_js.


louwers:
Should we maybe stick to package-lock.json so we don't have two separate lockfiles?

keith:
Unfortunately I think that's a requirement of these rules, but I'm asking around to see if that interpretation is true

platform/ios/src/MLNMapView+IBAdditions.h
```diff
@@ -1,6 +1,5 @@
 #import <Foundation/Foundation.h>
-
-@class MLNMapView;
```
keith, line 3: extending a forward declared class isn't allowed, i think this worked before by things being automatically imported?

platform/macos/src/MLNMapViewDelegate.h
```diff
@@ -1,12 +1,16 @@
+#import <AppKit/AppKit.h>
 #import <Foundation/Foundation.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class MLNMapView;
 @class MLNAnnotationImage;
+@class MLNMapCamera;
```
keith, line 7: more missing forward declarations I assume because of import pollution

keith:
ok once we merge the 7.x bump and I rebase this I think it will look good!

keith:
As is this fixes https://github.com/maplibre/maplibre-native/issues/2009 by replacing the umbrella header with one generated by bazel, and taking some part of https://github.com/maplibre/maplibre-native/pull/2010/. This probably results in the import being different for folks using `#import <MapLibre/MapBox.h>` syntax, but I think for folks using `@import MapLibre` (or from Swift) there is no noticeable difference in that. This does remove some headers as is that we might want to add back though:

```diff
--- /dev/fd/63	2024-01-29 14:06:43
+++ /dev/fd/62	2024-01-29 14:06:43
@@ -1,5 +1,4 @@
 MLNAnnotation.h
-MLNAnnotationContainerView.h
 MLNAnnotationImage.h
 MLNAnnotationView.h
 MLNAttributedExpression.h
@@ -10,15 +9,13 @@
 MLNCircleStyleLayer.h
 MLNClockDirectionFormatter.h
 MLNCluster.h
-MLNCompactCalloutView.h
 MLNCompassButton.h
 MLNCompassDirectionFormatter.h
 MLNComputedShapeSource.h
-MLNConversion.h
 MLNCoordinateFormatter.h
+MLNCustomDrawableStyleLayer.h
 MLNDefaultStyle.h
 MLNDistanceFormatter.h
-MLNFaux3DUserLocationAnnotationView.h
 MLNFeature.h
 MLNFillExtrusionStyleLayer.h
 MLNFillStyleLayer.h
@@ -32,14 +29,10 @@
 MLNLineStyleLayer.h
 MLNLocationManager.h
 MLNLoggingConfiguration.h
-MLNMapAccessibilityElement.h
 MLNMapCamera.h
 MLNMapProjection.h
 MLNMapSnapshotter.h
 MLNMapView+IBAdditions.h
-MLNMapView+Impl.h
-MLNMapView+Metal.h
-MLNMapView+OpenGL.h
 MLNMapView.h
 MLNMapViewDelegate.h
 MLNMultiPoint.h
@@ -47,6 +40,7 @@
 MLNOfflinePack.h
 MLNOfflineRegion.h
 MLNOfflineStorage.h
+MLNOpenGLStyleLayer.h
 MLNOverlay.h
 MLNPointAnnotation.h
 MLNPointCollection.h
@@ -55,10 +49,6 @@
 MLNRasterDEMSource.h
 MLNRasterStyleLayer.h
 MLNRasterTileSource.h
-MLNReachability.h
-MLNRendererConfiguration.h
-MLNRendererFrontend.h
-MLNScaleBar.h
 MLNSettings.h
 MLNShape.h
 MLNShapeCollection.h
@@ -67,7 +57,6 @@
 MLNSource.h
 MLNStyle.h
 MLNStyleLayer.h
-MLNStyleLayerManager.h
 MLNStyleValue.h
 MLNSymbolStyleLayer.h
 MLNTilePyramidOfflineRegion.h
@@ -77,30 +66,10 @@
 MLNUserLocation.h
 MLNUserLocationAnnotationView.h
 MLNUserLocationAnnotationViewStyle.h
-MLNUserLocationHeadingArrowLayer.h
-MLNUserLocationHeadingBeamLayer.h
-MLNUserLocationHeadingIndicator.h
-MLNValueEvaluator.h
 MLNVectorStyleLayer.h
 MLNVectorTileSource.h
+MapLibre.h
 Mapbox.h
-NSArray+MLNAdditions.h
-NSBundle+MLNAdditions.h
-NSCoder+MLNAdditions.h
-NSComparisonPredicate+MLNAdditions.h
-NSCompoundPredicate+MLNAdditions.h
-NSDate+MLNAdditions.h
-NSDictionary+MLNAdditions.h
-NSException+MLNAdditions.h
 NSExpression+MLNAdditions.h
-NSOrthography+MLNAdditions.h
 NSPredicate+MLNAdditions.h
-NSString+MLNAdditions.h
-NSURL+MLNAdditions.h
 NSValue+MLNAdditions.h
-NSValue+MLNStyleAttributeAdditions.h
-UIColor+MLNAdditions.h
-UIDevice+MLNAdditions.h
-UIImage+MLNAdditions.h
-UIView+MLNAdditions.h
-UIViewController+MLNAdditions.h
```

keith:
@louwers fun one for you

louwers:
This looks good to me.

> This does remove some headers as is that we might want to add back though:

I think these are intended to be private headers. If someone complains though, it will be pretty easy to fix.

> This probably results in the import being different for folks using #import <MapLibre/MapBox.h> syntax

I think before the 6.0.0 release it used to be `#import <Mapbox.h>`. It was not intended to change to `MapLibre/Mapbox.h`... But `@import MapLibre` should be the prefered way to consume the library.


One question though: what steps are needed to update npm packages now?

keith:
So when node packages are updated now the build will fail CI until this new file is re-sync'd with:

```
bazel run -- @pnpm//:pnpm --dir $PWD install --lockfile-only
```

louwers (approved):

bsudekum:
Hey folks, can we merge this one?

keith:
@louwers want to merge this one?

louwers:
@keith @bsudekum Yes! Thanks for the reminder (am out of office, back this Friday). 🙏

PR #1956, opened 2023-12-13
[bazel] Update to 7.0.0

keith:


louwers:
This should really not fail the tests built with CMake....

louwers:
I think we may also be able to replace this now?

https://github.com/maplibre/maplibre-native/blob/ff35e38c6fc37bc36392b0e6e089f93ca4d6ca5f/BUILD.bazel#L176-L206

keith:
Ok it looks like the linux issue was that with bazel 7.x the tests that rely on X cannot be sandboxed anymore. That doesn't really surprise me and I think it's low risk to opt them out.

I agree we can replace the custom platform logic but if it's ok with you I'd rather do that in a separate change

keith:
@louwers ok I think this one is ready for you!

louwers (approved):
Seems to work. Thanks!

louwers:
It seems that coverage generation is bugged. It counts a lot of new lines as missing coverage, even though they are executed.

For example:

https://app.codecov.io/gh/maplibre/maplibre-native/commit/12960e373f0965978a5e76f05fbc1abc27dc2818/blob/src/mbgl/programs/fill_program.cpp

image

keith:
interestingly I saw that as well but my most recent commit made that disappear, have you ever seen it be flaky before?

louwers:
@keith No it has been working pretty reliably I think.

keith:
Interestingly the lcov files I pulled from the CI jobs from the good and bad commits are roughly identical, the only differences is slight run counts on some functions, but not anything being covered or not.

bsudekum:
@louwers any ideas what we can do here? Eager to get this PR and https://github.com/maplibre/maplibre-native/pull/1948 in 🙏 .

louwers:
@bsudekum Open an upstream issue maybe? We need our coverage tests!

louwers:
OK... CI is all green. I'll keep an eye on it.

PR #1877, opened 2023-11-20
Design proposal: ambient cache initialization speed improvement

geolives:
Added design proposal for ambient cache initialization speed improvement to fix issue https://github.com/maplibre/maplibre-native/issues/1815

I recreated a pull request to isolate the design proposal for the implementation. Sorry for the inconvenience.
Original pull request here : https://github.com/maplibre/maplibre-native/pull/1839

PR #1959, opened 2023-12-14
Ambient cache initialization speed improvement

geolives-contact:
In order to solve https://github.com/maplibre/maplibre-native/issues/1815

Here is the matching design proposal : https://github.com/maplibre/maplibre-native/pull/1877

### DIFF
--- a/platform/default/src/mbgl/storage/offline_database.cpp
+++ b/platform/default/src/mbgl/storage/offline_database.cpp
@@ -484,7 +484,6 @@ std::optional<int64_t> OfflineDatabase::extractResourceDataSize(const Resource& 
 }
 
 std::optional<int64_t> OfflineDatabase::hasResource(const Resource& resource) {
-    
     // First, try to find the resource in the 'resources' table
 
     std::optional<int64_t> selectResourcesResult = extractResourceDataSize(
@@ -736,7 +735,6 @@ std::optional<int64_t> OfflineDatabase::hasTile(const Resource::TileData& tile) 
             "SELECT length(data) FROM ambient_tiles WHERE url_template = ?1 AND pixel_ratio = ?2 AND x = ?3 AND y = ?4 "
             "AND z = ?5");
         if (selectAmbientTilesResult) {
-
             mapbox::sqlite::Query insertQuery(
                 getStatement("INSERT INTO tiles (url_template, pixel_ratio, z, x, y, expires, modified, etag, data, "
                              "compressed, accessed, must_revalidate) "


geolives:
The OfflineDatabase.CorruptDatabaseOnQuery still fails.

We would need to create a corrupted database in version 7 of the database model, which opens without any error, but where the first query triggers error of corruption

(with test database in version 6, simply instantiating the object triggers the corruption error when it tries to create the new tables for the model upgrade, which makes the test fail)

louwers:
How did you make the databases?

platform/ios/app/MBXAppDelegate.m
```diff
@@ -11,6 +11,12 @@ @implementation MBXAppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+    NSArray *paths = NSSearchPathForDirectoriesInDomains(
+                                                         NSApplicationSupportDirectory,
+                                                         NSUserDomainMask, YES);
+    NSLog(@"%@/com.geolives.maplibre.app/.mapbox/cache.db",[paths objectAtIndex:0]);
```
louwers, line 17: This looks like it is leftover from debugging.

geolives: Yes, sorry, I have just fixed it.

geolives:
@louwers

For databases used for merging / sideloading tests (satellite_test.db, sideload_ambient.db, sideload_sat.db, and sideload_sat_multiple.db), I simply upgraded them with a SQLite database tool by running the upgrade script to model version 7.

For v6.db database (which is used to test the upgrade from version 6 to version 7), I made a copy of v5.db and it with a SQLite database tool by running the upgrade script to model version 6.

I still haven't managed to create the upgraded corrupt-delayed.db database (a corrupted database in version 7 of the database model, which opens without any error, but where the first query triggers error of corruption) to make the OfflineDatabase.CorruptDatabaseOnQuery test succeed :-(

geolives:
CorruptDatabaseOnQuery test has been temporarily disabled as discussed with @louwers 
Could you please proceed with merge? :-)
Thanks a lot.

platform/default/src/mbgl/storage/offline_database.cpp
```diff
@@ -574,135 +735,178 @@ std::optional<std::pair<Response, uint64_t>> OfflineDatabase::getTile(const Reso
...
+        // std::cout << "-------- HASTILE - FOUND IN TILES\n";
+        return selectTilesResult;
+    } else {
+        // If not found in 'tiles', try the 'ambient_tiles' table
+        std::optional<int64_t> selectAmbientTilesResult = extractTileDataSize(
+            tile,
+            "SELECT length(data) FROM ambient_tiles WHERE url_template = ?1 AND pixel_ratio = ?2 AND x = ?3 AND y = ?4 "
+            "AND z = ?5");
+        if (selectAmbientTilesResult) {
+            // std::cout << "-------- HASTILE - FOUND IN AMBIENT_TILES\n";
```
louwers, line 756: There are commented out lines that print used for debugging throughout this file.

Please remove them or change them to (debug) logging.

louwers:
I ran the `offline_database.benchmark.cpp` benchmarks:

```
$ python3 vendor/benchmark/tools/compare.py --no-color benchmarks ./mbgl-benchmark-runner-main ./mbgl-benchmark-runner --benchmark_filter='OfflineDatabase/.*'

Benchmark                                                     Time             CPU      Time Old      Time New       CPU Old       CPU New
------------------------------------------------------------------------------------------------------------------------------------------
OfflineDatabase/InvalidateRegion                           -0.0228         -0.0226        144618        141326        144231        140973
OfflineDatabase/DeleteRegion                               +0.0586         +0.0588      22971082      24316111      22911370      24259681
OfflineDatabase/InsertTileRegion                           -0.0590         -0.0587        236233        222301        235708        221865
OfflineDatabase/InvalidateAmbientCache                     -0.5156         -0.5156         81160         39313         81146         39308
OfflineDatabase/ClearAmbientCache                          +0.0617         +0.0620      23023497      24444877      22966607      24389660
OfflineDatabase/InsertTileCache                            -0.0130         -0.0131        218526        215695        218123        215269
OfflineDatabase/InsertBigTileCache                         -0.0167         -0.0166       2157706       2121650       2157331       2121517
OfflineDatabase/GetTile                                    +0.0092         +0.0092        138014        139284        138001        139266
OfflineDatabase/AddTilesToFullDatabase                     -0.0043         -0.0043        166905        166193        166889        166178
OfflineDatabase/AddTilesToDisabledDatabase                 +0.0309         +0.0309           932           961           932           961
OfflineDatabase/GetTileFromDisabledDatabase                +0.0375         +0.0376           744           772           744           772
OfflineDatabase/ResizeDatabase                             +0.0141         +0.0144            11            11            11            11
OVERALL_GEOMEAN                                            -0.0516         -0.0515             0             0             0             0
```

geolives:
Hello @louwers 

I have just removed the cout logs in comments.

About the performance test, the improvements are significant when there is a lot of offline data downloaded as offline regions (for example 1,5 GB). The startup of the app is much faster in that case :-)

platform/default/src/mbgl/storage/offline_database.cpp
```diff
@@ -173,7 +178,7 @@ void OfflineDatabase::createSchema() {
     db->exec("PRAGMA synchronous = FULL");
     mapbox::sqlite::Transaction transaction(*db);
     db->exec(offlineDatabaseSchema);
-    db->exec("PRAGMA user_version = 6");
+    db->exec("PRAGMA user_version = 7");
```
louwers, line 181: Doesn't this mean that `ambient_resources` and `ambient_tiles` is never created?

platform/default/src/mbgl/storage/offline_database.cpp
```diff
@@ -403,147 +469,224 @@ std::optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const
...
+        }
+        return selectResourcesResult;
+    }
 }
 
-bool OfflineDatabase::putResource(const Resource& resource,
-                                  const Response& response,
-                                  const std::string& data,
-                                  bool compressed) {
+bool OfflineDatabase::putResource(
```
louwers, line 518: There is too much code duplication in this function and the `putTile` function below.

Instead of adding a `bool` I would for example define an `enum class` that can be converted to a `string` to get the related database name using another utility function.

louwers:
I downloaded 4GB of offline data.

This is a build before this PR. It loads pretty quickly...?

https://youtube.com/shorts/CZqLHsIXg4I?si=abdOzSF47ZvxOeJ5

platform/default/src/mbgl/storage/offline_database.cpp
geolives: If I have understood the code correctly, the two tables are created in the schema creation script (specified by the offlineDatabaseSchema variable), which is located in offline_schema.hpp. Isn't that the case?

platform/default/include/mbgl/storage/offline_schema.hpp
```diff
@@ -35,6 +35,35 @@ static constexpr const char* offlineDatabaseSchema =
     "  must_revalidate INTEGER NOT NULL DEFAULT 0,\n"
     "  UNIQUE (url_template, pixel_ratio, z, x, y)\n"
     ");\n"
```
louwers, line 37: This file is generated and should not be modified manually.

platform/default/src/mbgl/storage/offline_database.cpp
louwers: Ah you are right, I missed that that was updated. Thanks!

platform/default/include/mbgl/storage/offline_schema.hpp
louwers: Could you try updating `offline_schema.js` and running that?

platform/default/src/mbgl/storage/offline_database.cpp
```diff
@@ -574,135 +718,172 @@ std::optional<std::pair<Response, uint64_t>> OfflineDatabase::getTile(const Reso
...
-        "WHERE url_template = ?1 "
-        "  AND pixel_ratio  = ?2 "
-        "  AND x            = ?3 "
-        "  AND y            = ?4 "
-        "  AND z            = ?5 ") };
-    // clang-format on
+    // First, try to find the tile in the 'tiles' table
+    std::optional<int64_t> selectTilesResult = extractTileDataSize(
+        tile,
+        "SELECT length(data) FROM tiles WHERE url_template = ?1 AND pixel_ratio = ?2 AND x = ?3 AND y = ?4 AND z = ?5");
```
louwers, line 728: This will always fail for users not using any offline functionality at all. I'm not comfortable proceeding without first ensuring the performance impact is minimal, otherwise we need to keep track of if the `tiles` table contains any tiles at all and skip this check.

louwers (changes requested):
Again apologies for taking so long for me to review this, but I wasn't at all familiar with this subsystem of the library.

It looks like the problematic `initAmbientCacheSize` method was only introduced so that cache size could be determined without taking into account offline data. https://github.com/mapbox/mapbox-gl-native/pull/15622

It seems these changed can be reverted after the split?

By the way, I was not able to reproduce the performance problems. Are you calling `setMaximumAmbientCacheSize` on startup? It seems somehow I was able to store 3GB of data with `OfflineActivity.kt`, without calling`setMaximumAmbientCacheSize` and without an `evict` being triggered (these are the only places that call `initAmbientCacheSize`).

platform/default/include/mbgl/storage/offline_schema.hpp
geolives: OK, I didn't know that the file was autogenerated, sorry, I will try to do that :-)

platform/default/src/mbgl/storage/offline_database.cpp
geolives: We agree with you about the code duplication issue, but we encountered difficulties in resolving it :-(

Initially, we attempted to minimize code duplication in methods like getTile(), getResource(), putTile(), putResource(), etc., by using lambdas and c_str() to concatenate the table name in the query. However, this approach led to memory issues (EXC_BAD_ACCESS) and erratic behavior, particularly errors during the insertion of tiles/resources, for example : 
_[DEBUG] {abort at 23 in [INSERT INTO ambient_tiles (url_template, pixel_ratio, x, y, z, modified, must_revalidate, etag, expires, accessed, data, compressed) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, ?12)]: NOT NULL constraint failed: ambient_tiles.compressed}[Database](1299): %s_

Consequently, we had to revert our changes, even though it resulted in increased code duplication.

You can find here an old version of the code here with our trials for getTile() and getResource() method: https://github.com/geolives/maplibre-native/commit/cb8cf110e3675b022b99b0424d1db44a7b1a86d3

As our expertise in C++ is not as strong as it is in other higher-level languages, we welcome any assistance from the community to help resolve this issue of code duplication without causing adverse effects :-)

Thanks in advance.

louwers:
@geolives Alright, I will try iOS as well.

What do you think about removing `initAmbientCacheSize()` and some of the [other changes](https://github.com/mapbox/mapbox-gl-native/pull/15622) introduced alongside it? It does not serve a purpose anymore after the database is split does it?

geolives:
> What do you think about removing `initAmbientCacheSize()` and some of the [other changes](https://github.com/mapbox/mapbox-gl-native/pull/15622) introduced alongside it? It does not serve a purpose anymore after the database is split does it?

It requires further investigation because we still need to compute the ambient cache size to avoid exceeding the maximum chosen size. However, you are right that we need to understand how it was done before integrating https://github.com/mapbox/mapbox-gl-native/pull/15622

platform/default/src/mbgl/storage/offline_database.cpp
geolives: We agree with you on this point, a boolean variable which would be initialized on startup (by counting tiles / resources, or even by checking if there is an offline_region in the database or not) and set when we add the first resource / tile is a good idea in our opinion, we will try to implement that.

louwers:
@geolives-contact Do you still want to get this merged?

louwers:
No problem and thanks for the update.

PR #1966, opened 2023-12-17
Add Bloaty iOS size test for PRs

### LINKED ISSUE #1797: Bloaty report for iOS dynamic library
Now that Bloaty is set up to report on the Linux renderer binary, we can expand its usage to iOS by generating a report for the iOS app size.

image

cc @mwilsnd

louwers:
Closes  #1797 to report size tests results generated with Bloaty on iOS. These tests should be a bit more reliable than the Linux size test we have now.

As always working on GitHub actions workflows may take a couple of tries to get it right. There is some duplication between `pr-bloaty_ios.yml` and `pr-tests.yml` but let's fix that once we decide if we want to keep the Linux size test (and this new one works).

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -164,6 +164,45 @@ jobs:
             ${{ env.ios_cpp_test_artifacts_dir }}/CppUnitTests.xctest.zip
             ${{ env.ios_cpp_test_artifacts_dir }}/CppUnitTestsApp.ipa
 
+      # Size test (Bloaty)
+
+      - name: Build dynamic library for size test (Bloaty)
+        if: matrix.renderer == 'drawable'
+        run: |
+          bazel build //platform/ios:MapLibre.dynamic --//:renderer=metal --//:maplibre_platform=ios --compilation_mode="opt" --copt -g --strip never --output_groups=+dsyms --apple_generate_dsym
+          bazel_bin="$(bazel info --compilation_mode="opt" bazel-bin)"
+          unzip "$bazel_bin"/platform/ios/MapLibre.dynamic.xcframework.zip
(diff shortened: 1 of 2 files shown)


.github/workflows/ios-ci.yml
```diff
@@ -164,6 +164,45 @@ jobs:
             ${{ env.ios_cpp_test_artifacts_dir }}/CppUnitTests.xctest.zip
             ${{ env.ios_cpp_test_artifacts_dir }}/CppUnitTestsApp.ipa
 
+      # Size test (Bloaty)
+
+      - name: Build dynamic library for size test (Bloaty)
+        if: matrix.renderer == 'drawable'
+        run: |
+          bazel build //platform/ios:MapLibre.dynamic --//:renderer=metal --//:maplibre_platform=ios --compilation_mode="opt" --copt -g --strip never --output_groups=+dsyms --apple_generate_dsym
```
mwilsnd, line 172: Build with `-Oz` here to prevent the inliner behavior from distorting the optimal size configuration. Otherwise things may get aggressively inlined which distorts the metric we're interested in.

Additionally, if we can, build only for armv8/AArch64. By default I think the xcframework should have both binaries in it (x86-64 and armv8). Bloaty gets confused by that and tends to pick one, usually x86-64 in my testing. Arm binaries are going to be bigger just by virtue of the instruction set, so make sure we're comparing arm.

mwilsnd (changes requested):

.github/workflows/ios-ci.yml
louwers: Yes it contains both:

```
$ ls MapLibre.xcframework
Info.plist*                  ios-arm64/                   ios-arm64_x86_64-simulator/
```

But I am extracting the armv8 dynamic library from the XCFramework:

```
cp MapLibre.xcframework/ios-arm64/MapLibre.framework/MapLibre MapLibre_dynamic
```

```
 $ file MapLibre_dynamic
MapLibre_dynamic: Mach-O 64-bit dynamically linked shared library arm64
```

So should be fine.

mwilsnd (approved):

PR #1967, opened 2023-12-17
Sunset non-Metal iOS builds on CI

louwers:
Switch all iOS builds and tests to use Metal on CI.


platform/darwin/test/MLNDocumentationExampleTests.swift
```diff
@@ -459,6 +459,8 @@ class MLNDocumentationExampleTests: XCTestCase, MLNMapViewDelegate {
     }
     
     func testMLNMapSnapshotter() {
```
louwers, line 461:
```suggestion
    func testMLNMapSnapshotter() throws {
```

stefankarschti (approved):

cgalvan (approved):

PR #1980, opened 2023-12-22
UBO by index instead of map

alexcristici:


TimSylvester (approved):

include/mbgl/gfx/uniform_block.hpp
```diff
@@ -101,31 +92,21 @@ class UniformBlockArray {
...
     /// Do something with each block
-    template <typename Func /* void(const StringIdentity, const UniformBlock&) */>
+    template <typename Func /* void(const size_t, const UniformBlock&) */>
     void visit(Func f) {
-        std::for_each(uniformBlockMap.begin(), uniformBlockMap.end(), [&](const auto& kv) {
-            if (kv.second) {
-                f(kv.first, *kv.second);
+        std::for_each(uniformBlockVector.begin(), uniformBlockVector.end(), [&](const auto& block) {
+            if (block) {
+                f(0, *block);
```
stefankarschti, line 99: Did you want to send the UBO index? If it's never used perhaps remove the first parameter entirely?

src/mbgl/shaders/gl/shader_program_gl.cpp
```diff
@@ -149,23 +151,21 @@ std::shared_ptr<ShaderProgramGL> ShaderProgramGL::create(Context& context,
...
-        GLsizei length = 0;
+    for (const auto& blockInfo : uniformBlocksInfo) {
+        GLint index = MBGL_CHECK_ERROR(glGetUniformBlockIndex(program, blockInfo.name.data()));
         GLint size = 0;
-        GLint binding = index;
-        MBGL_CHECK_ERROR(glGetActiveUniformBlockName(program, index, maxLength, &length, name.data()));
         MBGL_CHECK_ERROR(glGetActiveUniformBlockiv(program, index, GL_UNIFORM_BLOCK_DATA_SIZE, &size));
-        assert(length > 0 && size > 0);
+        assert(size > 0);
+        GLint binding = GLint(blockInfo.binding);
```
mwilsnd, line 159: Prefer `static_cast<GLint>`

include/mbgl/gfx/uniform_buffer.hpp
```diff
@@ -46,37 +42,30 @@ class UniformBuffer {
...
-
     UniformBufferArray() = default;
     UniformBufferArray(UniformBufferArray&&);
     // Would need to use the virtual assignment operator
     UniformBufferArray(const UniformBufferArray&) = delete;
     virtual ~UniformBufferArray() = default;
 
     /// Number of elements
-    std::size_t size() const { return uniformBufferMap.size(); }
+    std::size_t size() const { return uniformBufferVector.size(); }
```
mwilsnd, line 55: Expand on the comment here to explain that this size is actually an upper limit on the number of UBOs per array now.

include/mbgl/shaders/ubo_max_count.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <mbgl/shaders/line_layer_ubo.hpp>
+#include <mbgl/shaders/raster_layer_ubo.hpp>
+#include <mbgl/shaders/symbol_layer_ubo.hpp>
+
+#include <algorithm>
+
+namespace mbgl {
+namespace shaders {
+
+static constexpr auto maxUBOCountPerShader = std::max({static_cast<size_t>(backgroundUBOCount),
```
mwilsnd, line 22: So we're trading some level of maintainability here for the moment to do this. A follow up for me probably is to generate things like this as part of the revamped shader code generation.

src/mbgl/shaders/gl/shader_info.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+namespace mbgl {
+namespace shaders {
+
+UniformBlockInfo::UniformBlockInfo(std::string_view name_, std::size_t id_)
+    : name(name_),
+      id(id_),
+      binding(id_) {}
+
+const std::vector<UniformBlockInfo> ShaderInfo<BuiltIn::BackgroundShader, gfx::Backend::Type::OpenGL>::uniformBlocks = {
```
mwilsnd, line 13: For now vectors are fine here, once I can come back to the code generation PR these can be made into `std::array`s.

mwilsnd (changes requested):
Looking good! Guess I need to dust off that shader code generation PR when I have a chance and really update it for this.

include/mbgl/gfx/uniform_block.hpp
alexcristici: Fixed.

src/mbgl/shaders/gl/shader_info.cpp
alexcristici: I used vector here, because I am passing it here are param:
        ```
shader = ShaderProgramGL::create(glContext,
                                         programParameters,
                                         shaderName,
                                         firstAttribName,
                                         shaders::ShaderInfo<ShaderID, gfx::Backend::Type::OpenGL>::uniformBlocks,
                                         vert,
                                         frag,
                                         additionalDefines);
```

alexcristici: We can adjust it later and for now it's fine like this or should I find a way an change it now to arrays?

include/mbgl/shaders/ubo_max_count.hpp
alexcristici: Yes, that's true.

src/mbgl/shaders/gl/shader_info.cpp
mwilsnd: It's fine for now, I don't think there will be much difference moving to arrays. Just a thought that code generation can make a bit more maintainable.

stefankarschti (approved):

include/mbgl/gfx/uniform_buffer.hpp
alexcristici: Updated comment and changed the function name to `allocatedSize()`. Not sure if this is the best name, let me know if you have any other suggestions.

mwilsnd (approved):

Issue #1981, opened 2023-12-22
Drawable Renderer: some styles have missing layers at zoom level >= 15


louwers:
**Describe the bug**

MapTiler Outdoors and MapTiler Bright are missing some layers at zoom level >= 15 with the Drawable renderer on Android.

**To Reproduce**
1. Create drawable release build for Android (test app)
2. Open Debug Activity
3. Switch to second style (MapTiler Outdoors), zoom in to zoom level >= 15

**Expected behavior**
At zoom level >=15 buildings should continue to show

**Actual behavior**
Buildings disappear.

**Video**

https://github.com/maplibre/maplibre-native/assets/649392/0a56eccc-7c5e-4ebe-bb68-a6cd71bc4c77

louwers:
Legacy OpenGL Build (iOS) also has some problems!

![7A85DFE6-7F3C-4706-837D-2CA35E1EF810](https://github.com/maplibre/maplibre-native/assets/649392/9e341a47-fa76-445e-9011-cf6181be3ce6)

louwers:
I just tried doing a `git bisect`, but the long lived feature branch that yielded the drawable renderer was merged to preserve the commits, and in the early stages there was no drawable / legacy renderer split. So you will just find commit b052cb904ec75b0fb286c2ed467a8df98e508c1e that starts to apply the render pass to drawables.

Here is `--lat 52.377232239702366 --lon 4.8984822546846924 --zoom 15` with MapTiler Outdoors on an old commit:

![out-legacy](https://github.com/maplibre/maplibre-native/assets/649392/5b6f03a5-7008-4ab3-9fb0-79941bcb67d6)

Current main (legacy renderer):

![out](https://github.com/maplibre/maplibre-native/assets/649392/42e29ab8-3459-47fb-ab26-ffa6b9f3eef7)

Current main (drawable renderer):

![out](https://github.com/maplibre/maplibre-native/assets/649392/deeb5a3b-9bc7-44a4-a185-9eaeb431e0fb)

louwers:
I added a simple vector layer with a single triangle.

```
{
	"coordinates": [
		[
			[
				4.897674510435564,
		52.37767974030237
			],
			[
				4.900814033548812,
			52.37717322491926
			],
			[
				4.899872176615219,
			52.37836421131914
			],
			[
				4.897674510435564,
			52.37767974030237
			]
		]
	],
	"type": "Polygon"
}
```

Add to `sources`:

```
[...]

PR #1871, opened 2023-11-15
Bug/stencil tiles

TimSylvester:
Fixes incorrect ordering of tiles rendered to the stencil buffer, causing low-zoom tiles to be rendered over high-zoom tiles in some cases, particularly between the completion of loading prefetch tiles (see `setPrefetchZoomDelta`) and their removal.

This seems to be the root cause of #1845. ... and #1846

Also fixes incorrect use of `renderTiles` in the background layer and depth write options on fill drawables.

[https://osmus.slack.com/archives/C04SSR0S78B/p1700070024256889?thread_ts=1699920382.735459&cid=C04SSR0S78B](https://osmus.slack.com/archives/C04SSR0S78B/p1700070024256889?thread_ts=1699920382.735459&cid=C04SSR0S78B)

PR #1989, opened 2024-01-02
Render tile masks in order

TimSylvester:
Tile masks were being rendered into the stencil buffer in whatever order they ended up in `RenderTiles`.  When multiple sources are present, this can result in lower-detail tiles blocking higher-detail ones from rendering.

### DIFF
--- a/src/mbgl/renderer/paint_parameters.cpp
+++ b/src/mbgl/renderer/paint_parameters.cpp
@@ -188,11 +188,17 @@ void PaintParameters::renderTileClippingMasks(const RenderTiles& renderTiles) {
             // already present
             continue;
         }
+    }
 
-        if (tileUBOs.empty()) {
-            tileUBOs.reserve(count);
-        }
+    if (tileUBOs.empty()) {
+        tileUBOs.reserve(count);
+    }
 
+    // Render tile clipping masks in order of tile ID, by increasing
+    // z-value, so that more detailed tiles cover less detailed ones.
+    for (const auto& pair : tileClippingMaskIDs) {
+        const auto& tileID = pair.first;
+        const auto stencilID = pair.second;
         tileUBOs.emplace_back(shaders::ClipUBO{/*.matrix=*/util::cast<float>(matrixForTile(tileID)),
                                                /*.stencil_ref=*/stencilID,
                                                /*.pad=*/0,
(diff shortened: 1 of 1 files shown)


louwers:
This is to fix #1981 right? I am testing with maptiler-planet source from the MapTiler Outdoors style and a custom vector source with a single triangle. This is what is rendering...

this branch

out

826f17b65931cf4623711fc2e625b9ecebb408a5 (commit before 2e489ebd3e8119da0e8fed7aad4e4117c78906e8)

out-826f17b65931cf4623711fc2e625b9ecebb408a5

Prior to Metal project (`ios-v5.13.0`)

out

MapLibre GL JS

image

mwilsnd (approved):

TimSylvester:
This seems to improve things, and makes sense to me.  But looking way back, tiles have always been rendered to the stencil buffer in `RenderTiles` order which I'm pretty sure hasn't changed.

https://github.com/maplibre/maplibre-native/blob/71f6da27228c6b95ccd0e5d9e7229eba41724012/src/mbgl/renderer/paint_parameters.cpp

So I'm a little worried that I'm totally missing something here.  Perhaps the order in which sources are listed matters, and it only worked before if they were in order of increasing max-zoom?

sjg-wdw:
Might be a good idea to try a bunch of the styles visually and see if there are problems.

mwilsnd:
One observation I made - clipping masks were rendered during the opaque pass but the raster layer in the pastel style is translucent. That feels like it shouldn't work, since high zoom vector source tiles are possibly being rendered both in front and behind the raster layer. But everything is in the translucent pass. It feels like the most correct solution then would be to detect when visiting a new RenderItem with a different source and re-rendering the clipping mask for that source.

louwers (approved):
Let's merge this. On zoom level >= 14 the behavior is a bit different still (see comment), but this fixes the problem with MapTiler Outdoors.

Let's focus on extending the render tests instead of manual checking.

TimSylvester:
> ... detect when visiting a new RenderItem with a different source and re-rendering the clipping mask for that source.

I had been thinking along similar lines.  Re-considering what are the correct criteria for rebuilding the mask reminded me that the way we decide that has changed, and that may have been a mistake.

Specifically, a layer reuses the mask if the tiles are a subset of what's there, rather than set equality.  If the tiles don't overlap that makes sense, but if there can be overlaps then sometimes a subset should be considered not equivalent.

That was done because we were rebuilding the mask several times per frame in some cases, when some layers didn't produce any data for one tile or another.  But that has since changed to account for prefetch tiles, see #1871.

PR #1994, opened 2024-01-04
Stencil clear

### LINKED ISSUE #1981: Drawable Renderer: some styles have missing layers at zoom level >= 15

TimSylvester:
Reverts the subset-based check for stencil mask reset, which would consider a mask with a superset of the current render tile set valid for reuse.  When multiple sources are in use, this would lead to rendering with incorrect masks causing some tiles to not appear for some layers.  It's no longer necessary to account for different layers of the same source not generating the same set of tile drawables, per #1871, so this should not be a performance problem.

Replaces #1989
Resolves #1981

### DIFF
--- a/src/mbgl/renderer/paint_parameters.cpp
+++ b/src/mbgl/renderer/paint_parameters.cpp
@@ -101,34 +101,11 @@ using GetTileIDFunc = const UnwrappedTileID& (*)(const typename TIter::value_typ
 
 using TileMaskIDMap = std::map<UnwrappedTileID, int32_t>;
 
-// `std::includes` adapted to extract tile IDs from arbitrary collection iterators.
-// This is needed because, although it accepts iterators of different types, they must be:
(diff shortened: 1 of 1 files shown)


alexcristici (approved):

mwilsnd (approved):

louwers (approved):
Neat!

PR #1995, opened 2024-01-05
Update node release based on maplibre-js release workflow - v2 with changelog

acalcutt:
This PR updates the node release workflow so it no longer increments the version on it's own. This should allow it to work with the approval requirements on this repo. I've made this work more like the maplibre-js workflow at https://github.com/maplibre/maplibre-gl-js/blob/b7b86c1018842a139d4ced21a071ca4a3d9f24b5/.github/workflows/release.yml .

To increment the version now, someone would make a PR that changes the version in 'package.json' and release notes in 'platform\node\CHANGELOG.md'. Once the version in package.json has change, the workflow automatically checks that version against what is published on npm. If the version has not been published, the binaries get built and published to github using NODE_PRE_GYP_GUTHUB. Then the release is published to npm. 

I've also cherry picked some https://github.com/maplibre/maplibre-native/pull/1941 I made in the opengl-2 branch.

- update '@acalcutt/node-pre-gyp' to support node 20
- put '--fallback-to-build=false' in package.json. I had changed this a while back but it should be false since we don't have GYP build files
- fix 'runner.arch' that should be 'matrix.arch'
- update readme to reflect binaries being built
- remove unused node-ci-mac.yml
- I was also thinking about versioning. Maybe in the 'opengl-2' branch we can continue a 5.x release , but in main we can start with '6.0.0-pre.0' (once we get something together for a metal pre-release)

**Other fixes**
- Added automatic inserting of node changelog into release notes.
- Try to use GITHUB_TOKEN instead of NODE_PRE_GYP_GITHUB_TOKEN. Added 'contents: write' permission to allow this token to publish the release and add release notes.

**Testing**
I have created a test release (minus macos support) at
https://github.com/WifiDB/maplibre-native/actions/runs/7429441972
https://www.npmjs.com/package/@acalcutt/maplibre-gl-native-test/v/5.2.9
https://github.com/WifiDB/maplibre-native/releases/tag/node-v5.2.9

**Notes**
After this is approved I plan to cherry pick these changes over to 'opengl-2' so I could also make a updated release with macos support there, so a squash and merge may be easier if this is merged.

PR #2005, opened 2024-01-08
[node][opengl-2] Update node release based on maplibre-js release workflow - v2 with changelog

acalcutt:
This PR is a cherry pick of https://github.com/maplibre/maplibre-native/commit/43e52d054e234992ba030ffae3076b682d410324 which brings release workflow updates from https://github.com/maplibre/maplibre-native/pull/1995 into the 'opengl-2'

My goal is to create at least one more release (and pre-release) from the 'opengl-2' branch, which supports node 20 and macos

The only change I made other than the cherry pick was to update this to run on `branches: [opengl-2]` and to re-enable the macos in the workflow.


ntadej:
What should we do with the failing ARM test? I guess you need it for the release?

acalcutt:
For the arm releases, the binary is buiding but some of the tests are failing. So I was considering doing a pre-release and asking someone i know on macos to try it out. these tests have never been 100% on all platforms.

I was looking at the failed tests in https://github.com/maplibre/maplibre-native/pull/1942 . you can see how they are failing in the results at https://github.com/maplibre/maplibre-native/actions/runs/7183103165?pr=1942 . Some of the failing tests have been disable in the main branch, although the ones left seems like it is failing some type of calculation slightly.

most of the tests the failed visually don't looks that off, so I am unsure if it is a test issue or a build issue.

acalcutt:
We only usually run these tests on macos, but in that test PR I enabled them in linux too and one thing i notices is the arm64 linux version fails in a similar way with some type of lat/lon calculation slightly off. At least for that release, i can test a bit more myself once we have a pre-release of it, since I usually use the arm64 linux version locally.

louwers (approved):

louwers:
Approved based on the fact that it is a cherry-pick so already approved.

acalcutt:
@louwers Any suggestion on what I should do about the required test that isn't running required to merge this?

It looks like maybe the qt-ci build it is requiring, but it doesn't seem like it triggered. looking at that qt-ci it is set to run on this branch, so not sure what it didn't run like the node tests.

ntadej:
You can ignore that. I can merge it but I wonder if failing ARM test still allows you to make a release?

acalcutt:
yes, even with the failing arm test it would still allow a release. the release doesn't run any tests, it just builds the binary and posts it to github and publishes the npm. We can see from the tests that the build portion of this works, so that should succeed.

To publish a release I would create a new PR to change the version and CHANGELOG in this branch, like https://github.com/maplibre/maplibre-native/pull/2005/files#diff-98b4dc8f37a6a0ef8854dd55234f64ea248f90654c4027ee3fdba48e9b14a30a

PR #2007, opened 2024-01-10
[node][opengl-2] Release node-v5.3.0-pre.0 - test node release workflow

acalcutt:
This PR should push a release 'node-v5.3.0-pre.0' from the 'opengl-2' branch.  This should allow new node binaries for node 16/18/20

In the changelog, since this is a pre-release, I copied the unreleased changes from main. I wasn't sure if I should keep them under main also. i decided to leave them in both spots until a full release was made

I added two breaking changes
1.) The removal node 14 binary
2.) The change to ubuntu 22.04 may change OS requirements.

### DIFF
--- a/platform/node/CHANGELOG.md
+++ b/platform/node/CHANGELOG.md
@@ -2,15 +2,18 @@
 ## main
 
 * ...Add new stuff here...
-* [Breaking] Remove node 14 support and add node 20 support. We are now building binaries for node 16,18,20
-* [Breaking] Linux binary is now built on Ubuntu 22.04 instead of Ubuntu 20.04. it could require a OS update to use this update on linux.
+* [Note] This is a OpenGL-2 release. It does not include metal support.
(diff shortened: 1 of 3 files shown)


platform/node/CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
 
 ## main
 
+* ...Add new stuff here...
```
ntadej, line 4: Personally I would not duplicate changelog until we start making `main` and `opengl-2` releases in parallel.

ntadej (changes requested):

platform/node/CHANGELOG.md
acalcutt: I got rid of the duplication in main. 

the two things that we under 'main' before were part of the opengl-2 branch already, so this is actually the first release they will be included in.

ntadej (approved):

platform/node/CHANGELOG.md
acalcutt: I was thinking though that after the release, that maybe we changelog and package.json changes might be good to cherry pick back to main so everything is in sync version and changelog wise.

PR #2008, opened 2024-01-10
UBO fixes

alexcristici:


sjg-wdw (approved):

TimSylvester (approved):

include/mbgl/shaders/gl/drawable_fill_outline.hpp
```diff
@@ -11,7 +11,6 @@ struct ShaderSource<BuiltIn::FillOutlineShader, gfx::Backend::Type::OpenGL> {
     static constexpr const char* vertex = R"(layout (std140) uniform FillOutlineDrawableUBO {
     highp mat4 u_matrix;
     highp vec2 u_world;
-    highp vec2 pad;
```
stefankarschti, line 14: This is a needed padding field to keep the ubo size multiple of 16 bytes.

alexcristici: Then I probably need to add it in the glsl file to in sync.

src/mbgl/renderer/layers/render_line_layer.cpp
```diff
@@ -485,50 +485,92 @@ void RenderLineLayer::update(gfx::ShaderRegistry& shaders,
 
         // interpolation UBOs
         const float zoom = static_cast<float>(state.getZoom());
```
stefankarschti, line 487: `zoom` could change every frame, therefore the interpolation UBOs could also be different for data driven properties. Not updating these after the first compute could be affecting visuals?

alexcristici: This code is running per frame, so the zoom is updated per frame.

stefankarschti: correct, my bad.

stefankarschti (approved):

PR #2014, opened 2024-01-12
Enable incompatible_disallow_empty_glob

keith:
This just makes it easier to spot accidental bugs where we think we're globbing something when we're not


louwers:
Should we remove `allow_empty =False` from `glob`?

louwers (approved):

keith:
https://github.com/maplibre/maplibre-native/pull/2032

PR #2018, opened 2024-01-14
[node][main] Update latest release to node-v5.3.0

acalcutt:
This is a cherry pick of the node-v5.3.0 release on the 'opengl-2' branch. This is mainly so main package.json reflects the latest version and the CHANGELOG included the latest information.

In addition to updating the version, this will also make npm pull in the latest binaries when a user does 'npm install', since it uses the version in package.json to know which version to pull.

**Note**
This will not actually make a new release, since node-v5.3.0 is already published.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@maplibre/maplibre-gl-native",
-  "version": "5.2.0",
+  "version": "5.3.0",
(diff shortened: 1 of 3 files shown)


ntadej (approved):
Is there a reason this is still draft?

acalcutt:
It was late when I made this, so I wanted to wait until morning

Issue #2019, opened 2024-01-15
Wayland build: Could NOT find OpenGL (missing: GLX)


JonasVautherin:
*Note: opening an issue here under the advice of @louwers [on Slack](https://osmus.slack.com/archives/C02B2CBSNBU/p1705321862741179?thread_ts=1705082964.200539&cid=C02B2CBSNBU), hoping to get insights from @jwinarske.*

Trying to build maplibre-native on Wayland, I get the following error in the CMake configuration step:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenGL (missing: GLX)
```

I am running on Alpine 3.19 with Wayland.

I saw [this issue](https://github.com/maplibre/maplibre-native/pull/1018) and tried different things to remove the need for GLX (my understanding being that I don't need GLX on Wayland), like this:

```
cmake -DMBGL_WITH_EGL=ON -DMLN_WITH_WAYLAND=ON -DMLN_WITH_ELG=ON -DMBGL_WITH_WAYLAND=ON -Bbuild -S.
```

But it does not change the GLX error I get. Am I missing something to build with Wayland?

acalcutt:
If you are doing this on a headless system, maybe throwing 'xvfb-run --auto-servernum' before the command will allow it to work? I've never had to use that on cmake,

Also, Is this node binaries that you are building? I'd check out the node workflow steps at
https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/node-ci.yml#L206-L216
and
https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/node-ci.yml#L236-L243

Node needs the `DMLN_WITH_NODE=ON` now, which I you would then add the `DMLN_WITH_ELG=ON` to have it build in egl mode. I'm not familiar with that DMLN_WITH_WAYLAND flag.

acalcutt:
I remember seeing this old converation that makes me think xvfb is needed with EGL
https://github.com/maplibre/maplibre-native/pull/217#issuecomment-1132215477

Btw, I just noticed you have a type. the option is [DMLN_WITH_EGL](https://github.com/maplibre/maplibre-native/blob/main/CMakeLists.txt#L11) , not DMLN_WITH_ELG

JonasVautherin:
Thanks for the quick answer!

> If you are doing this on a headless system

No, that's my main Desktop machine :see_no_evil:.

[...]

PR #2022, opened 2024-01-15
cmake: enable MLN_WITH_EGL and OPENGL_USE_GLES for Wayland

JonasVautherin:
### What it achieves

Without these changes, I need to configure with the unintuitive:

```
cmake -DMLN_WITH_EGL=ON -DMLN_WITH_WAYLAND=ON -DOPENGL_USE_GLES3=ON -Bbuild -S.
```

With these changes, I can configure successfully with:

```
cmake -DMLN_WITH_WAYLAND=ON -Bbuild -S.
```

### The change

This change sets two values to TRUE in platform/linux/linux.cmake when `MLN_WITH_WAYLAND` is set:

* MLN_WITH_EGL: without this set, it will try to find the GLX component of OpenGL, which is an X11 thing. Without this change I get the following error:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenGL (missing: GLX)
```

* OPENGL_USE_GLES2: without this (or OPENGL_USE_GLES3) set, [the CMake module for OpenGL will include the GLVND library](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L432-L435) which is, again, for X11. Without this change I get the following error:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenGL (missing: OPENGL_opengl_LIBRARY)
```

### Further considerations 

One could ask: "is it not an issue to set OPENGL_USE_GLES2, in case someone wants OPENGL_USE_GLES3?", and my (admittedly limited) understanding is that the FindOpenGL module will still detect the right one, but we need one to be set for the check [here](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L432-L435) (otherwise it will look for the GLVND library). In my case, I have glfw-3 installed on my system, and it gets detected even with OPENGL_USE_GLES2 set. Not sure what would happen if I had both installed (my distro doesn't provide both, I can't test :see_no_evil:). Maybe it would be better to set OPENGL_USE_GLES3?

### Notes

This is my suggestion to improve the Wayland build, following my struggling with https://github.com/maplibre/maplibre-native/issues/2019.

@jwinarske: FYI

### DIFF
--- a/platform/linux/linux.cmake
+++ b/platform/linux/linux.cmake
@@ -15,6 +15,18 @@ find_package(Threads REQUIRED)
 pkg_search_module(WEBP libwebp REQUIRED)
 pkg_search_module(LIBUV libuv REQUIRED)
 
+if(MLN_WITH_WAYLAND)
+    # See https://github.com/maplibre/maplibre-native/pull/2022
+
+    # MLN_WITH_EGL needs to be set for Wayland, otherwise this CMakeLists will
+    # call find_package(OpenGL REQUIRED GLX), which is for X11.
+    set(MLN_WITH_EGL TRUE)
+
+    # OPENGL_USE_GLES2 or OPENGL_USE_GLES3 need to be set, otherwise
+    # FindOpenGL.cmake will include the GLVND library, which is for X11.
+    set(OPENGL_USE_GLES3 TRUE)
+endif()
+
 target_sources(
     mbgl-core
     PRIVATE


acalcutt:
You question about OPENGL_USE_GLES2 vs OPENGL_USE_GLES3 reminds me of https://github.com/maplibre/maplibre-native/pull/1079/files where the change to GLES3 had to be reverted because the package that included it was GLES2. I wonder if it is similar here.

jwinarske:
It's not clear to me why there are GLESv2 and GLESv3 flags.  The functions exist in the same file named libgles2.so.  There is no libgles3.so.  You can create an EGL config that works with both GLESv2 and GLESv3.  The resolver handles the differences.  Enabling config attributes should be exposed.  Apache-2.0 Example:
https://github.com/toyota-connected/ivi-homescreen/blob/main/shell/constants.h#L88

JonasVautherin:
> It's not clear to me why there are GLESv2 and GLESv3 flags.

They apparently mostly help deciding whether `find_package(OpenGL)` will include GLX and GLVND or not, [see here](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L420-L459).

The one thing that is not entirely clear to me is that it sets [OPENGL_gles2_LIBRARY](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L468) or [OPENGL_gles3_LIBRARY](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L473), which are later [searched](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L563-L564).

But Maplibre only requires the [EGL](https://github.com/maplibre/maplibre-native/blob/81fa32213de8ba68d3df292b4cac81341ff31cf3/platform/linux/linux.cmake#L69) component in the CMakeLists, not the [GLES2](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L32) or [GLES3](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L37) components.

If it did, then it would be easy to know which `OPENGL_USE_GLESX` to set. But because it doesn't, my feeling is that we just need to set one of them for [the GLX/GLVND checks](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L420-L459).

At least setting either `OPENGL_USE_GLES2` or `OPENGL_USE_GLES3` works for me, and I know I only have glfw-3 (which I had to install for this very configure step to pass).

acalcutt:
If it doesn't make much difference I would think it would be better to set "OPENGL_USE_GLES3", just because the code is ES 3.0 now https://github.com/maplibre/maplibre-native/pull/995 . pretty much everywhere else 'GLES2' got changed to 'GLES3' (there was just the one case I mentioned where that caused an issue, but most places it was fine.)

JonasVautherin:
Is there something more I should be doing on this PR? If you don't think it makes sense, we can just close it :+1:.

ntadej:
For my curiosity, where is `OPENGL_USE_GLES3` used? Because I can't find any reference in the repo, but I guess it's used in one of the dependencies?

JonasVautherin:
@ntadej: As described above, and as far as I can tell, it seems like it is mostly used [in FindOpenGL.cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L420-L459). Without OPENGL_USE_GLES2 or OPENGL_USE_GLES3 set, [the CMake module for OpenGL will include the GLVND library](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenGL.cmake#L432-L435) which is for X11.

JonasVautherin:
I don't think it will go further than this :see_no_evil:.

ntadej:
Why close, I think at the moment we just do not have anyone who can test this.

@louwers, @acalcutt, do you think this is safe to just merge as is?

acalcutt:
I personally think it is fine, since it only changes anything if that -`DMLN_WITH_WAYLAND=ON`  option is used. Thought I am not familiar with if MLN_WITH_WAYLAND option was actually an option before.

JonasVautherin:
> Can you add a comment to this PR?

Not sure what you mean :confused:. Do you want me to modify the description of the PR? :eyes:

louwers:
@JonasVautherin Sorry, what I mean is, could you add a comment to the `if()` in the CMake file that explains why it is there with a link to this PR?

JonasVautherin:
Oh, got it :blush:. I added comments there, let me know what you think!

louwers (approved):

louwers:
Thanks, looks good.

ntadej:
It seems codecov is not reporting for some reason ...

louwers:
According to https://app.codecov.io/gh/maplibre/maplibre-native/commits it is uploaded...

PR #2029, opened 2024-01-16
[opengl-2] Cherry pick "Add WebP decoding support to Linux and Windows"

acalcutt:
Cherry pick of https://github.com/maplibre/maplibre-native/pull/1314 to bring WebP decoding support to the opengl-2 branch.

WebP support has been requested a few times in TileServer-GL, like this thread https://github.com/maptiler/tileserver-gl/issues/426#issuecomment-1731116965


acalcutt:
@ntadej  I was trying to merge in this webp support cherry pick, but i notice the added test fails in QT . Do you know if this test was ignored in main, or if something was added to support webp for QT? it doesn't seem like it fails on main.

mwilsnd:
You should also pick #1662 which includes important security updates to libwebp.

acalcutt:
I already merged that vpkg update as part of https://github.com/maplibre/maplibre-native/pull/1940

acalcutt:
On the question of the qt webp  "Unsupported image type" error, it seems by default it uses this [QImage::fromData](https://github.com/maplibre/maplibre-native/blob/main/platform/qt/src/mbgl/image.cpp#L38-L40) to decode the image. But it looks like it does not support webp from https://doc.qt.io/qt-6/qimagereader.html#supportedImageFormats , so I guess an unsupported error may make sense.


I was looking at adding in code for when 'QT_IMAGE_DECODERS' is set to false, like
https://github.com/maplibre/maplibre-native/pull/2029/commits/5bdfe0b87e923a7feb347ac32b34581ddf06938e , but I think that needs a few libraries added and it doesn't seem like that is what was done in main.

acalcutt:
Guessing maybe qt might need a 'mage I/O Plugin' for WEBP?   https://doc.qt.io/qt-6/qtimageformats-index.html#the-image-i-o-plugins

ntadej:
The plugin is not built by default. The test was just disabled in main.

I would not update the GCC version in this branch unless strictly needed.

acalcutt:
Do you know where that test gets disabled?

test/util/image.test.cpp
```diff
@@ -78,6 +78,12 @@ TEST(Image, JPEGTile) {
     EXPECT_EQ(256u, image.size.height);
 }
 
+TEST(Image, WebPTile) {
```
ntadej, line 81: @acalcutt, just disable this one for Qt with `#if !defined(__QT__)`.

acalcutt:
Thanks. I'm going to close this and re-open a fresh PR without all the testing, and just apply that `#if !defined(__QT__)`

PR #2033, opened 2024-01-17
Part 1

TimSylvester:
Explicit move semantics on task scheduling.
Context retains the background scheduler so `Scheduler` doesn't recycle it. Combine `VectorTileTest` definitions.
Prevent edge case causing captures to prevent destruction defer. De-template scheduler classes for code size.
Reset global count in `Map.KeepRenderData` to avoid failures in repeated test runs.


TimSylvester:
Changing source to allow for stacking

PR #2036, opened 2024-01-17
Async cleanup fixes - Part 3

TimSylvester:
Wait for the background scheduler to be empty before releasing it

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
+++ b/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
@@ -84,7 +84,7 @@ void MapRenderer::schedule(std::function<void()>&& scheduled) {
     }
 }
 
-std::size_t MapRenderer::waitForEmpty(std::chrono::milliseconds timeout) {
+std::size_t MapRenderer::waitForEmpty(Milliseconds timeout) {
     try {
         android::UniqueEnv _env = android::AttachEnv();
         static auto& javaClass = jni::Class<MapRenderer>::Singleton(*_env);
--- a/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
+++ b/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
@@ -92,10 +92,13 @@ std::size_t MapRenderer::waitForEmpty(Milliseconds timeout) {
         if (auto weakReference = javaPeer.get(*_env)) {
             return weakReference.Call(*_env, waitForEmpty, static_cast<int64_t>(timeout.count()));
         }
-    } catch (const std::exception& exception) {
-        Log::Error(Event::Android, std::string("MapRenderer::waitForEmpty failed: ") + exception.what());
+        // If the peer is already cleaned up, there's nothing to wait for
+        return 0;
+    } catch (...) {
+        Log::Error(Event::Android, "MapRenderer::waitForEmpty failed");
+        jni::ThrowJavaError(*android::AttachEnv(), std::current_exception());
+        return 0;
     }
-    return -1;
 }
 
 void MapRenderer::requestRender() {
--- a/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
+++ b/platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
@@ -84,6 +84,20 @@ void MapRenderer::schedule(std::function<void()>&& scheduled) {
     }
 }
 
+std::size_t MapRenderer::waitForEmpty(std::chrono::milliseconds timeout) {
+    try {
+        android::UniqueEnv _env = android::AttachEnv();
+        static auto& javaClass = jni::Class<MapRenderer>::Singleton(*_env);
+        static auto waitForEmpty = javaClass.GetMethod<jni::jint(jni::jlong)>(*_env, "waitForEmpty");
+        if (auto weakReference = javaPeer.get(*_env)) {
+            return weakReference.Call(*_env, waitForEmpty, static_cast<int64_t>(timeout.count()));
+        }
+    } catch (const std::exception& exception) {
+        Log::Error(Event::Android, std::string("MapRenderer::waitForEmpty failed: ") + exception.what());
+    }
+    return -1;
+}
+
 void MapRenderer::requestRender() {
     try {
         android::UniqueEnv _env = android::AttachEnv();
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/renderer/glsurfaceview/MapLibreGLSurfaceView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/renderer/glsurfaceview/MapLibreGLSurfaceView.java
@@ -1035,7 +1035,7 @@ public class MapLibreGLSurfaceView extends SurfaceView implements SurfaceHolder.
 
     /**
      * Wait for the queue to become empty
-     * @param timeoutMillis Timeout in milliseconds
+     * @param timeoutMillis Timeout in milliseconds, zero for indefinite wait
      * @return Number of queue items remaining
     */
     public int waitForEmpty(long timeoutMillis) {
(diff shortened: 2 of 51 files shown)


test/include/mbgl/test/vector_tile_test.hpp
```diff
@@ -38,6 +38,12 @@ class VectorTileTest {
                                   imageManager,
                                   glyphManager,
                                   0};
+
+    ~VectorTileTest() {
+        // Ensure that deferred releases are complete before cleaning up
+        EXPECT_EQ(0, loop.waitForEmpty(std::chrono::milliseconds{100}));
```
mwilsnd, line 44: Having a time based condition here in a test isn't ideal, this would be a spot to either use an infinite timeout or a much longer delay.

test/util/thread.test.cpp
```diff
@@ -328,3 +328,46 @@ TEST(Thread, DeleteBeforeChildStarts) {
...
+
+    EXPECT_EQ(0, pool->waitForEmpty(std::chrono::milliseconds(500)));
+}
+
+TEST(Thread, PoolWaitTimeout) {
+    auto pool = Scheduler::GetBackground();
+
+    pool->schedule([&] { std::this_thread::sleep_for(std::chrono::milliseconds(100)); });
+
+    EXPECT_EQ(1, pool->waitForEmpty(std::chrono::milliseconds(10)));
```
mwilsnd, line 348: I'm wondering if we'd ever see the OS scheduler preempt this test in a way that could cause a failure.

mwilsnd (changes requested):
This has been running stable in a long run endurance stress test, looking good overall. Just unsure on the timeout mechanism. In particular it's theoretically possible the tests may fail if preempted by the OS.

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/renderer/textureview/TextureViewRenderThread.java
```diff
@@ -135,6 +135,31 @@ void queueEvent(@NonNull Runnable runnable) {
     }
   }
 
+  /**
+   * Wait for the queue to be empty.
+   * @param timeoutMillis Maximum time to wait, in milliseconds
+   * @return The number of items remaining in the queue
+   */
+  @UiThread
+  int waitForEmpty(long timeoutMillis) {
```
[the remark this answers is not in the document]
TimSylvester: Presumably applies to the C++ `duration` methods as well?

I'm used to allowing a wait for zero duration as a non-blocking check, a la Win32 `WaitForSingleObject`, but I can't think of a reason to use that here.  I would tend to lean toward a zero-argument overload for unbounded wait, but it means more code.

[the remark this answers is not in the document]
mwilsnd: The meaning of 0 is a bit ambiguous, so long as it's possible to wait forever I'd be satisfied.

test/include/mbgl/test/vector_tile_test.hpp
TimSylvester: I just made it an unlimited timeout under the assumption that the unit test process has a timeout that would handle any deadlock reasonably.

test/util/thread.test.cpp
TimSylvester: I made the task block on a mutex so there's no potential for order ambiguity.

platform/android/MapboxGLAndroidSDK/src/main/java/org/maplibre/android/maps/renderer/glsurfaceview/MapLibreGLSurfaceView.java
```diff
@@ -1023,6 +1033,31 @@ public void queueEvent(@NonNull Runnable r) {
       }
     }
 
+    /**
+     * Wait for the queue to become empty
+     * @param timeoutMillis Timeout in milliseconds
+     * @return Number of queue items remaining
+    */
+    public int waitForEmpty(long timeoutMillis) {
```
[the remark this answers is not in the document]
TimSylvester: I would not expect the timeout to be reset, that could lead to any length of wait when the caller would reasonably expect a best-effort to return within approximately the timeout period.

If we are going to handle things being inserted from multiple threads, I think we need to add a way for the caller to make the scheduler reject new tasks, since the point is to know it's empty so you it can be destroyed, and there would always be a window between unlocking and returning where new stuff could sneak in. 

My impression is that tasks are added on the same thread that creates/destroys the pool, but there's nothing to enforce that.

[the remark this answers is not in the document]
TimSylvester: Asserting that `schedule` is called on the thread that created the pool or on one of the pooled threads seems to mostly work, but fails from `SnapshotterRenderer` in tests, which apparently runs in a separate thread.

So yeah, I guess we need a `Scheduler::close` or something.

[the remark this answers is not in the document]
TimSylvester: On reflection, we can't use a `close` in `~Context` because the background scheduler is potentially being used by other things at the same time.

But I think it's valid to say that when a `Context` is being destroyed, no new loading should be initiated, so it's enough to wait for all the already-added tasks to complete.  Any other map, etc., using the background scheduler needs to maintain its own reference to keep the pool alive.

stefankarschti (approved):

platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
```diff
@@ -84,6 +84,20 @@ void MapRenderer::schedule(std::function<void()>&& scheduled) {
...
+        android::UniqueEnv _env = android::AttachEnv();
+        static auto& javaClass = jni::Class<MapRenderer>::Singleton(*_env);
+        static auto waitForEmpty = javaClass.GetMethod<jni::jint(jni::jlong)>(*_env, "waitForEmpty");
+        if (auto weakReference = javaPeer.get(*_env)) {
+            return weakReference.Call(*_env, waitForEmpty, static_cast<int64_t>(timeout.count()));
+        }
+    } catch (const std::exception& exception) {
+        Log::Error(Event::Android, std::string("MapRenderer::waitForEmpty failed: ") + exception.what());
+    }
+    return -1;
```
mwilsnd, line 98: This will underflow to `std::numeric_limits<size_t>::max()`., If intentional, return that instead of allowing underflow.

src/mbgl/util/thread_pool.cpp
```diff
@@ -40,20 +49,83 @@ std::thread ThreadedSchedulerBase::makeSchedulerThread(size_t index) {
...
+            queue.push(std::move(fn));
+        }
+        cvAvailable.notify_one();
     }
+}
 
-    cv.notify_one();
+std::size_t ThreadedSchedulerBase::waitForEmpty(Milliseconds timeout) {
+    // Must not be called from a thread in our pool, or we would deadlock
+    if (!isThreadOwned(std::this_thread::get_id())) {
```
mwilsnd, line 107: Thinking about this, why not store a pointer to the owner thread pool in TLS? Then we don't even need to check our own TID, just that the thread local pointer is a match. ie: (`util/thread_local.hpp`, `ThreadLocal<ThreadedSchedulerBase>`)

mwilsnd (changes requested):

platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
TimSylvester: Good catch, there really isn't a correct value there.  An exception would be more appropriate to the Java caller, I guess.

src/mbgl/util/thread_pool.cpp
TimSylvester: Yeah, that's a good idea.

platform/android/MapboxGLAndroidSDK/src/cpp/map_renderer.cpp
TimSylvester: If I understand the JNI stuff correctly the else condition there isn't really an error, but shouldn't happen on a live object.  Nothing else appears to throw in that case, so returning zero (i.e., nothing still pending) seems appropriate.

Other methods there aren't translating exceptions, though there's a facility for doing so, but that seems appropriate.

mwilsnd (approved):

PR #2042, opened 2024-01-19
UV timer hack

TimSylvester:
On some Linux systems (the AWS-hosted one provided by @louwers and @stefankarschti's local one) but not CI, the following minimal set of tests reliably fails:

    ./mbgl-test-runner --gtest_filter=Thread.InvokeBeforeChildStarts:Thread.DeleteBeforeChildStarts:Timer.Basic

Investigation shows that the run loop is running and the timer is calling the callback, but with no delay and always resulting in an actual elapsed time of zero milliseconds.

The comment in `Timer.Basic` implies that the timer is not reliable anyway, and a +/- 20% range is allowed.

Adding a call to [uv_update_time](https://docs.libuv.org/en/v1.x/loop.html#c.uv_update_time) seems to resolve the issue, and allows the range to be narrowed to +/- 1% on my system, so I conclude that the previous behavior documented in the comment is a less severe effect of the same problem, and I've narrowed the range to be more conservative.

I suspect that there's a deeper issue indicated by the interaction of the three tests listed, but this change at least improves the behavior of timers without any apparent drawbacks.

@louwers FYI, `pre-commit run --all-files` produces numerous updates to Bazel files, but I didn't commit them.

### DIFF
--- a/platform/default/src/mbgl/util/timer.cpp
+++ b/platform/default/src/mbgl/util/timer.cpp
@@ -1,10 +1,11 @@
 #include <mbgl/util/timer.hpp>
 
-#include <stdexcept>
-
 #include <mbgl/util/run_loop.hpp>
+#include <mbgl/util/string.hpp>
 
 #include <uv.h>
+
+#include <stdexcept>
 
 namespace mbgl {
 namespace util {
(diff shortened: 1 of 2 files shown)


sjg-wdw (approved):

stefankarschti (approved):

cgalvan (approved):

louwers:
@TimSylvester Yes pre-commit.ci doesn't run for the Bazel formatter because it failed sometimes.

mwilsnd (approved):
Pretty interesting, it does indeed looks like this should be getting called. `Timer::start` looks like a reasonable place to do so,

PR #2051, opened 2024-01-25
Asynchronous GeometryTile deletion

mwilsnd:
General performance augmentation. Defers destruction of a GeometryTile's mailbox to a worker thread. These tiles hold OpenGL objects which must be deleted on the correct thread, so those objects are deleted on the render thread.

A number of special concerns around synchronization had to be addressed.

The primary performance problem centers around the mailbox needing to take both mutex locks before it can be properly closed. As this was done on the render thread, it would stall the whole renderer when the map was in motion.

This started in #1928 and has since been moved to this development branch.


alexcristici (approved):
Looks great! The android benchmark runs very smooth now.

stefankarschti (approved):

src/mbgl/renderer/tile_pyramid.cpp
```diff
@@ -231,7 +233,14 @@ void TilePyramid::update(const std::vector<Immutable<style::LayerProperties>>& l
                     tilesIt->second->setNecessity(TileNecessity::Optional);
                     cache.add(tilesIt->first, std::move(tilesIt->second));
                 }
-                tiles.erase(tilesIt++);
+
+                if (tilesIt->second) {
+                    tilesIt->second->cancel();
+                    Scheduler::GetBackground()->schedule(
+                        [tile_{std::shared_ptr<Tile>(std::move(tiles.extract((tilesIt++)->first).mapped()))}]() {});
```
louwers, line 240: Somehow my brain can't parse this capture group.

mwilsnd: I had to look up how to do this actually, haha. It extracts a value from the container and move it to a shared pointer inside the lambda's capture group. Doing it this way means we can keep using the iterator for the rest of the container.

mwilsnd:
!benchmark android

src/mbgl/renderer/tile_pyramid.cpp
TimSylvester: I don't follow, why would it invalidate the iterator?

I somehow never noticed this, but I think it has the same temporary lambda copy retention issue as in `TileCache`.

TimSylvester: Is `tiles.extract((tilesIt++)->first)` important here, vs. `tiles.extract(tilesIt++)`?  It's doing an extra key search.

mwilsnd: From when I worked on it, I believe `extract().mapped()` was my best option to prevent a copy of the element while also removing it from the container. Other options didn't look like they'd work.

It may have that lambda issue, I never noticed the other instance in `TileCache`. I also haven't seen the destructor running on the main thread. To be safe, I'd say do the same here as in `TileCache`.

TimSylvester: It seems to be called somewhat rarely.  I couldn't trigger it from the benchmark or interactive, but it happens a few times in the render tests, apparently related to GeoJSON, so it might show up in performance in some scenario related to that.

mwilsnd: > Is `tiles.extract((tilesIt++)->first)` important here, vs. `tiles.extract(tilesIt++)`? It's doing an extra key search.

If it compiles and works correctly then no, I suppose it isn't.

mwilsnd: >It seems to be called somewhat rarely.

If it gets called at all then I say we should fix it, there is always a possibility of threading issues if that is allowed to happen.

TimSylvester (approved):
The additional locks in the managers are still a potential performance issue, I think, but I don't feel like we should mess around with it more unless they are observed to be relevant.

I'd prefer to further reduce direct use of the `Scheduler` static variables to help make it easier to reason about what can and must be using the same thread pool, but not enough to keep this open.

TimSylvester:
ok, I think this one is finally ready

PR #2058, opened 2024-01-26
Remove custom platforms setting

keith:
With bazel 7.x and rules changes we can use the native platforms instead


keith:
depends on https://github.com/maplibre/maplibre-native/pull/1956

louwers (approved):

PR #2060, opened 2024-01-27
Combine multiple segments into a drawable when `sortFeaturesByKey` is not used

TimSylvester:
Maintain the grouping of segments when they don't need to be sorted individually.

The good news is that this turned out to be pretty simple using a new data structure rather than trying to make use of the variant defined for the legacy renderer.  Everything is a reference, so it shouldn't have any significant runtime cost.

The bad news is that I have yet to find an example where the number of segments is greater than 1, so this doesn't seem to have any benefit in practice.

The other good news, though, is that the bucket does routinely combine numerous labels and icons into a single segment, so it's probably not as important as we thought.

Screenshot 2024-01-26 at 4 03 57 PM

&nbsp;
Screenshot 2024-01-26 at 4 01 54 PM

### DIFF
--- a/src/mbgl/mtl/drawable.cpp
+++ b/src/mbgl/mtl/drawable.cpp
@@ -24,6 +24,9 @@
 #include <simd/simd.h>
 
 #include <cassert>
+#if !defined(NDEBUG)
+#include <sstream>
+#endif
 
 namespace mbgl {
 namespace mtl {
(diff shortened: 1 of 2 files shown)


sjg-wdw (approved):
Well, at least we can take that off our list.
Glad it wasn't too crazy.

mwilsnd (approved):

alexcristici (approved):

PR #2066, opened 2024-01-28
remove unused member variable

nunojpg:
Hi, found this via UBSAN:
`src/mbgl/layout/symbol_instance.hpp:55:7: runtime error: load of value 190, which is not a valid value for type 'bool'`
As grep doesn't show any usage of this member variable I suggest the removal.

### DIFF
--- a/src/mbgl/layout/symbol_instance.hpp
+++ b/src/mbgl/layout/symbol_instance.hpp
@@ -113,7 +113,6 @@ public:
     std::array<float, 2> textOffset;
     std::array<float, 2> iconOffset;
     std::u16string key;
-    bool isDuplicate;
     std::optional<size_t> placedRightTextIndex;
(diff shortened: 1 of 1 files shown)


louwers (approved):

louwers:
@nunojpg How did you run USBAN? Might be good to get running on CI again.

nunojpg:
@louwers hi. I have my CMAKE project which includes maplibre-native-qt.
I set CXX flags with -fsanitize=undefined (and also address). Note it requires also RTTI (MLN_WITH_RTTI=ON)

louwers:
OK I see we have it set up in the CMake config already with `MBGL_WITH_SANITIZER`. Let me make an issue for it.

nunojpg:
@louwers if you don't use it for a long time it's actually quite impressive this was the only violation I found. Great work!

PR #2073, opened 2024-01-30
Use `enableDepth` option

### LINKED ISSUE #2072: Tile borders are truncated
The tile borders are drawn truncated on Metal

Image

TimSylvester:
The check for this option was apparently lost in a previous round of updates.

Resolves #2072

### DIFF
--- a/include/mbgl/gfx/drawable.hpp
+++ b/include/mbgl/gfx/drawable.hpp
@@ -143,7 +143,7 @@ public:
     void setDrawPriority(DrawPriority value) { drawPriority = value; }
 
     /// Whether to enable depth testing
-    bool getEnableDepth() { return enableDepth; }
+    bool getEnableDepth() const { return enableDepth; }
     virtual void setEnableDepth(bool value) { enableDepth = value; }
 
     /// Determines depth range within the layer for 2D drawables
--- a/include/mbgl/mtl/tile_layer_group.hpp
+++ b/include/mbgl/mtl/tile_layer_group.hpp
@@ -1,6 +1,11 @@
 #pragma once
 
+#include <mbgl/mtl/mtl_fwd.hpp>
 #include <mbgl/renderer/layer_group.hpp>
+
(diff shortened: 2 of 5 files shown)


alexcristici (approved):

louwers (commented):
Would it be possible to add a regression test for this?

mwilsnd (commented):
For correctness we probably need to add this check in layer_tweaker.cpp:48 as well for non-GL backends. Not calling depthModeForSublayer means glDepthRange isn't setup (on OpenGL) and defaults back to 0-1.

TimSylvester:
> Would it be possible to add a regression test for this?

There are tests that I would expect to catch it, e.g.:

image

I suppose these are accidentally working because the depth value it comes up with happens to put it above the extrusions.

It might be as easy as enabling tile outlines on one of the other tests and re-baselining its image, but I'm not sure how to find one that would be appropriate.

stefankarschti (approved):

stefankarschti:
Fixes #2072

cgalvan (approved):

stefankarschti:
> > Would it be possible to add a regression test for this?
> 
> There are tests that I would expect to catch it, e.g.:
> 
> image
> I suppose these are accidentally working because the depth value it comes up with happens to put it above the extrusions.
> 
> It might be as easy as enabling tile outlines on one of the other tests and re-baselining its image, but I'm not sure how to find one that would be appropriate.

the debug layers use their `type` as layer index, but also set `builder->setEnableDepth(false);`
I suppose any style that has more than 2 layers would hide the tile borders if the `enableDepth` flag is not correctly interpreted.

TimSylvester:
I added the same check to the matrix offset condition.

The 3D mode setup was also not paying attention to the depth option.  I'm not sure if that's ever used, but it ought to work.  

This was creating both state objects, with and without stenciling, and switching between them per-drawable.  Rather than creating all four combinations of depth and stencil, I made it create only the ones that actually get used, and keep the ones that don't change in the group instance.

PR #2076, opened 2024-01-31
iOS moving camera benchmark

alexcristici:
Changed iOS benchmark to use a moving camera instead of a static one.
The locations are similar with the ones from Android benchmark.

image


mwilsnd (approved):

stefankarschti:
This is really cool! However on my device the frames are not always fully loaded. Maybe we need to force the camera transitioning engine to only produce fully rendered frames if possible?
Screenshot 2024-02-01 at 18 05 11
Screenshot 2024-02-01 at 18 05 23
Screenshot 2024-02-01 at 18 05 37
Screenshot 2024-02-01 at 18 06 20

alexcristici:
> This is really cool! However on my device the frames are not always fully loaded. Maybe we need to force the camera transitioning engine to only produce fully rendered frames if possible?

This could be related to Benchmark app, not to camera transition engine, because I do a snapshot the GL buffer `auto image = frontend->readStillImage();` and display it in a view, only to have a visual feedback, the measurement is done on headless rendering.

TimSylvester (approved):

stefankarschti (approved):

Issue #2053, opened 2024-01-25
[iOS - Metal - v6.0.0]  - MLNAnnotationView delay with map panning


sepperousseau:
**Describe the bug**
I've been doing some tests with MapLibre Native iOS (6.0.0 (Metal)) And I was just testing some annotations.
Just wondering if this visible delay of annotation views was already there in 5.x.x or someone has the same behaviour ?
The blue ones are MLNAnnotationImage (UIView converted to Image) and the red ones are MLNAnnotationView.
And as you can see the red ones have a slight delay when panning the map. Just like the user location annotation.

**To Reproduce**
Create a MLNAnnotationView and add it to the map.
```
class TestAnnotationView: MLNAnnotationView {
    init(annotation: MLNAnnotation) {
        super.init(annotation: annotation, reuseIdentifier: "test-identifier")
        frame = CGRect(x: 0, y: 0, width: 40, height: 40)
        layer.cornerRadius = 20
        backgroundColor = .red
    }
}
```

**Expected behavior**
I would expect that the MLNAnnotationView and UserLocationAnnotation has slight to no delay as the MLNAnnotationImage?
MLNAnnotationImage will always have better performance but even with 1 annotation it has slight delay.

**Screenshots**
https://github.com/maplibre/maplibre-native/assets/10544377/d7f309cd-04b9-4e72-8b52-e1a1f3e51813

**Platform information (please complete the following information):**
 - OS: iOS
 - Version 6.0.0

Any help is welcome 🙏

stefankarschti:
do you have a fixed set of coordinates that you use for the blue and red circles?

sepperousseau:
> do you have a fixed set of coordinates that you use for the blue and red circles?

Not really. Just some random generated coordinates around a location.
It's the same with 1 point. I'll try to make a test project tomorrow.


https://github.com/maplibre/maplibre-native/assets/10544377/65948bdf-9506-4115-89b1-44b0bd44cbfa

stefankarschti:
Cool. I'd be happy to have a look at the test project!

sepperousseau:
> Cool. I'd be happy to have a look at the test project!

Ok I already created one :D quick and dirty.
I've noticed that the issue isn't there when using no styles url. But with like a standard maptiler vector styles it's there.
I didn't put my api key inside but I guess you have one maybe :D 

[TestMapLibre.zip](https://github.com/maplibre/maplibre-native/files/14057513/TestMapLibre.zip)

[...]

PR #2081, opened 2024-02-01
Fix issue with desynchronized annotation views on Metal

stefankarschti:
ref #2053 #2082


sjg-wdw:
I think we should revisit this after you're back.  We might want to do this with a display link.

PR #2093, opened 2024-02-07
Continued destruction of tile objects on the render thread

TimSylvester:
Resolves several cases where tiles could still be destroyed on the render thread, leading to delays from mutex contention.

That includes cases where the collection of cached tiles is cleared as well as a strange case, possibly a compiler bug, where the temporary lambda captures are not destroyed at the end of the statement.


src/mbgl/tile/tile_cache.cpp
```diff
@@ -22,6 +22,17 @@ void TileCache::setSize(size_t size_) {
     assert(orderedKeys.size() <= size);
 }
 
+namespace {
+template <typename T>
+struct CaptureWrapper {
+    CaptureWrapper(std::unique_ptr<T>&& item_)
+        : item(std::move(item_)) {}
+    CaptureWrapper(const CaptureWrapper& other)
```
TimSylvester, line 30: Comment out this definition, or replace the block with `= default`

mwilsnd (approved):

PR #2100, opened 2024-02-12
Upload new external data package for every benchmark run

### LINKED ISSUE #2087: Fix Android Benchmark on CI
The external data package expired, so the benchmark fails.

```
% aws devicefarm get-upload --arn "arn:aws:devicefarm:us-west-2:373521797162:upload:20687d72-0e46-403e-8f03-0941850665bc/c27174c2-63f4-4cdb-9af9-68957d75ebed"

An error occurred (NotFoundException) when calling the GetUpload operation: Upload file expired. The upload id is: c27174c2-63f4-4cdb-9af9-68957d75ebed
```

We should upload this external data package before the run so we get a new ARN.

https://github.com/maplibre/maplibre-native/blob/14bfb5df7257848ee0d14570e349a07996c9469f/.github/workflows/android-device-test.yml#L29-L33

I'll work on this next week @mwilsnd.

louwers:
Closes #2087

### DIFF
--- a/.github/workflows/android-device-test.yml
+++ b/.github/workflows/android-device-test.yml
@@ -26,12 +26,6 @@ jobs:
             appFile: "MapboxGLAndroidSDKTestApp-drawable-release.apk",
             name: "Android Benchmark",
             testFilter: "org.maplibre.android.benchmark.Benchmark",
-            # echo '{"styleNames": [...], "styleURLs": [...], "resultsAPI: "..." }' > benchmark-input.json 
-            # zip benchmark-input.zip benchmark-input.json 
-            # aws devicefarm create-upload --project-arn <project_arn> --type EXTERNAL_DATA --name benchmark-input.zip
-            # curl -T benchmark-input.zip <upload_url>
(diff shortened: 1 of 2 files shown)


.github/workflows/android-device-test.yml
```diff
@@ -83,6 +77,33 @@ jobs:
         with:
           files: "${{ matrix.test.testFile }}, ${{ matrix.test.appFile }}"
 
+      - name: Upload external data for benchmark
+        if: matrix.test.name == 'Android Benchmark'
+        run: |
```
HarelM, line 82: I would recommend putting this in a script in the repo and referencing this script instead of writing it all in the CI code, this way you can also test it locally.

HarelM (approved):

.github/workflows/android-device-test.yml
louwers: Like this?

HarelM: Yup

PR #2107, opened 2024-02-12
Texture by index instead of map

alexcristici:
I replaced StringIdentity idTexUniformName with

```
enum {
    idBackgroundImageTexture,
    backgroundTextureCount
};
```
Texture ids are in the same order in `Textures` as in `SamplerLocations`/`TextureBindings`, using above enum ordering.
In this way we match the texture for the right sampler location / texture binding.


sjg-wdw (approved):

TimSylvester (approved):

stefankarschti (approved):

include/mbgl/shaders/shader_defines.hpp
```diff
@@ -40,5 +40,73 @@ static constexpr auto maxUBOCountPerShader = std::max({static_cast<size_t>(backg
                                                        static_cast<size_t>(rasterUBOCount),
                                                        static_cast<size_t>(symbolUBOCount)});
 
+enum {
```
louwers, line 43: I think it would be better to put these in their own file, because unlike the above they don't require a ton of includes.

src/mbgl/gl/drawable_gl.cpp
```diff
@@ -240,27 +240,28 @@ gfx::StencilMode DrawableGL::makeStencilMode(PaintParameters& parameters) const
 }
 
 void DrawableGL::uploadTextures() const {
-    for (const auto& pair : textures) {
-        if (const auto& tex = pair.second) {
-            std::static_pointer_cast<gl::Texture2D>(tex)->upload();
+    for (const auto& texture : textures) {
+        if (texture) {
+            static_cast<gl::Texture2D&>(*texture).upload();
```
louwers, line 245: Just curious, but why is the cast needed here?

louwers (approved):

src/mbgl/gl/drawable_gl.cpp
alexcristici: No needed, I just removed it. 👍

include/mbgl/shaders/shader_defines.hpp
alexcristici: Let's postpone this split for my next PR: https://github.com/alexcristici/maplibre-native/pull/14/files, which is on top of this. Splitting the file now will cause some conflicts.

mwilsnd (approved):

include/mbgl/shaders/shader_defines.hpp
louwers: Alright!

src/mbgl/gl/drawable_gl.cpp
louwers: OK it occurs in a few other places as well.

src/mbgl/gl/drawable_gl.cpp
```diff
@@ -245,27 +245,28 @@ gfx::StencilMode DrawableGL::makeStencilMode(PaintParameters& parameters) const
...
 void DrawableGL::bindTextures() const {
     int32_t unit = 0;
-    for (const auto& pair : textures) {
-        if (const auto& tex = pair.second) {
-            const auto& location = pair.first;
-            std::static_pointer_cast<gl::Texture2D>(tex)->bind(location, unit++);
+    for (size_t id = 0; id < textures.size(); id++) {
+        if (const auto& texture = textures[id]) {
+            if (const auto& location = shader->getSamplerLocation(id)) {
+                static_cast<gl::Texture2D&>(*texture).bind(static_cast<int32_t>(*location), unit++);
```
louwers, line 260: Please verify if the cast is needed.

src/mbgl/mtl/drawable.cpp
```diff
@@ -386,27 +386,29 @@ void Drawable::bindUniformBuffers(RenderPass& renderPass) const noexcept {
...
 
 void Drawable::bindTextures(RenderPass& renderPass) const noexcept {
-    for (const auto& pair : textures) {
-        if (const auto& tex = pair.second) {
-            const auto& location = pair.first;
-            static_cast<mtl::Texture2D&>(*tex).bind(renderPass, location);
+    for (size_t id = 0; id < textures.size(); id++) {
+        if (const auto& texture = textures[id]) {
+            if (const auto& location = shader->getSamplerLocation(id)) {
+                static_cast<mtl::Texture2D&>(*texture).bind(renderPass, static_cast<int32_t>(*location));
```
louwers, line 392: Please verify if the cast is needed.

src/mbgl/mtl/drawable.cpp
```diff
@@ -386,27 +386,29 @@ void Drawable::bindUniformBuffers(RenderPass& renderPass) const noexcept {
...
 
 void Drawable::unbindTextures(RenderPass& renderPass) const noexcept {
-    for (const auto& pair : textures) {
-        if (const auto& tex = pair.second) {
-            const auto& location = pair.first;
-            static_cast<mtl::Texture2D&>(*tex).unbind(renderPass, location);
+    for (size_t id = 0; id < textures.size(); id++) {
+        if (const auto& texture = textures[id]) {
+            if (const auto& location = shader->getSamplerLocation(id)) {
+                static_cast<mtl::Texture2D&>(*texture).unbind(renderPass, static_cast<int32_t>(*location));
```
louwers, line 402: Please verify if the cast is needed.

src/mbgl/gl/drawable_gl.cpp
alexcristici: Already verified. `bind` and `unbind` are platform specific. Probably that's why upload got casted, copy paste.

alexcristici: https://github.com/maplibre/maplibre-native/blob/main/include/mbgl/gfx/texture2d.hpp

Issue #1363, opened 2023-07-13
Documentation for iOS contains a lot of references to "mapbox" / "Mapbox Maps SDK" / "Mapbox SDK"


JannikGM:
**There are probably many more instances,  but here are some examples:**

---

The index refers to names prefixed by `MGL`, rather than `MLN`, which makes me think the docs haven't been regenerated properly after #919 (to be fair, there has been an Android release since then, but no iOS release yet).

---

https://maplibre.org/maplibre-native/ios/api/index.html

- Headline refers to `[Mapbox Maps SDK for iOS](https://www.mapbox.com/ios-sdk/)` and clicking it sends you to mapbox.
- Multiple mentions of `Mapbox Maps SDK for iOS` everywhere except for the "Changes in version [...]".
- All links, including link to "full changelog" refer to mapbox ecosystem.
- Breadcrumbs are "Mapbox Reference > Mapbox Reference".

---

https://maplibre.org/maplibre-native/ios/api/working-with-geojson-data.html

- Tells user to store data on mapbox servers, but I think support for this feature might have been removed in maplibre? (mapbox URLs have at least been removed from maplibre-gl-js); should potentially refer to other tool providers.

---

https://maplibre.org/maplibre-native/ios/api/Other%20Constants.html#/c:@MapboxVersionNumber
https://maplibre.org/maplibre-native/ios/api/Other%20Constants.html#/c:@MapboxVersionString

- Variable names refer to mapbox explicitly, potentially also a bug (or backwards compatiblity feature?) in the code.

---

https://maplibre.org/maplibre-native/ios/api/Enums/MGLErrorCode.html

- Refers to the "Mapbox SDK"

wipfli:
I think there are also some Mapbox references in README files. I think it would be good to go through the entire repo and replace Mapbox with MapLibre in documentation where possible. And if it is not possible, then just remove the sentence or paragraph. This is also what we did in MapLibre GL JS over the course of 2 or so years...

louwers:
@wipfli There is a lot of cruft in the README files in general. I'll do a thorough sweep of them next month.

louwers:
The new documentation is hosted at https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre

I did a first pass with https://github.com/maplibre/maplibre-native/pull/2169 to remove the links to the examples hosted by Mapbox.

I'll create another issue for removing Mapbox references from README files. @wipfli

PR #2111, opened 2024-02-13
DocC Improvements

louwers:
- Fixes deployment
- Links to source now working:

image

- Adds some basic theming

image

- Cleans up MLNMapView for #1363
- Makes `Mapbox.template.h` into `Mapbox.h` since we only use Metal now for iOS.


HarelM (approved):
Not sure I'm the right person to review this though...

PR #2112, opened 2024-02-14
Wide line PoC (Metal Custom Layer)

stefankarschti:
Implement wide line vectors as a Custom (raw) Metal layer by adapting code from WhirlyGlobe.


stefankarschti:
image

louwers (changes requested):
It's a cool POC, but I don't think it practical to have source code with different licenses in the same root, even if it's just the sample app and not part of the library.

sjg-wdw:
> It's a cool POC, but I don't think it practical to have source code with different licenses in the same root, even if it's just the sample app and not part of the library.

Oh right, we need to address that.  I own that code and I can grant whatever rights you want.  What's the copyright for new MapLibre code look like?

platform/BUILD.bazel
```diff
@@ -190,7 +190,14 @@ objc_library(
         "//platform/ios:ios_app_srcs",
         "//platform/darwin:app/CustomStyleLayerExample.h",
         "//platform/darwin:app/CustomStyleLayerExample.m",
-    ],
+    ] + select({
+        "//:metal_renderer": [
+        "//platform/darwin:app/WideLineStyleLayer.h",
```
louwers, line 195: Why is the source for the iOS app in `darwin`?

louwers:
@sjg-wdw This is the license for MapLibre Native: https://github.com/maplibre/maplibre-native/blob/main/LICENSE.md

So the copyright specifically is _Copyright (c) 2021 MapLibre contributors_

platform/BUILD.bazel
louwers: Can we add these to the `ios_app_srcs` target and not use `exports_files`?

`exports_files` makes it harder to reason about dependencies.

sjg-wdw:
Stefan has my permission to change the copyright on those files to MapLibre contributors.  Let me know if you need that permission in some other form.

platform/BUILD.bazel
stefankarschti: That is where the original LimeGreenTriangle example was. We can move it to another location.

louwers: Ah alright, makes sense.

Maybe you can try to avoid using `exports_files`, but I can have a look after it is merged as well.

stefankarschti: acknowledged! 👍

louwers:
@stefankarschti Could you remove the copyright header, and then clarify with Steve that the following is true:

> Whenever you make a contribution to a repository containing notice of a license, you license your contribution under the same terms, and you agree that you have the right to license your contribution under those terms.

platform/BUILD.bazel
stefankarschti: I was trying to move the files to `ios` in the app and I realized the way it's done now is the `objc_library "iosapp"` is defined in `platform/` and it requires sources from `platform/ios` which needs the files in the `platform/ios` export list. The `ios_application` in `platform/ios` then adds the `//platform:iosapp` as dependency.

sjg-wdw:
> @stefankarschti Could you remove the copyright header, and then clarify with Steve that the following is true:
> 
> > Whenever you make a contribution to a repository containing notice of a license, you license your contribution under the same terms, and you agree that you have the right to license your contribution under those terms.

Yes, that's fine with me.  I grant permission on behalf of mousebird consulting.

platform/darwin/app/wideline.metal
```diff
@@ -0,0 +1,495 @@
...
+    const float2 dB = b0 - b1;
+
+    // Solve the system of equations formed by equating the lines to get their common point.
+    const float denom = wedge(dA, dB);
+    if (denom == 0.0) {
+        // If the denominator comes out zero, the lines are parallel and do not intersect
+        return { .valid = false };
+    }
+
+    const float tA = (a0.x * a1.y - a0.y * a1.x);
```
mwilsnd, line 49: Mostly the same shader source as before here? Refer to #2183 for previous nits I had in this shader.

platform/darwin/app/wideline.metal
```diff
@@ -0,0 +1,495 @@
...
+    const bool isInsideEdge = (isLeft == turningLeft);
+
+    // Current corner is the default result for all points.
+    float2 pos = corner;
+
+    // Texture position is based on cumulative distance along the line.
+    // Note that `inst[2].totalLen` wraps to zero at the end, and we don't want that.
+    // Texture coords are used for edge blending, so we need to set them up even if there are no textures.
+    float texY = 0;
+    float texX = 0;
```
mwilsnd, line 316: It looks like texture support was removed from this shader, I see some texcoord manipulation commented out in a few places. Could you remove these references?

platform/darwin/app/wideline.metal
```diff
@@ -0,0 +1,495 @@
...
+//        outVert.texCoord = -wideVec.texOffset + float2(texX, texY * texScale);
+    }
+
+    return outVert;
+}
+
+struct TriWideArgBufferFrag {
+    UniformDrawStateA uniDrawState      [[ id(WKSUniformDrawStateEntry) ]];
+    UniformWideVec wideVec              [[ id(WKSUniformWideVecEntry) ]];
+    bool hasTextures;
```
mwilsnd, line 473: Unused?

platform/darwin/app/wideline.metal
```diff
@@ -0,0 +1,495 @@
...
+        outVert.position = float4(pos, 0, 1);
+        // Opposite values because we're showing back faces.
+//        outVert.texCoord = -wideVec.texOffset + float2(texX, texY * texScale);
+    }
+
+    return outVert;
+}
+
+struct TriWideArgBufferFrag {
+    UniformDrawStateA uniDrawState      [[ id(WKSUniformDrawStateEntry) ]];
```
mwilsnd, line 471: Unused?

PR #2113, opened 2024-02-14
Track expression dependencies

TimSylvester:
When building expression trees, we assign and capture what categories of input they depend on, like feature data and zoom level.

When evaluating whether to update a layer, zoom changes only trigger an update if zoom is used in an expression somewhere within the style.

---

I discovered that the benchmarking I was doing was thrown off by some local changes lingering in my workspace, so I'm re-evaluating...

deps:

```
Average frame encoding time: 2.650 ms
Average frame rendering time: 6.487 ms

Average frame encoding time: 2.606 ms
Average frame rendering time: 6.392 ms

Average frame encoding time: 2.579 ms
Average frame rendering time: 6.360 ms

Average frame encoding time: 2.673 ms
Average frame rendering time: 6.564 ms

Average frame encoding time: 2.606 ms
Average frame rendering time: 6.536 ms

Average frame encoding time: 2.592 ms
Average frame rendering time: 6.444 ms
```

Mean encoding: 2.6177

main:
```
Average frame encoding time: 2.664 ms
Average frame rendering time: 6.47 ms

Average frame encoding time: 2.711 ms
Average frame rendering time: 6.451 ms

Average frame encoding time: 2.639 ms
Average frame rendering time: 6.372 ms

Average frame encoding time: 2.624 ms
Average frame rendering time: 6.472 ms

Average frame encoding time: 2.715 ms
Average frame rendering time: 6.536 ms

Average frame encoding time: 2.662 ms
Average frame rendering time: 6.427 ms
```

Mean encoding: 2.6692

Difference: 1.95%


sjg-wdw (approved):

stefankarschti (approved):

src/mbgl/map/map_impl.cpp
```diff
@@ -84,6 +86,33 @@ void Map::Impl::onStyleLoaded() {
         annotationManager.onStyleLoaded();
     }
     observer.onDidFinishLoadingStyle();
+
+#if !defined(NDEBUG)
```
louwers, line 90: Should be extracted in a separate function.

src/mbgl/style/properties.hpp
```diff
@@ -298,6 +339,23 @@ class Properties {
                                other.template get<Ps>().value))...});
             return result;
         }
+
+        Dependency getDependencies() const noexcept {
```
louwers, line 343: `std::make_tuple` is not `noexcept`.

louwers (changes requested):
Could you add tests?

src/mbgl/style/properties.hpp
TimSylvester: This is evaluated at compile time, though I forgot to put `constexpr` on it.  But I can do it more manually to eliminate any ambiguity.

image

louwers: Ah OK, but this is one compiler.

I would err on the side of safety in general when it comes to `noexcept`.

TimSylvester:
> Could you add tests?

Done.  I also extended some test cases because the modifications to existing code with low coverage made CodeCov mad, so overall coverage should be up a bit.

TimSylvester:
> Could you add tests?

Done.  I also expanded some existing test cases, as modifications to some areas with low coverage brought the average way down.

louwers (approved):

PR #2129, opened 2024-02-20
Get unit and render tests running on iOS simulator

TimSylvester:
Fixes Metal errors on iOS simulator related to special requirements on textures.

Also disables HTTP tests, because listening on port 3000 fails, and it seems impractical to make the port dynamic since it appears in a bunch of `.json` files.

I'm curious whether this is specific to my system/configuration, though, so perhaps it can be left enabled.  I don't find anyone else having the same issue on, e.g., StackOverflow.

48 of the 1165 render tests fail on the simulator, which seems not so bad.  Interestingly, a lot of them seem to show the same pattern of a vertical and horizontal stripe of errors.

image

### DIFF
--- a/src/mbgl/mtl/offscreen_texture.cpp
+++ b/src/mbgl/mtl/offscreen_texture.cpp
@@ -35,6 +35,8 @@ public:
                 ->setUsage(MTL::TextureUsageShaderRead | MTL::TextureUsageShaderWrite | MTL::TextureUsageRenderTarget);
         }
 
+        // On iOS simulator, the depth target is PixelFormatDepth32Float_Stencil8
+#if !TARGET_OS_SIMULATOR
         if (stencil) {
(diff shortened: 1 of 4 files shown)


cgalvan (approved):

louwers:
OK, I tried

```
 bazel test --//:renderer=metal //platform/ios/test/core:CppUnitTests
```

But it fails to compile with because of

```
#include <mbgl/gl/context.hpp>
```

in `test/map/map.test.cpp`.

Will try to enable it on CI with a separate PR, though I wonder why we are not hitting this with the other test.

louwers (approved):
Actually, since it is already

TimSylvester:
I see the same error when building from the command line, but not from the Xcode project.  But that header doesn't seem to be needed for anything, anyway.

PR #2132, opened 2024-02-22
Free GitHub runners

louwers:
The free GitHub runners got faster for open-source projects: https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

So this PR disables the paid runners.

We go back from 16-core to 4-core, let's see if CI time is acceptable.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,7 +1,6 @@
 self-hosted-runner:
   # Labels of self-hosted runner in array of strings.
   labels:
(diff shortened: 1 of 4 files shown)


cgalvan (approved):

louwers:
> getaddrinfo ENOTFOUND productionresultssa3.blob.core.windows.net

Sounds legit.

louwers:
Looks like I need to split the Android workflow up in parallel stages.

PR #2149, opened 2024-02-27
Always create check for Android render test, Take into account cancelled workflows

louwers:
- Makes sure that a check for Android render test is always created (so it can be made into a required check).
- When the parent workflow is cancelled (and not just skipped), some workflows need to be skipped

### DIFF
--- a/.github/workflows/android-device-test.yml
+++ b/.github/workflows/android-device-test.yml
@@ -7,21 +7,7 @@ on:
       - completed
 
 jobs:
-  pre_job:
-    runs-on: ubuntu-latest
-    outputs:
-      should_skip: ${{ steps.check_skip.outputs.should_skip }}
-    steps:
-      # if android-build in android-ci was skipped, the android-device-test job can be skipped entirely
-      - id: check_skip
-        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/android-device-test.yml
```diff
@@ -141,7 +129,7 @@ jobs:
           testSpecArn: ${{ matrix.test.testSpecArn }}
 
       - uses: LouisBrunner/checks-action@v2.0.0
-        if: always() && env.run_device_test == 'true'
+        if: always() && (env.run_device_test == 'true' || matrix.test.name == 'Android Render Tests')
```
HarelM, line 132: This file seems a bit long for yaml... 100+ lines long... I would consider moving the invocation to a separate script files and keep the minimum possible here...

cgalvan (approved):

.github/workflows/android-device-test.yml
louwers: I don't really see any opportunities to make it shorter. When you move stuff into a script you still need to pass a lot of variables, and this has the potential to make it even longer.

The matrix takes up a lot of space, but this is also not something you can move into another file.

PR #2152, opened 2024-02-27
Add optional split output location to codegen scripts

mwilsnd:
This PR adds a new `--out` option to the core and iOS/Darwin code generation tools. This enables out-of-source code generation, useful for custom build systems and caching artifacts. This parameter defaults to whatever the value of root is (aka the old behavior) if not specified.


louwers (commented):
We can maybe remove the other argument since we are the only ones that use this script.

mwilsnd:
> We can maybe remove the other argument since we are the only ones that use this script.

Yeah, just looked and it doesn't appear to be used anymore. I've gone ahead and removed it.

louwers (approved):

PR #2160, opened 2024-03-01
Add option to disable bundle assertion in Darwin platform

mwilsnd:
Adds a preprocessor option to disable the mapbox bundle assertion in the Darwin platform. This is useful for 1). Custom build systems that merge resources from multiple libraries into a unified bundle, and 2). Custom builds that statically link the application without having to build MapLibre as a framework.

### DIFF
--- a/platform/darwin/src/NSBundle+MLNAdditions.mm
+++ b/platform/darwin/src/NSBundle+MLNAdditions.mm
@@ -9,6 +9,7 @@ const MLNExceptionName MLNBundleNotFoundException = @"MLNBundleNotFoundException
 + (instancetype)mgl_frameworkBundle {
     NSBundle *bundle = [self bundleForClass:[MLNSettings class]];
 
+#ifndef MLN_CUSTOM_COMBINED_BUNDLE
     if (![bundle.infoDictionary[@"CFBundlePackageType"] isEqualToString:@"FMWK"]) {
         // For static frameworks, the bundle is the containing application
         // bundle but the resources are in Mapbox.bundle.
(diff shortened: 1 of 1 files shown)


platform/darwin/src/NSBundle+MLNAdditions.mm
```diff
@@ -9,6 +9,7 @@ @implementation NSBundle (MLNAdditions)
 + (instancetype)mgl_frameworkBundle {
     NSBundle *bundle = [self bundleForClass:[MLNSettings class]];
 
+#ifndef MLN_CUSTOM_COMBINED_BUNDLE
     if (![bundle.infoDictionary[@"CFBundlePackageType"] isEqualToString:@"FMWK"]) {
```
louwers, line 13: Since we are touching this code, could you replace "Mapbox" with "MapLibre"?

louwers (approved):

platform/darwin/src/NSBundle+MLNAdditions.mm
mwilsnd: Do we want to rename the expected bundle name too, or just the assertion for now?

louwers: Just the message is OK, although I don't see the harm in changing the bundle name too.

It's the `apple_resource_bundle` in `platform/os/BUILD.bazel` it looks like.

platform/darwin/src/NSBundle+MLNAdditions.mm
```diff
@@ -20,6 +21,7 @@ + (instancetype)mgl_frameworkBundle {
                         format:@"The Mapbox framework bundle could not be found. If using the Mapbox Maps SDK for iOS as a static framework, make sure that Mapbox.bundle is copied into the root of the app bundle."];
```
louwers, line 21:
```suggestion
                        format:@"The MapLibre framework bundle could not be found. If using the MapLibre Native for iOS as a static framework, make sure that MapLibre bundle is copied into the root of the app bundle."];
```

PR #2162, opened 2024-03-03
Add render tests to Gradle project root at `platform/android`

louwers:
And some other Android development tasks

- Update README
- Remove some Mapbox references
- Fix release workflow (it was failing because local.properties was present on CI)
- Lower acceptable delta of one instrumentation test

### DIFF
--- a/platform/android/DEVELOPING.md
+++ b/platform/android/DEVELOPING.md
@@ -1,12 +1,65 @@
 # Developing - MapLibre Native for Android
 
-These instructions are for developers interested in making code-level contributions to the SDK itself.
+These instructions are for developers interested in making code-level contributions to MapLibre Native for Android.
+
+## Getting the source
+
+Clone the git repository and pull in submodules:
+
+```bash
+git clone git@github.com:maplibre/maplibre-native.git
+cd maplibre-native
+git submodule update --init --recursive
+cd platform/android
+```
+
+## Requirements
+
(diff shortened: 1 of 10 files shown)


HarelM:
Someone else would need to review this as it's specific to Android...

platform/android/DEVELOPING.md
```diff
@@ -33,101 +13,64 @@ git submodule update --init --recursive
...
+Open the `platform/android` directory to get started.
 
-**Note**: We partially support C++14 because GCC 4.9 does not fully implement the
-final draft of the C++14 standard. More information in [DEVELOPING.md](../../DEVELOPING.md).
+## Setting an API Key
 
-**Note**: On macOS you can install clang with installing the [Apple command line developer tools](https://developer.apple.com/download/).
-
-## Opening the project
+_The test application (used for development purposes) uses MapTiler vector tiles, which require a MapTiler account and API key._
```
cgalvan, line 24: Could we have the test application use the MapLibre demotiles instead, so the developer wouldn't need to create a MapTiler account?

https://github.com/maplibre/demotiles/blob/gh-pages/README.md

cgalvan (approved):

louwers:
@HarelM OK I will keep that in mind!

platform/android/DEVELOPING.md
louwers: It uses the demotiles by default, but many features cannot be demoed with the demotiles style...

PR #2169, opened 2024-03-04
Remove references to Mapbox iOS examples

louwers:
That's a lot of examples.

### DIFF
--- a/platform/darwin/src/MLNOfflineStorage.h
+++ b/platform/darwin/src/MLNOfflineStorage.h
@@ -28,8 +28,7 @@ NS_ASSUME_NONNULL_BEGIN
  path.
  
  #### Related examples
- See the <a href="https://docs.mapbox.com/ios/maps/examples/offline-pack/">
- Download an offline map</a> example to learn how to calculate the progress
+ TODO:Download an offline map, learn how to calculate the progress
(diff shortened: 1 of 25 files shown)


ianthetechie (commented):
Do we care about comma splices and funny grammatical constructions at this stage? Assuming we don't, then this looks good. This actually gives me a good list of stuff to start working through, and is already an improvement ;)

platform/darwin/src/MLNOfflineStorage.h
```diff
@@ -28,8 +28,7 @@ NS_ASSUME_NONNULL_BEGIN
  path.
  
  #### Related examples
- See the <a href="https://docs.mapbox.com/ios/maps/examples/offline-pack/">
- Download an offline map</a> example to learn how to calculate the progress
+ TODO:Download an offline map, learn how to calculate the progress
```
cgalvan, line 31: Spacing nit
```suggestion
 TODO: Download an offline map, learn how to calculate the progress
```

cgalvan (approved):

louwers:
@ianthetechie I would say we do care, but we care more about getting this dead links removed for now. 😉 

Thanks for taking a look!

PR #2195, opened 2024-03-13
Minor improvements to developer onboarding + documentation of DocC

ianthetechie:
Clarifies a few things I noticed while attempting to start on porting the old examples to the new DocC format. I encountered a few rather strange errors that are probably not likely to be encountered by others, but deserve a mention (I had 10+ years of provisioning profiles, so a few surprises on the Bazel side).


louwers (approved):
Thanks!

Issue #2189, opened 2024-03-12
[iOS] MLNRasterStyleLayer gets cut off with mask in StyleURL file since v6.0


NicolasBrunnerSRF:
**Describe the bug**
MLNRasterStyleLayer is cut off/masked by the mask defined in StyleURL JSON file, even though it should not.

See demo app: [MapLibreMaskBug on Github](https://github.com/NicolasBrunnerSRF/mapLibreMaskBug)

This demo app sets a custom style to the MapView, which contains a mask to cut out Switzerland and mask all other countries. It then adds an ImageLayer on top of the map (to show precipitation data). This ImageLayer should not be cut/masked outside of Switzerland. This worked fine with the latest major release of MapLibre v5 but is broken since the introduction of v6. See ReadMe of demo app github repo for a detailed description.

**To Reproduce**
1. Download demo app
2. Build and run on simulator

**Expected behavior**
ImageLayer (MLNRasterStyleLayer) is not affected by StyleURL file.

**Screenshots**
See demo app github repo ReadMe.

**Platform information (please complete the following information):**
 - iOS 17.4
 - XCode 15.3
 - MapLibre iOS v6.2.0 (was fine in v5)

**Additional context**
We are currently in a deadlock situation in our project: Using MapLibre iOS v5 crashes our app on iOS 17.4 simulators with XCode 15.3. Also, we cannot use MapLibre iOS v6 due to the above bug. So we are currently stuck at using XCode 15.2

r3econ:
I'm facing the same issue 🤔

louwers:
Thanks for the detailed bug report! cc @sjg-wdw

sjg-wdw:
Do we have a test case that covers this one?

NicolasBrunnerSRF:
@louwers yes, good idea!

I updated the demo project with a minimal style and the issue is still reproducable:
![Simulator Screenshot - iPhone 15 Pro - 2024-03-13 at 09 09 49](https://github.com/maplibre/maplibre-native/assets/98830687/526be05e-9baf-492a-afc6-69e6f22d7d5b)

louwers:
@NicolasBrunnerSRF Perfect! Thanks.

@sjg-wdw The style is here: https://github.com/NicolasBrunnerSRF/mapLibreMaskBug/commit/582328fd8f05c7f945877b8939875845414c27e3

Armenm:
[...]

PR #2197, opened 2024-03-14
Makes `setURL` method on MLNImageSource public

manuelroth:
This PR makes the `setURL` method on the MLNImageSource public. This is required to be able to dynamically update the url of an existing image source. The Android SDK already provides such a method ([see setUri ](https://maplibre.org/maplibre-native/android/api/-map-libre%20-native%20for%20-android/com.mapbox.mapboxsdk.style.sources/-image-source/index.html?query=class%20ImageSource%20:%20Source#-682102938%2FFunctions%2F2110661137)for more details). Until now this use case could only be addressed by creating a new source when ever the url changed and deleting the "old" source which resulted in bad performance of the app. The `setURL` method would greatly improve the implementation of this use case.

Please let me know if I can improve anything. I am not an Objective-C developer. Therefore any hints on improvements are greatly appreciated.

### DIFF
--- a/platform/darwin/test/MLNImageSourceTests.m
+++ b/platform/darwin/test/MLNImageSourceTests.m
@@ -39,7 +39,7 @@
     XCTAssertNil(source.URL);
 }
 
-- (void)testSetURL {
+- (void)testMLNImageSourceSetURL {
     // Create a test instance of MLNImageSource
     MLNCoordinateQuad quad = { { 80, 37}, { 81, 37}, { 81, 39}, { 80, 39}};
     MLNImageSource *source = [[MLNImageSource alloc] initWithIdentifier:@"source-id" coordinateQuad:quad URL:[NSURL URLWithString:@"http://host/image.png"]];
--- a/platform/darwin/test/MLNImageSourceTests.m
+++ b/platform/darwin/test/MLNImageSourceTests.m
@@ -39,4 +39,19 @@
     XCTAssertNil(source.URL);
 }
 
+- (void)testSetURL {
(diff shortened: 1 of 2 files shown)


platform/darwin/test/MLNImageSourceTests.m
```diff
@@ -39,4 +39,19 @@ - (void)testMLNImageSourceWithImage {
     XCTAssertNil(source.URL);
 }
 
+- (void)testSetURL {
```
louwers, line 42: Nit:

```suggestion
- (void)testMLNImageSourceSetURL {
```

louwers, line 42:
```suggestion
- (void)testMLNImageSetURL {
```

manuelroth: Fixed in https://github.com/maplibre/maplibre-native/pull/2197/commits/8fdcaaad1008fe92772ae48170b7ae172c1bbe16

louwers (approved):
Thanks for your contribution. :)

manuelroth:
Thank you very much for reviewing my PR. It is awesome to contribute in this way. I really appreciate it. Unfortunately we are currently blocked from upgrading to MapLibre v6 by this issue reported by my teammate: https://github.com/maplibre/maplibre-native/issues/2189. Would it be feasible to create another v5 release with this changes in addition to a v6 release?- So we could already take advantage of this change while a solution for the issue #2189 can be found. If its too much effort it is also fine. Of course we want to upgrade to v6 as soon as possible, but for the moment we unfortunately cannot.

PR #2201, opened 2024-03-15
Android - Renaming the library generated .so files

Sideness:
# Main goal
The main goal of this PR is to rename the .so files that are generated.   
For now, it's impossible for an app to import Mapbox and Maplibre at the same time because the .so files have the same name.   
This PR allows an app to have a Mapbox AND a Maplibre instance.

## Why do we need that?   
My app have 2 seperate features (displaying maps and navigation both based on Mapbox and Mapbox-navigation) and I can't afford migrating both at the same time.   
I figured those files would be renamed in the future as MapLibre evolves but for me, the sooner the better.


Sideness:
!benchmark android

louwers (changes requested):
Thanks.

If we are renaming it, could you rename it to just `maplibre`?

Since the name of the library itself is now MapLibre Native (we don't have GL in the name anymore).

Sideness:
Ok, I'll do that today :+1:

louwers (approved):

Sideness:
Hi! Thanks for the approve.
It's the first time I open a PR for this project. What's the next step? Should I do anything or are you the one that will merge it when needed?

louwers:
@Sideness Just forgot to click auto-merge. ;-)

Thanks for your contribution!

louwers:
@Sideness The latest pre-release includes these changes.

PR #2205, opened 2024-03-18
macOS Metal + AppKit app

louwers:
- Remove unused macOS scripts.
- Add Bazel targets for macOS (main toolkit, AppKit app, GLFW app, cache tool, render tool)
- Update CI to build macOS (no tests yet).
- Switch to libuv instead of CFRunLoop on macOS.
- Modify `.clang-format` to make pre-commit happy. I can maybe apply Objective-C formatting to the entire codebase in a later PR.
- Updated Bazel.
- Improved config handling.
  - The config is shared between macOS and iOS
  - Location: `platform/darwin/bazel/config.bzl`
  - By default, the example config is used. This is useful for including this repo in another Bazel workspace.

I would like to get this merged because it's already getting quite a big change. To be addressed later:

- TODO: the AppKit app probably had it's own run loop, because it's missing one now. We need to either:
  - Get the CFRunLoop working (again) under macOS, or:
  - Kick off the libuv based run loop on macOS AppKit
- TODO: GLFW Bazel build for other platforms?
- TODO:: re-use `platform/glfw/metal_backend.mm` for other platforms (Qt, Node.js?) move to `platform/darwin`


platform/darwin/src/NSBundle+MLNAdditions.mm
```diff
@@ -9,19 +9,19 @@ @implementation NSBundle (MLNAdditions)
...
-        NSString *bundlePath = [bundle pathForResource:@"Mapbox" ofType:@"bundle"];
-        if (bundlePath) {
-            bundle = [self bundleWithPath:bundlePath];
-        } else {
-            [NSException raise:MLNBundleNotFoundException
-                        format:@"The MapLibre framework bundle could not be found. If using the MapLibre Native for iOS as a static framework, make sure that MapLibre bundle is copied into the root of the app bundle."];
-        }
-    }
-#endif
+// #ifndef MLN_CUSTOM_COMBINED_BUNDLE
```
louwers, line 12: TODO

mwilsnd (approved):

PR #2230, opened 2024-04-04
Enable Gradle dependencies Dependabot

louwers:


cgalvan (approved):

ntadej (changes requested):
Why not use Renovate for everything?

louwers:
@ntadej Dependabot is GitHub's official solution, and therefore I want to use it wherever possible.

ntadej:
> @ntadej Dependabot is GitHub's official solution, and therefore I want to use it wherever possible.

Please then remove Renovate config if not used anymore.

louwers:
@ntadej Renovate is used for dependencies not supported by Dependabot.

Right now, Bazel dependencies and Gradle wrapper cannot be updated by Dependabot.

ntadej:
Ah sorry, I thought that we also use it for Gradle. I use it already in my project. Sorry for the noise.

ntadej (approved):

PR #2234, opened 2024-04-04
Add debug symbols to iOS GitHub release

### LINKED ISSUE #2233: Add iOS debug symbols to release assets
It would be beneficial to users if we include the iOS debugging symbols (dSYM file) in our release assets. This would allow users to better debug stack traces from crashes they encounter when using the MapLibre library.

The release build process (for iOS) should include generating the debugging symbols (dSYM file) and the symbol file should be included in the published release assets.

The alternative would be for the user to download the source from the release and build the debug symbols, but they won't match the release binary exactly unless the UUID matches, which would take a number of factors (e.g. XCode version, Mac OS version)  as well as some non-deterministic factors such as build timestamps and number of cores on build system.

So I believe building/publishing the matching debug symbols for our release binaries is the best path forward for our users.

**Additional context**
This request is specifically for iOS, but we should look at doing the same for our other supported platforms as well.

louwers:
Closes #2233

Once this is merged I can attempt to make a pre-release.

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -209,8 +209,14 @@ jobs:
 
       - name: Build XCFramework
         run: |
-          bazel build --compilation_mode=opt --//:renderer=metal //platform/ios:MapLibre.dynamic --embed_label=maplibre_ios_"$(cat VERSION)"
+          bazel build --compilation_mode=opt --apple_generate_dsym --output_groups=+dsyms --//:renderer=metal //platform/ios:MapLibre.dynamic --embed_label=maplibre_ios_"$(cat VERSION)"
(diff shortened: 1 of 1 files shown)


louwers:
Looking at https://developer.apple.com/documentation/xcode/adding-identifiable-symbol-names-to-a-crash-report https://developer.apple.com/documentation/xcode/adding-identifiable-symbol-names-to-a-crash-report#Symbolicate-the-crash-report-in-Xcode and https://developer.apple.com/documentation/xcode/adding-identifiable-symbol-names-to-a-crash-report#Symbolicate-the-crash-report-with-the-command-line this should do what you want.

cgalvan (approved):

PR #2242, opened 2024-04-08
Fix projection matrix for raster image drawables

### LINKED ISSUE #2189: [iOS] MLNRasterStyleLayer gets cut off with mask in StyleURL file since v6.0

stefankarschti:
Fixes the projection matrix for raster image drawables that are not tile based. This results in correct depth output according to the layer index and sub-index.

Fixes #2189


mwilsnd (approved):

sjg-wdw:
Nice!  How hard would it be to get some test coverage on this?

sjg-wdw (approved):

stefankarschti:
> Nice! How hard would it be to get some test coverage on this?

I'm looking into that.

louwers (approved):
Nice one!

metrics/cache-style.db
TimSylvester, line 1: Is this supposed to be in git?

TimSylvester (approved):

alexcristici (approved):

metrics/cache-style.db
louwers: It already is.

stefankarschti: yes, I needed to add the radar2.png image referenced from the [style.json](https://github.com/stefankarschti/maplibre-native/blob/issue-2189/metrics/integration/render-tests/combinations/fill-opaque--image-translucent/style.json)

TimSylvester: Not the images, the `cache-style.db`.  I can't tell at a glance what it's for, but generally cache files shouldn't be kept around or transported between clients.  Seems like that might cause misleading results.  I see that you didn't add it, but I wonder why it changed and whether that change is important to capture, or if perhaps it should just be deleted.

stefankarschti: I agree that would be good to reexamine. Initially I copied `radar2.png` to `images/` but the test runner didn't manage to load it, therefore I found that it got the resources from that sqlite database and inserted there. 
As a side note, the fields contain additional information like expiration date, etc.

louwers: It is intentional that the cache is checked in. The render tests run offline.

TimSylvester: ah, ok

PR #2247, opened 2024-04-09
Layer group uniforms

alexcristici:
Implemented layer group uniforms, we had only drawable uniforms.
Moved evaluated props UBOs to layer level for all layers.

This will reduce the number of UBO binds on Metal and OpenGL and it will reduce also the bandwidth between CPU and GPU.
I did some measurements and the number of binds is reduced with approx. 10%.


src/mbgl/gl/drawable_gl.cpp
```diff
@@ -108,7 +108,7 @@ void DrawableGL::bindUniformBuffers() const {
             const auto& block = uniformBlocks.get(id);
             if (!block) continue;
             const auto& uniformBuffer = getUniformBuffers().get(id);
-            assert(uniformBuffer && "UBO missing, drawable skipped");
+            /*assert(uniformBuffer && "UBO missing, drawable skipped");
```
stefankarschti, line 111: do we want / can we still do this check?

src/mbgl/mtl/drawable.cpp
```diff
@@ -387,7 +387,7 @@ void Drawable::bindUniformBuffers(RenderPass& renderPass) const noexcept {
             const auto& block = uniformBlocks.get(id);
             if (!block) continue;
             const auto& uniformBuffer = getUniformBuffers().get(id);
-            assert(uniformBuffer && "UBO missing, drawable skipped");
+            // assert(uniformBuffer && "UBO missing, drawable skipped");
```
stefankarschti, line 390: can this be reenabled?

mwilsnd (approved):
Looks good to me, Stefan already caught the two things I saw.

src/mbgl/gl/drawable_gl.cpp
alexcristici: No, because now some UBO can miss at the drawable level, being set a layer group level. So I will remove this assert.

src/mbgl/mtl/drawable.cpp
alexcristici: No, because now some UBO can miss at the drawable level, being set a layer group level. So I will remove this assert.

TimSylvester (approved):

stefankarschti:
There is a reproducible crash after a style change and some pan / zoom.
Screenshot 2024-04-17 at 17 45 12

stefankarschti (changes requested):
crash with metal validation error
```
-[MTLDebugRenderCommandEncoder validateCommonDrawErrors:]:5446: failed assertion `Draw Errors Validation
Fragment Function(fragmentMain): missing buffer binding at index 2 for tileprops[0].
```

src/mbgl/shaders/mtl/symbol_icon.cpp
```diff
@@ -3,23 +3,23 @@
 namespace mbgl {
 namespace shaders {
 
+const std::array<UniformBlockInfo, 5> ShaderSource<BuiltIn::SymbolIconShader, gfx::Backend::Type::Metal>::uniforms = {
+    UniformBlockInfo{true, true, sizeof(SymbolDynamicUBO), idSymbolDynamicUBO},
+    UniformBlockInfo{true, true, sizeof(SymbolDrawableUBO), idSymbolDrawableUBO},
+    UniformBlockInfo{true, false, sizeof(SymbolTilePropsUBO), idSymbolTilePropsUBO},
```
stefankarschti, line 9: `tileprops` ubo is not bound into fragment

alexcristici: Fixed.

stefankarschti (approved):

sjg-wdw:
Would love to see some performance numbers on this.  Can we kick off a Device Farm run?

Issue #2256, opened 2024-04-11
Node.js Windows CI Broken


louwers:
I updated the Node.js packages, it silently broke the Linux and macOS builds. I fixed those, but now the Windows Node.js build is failing.

https://github.com/maplibre/maplibre-native/actions/runs/8638457928/job/23682950776

I think @acalcutt ran into this one as well.

acalcutt:
Just an FYI, I think this is a different issue I started to encounter last night when I was testing.

I had determined my issue was related to the macos cmake fix I did, which removed the default shell, it cause windows configure to run as powershell (which broke it). After a lot of tweaking I had figured out how to get the cmake fix to run without changing the default shell. Right as I figured that out last night, I started getting a new windows error which I think is what you are getting now.

acalcutt:
It looks like it worked for me in this commit anyway 
https://github.com/maplibre/maplibre-native/pull/2252/commits/8238c791b4f3c431009a49b5c1ad3af51e76b3b9

I think that might be the last I saw it work though.

louwers:
@acalcutt https://github.com/microsoft/vcpkg/issues/38116 👀

acalcutt:
Looking in the comments in some link threads it seems like maybe this is cause y something that will be reverted in 3.26.2 , so I wonder when that will be in the github runners

PR #2263, opened 2024-04-15
Fix node windows workflow

acalcutt:
Fixes https://github.com/maplibre/maplibre-native/issues/2256

The github runners are using cmake 3.29.1, which has been causing issues in macos and windows ci.  This PR uses jwlawson/actions-setup-cmake to make the workflow use 3.29.2 like suggested in https://github.com/actions/runner-images/issues/9680#issuecomment-2052018532 , which seems to fix the build issues

It looks like 'jwlawson/actions-setup-cmake' was already being used in the linux runner, so this changes it to also be used in all github runners. It did not work in the self-hosted runner, so I have restricted it from running in those workers

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -265,10 +265,10 @@ jobs:
         run: set -o pipefail && ./build/mbgl-render-test-runner --manifestPath metrics/macos-xcode11-release-style.json
 
       - name: Upload render test artifacts (MacOS)
-        if: runner.os == 'MacOS'
+        if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: render-query-test-results
+          name: render-query-test-results_${{ runner.os }}_${{ matrix.arch }}
           path: metrics/macos-xcode11-release-style.html
 
       - name: Test (Linux)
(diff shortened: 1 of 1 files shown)


acalcutt:
@louwers I think i fixed they windows workflow by upgrading cmake to 3.29.2 in all github runners. It seems like it works from the ci.

I do notice the pre-commit doesn't like `runner.environment`, but I had read that was valid and seems to work as expected (since that section does not run on the self-hosted runner like I wanted). Any idea how I can fix that?

louwers:
Thanks!

Do you still have the error? We can probably report it over at [actionlint](https://github.com/rhysd/actionlint).

But it seems like you found a solution.

louwers (approved):

acalcutt:
The error was

```
.github/workflows/node-ci.yml:168:13: property "environment" is not defined in object type {arch: string; debug: string; name: string; os: string; temp: string; tool_cache: string} [expression]
    |
168 |         if: runner.environment == 'github-hosted'
```

After looking a bit more on it, I feel like maybe 'runner.environment' isn't yet documented properly yet
https://github.com/orgs/community/discussions/118413

louwers (approved):

acalcutt:
Just an FYI, I did put in an actionlint PR at https://github.com/rhysd/actionlint/issues/412 and it sounds like they may add runner.environment in the future. However I think the workaround of using name should work fine for now.

PR #2265, opened 2024-04-15
Update Android application ID to match expected value

mueschm:
The package name being used everywhere is `org.maplibre.android`. This is expected throughout the app. Because the `android` is missing the back button on the test app does not work. Specifically the arrow in the top left corner. This fixes the issue.

### DIFF
--- a/platform/android/Makefile
+++ b/platform/android/Makefile
@@ -139,8 +139,8 @@ run-android-benchmark-$1-%: android-benchmark-$1
 # Run the test app on connected android device with specified abi
 .PHONY: run-android-$1
 run-android-$1:
-	-adb uninstall org.maplibre.testapp 2> /dev/null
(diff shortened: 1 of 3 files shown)


mwilsnd (commented):
Looks like there are a few other references to the old applicationId.
* build.gradle: testNamespace 'org.maplibre.testapp.test'
* Makefile (platform/android) multiple locations

Pretty sure CI is using the makefile so I know that needs to get updated

louwers (approved):
Thanks.

I replaced the other instances. Looks like testNamespace [should be set automatically](https://developer.android.com/build/configure-app-module#change-namespace-for-testing).

The Makefile is pretty crufty... It's not used to build the test app on CI.

louwers:
Hold the fort. The namespace is `org.maplibre.android.testapp`, but the application ID is `org.maplibre.testapp`. So the other changes need to be reverted.

fynngodau:
I do think we chose this name intentionally, and I do wonder why this would break the back button. In any case I'm pretty sure this is not the only solution. 🤔 But if this fixes it it's not worth the effort of finding such other solution.

PR #2269, opened 2024-04-16
[opengl-2][node] Cherry pick "Fix node windows workflow" and "update the node packages"

acalcutt:
Cherry pick https://github.com/maplibre/maplibre-native/pull/2263 into the opengl-2 branch. This fixes the macos and windows node workflows in opengl-2

I also cherry picked https://github.com/maplibre/maplibre-native/pull/2246 to update the node packages.

Note: I added package.json and package-lock.json to the triggers for the node and qt workflows so that the required tests would actually run. This will also be useful when I make the next node release, since the version increment with trigger the required tests, where before they didn't and required an admin to merge.


louwers (approved):

.github/workflows/node-ci.yml
```diff
@@ -4,7 +4,10 @@ on:
   workflow_dispatch:
   push:
     branches:
+      - main
       - opengl-2
+      - topic/drawable
```
louwers, line 9: This branch is no longer used.

PR #2280, opened 2024-04-16
Sanitizer fixes

TimSylvester:
Enables the undefined-behavior sanitizer when running the iOS UI tests.  Resolves some errors observed with that and the thread sanitizer, and adds suppression for some issues reported in and because of third-party code which are acceptable.

I was unable to get compile-time suppressions to work, so each app where sanitizers are to be run will need to enable the suppressions at startup.


sjg-wdw (approved):

louwers (commented):
From the metal-cpp [README.md](https://github.com/bkaradzic/metal-cpp/tree/metal-cpp_macOS14.2_iOS17.2?tab=readme-ov-file#nullptr):

> Similar to Objective-C, it is legal to call any method, including `retain()` and `release()`, on `nullptr` "objects". While calling methods on `nullptr` still does incur in function call overhead, the effective result is equivalent of a NOP.

> Conversely, do not assume that because calling a method on a pointer did not result in a crash, that the pointed-to object is valid.

I don't think this is needed.

alexcristici (approved):

louwers (changes requested):
See comment above.

TimSylvester:
It's almost certainly not causing problems in production, but it's technically not correct, and prevents use of UBSan by causing it to fail immediately.  An alternative would be to generate a suppressions file which would need to be put on the device and specified in the environment variables when running UBSan, or putting `__attribute((no_sanitize(...)))` all over .

That said, I'm fine with backing it off to just the mutex change.

@mwilsnd ?

mwilsnd:
Not being able to use ubsan is going to be really painful. The objc ABI may allow sending to null but if it's tripping ubsan then I feel it should be addressed. Wrapping the shared pointer is a good option to reduce the number of places ubsan will complain. I'm fine either keeping the null checks inside the wrapper or annotating the offending calls inside the wrapper to suppress ubsan. There could be arguments that a null check is faster compared to the redundant msg send call but I don't think they would be substantial enough to bias any decision here.

louwers:
Can we add ubsan to the Bazel config? Seems like it has support for it. https://github.com/bazelbuild/bazel/commit/8959dff512fe4505af786bcf2ef981ec7082a913

I don't want to add a wrapper class because a tool (albeit an important one) misclassified valid code as errors.

TimSylvester:
It would probably be a good idea to run the tests with the sanitizers, though apparently you can't use asan and ubsan together.  That would just make this more important, though.

I could probably make it so that the type aliases resolve to `NS::SharedPtr` in regular builds and `SafeSharedPtr` when ubsan is enabled...

louwers:
Let's exclude all metal-cpp code from UBSan?

louwers:
I don't like the idea of making the wrapper class conditional. Running different code with and without a sanitizer defeats the purpose of using one in my opinion.

Since the error includes the source path it should be possible to configure UBSan instead.

```
vendor/metal-cpp/Foundation/NSSharedPtr.hpp:219:16: runtime error: member call on null pointer of type 'NS::Referencing<MTL::DepthStencilState>'
```

Looking at `scripts/usban.blacklist` this is also what Mapbox did.

TimSylvester:
I think the practical way to do excludes would be to add a suppression file to the assets, and then find its path at startup and pass it to `__ubsan_default_options` (in `sanitizer/ubsan_interface.h`).

TimSylvester:
Oh, it can be specified at compile time, that's much easier.  I guess we can use `--copt` when using `--features=ubsan`, or maybe detect the latter in one of the Bazel files and add it automatically.

https://clang.llvm.org/docs/SanitizerSpecialCaseList.html

louwers:
Yes adding the config to the assets is a good idea.

We can pass the location as an environment variable it looks like:

https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#runtime-suppressions

By the way, could you add `--features=ubsan` to `//platform/ios/iosapp-UITests` in `ios-ci.yml`? That may be a good place to start running it on CI.

I think you can also pass `--features=default_sanitizer_flags_feature` to run the whole bunch.

TimSylvester:
Wouldn't it be added to the "test host" `//platform/ios:App`?

stefankarschti (approved):

TimSylvester:
I haven’t been able to get -fsanitize-ignorelist to have any effect, nor does setting UBSAN_OPTIONS environment options seem to have any effect on the output, so I suspect that even if I put the ignore list in the assets and figure out its path from the app, that wouldn’t work either.  I’m going to try using __ubsan_default_options at startup, next…

louwers (changes requested):
Any chance we can address the outstanding remove the `__has_feature(undefined_behavior_sanitizer)` preprocessor macros in favor of the config?

TimSylvester:
The render test pass either way, so I left it on clamp, which should eliminate the need for an exception.

mwilsnd (approved):

louwers (approved):

PR #2151, opened 2024-02-27
[opengl-2] Cherry pick - Implementation of multi-sprites in mbgl-core

acalcutt:
Cherry pick - Implementation of multi-sprites in mbgl-core

PR #2286, opened 2024-04-17
[opengl-2][node] Release node-v5.4.0-pre.0

acalcutt:
This PR is to Release node-v5.4.0-pre.0

This release adds support for multi-sprites, which have been cherry picked from the main branch in https://github.com/maplibre/maplibre-native/pull/2151

I also updated the release workflow to match the ci, so this release will actually work. I can cherry pick this to main later when the version gets updated there. Note that this also changes the macos arm64 binary to be built on the github 'macos-13-xlarge' runner, so this is a slighly newer version of macos that could need testing for macos arm64 users, which this prerelease will be useful for. I do wonder though if we should change macos-12 to macos-13, for the AMD64 runner.

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -49,7 +49,7 @@ jobs:
             arch: arm64
           - runs-on: macos-12
             arch: x86_64
-          - runs-on: [self-hosted, macOS, ARM64]
+          - runs-on: macos-13-xlarge
             arch: arm64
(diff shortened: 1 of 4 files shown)


acalcutt:
@ntadej  @louwers  I still need an approval on this even though it says it is not needed right? Just wanting to make this release while we know the workflow is working and get this sprite feature parity with the rest of maplibre

PR #2292, opened 2024-04-18
Global UBO

alexcristici:
Implemented global UBO, for paint parameters that are same for all shaders.
Optimised shaders UBOs by using the global UBO.

This will reduce the number of UBO binds and it will reduce also the bandwidth between CPU and GPU.

The shaders have the following UBOs:
- paint parameters UBO, it's set at **global level** and contains parameters that are the same for all shaders.
- evaluated properties UBO, it's set at **layer group level** and contains the layer properties.
- drawable UBO, it's set at **drawable level** and contains the `matrix` and other properties that are drawable dependent.
- interpolate UBO and tile properties UBO are set at **drawable level** in the render layer.


TimSylvester (approved):

louwers:
!benchmark android

mwilsnd (approved):

sjg-wdw (approved):

PR #2293, opened 2024-04-18
iOS - Allow changing of User-Agent

hactar:
MapLibre allows setting additional headers to be sent in network requests via `MLNNativeNetworkManager.sharedManager.sessionConfiguration`. Here a user could set up a `User-Agent` to be sent during requests. Unfortunately, in `http_file_source.mm`, we set a `User-Agent` without checking if a user might have set one too.

This prevents users from being able to set a `User-Agent` at all. In my use case for example, I have to talk to a service which expects a certain `User-Agent`, and currently I have no way to get MapLibre to use the required `User-Agent`.

This PR resolves the issue by checking if a custom `User-Agent` has been set in `MLNNativeNetworkManager.sharedManager.sessionConfiguration` first - if one is set, then that one is used, else MapLibre defaults to its own.


louwers (approved):
Sounds reasonable.

PR #2296, opened 2024-04-18
Always use blending for basic fill layers

mwilsnd:


TimSylvester (approved):
Makes sense, if an alpha value is specified it's hard to imagine how we could get correct results without using it, but there is a clear intent to do that sometimes in legacy so I have a suspicion that there might be an even more obscure bug lurking there.

louwers (approved):

louwers:
Seems to be fixed!!

PR #2297, opened 2024-04-19
Update iOS BAZEL.md

hactar:
If the commands are executed the way described in this file, the command fails as on `bazel-package.sh`

```
if [ ! -d platform/ios/build ]; then
   mkdir platform/ios/build
fi
```

as there is no `platform/ios` folder in `platform/ios/scripts` and mkdir only would create the top directory `build`, but not `platform` and `ios` and then `build`. I've changed the documentation so that it is in reference to the root of the repo, which appears to be the way the scripts are set up. By executing the commands from there they start successfully.

Screenshot 2024-04-19 at 13 22 14

### DIFF
--- a/platform/ios/BAZEL.md
+++ b/platform/ios/BAZEL.md
@@ -10,25 +10,23 @@ From there you can use the script in platform/ios/platform/ios/scripts/package-b
 
 ## There are 4 options:
 
-`cd platform/ios/platform/ios/scripts`
-
 Static xcframework compiled for release (this is default if no parameters are provided):
-`./bazel-package.sh --static --release`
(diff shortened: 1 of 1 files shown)


louwers:
Maybe we should get rid of both `BAZEL.md` and `bazel-package.sh`. They were checked in with the first Bazel PR, and not updated since. Bazel is now the only supported build method on iOS. We don't use the script on CI to build the XCFramework.

hactar:
Alright, I'm fighting some issues with build & run and was investigating, stumbled across this file and that script. If it iss unsupported then yes, let us remove it to not confuse new developers with outdated info. I'll close this PR and create a new one to remove the files.

PR #2299, opened 2024-04-19
Remove bazel-package.sh and BAZEL.md

louwers:
This removes `bazel-package.sh`. I want to avoid having scripts scripts checked in that are not used on CI, because they tend to go out of date quickly. I also think developer should use Bazel directly as much as possible instead of using wrapper scripts, because this makes it easier to debug issues.

We could move the information from `BAZEL.md` to the Wiki or in docstrings on the top of `BUILD.bazel` files, but I don't see much added value.

Added `--features=dead_strip --objc_enable_binary_stripping` to the XCFramework build on CI. Not sure what the effect will be.


hactar:
@louwers ah thanks, was about to create a PR for this ☺️

mwilsnd (approved):

PR #2304, opened 2024-04-21
Android annotations: Implement `CollisionGroup`

### LINKED ISSUE #1495: Android Annotations: `CollisionGroup`

---

Add the `CollisionGroup` object as specified in the proposal.

* `CollisionGroup` should be implemented.

fynngodau:
[Android annotations API proposal](https://github.com/maplibre/maplibre-native/blob/main/design-proposals/2023-06-17-android-annotations.md) implementation PR as a part of https://github.com/maplibre/maplibre-native/issues/1491. Closes #1495.

---

* `CollisionGroup` is implemented per specification. In effect, it is somewhat of a wrapper around `SymbolManager`.
* The `CollisionGroupTest` contains some tests I found to be helpful.
* One annoyance is that I couldn't add an observable list as I hoped. Instead, the list of `symbols` is currently immutable, and can be set to a new value through assignment (which we listen to using `Delegates.observable`). We could decide to add `add(Symbol)`, `delete(Symbol)`, etc. instead.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/annotations/CollisionGroup.kt
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/annotations/CollisionGroup.kt
@@ -7,6 +7,7 @@ import kotlin.properties.Delegates
 
 class CollisionGroup(
     symbols: List<Symbol> = emptyList(),
+    val zLayer: Int = Defaults.Z_LAYER,
     symbolSpacing: Float = Defaults.COLLISION_GROUP_SYMBOL_SPACING,
     symbolAvoidEdges: Boolean = Defaults.COLLISION_GROUP_SYMBOL_AVOID_EDGES,
     iconAllowOverlap: Boolean = Defaults.COLLISION_GROUP_ICON_ALLOW_OVERLAP,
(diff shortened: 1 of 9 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/annotations/CollisionGroup.kt
```diff
@@ -0,0 +1,121 @@
+package org.maplibre.android.annotations
+
+import okhttp3.internal.toImmutableList
```
louwers, line 3: We should probably not use internal packages.

louwers (changes requested):
Could you add an example to the test app or modify an existing example?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -1090,6 +1096,10 @@ public void removeAnnotations(@NonNull KAnnotation... annotations) {
     }
   }
 
+  public void removeCollisionGroup(@NonNull CollisionGroup collisionGroup) {
```
louwers, line 1099: Needs a documentation comment.

PR #2309, opened 2024-04-22
Run Android Instrumentation Tests as part of PR workflow

louwers:
The Android Instrumentation Tests have been failing for a while. What doesn't help is that they don't run as part of PRs.

This PR refactors the workflows a bit so that they do. This adds a failing check to each PR (temporarily), but we have 'Required' checks that should prevent a PR from being merged if those checks do not pass.


louwers:
!benchmark android

HarelM (approved):

HarelM (approved):

HarelM:
Just out of curiosity, how does one protect the aws device runners from being used by other people? I didn't see any secret in the lines that were changed and if this should run as part of a PR it means that it can't be protected...?

louwers:
@HarelM I store the build as an artifact and then use a privileged `workflow_call` workflow that gets run as soon as the unprivileged workflow is finished.

https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/android-device-test.yml#L4-L7

I use this pattern a lot.

PR #2314, opened 2024-04-23
Replace mapboxMap with maplibreMap

louwers:
- Remove "mapboxMap" and variations in a bunch of variable names and comments.
- Remove dead code in `CameraUpdateFactory.kt`
- The only potentially API breaking change is the changed FRAG_ARG_MAPLIBREMAPOPTIONS. But I don't the value of this value is persisted between app runs.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/annotations/Annotation.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/annotations/Annotation.java
@@ -70,18 +70,18 @@ public abstract class Annotation implements Comparable<Annotation> {
   /**
    * Do not use this method, used internally by the SDK.
    *
-   * @param maplibreMap the hosting mapbox map
+   * @param maplibreMap the hosting MapLibreMap
    */
-  public void setMapboxMap(MapLibreMap maplibreMap) {
+  public void setMapLibreMap(MapLibreMap maplibreMap) {
     this.maplibreMap = maplibreMap;
   }
 
   /**
-   * Gets the hosting mapbox map.
+   * Gets the hosting MapLibreMap.
    *
-   * @return the MapboxMap
+   * @return the MapLibreMap
    */
-  protected MapLibreMap getMapboxMap() {
+  protected MapLibreMap getMapLibreMap() {
     return maplibreMap;
   }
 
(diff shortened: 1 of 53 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/annotations/Annotation.java
```diff
@@ -70,18 +70,18 @@ public void setId(long id) {
   /**
    * Do not use this method, used internally by the SDK.
    *
-   * @param maplibreMap the hosting mapbox map
+   * @param maplibreMap the hosting MapLibreMap
    */
-  public void setMapboxMap(MapLibreMap maplibreMap) {
```
louwers, line 75: I mean... The comment is pretty clear right.

Maybe we want to deprecate this first though.

fynngodau: The whole class is deprecated so I'm not sure what use deprecating its individual methods is :)

cgalvan (approved):

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/style/SymbolGeneratorActivity.kt
```diff
@@ -139,7 +139,7 @@ class SymbolGeneratorActivity : AppCompatActivity(), OnMapReadyCallback {
      * Utility class to generate Bitmaps for Symbol.
      *
      *
-     * Bitmaps can be added to the map with [com.mapbox.mapboxsdk.maps.MapboxMap.addImage]
+     * Bitmaps can be added to the map with [MapLibreMap]
```
mwilsnd, line 142: Was this change fully intentional?

louwers: Yes. It works this way.

louwers: Oh I see. This is not the same.

louwers: I don't think this method ever existed. 😛

PR #2321, opened 2024-04-24
Android Instrumentation Tests Improvements

louwers:
I want to make the Android Instrumentation Tests pass on CI so we can rely on them for PRs.

In this PR I am pulling the logs from AWS Device Farm and log them on GitHub (for debugging in case a test fails).

Had to ignore quite a few tests because they are flaky.

`Style.getPredefinedStyle("Streets")` was failing unpredictably because the MapTiler styles weren't loaded yet. I replaced it with `Style.getDefaultStyles()[0].url` wherever possible. Maybe I will define the styles needed for the tests near the tests instead of relying on the pre-defined styles in the core (that we want to get rid of at some point).

I'll try to improve things further in another PR.

### DIFF
--- a/platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/offline/OfflineDownloadTest.kt
+++ b/platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/offline/OfflineDownloadTest.kt
@@ -2,6 +2,7 @@ package org.maplibre.android.offline
 
 import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
 import androidx.test.rule.ActivityTestRule
+import org.junit.Ignore
 import org.maplibre.geojson.Point
 import org.maplibre.android.log.Logger
 import org.maplibre.android.maps.Style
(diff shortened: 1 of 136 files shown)


platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/offline/OfflineDownloadTest.kt
```diff
@@ -16,6 +17,7 @@ import java.util.concurrent.TimeoutException
 /**
  * Integration test that validates downloading an offline region from a point geometry at zoomlevel 17
  */
+@Ignore("https://github.com/maplibre/maplibre-native/issues/2318")
```
cgalvan, line 20: Thanks for creating these issues so we don't forget to come back and look at these flaky tests :)

scripts/aws-device-farm/s3-client.mjs
```diff
@@ -0,0 +1,7 @@
+import { S3Client } from "@aws-sdk/client-s3";
+
+export function getS3Client() {
+  return new S3Client({
+    region: "eu-central-1",
```
cgalvan, line 5: Just wanted to call out that the device farm region is `us-west-2`. Should we have them on the same region, or do we want to test with them in different regions for more coverage?

cgalvan (approved):

scripts/aws-device-farm/s3-client.mjs
louwers: Our S3 bucket is in eu-central-1.

louwers: Yeah the AWS Device Farm client uses `us-west-2`. 👍

Issue #2082, opened 2024-02-01
Annotation view desynchronization on Metal while multiple frames in flight


stefankarschti:
Annotation Views desynchronize from the map on Metal.

ref #2053

louwers:
@stefankarschti Can this one be closed now that https://github.com/maplibre/maplibre-native/commit/9614b566d8e497cf8afffc3d26b004a1518baefe is merged?

stefankarschti:
fixed by #2334

PR #2334, opened 2024-04-25
Synchronous rendering on Metal when annotation views are visible

### LINKED ISSUE #2053: [iOS - Metal - v6.0.0]  - MLNAnnotationView delay with map panning

stefankarschti:
Implement conditional synchronous rendering on Metal backend when annotation views are visible, that is waiting for frames to fully complete by invoking `[commandBuffer waitUntilCompleted];` in the `swap` method of the `MLNMapViewMetalRenderableResource`.

This is needed because the map transformation is set as soon as the camera is mutated, and annotations are positioned on the screen according to that in `MLNMapView`::`renderSync` -> `updateAnnotationViews`, while the frame rendering is completing asynchronously on Metal - which causes desynchronization.

ref: #2082 
Fixes #2053

### DIFF
--- a/platform/ios/src/MLNMapView+Metal.h
+++ b/platform/ios/src/MLNMapView+Metal.h
@@ -50,8 +50,8 @@ public:
     MLNBackendResource getObject() override;
     // End implementation of MLNMapViewImpl
 
-    virtual void setSynchronous(bool value) override { synchronousFrame = value; };
-    virtual bool getSynchronous() const override { return synchronousFrame; };
+    void setSynchronous(bool value) override { synchronousFrame = value; };
+    bool getSynchronous() const override { return synchronousFrame; };
 private:
     bool presentsWithTransaction = false;
     bool synchronousFrame = false;
--- a/platform/ios/src/MLNMapView+Metal.h
+++ b/platform/ios/src/MLNMapView+Metal.h
@@ -50,6 +50,9 @@ public:
     MLNBackendResource getObject() override;
     // End implementation of MLNMapViewImpl
 
+    virtual void setSynchronous(bool value) override { synchronousFrame = value; };
+    virtual bool getSynchronous() const override { return synchronousFrame; };
 private:
     bool presentsWithTransaction = false;
+    bool synchronousFrame = false;
 };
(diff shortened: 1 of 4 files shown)


platform/ios/src/MLNMapView+Metal.h
```diff
@@ -50,6 +50,9 @@ class MLNMapViewMetalImpl final : public MLNMapViewImpl,
     MLNBackendResource getObject() override;
     // End implementation of MLNMapViewImpl
 
+    virtual void setSynchronous(bool value) override { synchronousFrame = value; };
```
mwilsnd, line 53: nit: `virtual` isn't required when the `override` keyword is present, high compiler warning levels can complain if they follow the core guidelines on [virtual, override and final](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final)

mwilsnd (changes requested):

mwilsnd (approved):

michaelkirk:
Thank you so much, it looks much better now! ❤️

PR #2343, opened 2024-04-29
Move Node.js package.json to platform/node

### LINKED ISSUE #1334: Separate `package.json` for script dependencies
I want to create a separate `package.json` for dependencies of the various scripts.

We can move the `package.json` for Node.js to `platform/node`.

@acalcutt What do you think?

louwers:
I want to split the `package.json` for the Node.js package and the `package.json` used to define the dependencies for the various scripts used. Will probably fail.

@acalcutt What do you think about this?

I probably need to update a few other paths.

Closes #1334


platform/node/package.json
```diff
@@ -0,0 +1,79 @@
...
+  ],
+  "directories": {
+    "lib": "lib"
+  },
+  "files": [
+    "lib",
+    "platform/node/index.d.ts"
+  ],
+  "main": "platform/node/index.js",
+  "types": "platform/node/index.d.ts",
```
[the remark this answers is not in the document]
louwers: Oh yeah of course.

platform/node/index.js
```diff
@@ -2,7 +2,7 @@
 
 // Shim to wrap req.respond while preserving callback-passing API
 
-var mbgl = require('../../lib/node-v' + process.versions.modules + '/mbgl');
+var mbgl = require('lib/node-v' + process.versions.modules + '/mbgl');
```
louwers, line 5:
```suggestion
var mbgl = require('./lib/node-v' + process.versions.modules + '/mbgl');
```

acalcutt:
I added some changes to merge into this in your repo
https://github.com/louwers/maplibre-native/pull/12

This fixes the node-release workflow so it will work with these changes

louwers:
@acalcutt Awesome!

acalcutt:
One thing I just noticed is how does the ci pass if there is no new package-lock.json?

.github/workflows/node-release.yml
```diff
@@ -119,6 +120,7 @@ jobs:
           node-version-file: '.nvmrc'
 
       - name: npm ci
+        working-directory: platform/node
```
acalcutt, line 123: I am unsure if this npm ci is needed or not. in the node-ci it does not included this working-directory, so it seems it would behave differently... but that didn't seem to make a difference in the build so maybe this step isnt needed?

acalcutt: Thinking now , the npm dev dependencies might be needed for the tests, which don't run in the release...so maybe that explains why it didn't make a differenct.

I think at minimum this `working-directory: platform/node` line I added should be removed

louwers, line 123:
```suggestion
```

acalcutt:
The only other thing I see is maybe there should be a package-lock.json in 'platform/node'. I had that in my directory when testing the release. I think it is needed because the publish step does a 'npm ci --ignore-scripts', which would check that file.

birkskyum:
Great idea. Alternatively to a package lock inside the node, we could add a [pnpm-workspace.yml](https://pnpm.io/pnpm-workspace_yaml) and use pnpm in ci as well

PR #2353, opened 2024-05-02
Extending Within expression test case to complex geometries

TimSylvester:
I reworked the "Within expression" Android test/demo case to help investigate a reported crash related to Earcut and Within.  Fill-extrusion does an extra Earcut call, so that's used instead of fill, and the points are shuffled randomly to produce arbitrary geometries.

Steve suggested that perhaps this code should be saved for future use.

### DIFF
--- a/platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/turf/WithinExpressionActivity.kt
+++ b/platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/turf/WithinExpressionActivity.kt
@@ -240,6 +240,7 @@ class WithinExpressionActivity : AppCompatActivity() {
     )
     
     private fun bufferLineStringGeometry(coordinates: List<Point>): Polygon {
+        // TODO replace static data by Turf#Buffer: mapbox-java/issues/987
         return Polygon.fromLngLats(listOf(coordinates))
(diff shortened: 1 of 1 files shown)


louwers:
@TimSylvester Could you fix the merge conflict or invite me to the MapLibre Native fork of Wet Dog?

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/turf/WithinExpressionActivity.kt
```diff
@@ -126,17 +107,47 @@ class WithinExpressionActivity : AppCompatActivity() {
                     "poi-label"
                 )
         )
+
+        // Un-comment to repeatedly shuffle the points of the bound as a test of fills and earcut
```
louwers, line 111: Can we turn this into an automated regression test somehow?

TimSylvester: Yeah, it probably makes more sense as a unit test.  We never turned up any actual problem with earcut, I think the reports were a red herring.  The handling of degenerate polygons is likely part of the test suite for that project anyway, so this may be fully superfluous, anyway.

PR #2362, opened 2024-05-06
Reduce memory use of `GridIndex`

TimSylvester:
The `GridIndex` class uses a 64-bit integer as the unique identifier for a feature within a grid, which is per-bucket or feature query.  I think we can safely assume that it's not necessary to support more than 2^32 features.

These are not huge, but are among the largest groups of persistent allocations, behind HTTP responses and the largest vertex buffers.  

image

Also adds some reserve calls based on observations of large numbers of allocations.

### DIFF
--- a/src/mbgl/util/grid_index.hpp
+++ b/src/mbgl/util/grid_index.hpp
@@ -4,14 +4,14 @@
 #include <mapbox/geometry/box.hpp>
 #include <mbgl/math/minmax.hpp>
 
-#include <unordered_set>
+#include <cassert>
 #include <cmath>
-#include <cstdint>
 #include <cstddef>
-#include <vector>
+#include <cstdint>
 #include <functional>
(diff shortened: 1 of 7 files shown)


sjg-wdw (approved):

src/mbgl/util/grid_index.hpp
```diff
@@ -103,8 +103,9 @@ class GridIndex {
     std::vector<std::pair<T, BBox>> boxElements;
     std::vector<std::pair<T, BCircle>> circleElements;
 
-    std::vector<std::vector<size_t>> boxCells;
-    std::vector<std::vector<size_t>> circleCells;
+    using uid_t = uint32_t;
```
louwers, line 106: Leaving this out is clearer in my opinion.

louwers (approved):

PR #2364, opened 2024-05-06
Reduce memory use of `FeatureIndex`

TimSylvester:
`FeatureIndex` stores a large number of `IndexedSubfeature` for fast searching of intersecting features by grid cell.  Each subfeature contains a copy of two strings, which is quite wasteful.  The benchmark results in a large number of buckets for which between 10K and 50K items are created, among which there are only 8 and 30 unique values for those strings.

So by storing these values only in the index and referencing them in the sub-features we can eliminate a large number of allocations and copies.  Unfortunately, these values are sometimes used outside of the index, so a separate class which does copy the strings is used when, e.g., returning search results, making it somewhat more complex.

No significant performance effect.  Peak memory utilization is only reduced from 426 to 419MB, but differences of about 40MB appear in some places.

before:

Average frame encoding time: 2.51 ms
Average frame rendering time: 6.61 ms

Average frame encoding time: 2.54 ms
Average frame rendering time: 6.65 ms

after:

Average frame encoding time: 2.52 ms
Average frame rendering time: 6.66 ms

Average frame encoding time: 2.53 ms
Average frame rendering time: 6.64 ms


sjg-wdw (approved):
I'd be interested it know how much memory we could save by eliminating the ability to look up features entirely.  We might want to, opt-in, allow that.

TimSylvester:
yeah, that's the next step

TimSylvester:
Disabling `FeatureIndex` completely doesn't make a big difference:



Interestingly, the branch that unconditionally inserts every (non-filtered) feature into the index is never used in either the LM or MapLibre Basic styles, apparently every layer is "layout required" and as a result most layers don't add features to the index.  It seems that it's not actually used for symbols as I was thinking.

image

sjg-wdw:
Well that's wild.

louwers (commented):
> No significant performance effect. Peak memory utilization is only reduced from 426 to 419MB, but differences of about 40MB appear in some places.

How did you measure this?

TimSylvester:
From the allocations instrument UI in iOS profiling.  The image is two screenshots from that subtracted from one another.

stefankarschti (approved):

alexcristici (approved):

Issue #2357, opened 2024-05-04
CURL_POLL_REMOVE socket action is unhandled and causes runtime error


glebpinigin:
**Describe the bug**
mbgl-glfw often fails with network requests using curl 8.7.1 version
After opening mbgl-glfw application it fails with runtime error "Unhandled CURL socket action"

**To Reproduce**
Build mbgl-glfw target and run it with style url. It may crash immediately, or it can success first several requests and then crash.

**Platform information (please complete the following information):**
 - OS: Archlinux
 - Platform GLFW
 - Version upstream (https://github.com/maplibre/maplibre-native/commit/cd6a0ab11fcd9b5bbfee87896a14159d0ed56256)
 - CURL version 8.7.1

**Additional context**
I have built this target also under debian with curl version 7.88.1 and everything worked fine. What is more, qt example which also fetches tiles by the URL works well on my setup with system installed curl 8.7.1

Crash happens here: https://github.com/maplibre/maplibre-native/blob/cd6a0ab11fcd9b5bbfee87896a14159d0ed56256/platform/default/src/mbgl/storage/http_file_source.cpp#L210

I have modified this line adding value of action to output
```
        default: {
            std::stringstream ss;
            ss << "Unhandled CURL socket action: " << action;
            throw std::runtime_error(ss.str());
        }
``
And it crashed with following notice after rebuild: `Unhandled CURL socket action: 3`. Which means that I received CURL_POLL_INOUT socket action.

glebpinigin:
I have fixed this behavior by adding CURL_POLL_INOUT action handling as following. I have no idea what is actually going on here but it is quiet intuitive that INOUT action should be handled with ReadWrite Event (luckily this type of event exists in mbgl). If this approach is OK, I can open pull request.
`
int HTTPFileSource::Impl::handleSocket(
    CURL * /* handle */, curl_socket_t s, int action, void *userp, void * /* socketp */) {
    assert(userp);
    auto context = reinterpret_cast<Impl *>(userp);

    switch (action) {
        case CURL_POLL_IN: {
            using namespace std::placeholders;
            util::RunLoop::Get()->addWatch(
                static_cast<int>(s), util::RunLoop::Event::Read, std::bind(&Impl::perform, context, _1, _2));
            break;
        }
        case CURL_POLL_OUT: {
            using namespace std::placeholders;
            util::RunLoop::Get()->addWatch(
[...]

PR #2365, opened 2024-05-07
CURL_POLL_INOUT action handle added

glebpinigin:
Now HTTPFileSource handles CURL_POLL_INOUT socket action with ReadWrite event.
Issue #2357 mbgl-glfw fails with curl 8.x.x
After this fix build succeeds and app does not fail with my setup.

Please beware of merging this without proper control: I am not a C++ developer and barely understand what exactly my proposal does.


louwers:
> I am not a C++ developer 

You are now!

louwers (approved):

Issue #2326, opened 2024-04-24
Rendering Glitch on older iOS devices


sjg-wdw:
Seeing the occasional long frame glitch on old iOS devices.
This is a placeholder for the investigation Stefan is doing on an issue he found.

sjg-wdw:
Creating pipeline descriptors on older devices is very expensive.  Should cache them.

stefankarschti:
fixed by #2379

PR #2379, opened 2024-05-10
Fix Metal frame glitch on older devices

stefankarschti:
Creating the pipeline state for all the drawables can amount to several milliseconds on older devices and thus produce a longer frame. Implementing a cache on the shader should fix this issue.

Ref: #2326


include/mbgl/shaders/mtl/shader_program.hpp
```diff
@@ -62,7 +62,7 @@ class ShaderProgram final : public gfx::ShaderProgramBase {
 
     MTLRenderPipelineStatePtr getRenderPipelineState(const gfx::Renderable&,
                                                      const MTLVertexDescriptorPtr&,
-                                                     const gfx::ColorMode& colorMode) const;
+                                                     const gfx::ColorMode& colorMode);
```
louwers, line 65: This would be a good use of the `mutable` keyword. You can use it on the (internal) cache to keep the method `const`.

sjg-wdw (approved):
What sort of improvements are you seeing in pipeline state numbers?

stefankarschti:
The glitch frames are gone while testing. I'm running the benchmark now.

TimSylvester (approved):

TimSylvester:
What are the typical cache sizes and number of shader instances?  I wonder if we need to be concerned about limiting them.  Unordered maps should typically be `reserve`d as well.

stefankarschti:
```
iPad 5th generation

main

Average frame encoding time: 22.88 ms
Average frame rendering time: 10.38 ms

Average frame encoding time: 23.23 ms
Average frame rendering time: 10.46 ms

Average frame encoding time: 23.24 ms
Average frame rendering time: 10.43 ms

#2379

Average frame encoding time: 3.89 ms
Average frame rendering time: 9.07 ms

Average frame encoding time: 3.77 ms
Average frame rendering time: 8.91 ms

Average frame encoding time: 3.91 ms
Average frame rendering time: 8.98 ms
```

cgalvan (approved):

sjg-wdw:
Makes sense the render time went down too.  Nice work!

stefankarschti:
> What are the typical cache sizes and number of shader instances? I wonder if we need to be concerned about limiting them. Unordered maps should typically be reserved as well.

Looks like there's not too many entries in the `Streets` style. I'm trying it with others.

```
ClippingMaskProgram#42b00bf2bf2c6069 1
FillShader#cebf426f1e2d36cb 1
FillOutlineTriangulatedShader#2d87de870ed42302 1
SymbolSDFIconShader#c142b455133a0218 1
LineShader#9665b4457a71986f 1
LineSDFShader#9665b4457a71986f 1
SymbolIconShader#c142b455133a0218 1
FillPatternShader#cebf426f1e2d36cb 1
FillOutlinePatternShader#cebf426f1e2d36cb 1
FillExtrusionShader#939482c3288476cf 1
FillExtrusionShader#939482c3288476cf 2
```

sjg-wdw:
About what I'd expect.  Nice!

alexcristici (approved):

PR #2382, opened 2024-05-10
Do text/sdf halos and fills in the same shader ...

TimSylvester:
This results in slightly different results because it composites onto the background only once instead of twice.  If I understand this correctly, it's not possible to produce exactly the same results without knowing the color onto which the fragment will be composited, but rendering the halo and fill at the same time is as reasonable as doing them separately.

I haven't done the OpenGL version yet.

Credit to @stefankarschti for perfecting the estimate of composited RGBAs.

@sjg-wdw, @louwers: Is this worth pursuing, even if we'll have to accept the differences and re-baseline those tests?  My feeling is that nobody will be bothered by such minor differences.


louwers:
What are the advantages of this?

sjg-wdw:
For symbols with halos we drop half the rendering cost.  Or nearly half.

TimSylvester:
Yes, we currently produce two drawables per "renderable item" that the layout produces, which is an arbitrary group of labels/SDF icons with the same properties.  This allows us to produce one drawable and only do one `drawElements` (per segment), and avoid running the shaders twice with just one boolean value being different.

To answer a notification go by... yes, the results will be slightly different any time the "background" is different, regardless of whether that's s different layer or another overlapping label.  But it's arguably less correct for the background to be blended into the result twice.

I didn't get to running the benchmarks yet.

wipfli:
It is an interesting idea to do both halo and fill of glyphs in the same draw call.

If we change it here in Native, I would suggest that we do the same in MapLibre GL JS, cc @HarelM.

One option could also be to make this behavior opt-in.

wipfli:
> For symbols with halos we drop half the rendering cost. Or nearly half.

It is true that the symbol shaders only have to be called once instead of twice if we merge this pull request. However I wonder how much time is actually spent to run the shaders vs time spend to compute symbol collisions and prepare the data that will be sent from the CPU to the GPU. @TimSylvester can you give an estimate on the time spent for the two?

TimSylvester:
I'll measure it on my iPad on Monday.  I expect relatively little difference overall, but a view which is particularly label-heavy might be more compelling.

The main layout occurs with parsing on background threads, so mainly affects load latency.  The effective ratio of work on the render thread; updating visibility based on collision maps, evaluating attribute arrays, encoding to rendering ultimately depends on the hardware.  Fewer drawables also means fewer bindings, so I'd expect some benefit all around on a device where we're not fully utilizing the GPU.

TimSylvester:
```
main:

Average frame encoding time: 1.24 ms
Average frame rendering time: 3.63 ms

Average frame encoding time: 1.23 ms
Average frame rendering time: 3.63 ms

Average frame encoding time: 1.22 ms
Average frame rendering time: 3.63 ms

---

symbol-halo:

Average frame encoding time: 1.20 ms
Average frame rendering time: 3.64 ms

Average frame encoding time: 1.14 ms
Average frame rendering time: 3.61 ms

Average frame encoding time: 1.14 ms
Average frame rendering time: 3.60 ms
```

I'd need a bunch more runs to conclude anything with real confidence, but maybe 5% less encoding time, if it's not a fluke, with no significant additional rendering time on this device.

sjg-wdw:
That's one we kind of hoped would reduce the rendering time.  Curious.
But less encoding time is always welcome, particularly when we get to Android.

wipfli:
Are these numbers for three times the same data and style?

TimSylvester:
Yes, three consecutive runs of the benchmark app.  Though I'm still using the LM style rather than the default Basic.

xabbu42:
Does this also improve the rendering of overlapping labels as recently reported for maplibre-gl-js on slack?

TimSylvester:
I don't know, perhaps for some definition of improve.

On closer inspection, the halo blending isn't quite right and is cutting off the edge of the glyphs.  But I can't get the debugger to work at the moment.  That would need to be fixed if we're going to proceed.

image

image

xabbu42:
I mean, will the halo of the top label be drawn over the text of the labels below to make the top label readable?

TimSylvester:
I wouldn't expect it to have much effect, as the rendering was interleaved before: each label's halo was drawn immediately before the label, as opposed to drawing all the halos and then all the labels.  So while a previously rendered label will only be sampled once, the difference is only in the antialiased edge of the glyphs where two blends happened.

TimSylvester:
I was finally able to get frame captures, and it looks like those stubborn remaining errors occur because the glyph quads overlap.

In the example above, when the leftmost pixel of "C" is rendered, we can't tell that we're within the anti-aliasing edge of "B".

image

Buthrakaur:
We just migrated from Mapbox to Maplibre and discovered degraded performance especially on iOS (React Native app for both platforms) when zooming in/out or quickly panning the map. We have 3 symbol layers with small/medium/large variants of POI markers to let collision use the largest possible icon available for a nice map UX with hundreds of POIs in a city. I discovered the rendering performance goes massively down when I add a text in the style and especially when I set text halo (`textHaloWidth`, `textHaloColor`) so I had to turn off text halo effects to keep at least some acceptable performance so I believe this PR could help even though the Mapbox component seems to have some more performance optimization around text collision detection or text rendering in general. It's surprising the Android performance is quite ok though and it's sluggish on iPhones.

I'd appreciate any recommendations on how to improve the performance by changing the style - I ended up with this:
```jsx
const useTextHalo = Platform.OS === 'android';

const baseStyle: SymbolLayerStyle = {
  textHaloWidth: 2,
  textHaloBlur: 0,
  textAnchor: 'top',
  textMaxWidth: 10,
  textFont: ['get', 'textFont'],
  iconAllowOverlap: false,
  textAllowOverlap: false,
  iconOptional: false,
  iconAnchor: ['get', 'iconAnchor'],
  symbolSortKey: ['get', 'priority'],
  textOffset: [0, 0.1],
};

export const pinLargeStyle: SymbolLayerStyle = {
  ...baseStyle,
  textField: ['get', 'title'],
  textSize: SIZE_L,
  textColor: ['get', 'textColor'],
  textHaloColor: ['get', 'textHaloColor'],
  iconImage: ['get', 'large'],
  iconSize: ICON_SIZE_MULTIPLIER,
};

export const pinMediumStyle: SymbolLayerStyle = {
  ...baseStyle,
  textField: ['get', 'title'],
  iconImage: ['get', 'medium'],
  textSize: ['case', ['==', ['get', 'type'], 'tour'], SIZE_L, SIZE_M],
  textColor: ['get', 'textColor'],
  iconColor: ['get', 'iconColor'],
  textHaloColor: ['get', 'textHaloColor'],
  iconSize: ICON_SIZE_MULTIPLIER,
};

export const pinSmallStyle: SymbolLayerStyle = {
  iconImage: ['get', 'small'],
  iconColor: ['get', 'iconColor'],
  symbolSortKey: ['get', 'priority'],
  iconAnchor: 'center',
  iconSize: ICON_SIZE_MULTIPLIER,
};

if (!useTextHalo) {
  [pinLargeStyle, pinMediumStyle, pinSmallStyle].forEach((style) => {
    delete style.textHaloWidth;
    delete style.textHaloBlur;
    delete style.textHaloColor;
    delete style.textColor;
  });
}


...

  return (
    <>
      <Images images={images} />
      <ShapeSource id={props.layerId} shape={featuresCollection} onPress={onFeaturePress} hitbox={useMemo(() => ({ width: 0, height: 0 }), [])}>
        <SymbolLayer
          id={`SymbolLayer${props.layerId}Small`}
          minZoomLevel={props.minZoomLevel}
          maxZoomLevel={props.maxZoomLevel}
          style={pinSmallStyle}
        />
        <SymbolLayer
          id={`SymbolLayer${props.layerId}Medium`}
          minZoomLevel={props.minZoomLevel}
          maxZoomLevel={props.maxZoomLevel}
          style={pinMediumStyle}
        />
        <SymbolLayer
          id={`SymbolLayer${props.layerId}Large`}
          minZoomLevel={props.minZoomLevel}
          maxZoomLevel={props.maxZoomLevel}
          style={pinLargeStyle}
        />
      </ShapeSource>
    </>
  );
```

This is the intended result which worked well with Mapbox in terms of performance but is not acceptable with MapLibre on iOS:

image

PR #2385, opened 2024-05-11
Format Swift files with SwiftFormat

louwers:
Add a formatter for Swift.


HarelM (approved):
I assume this was done automatically...

louwers:
Had to use Swift 5.8 instead of 5.11 because CI is still on that version.

birkskyum:
> Had to use Swift 5.8 instead of 5.11 because CI is still on that version.

.. and because there is no Swift 5.11 ;)

PR #2388, opened 2024-05-12
Compile iOS Examples

louwers:
The code samples in the documentation should be compiled on CI so they do not go out of date.

I wrote a script `scripts/update-ios-examples.mjs` that can update documentation based on examples in the codebase.

Examples need to be fenced with
```swift
// #-example-code(LineTapMapView)
...
// #-end-example-code
```

Documentation code blocks that are prefixed with

````md
<!-- include-example(LineTapMapView) -->

```swift
...
```
````

are pulled from actual code.

- Add sample (Swift) app and compile it on CI
- Update `CONTRIBUTING.md` with info how to add examples
- Add script for updating examples
- Run script (so current examples are updated)

### DIFF
--- a/scripts/update-ios-examples.mjs
+++ b/scripts/update-ios-examples.mjs
@@ -1,18 +1,31 @@
+/**
+ * This script updates the iOS examples part of the DocC documentation.
+ * See platform/ios/CONTRIBUTING.md for the details.
+ */
+
 import fs from "fs/promises";
 
+/** documentation files that will be updated */
 const documentationPaths = [
   "platform/ios/MapLibre.docc/GettingStarted.md",
   "platform/ios/MapLibre.docc/LineOnUserTap.md",
 ];
 
-// source files that contain examples
+/** source files that contain examples */
 const examplePaths = [
   "platform/ios/app-swift/Sources/SimpleMapView.swift",
   "platform/ios/app-swift/Sources/LineTapMapView.swift"
 ];
 
 /**
+ * Parses a source file and looks for examples fenced in
+ * ```
+ * // #-example-code(ExampleId)
+ * ...
+ * // #-end-example-code
+ * ```
  * @param {string} examplePath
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: scripts/update-ios-examples.mjs [hunk 1]
```diff
@@ -0,0 +1,136 @@
...
+ * ...
+ * // #-end-example-code
+ * ```
+ * @param {string} examplePath
+ * @returns Mapping from example id to example, e.g. {"ExampleId": "..."}
+ */
+async function parseExampleCode(examplePath) {
+  const content = await fs.readFile(examplePath, "utf8");
+  const lines = content.split("\n");
+  /** @type {Record<string, string>} */
```
HarelM, line 33: Why not write it in typescript instead of this comment?

HarelM:
Are there any tips that are doing it out of the box? 
Are there standards to adding examples inside comments like tsdoc has? 
It feels weird to define a "standard"...

louwers:
> Are there any tools that are doing it out of the box?

I could not find anything. DocC does not support it at least.

scripts/update-ios-examples.mjs
louwers: I don't want to complicate our build setup further by requiring TypeScript compilation.

louwers: Not sure if a lot of people are familiar with using TypeScript with JSDoc, but plain JavaScript remains an option. We're not building a library with JavaScript/TypeScript, just some auxiliary scripts.

HarelM: Since it's a single script you can run it using ts-node without something special, but your call, I added this remark because of the comment here. 
Your call.

HarelM (approved):

louwers:
@HarelM I'll put using TypeScript directly on the agenda for the next TSC Meeting.

PR #2389, opened 2024-05-13
User Location and Location Privacy Example iOS

louwers:
- Add Navigation to the Swift example app.
- Add new example for requesting (precise) user location.

@ianthetechie I have no idea what I am doing with SwiftUI, maybe you can have a look at it sometime. But I am happy I got it to work...

image

### DIFF
--- a/platform/ios/MapLibre.docc/LocationPrivacyExample.md
+++ b/platform/ios/MapLibre.docc/LocationPrivacyExample.md
@@ -74,7 +74,6 @@ class Coordinator: NSObject, MLNMapViewDelegate {
             print("User location is currently not available.")
             return
         }
-        mapView.setCenter(userLocation.coordinate, zoomLevel: 8, animated: true)
         mapView.fly(to: MLNMapCamera(lookingAtCenter: userLocation.coordinate, altitude: 100_000, pitch: 0, heading: 0))
     }
 }
--- a/platform/ios/MapLibre.docc/LocationPrivacyExample.md
+++ b/platform/ios/MapLibre.docc/LocationPrivacyExample.md
@@ -1,0 +1,133 @@
+# User Location & Location Privacy
+
+Requesting precise location with ``MLNLocationManager``.
+
+## Overview
+
+This example shows how to request a precise location with ``MLNLocationManager``.
+
+First of all you need to prepare your `Info.plist`. You need to provide a description why your app needs to access location:
+
+```plist
+<key>NSLocationWhenInUseUsageDescription</key>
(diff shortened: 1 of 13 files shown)


platform/ios/MapLibre.docc/LocationPrivacyExample.md
```diff
@@ -0,0 +1,132 @@
...
+![](ImpreciseLocation.png)
+
+When the button is pressed a pop-up with the description we set in `Info.plist` will be shown:
+
+![](PreciseLocationRequestPopup.png)
+
+<!-- include-example(LocationPrivacyExample) -->
+
+```swift
+enum LocationAccuracyState {
```
HarelM, line 37: Consider moving this example to the relevant class definition, this way you'd get it checked as well, and IDEs might be able to show it, hopefully. 
Although I'm not familiar enough with xcode, I hope some of the comments here are shown as oart of a tooltip when hovering over a class or a method.

scripts/update-ios-examples.mjs
```diff
@@ -9,12 +9,14 @@ import fs from "fs/promises";
 const documentationPaths = [
   "platform/ios/MapLibre.docc/GettingStarted.md",
   "platform/ios/MapLibre.docc/LineOnUserTap.md",
+  "platform/ios/MapLibre.docc/LocationPrivacyExample.md",
```
HarelM, line 12: Can't you use glob or something that doesn't require adding a page every time?

scripts/update-ios-examples.mjs
```diff
@@ -9,12 +9,14 @@ import fs from "fs/promises";
...
 ];
 
 /** source files that contain examples */
 const examplePaths = [
   "platform/ios/app-swift/Sources/SimpleMapView.swift",
-  "platform/ios/app-swift/Sources/LineTapMapView.swift"
+  "platform/ios/app-swift/Sources/LineTapMapView.swift",
```
HarelM, line 18: Same as above.
In theory, this code doesn't need to change when you change the content of the documentation.

platform/ios/MapLibre.docc/LocationPrivacyExample.md
louwers: This is compiled. The script copies the code section into the Markdown file.

scripts/update-ios-examples.mjs
louwers: My idea was not parsing files unnecessarily, since some of them will not contain examples or do not need examples replaced, but since there are not that many files it's fine to go through them all.

HarelM (approved):

PR #2393, opened 2024-05-13
Improvements & Fixes: User Location and Location Privacy Example iOS

voncannon:
A continuation of PR https://github.com/maplibre/maplibre-native/pull/2389

1. User location and privacy example uses now uses a view model with published properties.
2. No strong references MLNMapView with @State.
3. Added documentation around updateUIView, dismatleUIView
4. Renamed LocationPrivacyExample to MapLibreViewRepresentable so it's a tad more clear that its primary purpose is just a wrapper around UIKit for SwiftUI representation.
5. Fixed a bug where pannedToUserLocation was never set.


voncannon:
@louwers Made a few updates

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
-    @Binding var mapView: MLNMapView
-    @Binding var locationAccuracy: LocationAccuracyState
-    var pannedToUserLocation = false
-
-    init(mapView: Binding<MLNMapView>, locationAccuracy: Binding<LocationAccuracyState>) {
-        _mapView = mapView
-        _locationAccuracy = locationAccuracy
+class MapViewModel: NSObject, ObservableObject {
+    // Weak because the coordinator has a strong reference to the view model
+    weak var mapCoordinater: MapLibreRepresentableCoordinator?
```
hactar, line 14: typo: coordinator.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
+        // Set delegate first, otherwise delegate callbacks can be missed. For example, setting showsUserLocation calls mapView:didUpdate:userLocation
         mapView.delegate = context.coordinator
-
+        mapView.showsUserLocation = true
         return mapView
     }
 
-    func updateUIView(_: MLNMapView, context _: Context) {}
+    func updateUIView(_: MLNMapView, context _: Context) {
+        // Be careful with doing any real work in updateUIView, this is called a lot.
```
hactar, line 101: https://developer.apple.com/documentation/swiftui/uiviewrepresentable/updateuiview(_:context:) states that all UI updates are to be made here. we are provided with the view and the context here. By performing updates here no own reference to the mapview has to be held.

Yes it is called, but "a lot" is relative. Its not called 120 times a second, its called when SwiftUI has been informed that something regarding the state of this view has been changed, so its the time to see if one needs to update this view.

I don't think it's wise to discourage an Apple best practice in the MapLibre docs.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
-
-    init(mapView: Binding<MLNMapView>, locationAccuracy: Binding<LocationAccuracyState>) {
-        _mapView = mapView
-        _locationAccuracy = locationAccuracy
+class MapViewModel: NSObject, ObservableObject {
+    // Weak because the coordinator has a strong reference to the view model
+    weak var mapCoordinater: MapLibreRepresentableCoordinator?
+    @MainActor @Published var locationAccuracy: LocationAccuracyState = .unknown
+
+    @MainActor func requestTemporaryLocationAuthorization() {
```
hactar, line 17: Instead of an function here, which is event driven, why not have a @State for the mapview that users can change, called showTemporaryLocationAuthorization? When this state changes, updateUIView would be called, including a reference to MLNMapview.

louwers (changes requested):
Thanks @voncannon 

Can you perhaps nest the classes or give them less generic names? Because I want to include a lot of examples in the same namespace.

Can you re-run this script mentioned here https://github.com/maplibre/maplibre-native/blob/main/platform/ios/CONTRIBUTING.md#examples
```
node scripts/update-ios-examples.mjs
```

Then the examples in the documentation will be updated.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
+        MapLibreRepresentableCoordinator(mapViewModel: mapViewModel)
     }
 
     func makeUIView(context: Context) -> MLNMapView {
         let mapView = MLNMapView()
-        mapView.showsUserLocation = true
+        // Set delegate first, otherwise delegate callbacks can be missed. For example, setting showsUserLocation calls mapView:didUpdate:userLocation
         mapView.delegate = context.coordinator
-
+        mapView.showsUserLocation = true
```
hactar, line 96: As far as I remember, setting showUserLocation to true immediately triggers a permission request?

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
-    @Binding var locationAccuracy: LocationAccuracyState
-    var pannedToUserLocation = false
-
-    init(mapView: Binding<MLNMapView>, locationAccuracy: Binding<LocationAccuracyState>) {
-        _mapView = mapView
-        _locationAccuracy = locationAccuracy
+class MapViewModel: NSObject, ObservableObject {
+    // Weak because the coordinator has a strong reference to the view model
+    weak var mapCoordinater: MapLibreRepresentableCoordinator?
+    @MainActor @Published var locationAccuracy: LocationAccuracyState = .unknown
```
louwers, line 15: Should the whole class be marked with `@MainActor`?

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
+        guard !pannedToUserLocation else { return }
         guard let userLocation = mapView.userLocation else {
             print("User location is currently not available.")
             return
         }
         mapView.fly(to: MLNMapCamera(lookingAtCenter: userLocation.coordinate, altitude: 100_000, pitch: 0, heading: 0))
+        pannedToUserLocation = true
+    }
+
+    func mapView(_ mapView: MLNMapView, didFinishLoading _: MLNStyle) {
```
louwers, line 68: Are these two methods strictly required?

I'd like to keep the example as focused as possible.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
+    weak var mapCoordinater: MapLibreRepresentableCoordinator?
+    @MainActor @Published var locationAccuracy: LocationAccuracyState = .unknown
+
+    @MainActor func requestTemporaryLocationAuthorization() {
+        print("Requesting precice location")
+
+        switch locationAccuracy {
+        case .reducedAccuracy:
+            let purposeKey = "MLNAccuracyAuthorizationDescription"
+            mapCoordinater?.mapView?.locationManager.requestTemporaryFullAccuracyAuthorization!(withPurposeKey: purposeKey)
```
hactar, line 23: Force unwrap is usually a bad idea. Also, should we really be using the locationManager within mapView and not one that the app initializes for its own purposes?

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
+            break
+        }
     }
+}
 
-    func mapView(_: MLNMapView, didChangeLocationManagerAuthorization manager: MLNLocationManager) {
+class MapLibreRepresentableCoordinator: NSObject, MLNMapViewDelegate {
+    private var mapViewModel: MapViewModel
+    // Weak reference because SwiftUI owns the strong reference
+    private(set) weak var mapView: MLNMapView?
```
hactar, line 33: Apple discourages keeping references to the UIView - the idea behind it is that the UIViewRepresentable should be written in a way so that iOS can discard the UIView at any time, and recreate it back to the state it needs to be in with calls to makeUIView and updateUIView at any time. So the UIView one might be keeping a reference to might no longer be the one that is actually on screen.

Sometimes it is impossible to adhere to this, and then one has to live with the risk (it rarely actually happens that iOS creates the UIView at "random"), but I don' think we need this workaround in this specific case? See my  `showTemporaryLocationAuthorization` State suggestion.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
-        mapView.showsUserLocation = true
+        // Set delegate first, otherwise delegate callbacks can be missed. For example, setting showsUserLocation calls mapView:didUpdate:userLocation
         mapView.delegate = context.coordinator
-
+        mapView.showsUserLocation = true
         return mapView
     }
 
-    func updateUIView(_: MLNMapView, context _: Context) {}
+    func updateUIView(_: MLNMapView, context _: Context) {
```
louwers, line 100: Since these two methods are empty, let's leave them out of this example.

You could create a new guide that explains these methods.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
```diff
@@ -9,75 +9,114 @@ enum LocationAccuracyState {
...
+        }
+    }
+
+    func mapViewDidFinishLoadingMap(_ mapView: MLNMapView) {
+        if self.mapView != mapView {
+            self.mapView = mapView
+        }
+    }
+
+    deinit {
```
louwers, line 80: Can be removed since it is empty?

voncannon:
Thanks for the comments, working on getting all this updated now 👍

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: In this case, yes, done.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: Nope, removing, done.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: Great catch 😄 . Fixed (removed)

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: I agree for this PR and this example that is out of scope. Updated the PR to follow Apple's doc. 

I need to write a blog post about all this stuff I've found out with `UIViewRepresentable` and put my options there.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: 💯 Agreed, updated.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: Yes that is correct. This is something that was already here with @louwers example so I left it alone. I think the example was to just show requesting temporary access. But if we wanted to expand this example we could add two buttons on this example. But I'll let @louwers make that call.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: Great catch, fixed.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: I need to write a blog post about all this stuff I've found out with UIViewRepresentable and put my options there. For this example I have followed Apple's doc for now :) Appreciate the feedback.

voncannon:
@louwers 

- [x] Added a prefix on these examples of "PrivacyExample" so the classes/structs don't clash.
- [x] Updated the examples: `node scripts/update-ios-examples.mjs`

@hactar Appreciate all the feedback. I didn't mark the conversations above as resolved so you can review them. Let me know if you see anything else.

🚀

platform/ios/MapLibre.docc/LocationPrivacyExample.md
```diff
@@ -40,75 +40,81 @@ enum LocationAccuracyState {
...
-
+        mapView.showsUserLocation = true
         return mapView
     }
 
-    func updateUIView(_: MLNMapView, context _: Context) {}
+    func updateUIView(_ mapView: MLNMapView, context _: Context) {
+        if mapViewModel.showTemporaryLocationAuthorization {
+            let purposeKey = "MLNAccuracyAuthorizationDescription"
+            mapView.locationManager.requestTemporaryFullAccuracyAuthorization?(withPurposeKey: purposeKey)
```
hactar, line 102: Nothing sets showTemporaryLocationAuthorization to false again, so any update will call that request function again once set to true. According to https://developer.apple.com/documentation/corelocation/cllocationmanager/3600217-requesttemporaryfullaccuracyauth there is a completion function, I think we should use that and set it back to false?

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
hactar: If you've tested this and it doesn't clash with that temporary request then fine, I was just worried that this would trigger an auth request on app launch, making your button obsolete...

hactar (changes requested):
Nice changes. I think I found a little issue regarding a missing completion block, please check ☺️

platform/ios/MapLibre.docc/LocationPrivacyExample.md
voncannon: That is part of the problem I have with `UIViewRepresentable` 

We have a published prop `showTemporaryLocationAuthorization` that will trigger a `updateUIView` update, but then inside `updateUIView` if we yet set `showTemporaryLocationAuthorization` back to false, it technically triggers another `updateUIView`. Now in this case, we have a bool we can check, but still. It just feels "meh" to me.

So let's take this as an example of setting `showTemporaryLocationAuthorization` back to false:

```
func updateUIView(_ mapView: MLNMapView, context _: Context) {
        if mapViewModel.showTemporaryLocationAuthorization {
            let purposeKey = "MLNAccuracyAuthorizationDescription"
            mapView.locationManager.requestTemporaryFullAccuracyAuthorization?(withPurposeKey: purposeKey)
            mapViewModel.showTemporaryLocationAuthorization = false
        }
    }
```
When try that, we even get a warning about it:

tears

I have read so people dealing and fighting with `UIViewRepresentable` like this, examples:

1. https://stackoverflow.com/a/76919202
2. https://www.reddit.com/r/SwiftUI/comments/kti9r9/uiviewrepresentable_how_to_update_bindings/
3. there are millions more I won't link to

So to fix that warning people are wrapping it in a dispatch to main, which I don't like personally:

```
    func updateUIView(_ mapView: MLNMapView, context _: Context) {
        if mapViewModel.showTemporaryLocationAuthorization {
            let purposeKey = "MLNAccuracyAuthorizationDescription"
            mapView.locationManager.requestTemporaryFullAccuracyAuthorization?(withPurposeKey: purposeKey)
            DispatchQueue.main.async {
                mapViewModel.showTemporaryLocationAuthorization = false
            }
        }
    }
```


And also, to your comment about `CLLocationManager` having a completion callback, this example uses `MLNLocationManager` which wraps `CLLocatonManager`, and maplbire MLNLocationManager's version of  `requestTemporaryFullAccuracyAuthorization` hasn't been updated to support the completion callback from what I can tell... unless I'm just missing it for some reason... can you double check please?

Also, this brings up one of your points in  our [Slack discussion](https://osmus.slack.com/archives/C06U5MM097B/p1715614940371019?thread_ts=1715612553.323529&cid=C06U5MM097B) about most apps they have their own Location Manager functionality, but I think @louwers intention was to keep these examples small and to the point and adding yet another class / code to handle location manager increases the complexity of the example.

platform/ios/app-swift/Sources/LocationPrivacyExample.swift
voncannon: I have tested it, but I see your point. I will double check and I think this will come down to what is in the host apps Info.plist. If the host app by default has `nslocationdefaultaccuracyreduced` set.

https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationdefaultaccuracyreduced

This example was originally created by @louwers , but maybe we should just change the example from requesting temporary precise location access we instead change this example to requesting and showing their location on the map

Let me know @louwers

platform/ios/MapLibre.docc/LocationPrivacyExample.md
voncannon: But to your comment regarding `I think we should use that and set it back to false?`, I agree 💯  ... but what am I missing with `UIViewRepresentable`?

This is why In my app I do not use `updateUIView`, but instead keep a weak reference to the underlying view. I know that seems worse, but it actually makes it so much cleaner.

For example, [look at this where people are using MKMapView](https://stackoverflow.com/a/73041182) and their implementation removes/adds annotations each update!

I mean that will work, but that seems highly unnecessary. 

And in most large apps we are going to have so many computed props and bindings that need to interface with `MLNMapView`, imagine how much unnecessary work will be done.

Thoughts? Suggestions? 

😄

voncannon: More links:

1. https://stackoverflow.com/questions/58460847/calling-a-local-uiviewcontroller-function-from-a-swiftui-view#comment113325151_61099536
2. https://stackoverflow.com/questions/58039286/how-do-you-call-a-method-on-a-uiview-from-outside-the-uiviewrepresentable-in-swi
3. https://stackoverflow.com/questions/56584059/send-tapaction-from-swiftui-button-action-to-uiview-function

hactar: This "inefficient" recreation is completely normal in SwiftUI. At the end of the day, it doesn't matter if update is called a few times - per runloop iteration its grouped together, so will only be called once - and as long as the recreation takes microseconds, it will have no impact on users. I know at every developers heart is efficiency - it took me quite a while to get used to this new paradigm too, but SwiftUI takes the "premature optimization is the root of all evil" approach: don't worry about it until you actually have a perceivable impact on performance, and if you ever do, then you can start optimizing - in our case this will have no perceivable impact.

But you are right, we need to remove that "Publishing changes from within view updates is not allowed, this will cause undefined behavior." issue. I assume that if we did have access to the "completion" version of requestTemporaryFullAccuracyAuthorization and call showTemporaryLocationAuthorization = false there, this error would not appear. So we either need to expose that function or, if we're looking for a quick and dirty solution, then yes we need to do the workaround with DispatchQueue.main.async.

As mentioned, I use my own location manager instead of maplibre's - a modified version of https://github.com/malcommac/SwiftLocation, this allows me to create this code for permission requests:

```Swift
 MapView(styleURL: styleURL, camera: $camera) // this is based on maplibre-swiftui-dsl-playground
            .task {
                for await event in await locationManager.startMonitoringAuthorization() {
                    print("Authorization status did change: \(event.authorizationStatus)")
                    // break to interrupt the stream
                    if event.authorizationStatus == .authorizedWhenInUse {
                        showUserLocation = true
                    } else {
                        showUserLocation = false
                    }
                }
            }
            .task {
                showUserLocation = locationManager.authorizationStatus == .authorizedWhenInUse
            }
```

And this is then a button to get the current location:

```Swift
        Button {
            Task {
                do {
                    location.accuracy = .threeKilometers
                    try await location.requestPermission(.whenInUse) // obtain the permissions
                    let userLocation = try await location.requestLocation()
                    
                    if let coordinates = userLocation.location?.coordinate {
                        withAnimation {
                            camera = MapViewCamera.center(coordinates, zoom: 17)
                            
                        }
                    } else {
                        print("location error: got no coordinates")
                    }
                } catch {
                    print("location error: \(error)")
                }
            }
        } label: {
            Image(systemName: "location")
        }
```
The code is clean, uses async/await, and changes "showUserLocation" based on changes to permissions - with permissions its on, without permissions its off and does not trigger a permission request.

I'm not suggesting we use https://github.com/malcommac/SwiftLocation in the official maplibre examples, but if we have an example, we need to decide if its supposed to be a "best practice" example or a "quick and dirty get it up and running" example. Personally I'd argue for "best practice, even if more complicated" because people will copy and paste our examples into their code and location management is not an easy task. But I do understand that complicated examples could scare off devs too. 🤷‍♂️

louwers: > Personally I'd argue for "best practice, even if more complicated" because people will copy and paste our examples into their code and location management is not an easy task.

Fully agreed. Even an example with an external dependency would be fine by me. Let's not forget the iOS MapLibre codebase is quite rusty, so it should not be surprising some things are not very modern anymore.

I'm just transplanting some examples from https://github.com/mapbox/ios-sdk-examples, I'm not a veteran iOS developer like you, so I am happy for your critical eye!

voncannon:
@louwers @hactar I believe this PR's current state / goals have been met which removes the UIView from @State and follows some of Apple's best practices. 

As far as the pattern we are going to demonstrate in `updateUIView`, that is going to apply to all of our SwiftUI examples and maybe we should breakout a discussion in slack on our next steps on that.

Also, I have tested the example with the last commit where I added `NSLocationDefaultAccuracyReduced` as well and this example feels much better now.

Let me know if there is any other items for this PR.

hactar (approved):
I'd agree that this example is an improvement to what we previously had. A further iteration to fulfill all best practices would be great and maybe I'll have some time for this during the next weeks, but for the time being this would already help devs be closer to what they need to be.

voncannon:
> I'd agree that this example is an improvement to what we previously had. A further iteration to fulfill all best practices would be great and maybe I'll have some time for this during the next weeks, but for the time being this would already help devs be closer to what they need to be.

Agreed. 😄  I still need to wrap my head around `UIViewRepresentable` best practices. Also, we are only a few weeks from wwdc so maybe there are some incremental improvements or best practices coming down the pike as well from Apple.

louwers (approved):
Thanks @hactar and @voncannon for the improvements!

PR #2401, opened 2024-05-19
[qt] Use GCC 13 on Ubuntu 24.04 to fix the CI

ntadej:
`gcc-13` somehow disappeared from the build notes/repos. Try to use `gcc-13` on Ubuntu 24.04.


louwers (approved):

ntadej:
I moved back to GCC 13 as I can not reproduce the warning myself.

PR #2419, opened 2024-05-22
Update SQLite, Android dependencies

louwers:
Update Android dependencies except `io.mockk.mockk`.

Reported a problem with it [here](https://github.com/mockk/mockk/discussions/1256). Might be our fault. 

After this is merged people should update to Android Studio Jellyfish.

## SQLite Update

This pull request also updates SQLite, because this warning now became an error:

>   /home/runner/work/maplibre-native/maplibre-native/vendor/sqlite/src/sqlite3.c:38937:8: error: incompatible integer to pointer conversion assigning to 'char *' from 'int' [-Wint-conversion]

This was [patched](https://sqlite.org/forum/forumpost/67c604d7c65a2f5b) but this patch was not included in a release yet.

**Note for reviewer**: updating `sqlite.c` is a bit tricky, because it's impossible to review the changes. Something malicious could hide in there.

You can run

```
scripts/vendor/sqlite.sh
```

to verify that the working directory stays clean, i.e. that this is just the released version of SQLite plus the patch applied.


ntadej:
Looks OK to me, but I do not use Android studio so I do not want to give the defining approval 😊

cgalvan (approved):

PR #2420, opened 2024-05-22
Speculative crash detection and prevention

TimSylvester:
External call stacks indicate several crashes that seem like they must be caused by members from the pointer portion of `key` to `placedVerticalIconIndex` becoming unexpectedly invalid.

We insert fixed values between each member and check them on each access in hopes of narrowing down the range of potential places from which this could be caused and skip further processing of the symbol instance once any invalid value has been detected.

Also includes the benchmarking activity from #1850 to confirm that there's no major effect on performance.

```
10.3.1

Result(average=1137.1406652094868, low1p=206.20395674648492),
Result(average=972.6569705258048, low1p=195.68381800358537),
Result(average=1006.3422475857232, low1p=196.21148643908523),
Result(average=979.1130397712535, low1p=195.33316971831644),
Result(average=999.3013716059238, low1p=192.15926101090554),
Result(average=967.7307632612294, low1p=191.03237983679773)]})

10.3.1.1

Result(average=1102.1602248079416, low1p=203.00198974531042),
Result(average=997.9265047694663, low1p=189.26893108387634),
Result(average=1043.7275178877694, low1p=194.42137487148483),
Result(average=1049.3060329936836, low1p=198.54736418525866),
Result(average=1036.1822841437606, low1p=192.19624842621522),
Result(average=1043.3555232138212, low1p=196.9277046632644)]})
```

`clang-format` modifies almost the entire contents of every file, so I'll do that after people have had a chance to look at the meaningful differences.

### DIFF
--- a/include/mbgl/util/logging.hpp
+++ b/include/mbgl/util/logging.hpp
@@ -40,6 +40,7 @@ public:
     Log();
     ~Log();
 
+    static bool useLogThread();
     static void useLogThread(bool enable);
 
     template <typename ...Args>
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
@@ -1168,6 +1168,20 @@
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value=".activity.FeatureOverviewActivity" />
         </activity>
+        <!-- Benchmark -->
+        <activity
+            android:name=".activity.benchmark.BenchmarkActivity"
+            android:description="@string/description_world_tour_benchmark"
+            android:exported="true"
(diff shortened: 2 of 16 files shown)


louwers (commented):
Shall we release this as `10.3.2-pre0`? Not sure if it makes sense to roll this out to everyone. I think we should verify the fix before cutting an (official) release out of it.

Just update `android/MapLibreAndroid/gradle.properties`, push it to a branch on `maplibre/maplibre-native`, and I can release it.

TimSylvester:
I assume a pre-release will suffice, unless they care what it's called.

I don't think this is really appropriate for a "real" public release, and it can all be discarded once we find the real source of the problem.

If the crash disappears in the field, it will be because some symbols are not displaying correctly, and we'll need to figure out how to bring the problem back in testing to track it down.

PR #2425, opened 2024-05-24
[node] Temporarily Exclude ABI 127

acalcutt:
Node 22 binaries are failing to build with the following error. This temporarily explode it from building until it can be fixed
image


acalcutt:
Closed in favor of https://github.com/maplibre/maplibre-native/pull/2426 , which seems to fix the node 22 error by upgrading nan

PR #2438, opened 2024-05-27
Remove MapTiler dependency Instrumentation Tests

louwers:
Swaps out Maptiler for Versatiles and Demotiles, which do not need an API key.

Finally the Instrumentation tests for Android are green.

image

### DIFF
--- a/src/mbgl/renderer/render_orchestrator.cpp
+++ b/src/mbgl/renderer/render_orchestrator.cpp
@@ -144,7 +144,12 @@ RenderOrchestrator::~RenderOrchestrator() {
     // to the scheduler because it cannot be destroyed from one of its own pool threads.
     constexpr auto deferredCleanupTimeout = Milliseconds{1000};
     [[maybe_unused]] const auto remaining = threadPool->waitForEmpty(deferredCleanupTimeout);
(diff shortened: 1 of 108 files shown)


src/mbgl/renderer/render_orchestrator.cpp
```diff
@@ -144,7 +144,12 @@ RenderOrchestrator::~RenderOrchestrator() {
     // to the scheduler because it cannot be destroyed from one of its own pool threads.
     constexpr auto deferredCleanupTimeout = Milliseconds{1000};
     [[maybe_unused]] const auto remaining = threadPool->waitForEmpty(deferredCleanupTimeout);
+// this assert is causing Android Instrumentation tests to fail
+// since they need to run in on a debug build
+// ignore it for now, see issue https://github.com/maplibre/maplibre-native/issues/2187
```
cgalvan, line 149: Nice to see we have a backlog task to check this out

cgalvan (approved):

src/mbgl/renderer/render_orchestrator.cpp
louwers: Marc already has a PR open.

PR #2452, opened 2024-05-30
Change naming in some platform files

TimSylvester:
Eliminates some additional vestiges of Mapbox.


src/mbgl/renderer/sources/render_tile_source.cpp
```diff
@@ -223,6 +228,17 @@ void TileSourceRenderItem::updateDebugDrawables(DebugLayerGroupMap& debugLayerGr
 
                 // We would need to set up `idLineExpressionUBO` if the expression mask isn't empty
                 assert(linePropertiesUBO.expressionMask == LineExpressionMask::None);
+
+                const LineExpressionUBO exprUBO = {
```
louwers, line 232: Is this unrelated change intentional?

louwers (approved):

src/mbgl/renderer/sources/render_tile_source.cpp
TimSylvester: It is related.  I got a render-test failure because the tile outlines were missing.  When I brought it up in the debugger, I got an Metal error which is fixed by that change.  The render test still failed, however, due to the `render_tile_source` change.

I put some explanation in the commit, but I'll copy it to the PR, I'm not sure how long those commits stay around after a squash.

PR #2472, opened 2024-06-04
Move all `WeakPtrFactory` members to the end of the containing class

TimSylvester:
The `mapbox::base::WeakPtrFactory` class requires that it's the last member of any class in which it's used, but two of those uses violated that requirement.

Also adds comments on each use to make it harder to repeat.


sjg-wdw (approved):
Doesn't seem all that plausible this could be the IndexedSymbol crash.

cgalvan (approved):

PR #2473, opened 2024-06-04
Make `useLogThread` a per-severity option and default `Error` to false

TimSylvester:
This should make any error events more likely to be written out before a crash.  Such events are rare, so there should be no significant effect on performance.

### DIFF
--- a/platform/darwin/src/MLNLoggingConfiguration.mm
+++ b/platform/darwin/src/MLNLoggingConfiguration.mm
@@ -29,6 +29,9 @@ public:
             case EventSeverity::Error:
                 MLNLogError(message);
                 break;
+            default:
+                assert(false);
+                return false;
         }
         return true;
     }
--- a/platform/darwin/src/MLNLoggingConfiguration.mm
+++ b/platform/darwin/src/MLNLoggingConfiguration.mm
@@ -31,7 +31,6 @@ public:
                 break;
             default:
                 assert(false);
-                return false;
         }
         return true;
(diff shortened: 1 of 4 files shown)


sjg-wdw (approved):

platform/darwin/src/MLNLoggingConfiguration.mm
```diff
@@ -29,6 +29,9 @@ bool onRecord(EventSeverity severity, Event event, int64_t code, const std::stri
             case EventSeverity::Error:
                 MLNLogError(message);
                 break;
+            default:
+                assert(false);
+                return false;
```
louwers, line 34: Nit-picking, but I think it's better to fall through here because otherwise the erroneous log message will get logged with `platformRecord`.

louwers (approved):

src/mbgl/util/logging.cpp
```diff
@@ -14,7 +15,8 @@ namespace mbgl {
 namespace {
 
 std::unique_ptr<Log::Observer> currentObserver;
-std::atomic<bool> useThread(true);
+static constexpr auto SeverityCount = underlying_type(EventSeverity::SeverityCount);
```
louwers, line 18:
```suggestion
constexpr auto SeverityCount = underlying_type(EventSeverity::SeverityCount);
```

louwers: Linux CI complaining about this.

Issue #2477, opened 2024-06-06
Label Blinking Issue at Integer Zoom Levels with animateCamera


geolives-contact:
**Describe the bug**
Starting with MapLibre v11, when the map zoom is set to an exact integer value (such as 15.0, 16.0 or 17.0 for instance), some labels blink when `animateCamera` is called.
The issue does not occur when using `moveCamera` or when the zoom is set between integer values (e.g., 16.4 or 17.1).

Here are some videos to demonstrate the issue:
* [First video]( https://drive.google.com/file/d/17WVr15KOfLPlXFBWX2onL0JVjPZUGXZR/view?usp=sharing): Shows the issue with labels for roads blinking during the animation.
* [Second video](https://drive.google.com/file/d/1UcCK5TUXKJh47Ra12Y4xhmaPRIhQAH9K/view?usp=sharing): Shows the labels not blinking when the zoom is set between integer values.

**To Reproduce**

1. Download and use this sample activity file: [SimpleMapActivity.kt.zip](https://github.com/user-attachments/files/15613146/SimpleMapActivity.kt.zip)
2. This activity file contains code to run animateCamera at short intervals with the zoom set exactly to 15.0.
3. If needed, we can create a branch with this edit in our fork.

**Expected behavior**
On the MapLibre v10 version, labels didn't blink when zoom are set at integer values.

**Platform information (please complete the following information):**
 - Platform: Android, iOS not tested
 - Version 11

louwers:
Thanks for the great bug report with reproduction!

That's a really strange bug, but we'll have a look at it.

louwers:
I tried iOS, doesn't seem to be a problem there:


```swift
struct SimpleMap: UIViewRepresentable {

    func makeUIView(context _: Context) -> MLNMapView {
        let mapView = MLNMapView(frame: .zero, styleURL: VERSATILES_COLORFUL_STYLE)
        mapView.zoomLevel = 15.0
        
        mapView.centerCoordinate = CLLocationCoordinate2D(latitude: 50.15, longitude: 5.2)
        
        startTimer(mapView: mapView)
        return mapView
    }

    func updateUIView(_: MLNMapView, context _: Context) {}
    
    private func startTimer(mapView: MLNMapView) {
        Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true) { i in
            // Update coordinates
            var currentCenter = mapView.centerCoordinate
            
            // Update coordinates
            currentCenter.latitude += 0.001
[...]

PR #2479, opened 2024-06-06
Do not interpolate zoom during flyTo when start and end zoom are same

louwers:
When using flyTo zoom should not be interpolated during the transition. This was causing flickering of labels. Fixes https://github.com/maplibre/maplibre-native/issues/2477

https://github.com/maplibre/maplibre-native/assets/649392/e8306a18-2da7-4dac-b546-66ee73bd30b3

### DIFF
--- a/platform/android/CHANGELOG.md
+++ b/platform/android/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ### 🐞 Bug fixes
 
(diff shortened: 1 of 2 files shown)


alexcristici (approved):

louwers:
Oops, this defeats the purpose of the flyTo algorithm. I hope the tests fail otherwise we need to add some test coverage for it.

Will look for another way to fix the problem.

PR #2494, opened 2024-06-11
Use cache for install-qt-action

louwers:
The Qt workflows are sometimes failing due to temporary network failures.

Caching should help reduce this to a minimum.

https://github.com/maplibre/maplibre-native/actions/runs/9469806144/job/26089733188?pr=2398

https://github.com/maplibre/maplibre-native/actions/runs/9355654961/job/25751324953


louwers:
I think we should cache these files somewhere else, the GitHub cache (10GB) is too small for them...

ntadej:
We should hardcode a more reliable mirror I guess.

louwers:
@ntadej Good idea. I can't find a list of mirrors.

PR #2500, opened 2024-06-12
Fix GL Fence signaling status

alasram:
The OpenGL renderer function `Fence::isSignaled` always returns true (signaled) and this bug can overrun the uniform buffer allocator. `glClientWaitSync` is interpreted as if it returns true (signaled) or false (unsignaled) but the function actually returns the following 4 enums:
```
GL_ALREADY_SIGNALED 0x911A
GL_TIMEOUT_EXPIRED 0x911B
GL_CONDITION_SATISFIED 0x911C
GL_WAIT_FAILED 0x911D
```
This PR ensures the returned value is correctly checked

### DIFF
--- a/src/mbgl/gl/fence.hpp
+++ b/src/mbgl/gl/fence.hpp
@@ -12,7 +12,7 @@ public:
     ~Fence();
 
     void insert() noexcept;
-    bool isSignaled() const noexcept;
+    bool isSignaled() const;
 
 private:
     platform::GLsync fence{nullptr};
--- a/include/mbgl/gl/buffer_allocator.hpp
+++ b/include/mbgl/gl/buffer_allocator.hpp
@@ -54,7 +54,7 @@ public:
     virtual void release(BufferRef* ref) noexcept = 0;
 
     /// Defragment the allocator's underlying buffer pool.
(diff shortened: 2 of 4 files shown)


louwers (changes requested):
We should keep the `noexcept` or modify functions that call `isSignaled` (and the functions that call those functions) not to be `noexcept` . Otherwise the program will crash.

src/mbgl/gl/fence.hpp
```diff
@@ -12,7 +12,7 @@ class Fence {
     ~Fence();
 
     void insert() noexcept;
-    bool isSignaled() const noexcept;
+    bool isSignaled() const;
```
alasram, line 15: @louwers the caller of this function is not `noexcept`

louwers: If I am not mistaken `updateInFlight` calls it which is used in 

```
void defragment(const std::shared_ptr<gl::Fence>& fence) noexcept override
```

louwers: I think `defragment` is erroneously marked `noexcept`.

louwers (approved):
Looks good to me, can be merged if one of the graphics engineers have a look as well.

mwilsnd (approved):

Issue #2498, opened 2024-06-12
Missing `com.mapbox` replacement on annotation layer ID


Fabi755:
Is there a reason why the [MapLibreConstants.LAYER_ID_ANNOTATIONS](https://github.com/maplibre/maplibre-native/blob/e755646eae8e248756db8a8a207635f03fe043ca/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/constants/MapLibreConstants.java#L200) is always `com.mapbox.annotation.points` and not migrated to `org.maplibre.annotations.points`.

Is see that the native code has already migrated to MapLibre naming. For example the [AnnotationManager](https://github.com/maplibre/maplibre-native/blob/e755646eae8e248756db8a8a207635f03fe043ca/src/mbgl/annotation/annotation_manager.cpp#L36)

By [migrating the navigation SDK to v11](https://github.com/maplibre/maplibre-navigation-android/pull/110), I have some issue with the layer order, and I'm not sure if this is the cause of this.

Fabi755:
I see there are still a lot of other constant values that contains the *mapbox* naming. Is this not converted yet in the full Java code?

louwers:
Looks like it could be problematic. Thanks for reporting.

PR #2505, opened 2024-06-13
Rename layer id annotations

louwers:
I think @TimSylvester missed this one in https://github.com/maplibre/maplibre-native/pull/2452

This PR also renames save instance state keys while we're at it.

Closes https://github.com/maplibre/maplibre-native/issues/2498 thanks for the report @Fabi755, appreciate it.


HarelM (approved):
Is this a breaking change?

louwers:
@HarelM No it's just an opaque id meant not to conflict. But we only changed it in the C++ not in Java, so that would introduce a bug.

Issue #1318, opened 2023-07-06
Tiles outside of metadata bounds not loading


nnhubbard:
**Describe the bug**
When using an offline .mbtiles file in iOS, zooming out to world view and then zooming into areas that are not within the mbtiles bounds, some tiles are not shown. The tiles will flash but not load. 

**To Reproduce**
1. View a .mbtiles file
2. Zoom around and away from the target map area

**Expected behavior**
Entire world should be shown in low resolution, even when zooming in. Tiles in those locations should not disappear.

**Actual behavior**
Tiles dissapear or flash in locations outside of the mbtiles target area. This didn't used to happen.

**Screenshots**
**What is happening:**
![54715007-9a7d3400-4b0f-11e9-80df-c3efc5669d23](https://github.com/maplibre/maplibre-native/assets/30858/83bb1d8f-1af9-4d32-8965-d782da19521c)

**What it should look like:**
![54714973-89ccbe00-4b0f-11e9-816a-4269d0891cf1](https://github.com/maplibre/maplibre-native/assets/30858/3aa67573-cda2-42db-9962-a3eee67c03af)

**Platform information (please complete the following information):**
 - OS: iOS
 - Platform: iOS
 - Version: [ios-v5.13.0](https://github.com/maplibre/maplibre-native/releases/tag/ios-v5.13.0)

**Additional context**
This issue was fixed in the MapTiler fork, but that is private so I am not sure I can sure the code here.

nnhubbard:
@louwers can this be looked into at some point?

nnhubbard:
@roblabs Have you ever had this issue?

acalcutt:
Is this really an issue with maplibre? it seems normal to me when you trim a mbtiles file to certain bounds. Some software seems like they show the full tile that contain your bounds, as you lower the zoom the tiles get bigger so it shows more as you zoom out.

Take for example this swiss terrain rgb I made recently. When you are zoomed out the is tile covers a much bigger area, but as you zoom in that area gets smaller because the tiles get smaller
https://tiles.wifidb.net/data/swissalti3d_terrarium_0-17/#1/41/8

acalcutt:
You can also see what I am talking about in this OMT demo that I had created. 
https://demotiles.maplibre.org/tiles-omt/#14/47.26475/11.40416 . When i made the the demo omt tiles there I bound them to the area around innbruck, but you can see as you zoom out the tiles get bigger and more area is covered.

louwers:
[...]

PR #2520, opened 2024-06-20
Tiles not loading outside of metadata bounds

nnhubbard:
Fixes issue #1318 where tiles outside of metadata bounds are not loading. This fix allows over zooming so that map features that are not on higher zoom levels will still show from their lower zoom level tiles.

For example, the offline mbtiles file might include world tiles in `z1` only, but when you zoom in farther those tiles are blank unless you move to the metadata bounds. With this fix, if you zoom into an area outside of the bounds you will get over zoom from `z1` which makes for a much nicer looking map without blank tiles.

### DIFF
--- a/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
+++ b/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
@@ -255,6 +255,10 @@ private:
 
     // Multiple databases open simultaneoulsy, to effectively support multiple .mbtiles maps
     mapbox::sqlite::Database &get_db(const std::string &path) {
+        
+        // Close the DB path before using to prevent a crash that happens when you update the mbtiles file on the filesystem
+        close_db(path);
+        
         auto ptr = db_cache.find(path);
         if (ptr != db_cache.end()) {
             return ptr->second;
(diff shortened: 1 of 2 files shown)


platform/default/src/mbgl/storage/mbtiles_file_source.cpp
```diff
@@ -209,6 +211,7 @@ class MBTilesFileSource::Impl {
                 response.noContent = false;
                 response.expires = Timestamp::max();
                 response.etag = resource.url;
+                response.error.release();
```
louwers, line 214: The memory pointed to by the unique_ptr is not cleaned up after `.release()`.

louwers (changes requested):
Glad you found the problem. Please see my comment.

Any chance you can add a test?

nnhubbard:
I need to rework this pull request. Closing for now.

platform/default/src/mbgl/storage/mbtiles_file_source.cpp
nnhubbard: I guess I don't know how to do that. Any tips?

louwers: I think you can just `reset()` instead.

nnhubbard: Use `response.error.reset();` ?

PR #2521, opened 2024-06-20
Tiles not loading outside of metadata bounds

nnhubbard:
Fixes issue #1318 where tiles outside of metadata bounds are not loading. This fix allows over zooming so that map features that are not on higher zoom levels will still show from their lower zoom level tiles.

For example, the offline mbtiles file might include world tiles in `z1` only, but when you zoom in farther those tiles are blank unless you move to the metadata bounds. With this fix, if you zoom into an area outside of the bounds you will get over zoom from `z1` which makes for a much nicer looking map without blank tiles.

### DIFF
--- a/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
+++ b/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
@@ -201,6 +201,8 @@ public:
 
         Response response;
         response.noContent = true;
+        response.error = std::make_unique<Response::Error>(Response::Error::Reason::Connection,
+                                                           "Not found in mbtile database");
 
         for (mapbox::sqlite::Query q(stmt); q.run();) {
             std::optional<std::string> data = q.get<std::optional<std::string>>(0);
@@ -209,6 +211,7 @@ public:
                 response.noContent = false;
                 response.expires = Timestamp::max();
                 response.etag = resource.url;
+                response.error.reset();
 
                 if (is_compressed(*response.data)) {
                     response.data = std::make_shared<std::string>(util::decompress(*response.data));


wipfli:
Thanks for working on this. The behavior you describe sounds like the correct one, i.e., only show data from inside the bounds.

@louwers you might want to coordinate with @HarelM on how this behavior is done in MapLibre GL JS.

louwers:
@wipfli Does MapLibre GL JS support reading local MBTiles?

louwers (commented):
I don't understand why your changes implements the behavior you say it does. Could you explain it?

wipfli:
MapLibre GL JS does not have offline mbtiles support. My think is that independent of the source - be it an offline mbtiles file or a zxy http endpoint - the behavior for unavailable tiles should be the same.

nnhubbard:
`openmaptiles`, which many people use to create offline `mbtiles` files will create a a world tile at `z1` where you can see all continents, countries, etc. It makes sense to continue to show these as you zoom into different parts of the world using over zoom. Without that, an `mbtiles` file would need to be created with world data on every other zoom level making the file significantly larger.

nnhubbard:
> I don't understand why your changes implements the behavior you say it does. Could you explain it?

This small code change will, when receiving an error for no tile, instead allow the previous zooms tile to show instead of showing blank. E.g. overzoom.

louwers:
This would be a pretty significant change. Not all users will want to have this behavior in all use cases, so I think it would need to be configurable.

Please create a discussion or a design proposal so we can discuss this problem first.

nnhubbard:
> This would be a pretty significant change to the tile loading behavior. Not all users will want to have this behavior in all use cases, so I think it would need to be configurable.
> 
> Please create a discussion or a design proposal so we can discuss this problem first.

That is fine. I have fixed it in my fork for my purposes.

Issue #100, opened 2021-06-25
Updating local mbtiles file causes crash


nnhubbard:
I am using the `mbtiles://` local file feature and I am running into an issue. In my app, I allow downloading of world-wide maps in `mbtiles` format. Updated maps are generated on my server, so users can choose to download an updated map.

The crash occurs when user has a specific map loaded, and then they download an updated map of that loaded location and the file on the file system gets replaced. This causes a crash because the database for the `mbtiles` file is not closed before updating the file.

Crash:

```
com.mapbox.mbgl.MaptilerFileSource (12): EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
terminating with uncaught exception of type mapbox::sqlite::Exception
```

Replicating Crash:

- Add a `mbtiles` file to your project such as from openmaptiles.org.
- Add a `style.json` file to your project such as https://github.com/openmaptiles/maptiler-basic-gl-style
- Load the styles and write the `mbtiles` path into the style:

```
- (NSURL *)writeMbtilesToCreatePath:(NSData *)jsonData {
    
    NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
    NSString *documentsDirectory = [paths objectAtIndex:0]; // Get documents directory
    
    NSString *styleURL = [documentsDirectory stringByAppendingPathComponent:@"mbtileStyle.json"];
    BOOL succeed = [jsonData writeToFile:styleURL atomically:YES];
    if (succeed){
        
        styleURL = [@"file:///" stringByAppendingString:styleURL];
        return [NSURL URLWithString:styleURL];
        
    }
    
    return nil;
    
}

- (NSData *)mbtilesStyleJSONWithFilenames:(NSString *)filename {
    
    // Style Path
    NSString *masterStyles = [[NSBundle mainBundle] pathForResource:@"style" ofType:@"json"];
    if (DEBUG) {
        NSLog(@"Mapbox Style: %@", masterStyles);
    }
    
    // Create dictionary of style JSON
    NSError *error = nil;
    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:[NSData dataWithContentsOfFile:masterStyles] options:NSJSONReadingAllowFragments error:&error];
    if (error) {
        NSLog(@"Load Style JSON Error: %@", error);
    }
    
    // JSON Source
    NSMutableDictionary *mutableJson = [json mutableCopy];
[...]

PR #2523, opened 2024-06-20
Updating local mbtiles file causes crash

nnhubbard:
This is a fix for issue #100 where updating a local `mbtiles` file with a new version will cause a crash.

The use case for this is for users that have `mbtiles` files on the filesystem an update that with a newer version of the file, such as downloading an updated map file. Previously this would crash.

### DIFF
--- a/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
+++ b/platform/default/src/mbgl/storage/mbtiles_file_source.cpp
@@ -252,6 +252,10 @@ private:
 
     // Multiple databases open simultaneoulsy, to effectively support multiple .mbtiles maps
     mapbox::sqlite::Database &get_db(const std::string &path) {
+        
+        // Close the DB path before using to prevent a crash that happens when you update the mbtiles file on the filesystem
+        close_db(path);
+
         auto ptr = db_cache.find(path);
         if (ptr != db_cache.end()) {
             return ptr->second;


louwers (commented):
Could you add a test?

nnhubbard:
> Could you add a test?

I would be happy to, but is there documentation somewhere on how to do this?

louwers:
You would need to add it here: https://github.com/maplibre/maplibre-native/blob/main/test/storage/mbtiles_file_source.test.cpp

What platform are you on? For CMake there is the `mbgl-test-runner` target (Linux, Windows), for iOS you need to run the CppUnitTests.

timonmw (approved):

nnhubbard:
> You would need to add it here: https://github.com/maplibre/maplibre-native/blob/main/test/storage/mbtiles_file_source.test.cpp
> 
> What platform are you on? For CMake there is the `mbgl-test-runner` target (Linux, Windows), for iOS you need to run the CppUnitTests.

I am on iOS. Thanks for the tip on where to start.

michalgwo:
Any updates on this?

louwers:
We should really add a test for this. It should be relatively straightforward. If we don't, it might break again tomorrow.

@nnhubbard Do you still want to add a test? Otherwise I can also look into it.

nnhubbard:
> We should really add a test for this. It should be relatively straightforward. If we don't, it might break again tomorrow.
> 
> @nnhubbard Do you still want to add a test? Otherwise I can also look into it.

To be honest, I have never built a test for another project before, if you are quick at it that might be easiest.

platform/default/src/mbgl/storage/mbtiles_file_source.cpp
```diff
@@ -252,6 +252,10 @@ class MBTilesFileSource::Impl {
 
     // Multiple databases open simultaneoulsy, to effectively support multiple .mbtiles maps
     mapbox::sqlite::Database &get_db(const std::string &path) {
+        // Close the DB path before using to prevent a crash that happens when you update the mbtiles file on the
+        // filesystem
+        close_db(path);
```
louwers, line 257: This means the db_cache is useless because it will always be erased if it existed before.

PR #2535, opened 2024-06-21
Add iOS example data from API with dynamic styling

louwers:
Adapted from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/WebAPIDataExample.swift


HarelM (approved):
I reviewed it without getting into all the details...

PR #2537, opened 2024-06-21
More context

TimSylvester:
Pass the file and line number of the original call through `SymbolInstance` accessor methods in addition to explicit checks, since the observed failures are from those calls.

Add more explicit checks, particularly before and after the buckets are transferred from the worker threads to the render thread, and avoid using the results when checks fail in hopes of preventing crashes.

Avoid short-circuiting on checks so that if multiple guard values are modified they will all be listed, in order to provide a better sense of the extent of the unexpected changes.


artakka (approved):

sjg-wdw (approved):
Looks legit.

PR #2538, opened 2024-06-21
iOS Documentation: Add Marker Example

louwers:
- Use DocC AutomaticArticleSubheading option to disable automatic 'Overview' heading https://www.swift.org/documentation/docc/automaticarticlesubheading
- Add marker example adapted from https://github.com/mapbox/ios-sdk-examples/blob/4fca7b8eb4004ded92ff5d9ed71f0f5f07b1f584/Examples/Swift/AddMarkerSymbolExample.swift
- Add a note to each example if it uses UIKit or SwiftUI. I am probably not going to add the same example for both UIKit and SwiftUI.

PR #2539, opened 2024-06-22
iOS Documentation: Animated Marker Example, Custom Annotation Example

louwers:
Builds on #2538

- Add Animated Marker Example, adapted from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/AnimatedLineExample.swift and Custom Annotation Example
- Updates to `docc.sh` to generate documentation faster.
- Add internal links to other documentation pages in `platform/darwin/src/MLNPolyline.h` and some other, need to update a lot of other places like this.


HarelM (approved):
I didn't dive deep into the code.

PR #2543, opened 2024-06-23
iOS Documentation: Snapshotter and Vector Layer

### LINKED ISSUE #2557: iOS Documentation: POI Along Route Example
Adapt from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/POIAlongRouteExample.swift

louwers:
- Adapted snapshotter example from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/StaticSnapshotExample.swift
- Adapted vector layer example from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/DDSCircleLayerExample.swift
- Adapted POI along route example from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/POIAlongRouteExample.swift Closes #2557
- Categorize examples
- Allow not showing logo on snapshots by making property public

### DIFF
--- a/platform/darwin/src/MLNSource.h
+++ b/platform/darwin/src/MLNSource.h
@@ -21,7 +21,7 @@ FOUNDATION_EXTERN MLN_EXPORT MLNExceptionName const MLNInvalidStyleSourceExcepti
 
  Create instances of `MLNShapeSource`, `MLNComputedShapeSource`,
  `MLNImageSource`, and the concrete subclasses of `MLNTileSource`
- (`MLNVectorTileSource` and `MLNRasterTileSource`) in order to use `MLNSource`’s
+ (``MLNVectorTileSource`` and `MLNRasterTileSource`) in order to use `MLNSource`’s
  properties and methods. Do not create instances of `MLNSource` directly, and do
  not create your own subclasses of this class.
  */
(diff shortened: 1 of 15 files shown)


hactar (approved):
Personally not the biggest fan of the `var mapView: MLNMapView!` syntax, as in real world applications, there's a lot of room for a developer to try to access mapView between init and viewDidLoad being called.

But that said, I know its still a common practice for many UIKit projects, so not a blocker for me.

HarelM (approved):

platform/darwin/src/MLNSource.h
```diff
@@ -21,7 +21,7 @@ FOUNDATION_EXTERN MLN_EXPORT MLNExceptionName const MLNInvalidStyleSourceExcepti
 
  Create instances of `MLNShapeSource`, `MLNComputedShapeSource`,
  `MLNImageSource`, and the concrete subclasses of `MLNTileSource`
- (`MLNVectorTileSource` and `MLNRasterTileSource`) in order to use `MLNSource`’s
+ (``MLNVectorTileSource`` and `MLNRasterTileSource`) in order to use `MLNSource`’s
```
michaelkirk, line 24: What's the significance of double ticks?

michaelkirk (commented):
:+1: thanks for adding these examples!

platform/darwin/src/MLNSource.h
louwers: Internal links: https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnsource

PR #2553, opened 2024-06-26
[opengl-2][node] Release node-v5.4.1-pre.0

acalcutt:
Releases node-v5.4.1-pre.0. 

This also adds publishing the node 22 binary so that can be tested in this pre-release. I did not remove Node 16 at the same time, so this isn't a breaking change.

Note, I want to merge this after https://github.com/maplibre/maplibre-native/pull/2552

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "BSD-2-Clause",
       "dependencies": {
-        "@acalcutt/node-pre-gyp": "^1.0.14",
+        "@acalcutt/node-pre-gyp": "^1.0.15",
         "@acalcutt/node-pre-gyp-github": "1.4.8",
         "minimatch": "^9.0.4",
         "npm-run-all": "^4.1.5"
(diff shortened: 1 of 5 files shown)


ntadej:
Looks OK to me.

acalcutt:
@ntadej do you have any ideas for the failed qt test. I don't think it has to do with this PR

ntadej:
Probably just a glitch, let's see now.

acalcutt:
nice, looks like it passed on the re-run

acalcutt:
This seems to work, but I had some trouble in windows at first. I ended up having to update my c++ redistributable. Without the update, when i tried to use it in tileserver-gl, the application just ended with no error.

I had the c++ 2015-2022 redistributable, but there was a slightly newer version of it that musts have been needed with the vpkg update. I should probably add a note for that

I haven't gotten to test node 22 yet, but mainly because I am having trouble getting 'canvas' to work with node 22 , so my usually *throw it into tileserver* test isn't working yet.

PR #2580, opened 2024-07-05
Avoid logging error for onMove(0,0) on Android

### LINKED ISSUE #2575: Could not call onMove with parameters 0.0,0.0
**Describe the bug**
We are seeing lots of automatic error report from our app with the message "Could not call onMove with parameters 0.0,0.0" coming from MapLibre. We get these report, because both our app and MapLibre use Timber for logging, and we use a Timber to automatically send error level logs to Sentry.

At first we thought these were somewhat random and there is some odd edge case that produces these messages, however, we are now able to consistently reproduce the issue.

**To Reproduce**
1. Rotate the map without moving it

**Expected behavior**
No _error level_ log messages when performing legitimate actions like rotating the map.

Ideally MapLibre would let us configure logging via a custom logging interface instead of using Timber directly, but that is a different issue.

 - OS: Android
 - Version: any Android Version, MapLibre currently 11.0.0, but happens with all versions we tried

**Additional context**
This is where the error is logged:

https://github.com/maplibre/maplibre-native/blob/5d5bd84d95f3692fa1bf05931cbf2fd1e222a6ce/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapGestureDetector.java#L490-L509

There is even a comment, that a move event without distance is expected and a check against 0. But then the error is still logged, if both `distanceX` and `distanceY` are zero. So unless there is a reason why the function should never be called with `0, 0`, the solution seems to be to change `else` into

```java
else if (distanceX != 0 || distanceY != 0)
```

kodebach:
Fixes #2575

### DIFF
--- a/platform/android/CHANGELOG.md
+++ b/platform/android/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### ✨ Features and improvements
 
+- Avoid logging error for onMove(0,0) on Android ([#2580](https://github.com/maplibre/maplibre-native/pull/2580)).
+
 ### 🐞 Bug fixes
 
 ## 11.0.2-pre0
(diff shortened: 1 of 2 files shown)


louwers:
@kodebach I replaced `Timber` here. Could you have a look?

Will merge if it looks OK to you.

louwers (approved):

kodebach:
LGTM using `if (Float.isNaN` makes it a lot clearer what's going on.

PR #2590, opened 2024-07-10
Experimental API to toggle tile cache in map view

alexcristici:


sjg-wdw (approved):

alasram:
Is there an equivalent Java API to turn on/off the cache on Android? Also, are there plans to expose controlling the size of the cache and not just turn it on/off?

alexcristici:
> cache on Android

This is the tile cache in memory. Yes, we will expose also for Android in the near future. First I would like to confirm that this is working well on iOS. Controlling the size of this cache could be a next step.

alasram:
I turned off the cache on Android at the C++ level and it worked just fine. The memory saving is huge with the complex map I'm rendering. I think exposing on Android should be easy in terms of testing

sjg-wdw:
Good feedback, thanks.

src/mbgl/renderer/tile_pyramid.cpp
```diff
@@ -215,12 +215,14 @@ void TilePyramid::update(const std::vector<Immutable<style::LayerProperties>>& l
...
     }
 
-    if (type != SourceType::Annotations) {
+    if (type != SourceType::Annotations && cacheEnabled) {
         auto conservativeCacheSize = static_cast<size_t>(
             std::max(static_cast<double>(parameters.transformState.getSize().width) / tileSize, 1.0) *
             std::max(static_cast<double>(parameters.transformState.getSize().height) / tileSize, 1.0) *
             (parameters.transformState.getMaxZoom() - parameters.transformState.getMinZoom() + 1) * 0.5);
         cache.setSize(conservativeCacheSize);
+    } else {
```
louwers, line 224: Should this be `else if (!cacheEnabled)`. Otherwise the cache size will get set to 0 when `type == SourceType::Annotations`. Not sure if that is problematic.

include/mbgl/renderer/renderer.hpp
```diff
@@ -108,6 +108,7 @@ class Renderer {
     const std::vector<PlacedSymbolData>& getPlacedSymbolsData() const;
 
     // Memory
+    void enableTileCache(bool);
```
louwers, line 111: Perhaps this should (also) be called `setTileCacheEnabled(bool)` because it is a bit clearer that it actually disables the tile cache when passing `false`.

louwers (changes requested):
We should keep `main` in a releasable state, so we should clearly indicate this is not a stable API for now.

We have another cache that is exposed (the 'ambient cache'). The documentation comment could perhaps offer some more clarification what cache this controls.

src/mbgl/renderer/tile_pyramid.cpp
alexcristici: When `type == SourceType::Annotations` cache is 0, because cache size is set only here and the default is 0.

include/mbgl/renderer/renderer.hpp
alexcristici: Renamed. 👍

TimSylvester (approved):

platform/ios/src/MLNMapView.mm
```diff
@@ -7501,4 +7502,15 @@ - (void)setShowsHeading:(BOOL)showsHeading
     self.showsUserHeadingIndicator = showsHeading;
 }
 
+// MARK: Tile Cache
+
+- (void)setTileCacheEnabled:(BOOL)enable {
```
louwers, line 7507: @alexcristici Should we just prefix this with `_experimental_`?

alexcristici: Yes, I think is the a good way to make sure everybody is aware.

Issue #2556, opened 2024-06-26
iOS Documentation: Offline Data Example


louwers:
Adapt from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/ManageOfflineRegionsExample.swift and https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/OfflinePackExample.swift

PR #2597, opened 2024-07-11
iOS Documentation: Offline Capabilities

louwers:
Adapted from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/OfflinePackExample.swift and https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/ManageOfflineRegionsExample.swift

Closes https://github.com/maplibre/maplibre-native/issues/2556


platform/ios/app-swift/Sources/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+private extension MLNOfflinePackProgress {
+    var percentCompleted: Float {
+        guard countOfResourcesExpected != 0 else {
+            return 0
+        }
+
+        let percentage = Float((countOfResourcesCompleted / countOfResourcesExpected) * 100)
```
ianthetechie, line 165:
```suggestion
        let percentage = Float(countOfResourcesCompleted) / Float(countOfResourcesExpected) * 100
```

I'm type checking in my head, but I _think_ both of the values in the division are unsigned integers (source: https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnofflinepackprogress). So I think the result of this division will always be either zero or one (Swift doesn't implicitly convert integers to floating point values).

louwers: You are completely right and nice catch.

louwers:
I know I need to start grouping that navigation view. I'll do it with the next one I add.

ianthetechie (approved):

ianthetechie:
It appears that SwiftFormat failed, but looks good otherwise.

Issue #2601, opened 2024-07-11
MapLibre version incorrect in action sheet


baubrey91:
**Describe the bug**
When long pressing on the attribution button, map libre shows its version. However the version number is pulled from the bundles MLNSemantivVersionString which is hard coded to 0.0.0

**To Reproduce**
1. Long press attribution indicator on bottom right
2. Notice action sheet says "MapLibre Maps SDK for iOS 0.0.0"

**Expected behavior**
Text should have correct version number ie "MapLibre Maps SDK for iOS 6.5.0"

**Screenshots**
![image](https://github.com/user-attachments/assets/8cb32fd9-06d2-4d5e-8144-859f17aa0800)

**Platform information (please complete the following information):**
 - OS: iOS
 - Platform: iPhone, iPad
 - Version: all

**Additional context**
So currently there is a version file that has version number. Possible solutions for this fix is to either update MLNSemanticVersionString in all plists with proper version number(and possible use fast lane or script to auto increment on builds) ad a new field in plists, pull from version file, or just remove the long press functionality.

louwers:
Confirmed.

louwers:
We should just use CFBundleShortVersionString.

baubrey91:
> We should just use CFBundleShortVersionString.

Ok PR made [here](https://github.com/maplibre/maplibre-native/pull/2603),  however the current CFBundleShortVersionString is 5.9.0 (I believe you guys are on 6.5.0). I can update it in my PR but it might be worth the effort to use fastlane or build a script that increments it each build

PR #2603, opened 2024-07-11
[iOS] Version number string bug in action sheet #2601

baubrey91:
This resolves issue #2601  When doing a long press on the attribution button the action sheet should show the version number. Currently it shows 0.0.0 since it is looking at the "MLNSemanticVersionString". I have changed code to look at "CFBundleShortVersionString".

## Before
image

## After
image


louwers:
Thanks but I already made a PR here: https://github.com/maplibre/maplibre-native/pull/2602

If you don't mind, I'd like to merge that one.

Issue #331, opened 2022-06-15
NSExpression function is forbidden


nvanfleet:
I've started getting an error log on device when using iOS 15.5. I'm not sure if it would happen in simulator since Maplibre currently crashes for iOS 15. 

Xcode 13.4

`2022-06-06 11:34:47.472457-0700 Lyft[42141:5886591] [general] NSPredicate: Use of 'mgl_interpolate:withCurveType:parameters:stops:' as an NSExpression function is forbidden.`

I tried to look around at things in terms of how we use this Expression in our codebase. We basically use:
`mgl_interpolate:withCurveType:parameters:stops:($lineProgress, 'linear', nil, %@)`

I was able to find some things online that said you can wrap the function here https://github.com/mapbox/mapbox-gl-native-ios/issues/271

So I tried that by doing `(mgl_interpolate:withCurveType:parameters:stops:($lineProgress, 'linear', nil, %@))` but it still issues errors.

I also tried to migrate to use the NSExpression conveniences from [Maplibre](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/ios/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L604). But I still get errors issues from those, so it seems they are also mis-formatted (?).

This is definitely polluting the logs but I am worried that somehow this will not be supported in the future by Apple. I think I was also getting some failures in my tests because it caused an unhandled exception.

wipfli:
Did the error not appear in older versions?

nvanfleet:
I'd be surprised if this wasn't something that could be repro'd by others 🤔

xD3CODER:
I got same error on iOS 15.5 too

roblabs:
In Xcode there are XCTests in the MapLibre CI scheme that can replicate this issue of *NSExpression function is forbidden*.  

To replicate in the Mapbox Unit Tests -> switch to CI scheme -> Set simulator SDK -> ⌘U (Test).

| Simulator SDK    | `NSExpression` Tests in CI Scheme     |
| :------------- | :------: |
| 14.5 | ✅ |
| 15.4 | ✅ |
| 15.5 | ❌  — (30/68 `NSExpression` tests failing)|

---

Inspecting the logs shows there are two types of Exceptions.

1. ***is not a supported method***. (NSInternalInconsistencyException)

[...]

PR #2607, opened 2024-07-13
iOS Documentation: Add example adding multiple images

louwers:
Fun (I think) example that uses an MBTiles file added as a resource to the app, and shows images for different kinds of POIs.

The resulting `.mbtiles` file is available on our S3 bucket but the example also contains information how to generate it with the open data provided by the National Park Service.

Adapted from https://github.com/mapbox/ios-sdk-examples/blob/main/Examples/Swift/MultipleImagesExample.swift which just said "drag and drop this shapefile to Mapbox Studio".


platform/ios/app-swift/Sources/MultipleImagesExample.swift
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        let imageLayer = MLNSymbolStyleLayer(identifier: "npc-poi-images", source: source)
+        imageLayer.sourceLayerIdentifier = "pois"
+        imageLayer.iconImageName = NSExpression(mglJSONObject: [
+            "match", ["get", "POITYPE"],
+            "Restroom", "restrooms",
+            "Trailhead", "trailhead",
+            "Viewpoint", "viewpoint",
+            "",
```
andresinaka, line 55: Does this work without doing the `["image", "restrooms"]`?

Related to: https://github.com/maplibre/maplibre-native/issues/331#issuecomment-2220468477

Documentation make is sound that we need to use the `["image", "image-name"]`. [Docs](https://maplibre.org/maplibre-style-spec/expressions/#image)

I'm confused! 🤔

andresinaka (approved):

platform/ios/app-swift/Sources/MultipleImagesExample.swift
louwers: That would be true if there was a property `iconImage`, however in iOS you set `iconImageName` directly.

louwers: And yes it works...

louwers: Both work actually.

ianthetechie (commented):
Just some minor wordsmithing nits; content looks good.

platform/ios/MapLibre.docc/MultipleImagesExample.md
```diff
@@ -0,0 +1,98 @@
...
+
+The resulting `.mbtiles` file can be hosted with a tile server such as [Martin](https://martin.maplibre.org/) or embedded in the app bundle as a resource. Since the file is quite small in this case we will use that last option in this example. Martin comes with a [`mbtiles` binary](https://maplibre.org/martin/mbtiles-meta.html) that allows us to inspect what the MBTiles file contains from the command line.
+
+```
+mbtiles meta-all pois-nps.mbtiles
+```
+
+The important thing to note is that there is a `pois` layer whose features include a `POITYPE` attribute. While we could add icons for all kinds of `POITYPE` values included in the dataset, we will only add icons for restrooms, trailheads and viewpoints, and leave the rest as an exercise to the reader.
+
+The icons need to be extracted with a tool like [Inkscape](https://inkscape.org/) because the National Park Service includes all icons in a big vector file. This is outside the scope of this example.
```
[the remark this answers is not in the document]
louwers: Explained here: https://github.com/maplibre/maplibre-native/blob/main/platform/ios/CONTRIBUTING.md#resources

ianthetechie (approved):

HarelM (approved):

PR #2609, opened 2024-07-14
Make minor cleanup improvements in scheduler.hpp

bilal614:
- follow through on using perfect forwarding instead of changing method signature of scheduleAndReplyValue
- make previously protected scheduleAndReplyValue method private as all current use-cases do not require it
- add trivial compile-time check for TaskFn to be invokable

### DIFF
--- a/include/mbgl/actor/scheduler.hpp
+++ b/include/mbgl/actor/scheduler.hpp
@@ -7,6 +7,7 @@
 
 #include <functional>
 #include <memory>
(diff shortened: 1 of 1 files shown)


louwers (approved):
Thanks! LGTM.

cgalvan (approved):

mwilsnd (approved):

louwers:
Thanks for your contribution @bilal614!

PR #2614, opened 2024-07-15
Allow changing color of scalebar on iOS

JesseCrocker:
MLNScale bar has all the properties needed for setting custom colors, but for unknown reasons it is blocked at several levels.
Changes in here:
* make MLNScaleBar header public
* Move color properties to public header
* change type of map scaleBar property to MLNScaleBar from UIView

Simulator Screenshot - iPhone 15 - 2024-07-15 at 10 07 29


louwers (approved):
Thank you! Will be included in the next release.

PR #2615, opened 2024-07-15
Eliminate redundant attribute updates

TimSylvester:
- Add some trace points
- Add alternate GL extension function names to allow for Tracy profiling on Pixel 8
- Reset the dirty flag on attributes even when they aren't used by the shader, e.g., `FloorWidth` on line layers, or `FillColor` on fill outlines (which share the attribute buffer with fills).


sjg-wdw (approved):

TimSylvester:
Replaced by #2629

PR #2620, opened 2024-07-16
Disabled iOS 14 metal options

alexcristici:


louwers (approved):
Looks good to me, we can add a runtime check later if we need it.

PR #2629, opened 2024-07-17
Use timestamps for attribute updates

### LINKED ISSUE #2628: Symbol attributes not updated
- The buckets generate attribute vectors and, in the case of symbols, update them when applying placement.
- The resulting values are shared among multiple drawables using `shared_ptr`.
- When an attribute vector is modified by the bucket its dirty flag is set.
- The first drawable to update will see that and reset it, causing the rest not to update.

Instead, the attribute vectors and buffers should track the last time at which they were updated.

TimSylvester:
Attribute vectors and buffers should track the last time at which they were updated.  For efficiency, the vectors still use a private dirty flag internally, to avoid querying the time for each individual value added, which is checked at the end of the update process and the update time adjusted appropriately.  And non-shared attribute data still uses a simple dirty flag.

Also incorporates #2615

Resolves #2628


TimSylvester:
The codecov results seem spurious

mwilsnd (approved):

PR #2630, opened 2024-07-17
Don't unbind texture units in release builds

TimSylvester:


sjg-wdw (approved):

TimSylvester:
> What is the idea behind this? Why keep it for debug builds?

It helps to find and prevent GL state bugs.  If some code fails to bind something and attempts to use it, it fails rather than using whatever previous value happened to be bound.

louwers:
> > What is the idea behind this? Why keep it for debug builds?
> 
> It helps to find and prevent GL state bugs. If some code fails to bind something and attempts to use it, it fails rather than using whatever previous value happened to be bound.

Could you add that as a comment?

src/mbgl/gl/context.cpp
```diff
@@ -838,6 +838,9 @@ void Context::performCleanup() {
     MLN_TRACE_FUNC();
 
 #ifndef NDEBUG
+    // In debug builds, un-bind all texture units so that any incorrect use results in an
+    // error rather than using whatever texture happened to have been bound previously.
+    // This takes some time, however, so don't do it in release builds.
     // TODO: Find a better way to unbind VAOs after we're done with them without
```
louwers, line 844: Is this TODO still relevant?

louwers (approved):

louwers:
Another look at this from @alexcristici @adrian-cojocaru or @mwilsnd would be good.

src/mbgl/gl/context.cpp
TimSylvester: I don't really know.  Assuming it means better in the sense of performance, so long as it's only in the debug build there's no need to improve it.

TimSylvester: Looks like it goes back pre-fork, so whatever.

Issue #2579, opened 2024-07-05
SIGSEGV in MapRenderer


kodebach:
We are seeing an issue similar to #2206 in our app. The crash occurs a lot, but so far we have not heard from users that it impacts them. So the crash might occur in the background or while leaving the app.

This is the current crash statistic from our Sentry instance

device model | crashes in last 90 days
-- | -- 
SM-A600FN | 108
SM-A336B | 39
SM-A405FN | 36 
SM-A325F | 56 
SM-A320FL | 212
SM-G525F | 584

To frame the number of crashes seen per device: This is a private app deployed to fewer than 500 devices, all of them Samsung phones. The list above is the complete list of devices in use, so as far as we can tell, it is not hardware specific (unless it is a general Samsung issue).

With Maplibre 11 we finally got a more detailed stack trace

```
OS Version: Android 13 (TP1A.220624.014.G525FXXSACXCA)
Report Version: 104

Exception Type: Unknown (SIGSEGV)

Application Specific Information:
Segfault

Thread 0 Crashed:
0   split_config.arm64_v8a.apk      0x70a5f1d3ec        std::__ndk1::__split_buffer<T>::~__split_buffer
1   split_config.arm64_v8a.apk      0x70a5dfcacc        mbgl::android::MapRenderer::render
2   split_config.arm64_v8a.apk      0x70a5dff5fc        jni::MakeNativeMethod<T>::lambda::__invoke<T>
3   base.odex                       0x70bea5c2d4        <unknown> + 484234871508
```

Not sure if that is helpful in any way, `__split_buffer` seems to be a an internal class of the C++ stdlib used in `std::vector` and possibly others. But I did not see any `std::vector` instances in `MapRenderer`.

Sentry also shows the `__split_buffer` frame as (although I'm not sure how it got that information)

```
std::__ndk1::__split_buffer<mbgl::util::DefaultStyle, std::__ndk1::allocator<mbgl::util::DefaultStyle>&>::~__split_buffer()
```

That would point to

https://github.com/maplibre/maplibre-native/blob/5d5bd84d95f3692fa1bf05931cbf2fd1e222a6ce/src/mbgl/util/tile_server_options.cpp#L34

But I did not check how this is connected to `MapRenderer` (if at all).

_Originally posted by @kodebach in https://github.com/maplibre/maplibre-native/issues/2206#issuecomment-2208392826_

louwers:
[...]

PR #2631, opened 2024-07-17
Fix null-pointer dereference MapRenderer Android

louwers:
In the Android-specific `MapRenderer` class a `std::unique_ptr<AndroidRendererBackend>` is dereferenced that may be a `nullptr`. This can of course lead to a crash, which was reported in https://github.com/maplibre/maplibre-native/issues/2579

I am surprised this has not resulted in more crashes being reported, because I think the null-pointer dereference happens every time `onSurfaceCreated` is (initially) called, and a method is even called on the dereferenced null-pointer...

We should probably release 11.1.0 once this is released.

### DIFF
--- a/platform/android/CHANGELOG.md
+++ b/platform/android/CHANGELOG.md
@@ -6,8 +6,11 @@
 
 - Avoid logging error for onMove(0,0) on Android ([#2580](https://github.com/maplibre/maplibre-native/pull/2580)).
 - Experimental API to toggle tile cache in map view ([#2590](https://github.com/maplibre/maplibre-native/pull/2590)). This can reduce memory usage at the cost of having to parse tile data again when the zoom level changes.
+- Add TaggedScheduler, couple lifetime of tasks and orchestrator ([#2398](https://github.com/maplibre/maplibre-native/pull/2398)).
 
 ### 🐞 Bug fixes
+
+- Fix null pointer dereference MapRenderer Android.
 
 ## 11.0.2-pre0
 
--- a/platform/android/MapLibreAndroid/src/cpp/map_renderer.cpp
+++ b/platform/android/MapLibreAndroid/src/cpp/map_renderer.cpp
@@ -182,6 +182,7 @@ void MapRenderer::render(JNIEnv&) {
     }
 
     // Activate the backend
+    assert(backend);
     gfx::BackendScope backendGuard{*backend};
 
     // Ensure that the "current" scheduler on the render thread is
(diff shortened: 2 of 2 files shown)


TimSylvester:
Interesting, it looks like that means a render can occur before `onSurfaceCreated` is called or after a `resetRenderer`.

louwers:
> Interesting, it looks like that means a render can occur before `onSurfaceCreated` is called or after a `resetRenderer`.

I think onSurfaceCreated dereferences `backend` before it is initialized in that same method.

kodebach:
> I think onSurfaceCreated dereferences `backend` before it is initialized in that same method.

I rather suspect it's the other end of the lifecycle, i.e. `render` is called _after_ Android has destroyed the surface. (see also https://github.com/maplibre/maplibre-native/issues/2579#issuecomment-2236232406)

I don't know the maplibre code, so no idea if Android destroying the surface would cause the `backend` reference to become invalid. But based on our crash data that seems most likely to me.

louwers:
Right. When `render` is called after `resetRenderer`, the `backend` is again a nullptr.

louwers:
When the app goes to the background and `onSurfaceDestroyed` is called, the backend and renderer are reset. `onSurfaceChanged` calls `onSurfaceCreated` when no renderer exists, but `onSurfaceCreated` needs a valid backend to exist. I updated the `reset()` method to create a new backend to make sure there is always a valid one.

TimSylvester (approved):

mwilsnd (approved):

elican-doenyas:
@louwers I experienced the same issue when I upgraded to use the latest map libre android sdk (v11.6.1). Looking into the issue, there seems to be a nullptr dereference on MapRenderer:render method. I believe this happens when the style.json does not include a renderer property in its metadata. I was able to resolve the issue by adding;

"maputnik:renderer": "mbgljs"

value to the metadata section of the style.json I am using. 

Can you please verify this and possibly introduce a fix so it does not crash in the absence of this property? We were not aware that it was required as the metadata is listed as 'optional' in the style specs here: https://maplibre.org/maplibre-style-spec/root/#metadata

Note 1: Previously, our company was using the v10.2.0 maplibre android sdk, and it was working fine without this property in the style.

Note 2: The relevant crash log is attached.
[maplibre-nullptr.txt](https://github.com/user-attachments/files/17831870/maplibre-nullptr.txt)

louwers:
@elican-doenyas That is a different issue, please open a separate issue.

PR #2639, opened 2024-07-22
Improve linux build docs

nyurik:
* ensure docker image uses host's files -- much better for IDE and other regular work
* ensure docker does not create files as `root` -- harder to cleanup
* improve docs


platform/linux/README.md
```diff
@@ -2,77 +2,85 @@
...
+
+![Sample image of world from mbgl-render command](/misc/sample-maplibre-style-mbgl-render-out.png)
+
+### Outside of Docker
+
+Render output image using default MapLibre demo tiles server. You can run this binary on your host machine even if you built it inside a Docker container.
 
 ```bash
 ./build/bin/mbgl-render --style https://raw.githubusercontent.com/maplibre/demotiles/gh-pages/style.json --output out.png
+open out.png
```
louwers, line 69: This is not a standard program as far as I am aware, maybe use `xdg-open`.

louwers (approved):

platform/windows/README.md
```diff
@@ -83,7 +83,7 @@ If all went well and target `mbgl-render` or `ALL_BUILD` was chosen, there shoul
 
 This should produce an `out.png` map tile image with the default MapLibre styling from [the MapLibre demo](https://maplibre.org/).
 
-![Sample image of world from mbgl-render command](/misc/sample-maplibre-style-mbgl-render-out.png)
+![Sample image of world from mbgl-render command](https://github.com/user-attachments/assets/b1a9355f-731e-4a0f-ab90-68c5ae4beb65)
```
nyurik, line 86: I don't think this is a good path: repo's content should not depend on external services like this. When all content is together, cloning it is enough -- and we can easily do any kind of markdown book packaging to create a site. When it's external, it can't not be part of the docs.  You could do this as a separate PR and discuss it there of course.

louwers: I reverted the change.

PR #2643, opened 2024-07-22
Improve Dockerfile with cache, Bazelisk, and cmake support

nyurik:
Allow `bazel` and `cmake` to be used with docker by simply adding them after `docker run ...`. Moreover it is designed to keep build and bazel cache between runs, which means no re-downloads between separate `docker run` commands.


louwers:
This looks a bit weird

```
I have no name!@f862704c4754:/app$ 
```

I also followed the instructions and I get

```
$ bazel build //:mbgl-core
2024/07/23 12:36:03 could not create directory /.cache/bazelisk: mkdir /.cache: permission denied
```

Running Docker as non root user.

platform/linux/README.md
```diff
@@ -26,18 +26,20 @@ Optional: `libsqlite3-dev` (also available as vendored dependency).
 
 ### Building with Docker
 
-You can use a Docker container to build MapLibre Native. A `Dockerfile` that installes the required dependencies when the image is built is provided in this directory.
+You can use a Docker container to build MapLibre Native. Note that you cannot build using both Docker and host methods at the same time. If you want to switch, you need to clean the repository first, e.g. by using `git clean -dxfi -e .idea -e .clwb -e .vscode`
```
louwers, line 29: This should be (made) possible.

nyurik: @louwers I don't think it can be -- the `cmake -B build ...` step creates absolute paths in the `build/` dir, so you should use just one method. If you can build locally, why would you use docker?

One workaround could be to set some env var inside docker container to force the output to go to another dir instead of `build/` or we could even remap that dir to some other location (although that's kinda tricky because mapping a dir and a subdir of the same dir is not a good practice i think)

nyurik:
@louwers please try again - I changed the approach, should work now, same as in vscode

nyurik:
@louwers maybe your `.docker_cache/` got messed up due to earlier docker tests.  See if you can clean up your repo with `git clean -dxfi -e .idea -e .clwb -e .vscode`

platform/linux/README.md
louwers: Can you build with Docker and locally if you use a different build directory? If yes, then that comment can be removed I think.

platform/linux/startup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+if [ ! -d /app/.github ] || [ ! -d /home/user/.cache ]; then
```
louwers, line 3: Can we move all docker related files and instructions to a `docker` directory? I think it is getting a bit cluttered now, and for local development of a graphics toolkit you would almost never use Docker, so while it is good to have it, it somewhat of a fringe use case.

platform/linux/README.md
```diff
@@ -26,25 +26,22 @@ Optional: `libsqlite3-dev` (also available as vendored dependency).
...
-docker run --rm -it -v "$PWD:/code/" -u $(id -u):$(id -g) maplibre-native-image
-```
-
-You can safely ignore this type of message and a missing username. It happens if your linux user was not the first one created on the system.
-
-```
-groups: cannot find name for group ID ....
-I have no name!@...:/root$ `
+# Run all build commands using the docker container.
+# You can also execute build commands from inside the docker container by starting it without the build command.
```
louwers, line 43: If we have to choose one, just include the command without the `___any_build_command___`, because it's nicer to have a valid command that people can copy and paste.

Optionally mention that you can also run build commands directly.

docker/Dockerfile
```diff
@@ -36,7 +36,7 @@ ARG USERNAME=user
 ARG USER_UID=1000
```
louwers, line 36: Thanks for this but can we make it under `platform/linux`?

louwers (approved):

PR #2647, opened 2024-07-24
Rust css color

nyurik:
This is a totally incomplete proof of concept how a Rust css parser can be used from C++

This PR uses changes from #2643


louwers:
I ran the Bloaty size test manually on Linux. It is reporting a +364% binary size increase. Diff here: https://gist.github.com/louwers/809e971d9ae3459bcff450487c29249c

louwers:
Something to consider is platform support by `rustc` and the Rust standard library. ARM64 iOS and ARM64 Android are classified as having 'Tier 2' support.

> Tier 2 targets can be thought of as "guaranteed to build". The Rust project builds official binary releases of the standard library (or, in some cases, only the core library) for each tier 2 target, and automated builds ensure that each tier 2 target can be used as build target after each change.

> Tier 2 target-specific code is not closely scrutinized by Rust team(s) when modifications are made. Bugs are possible in all code, but the level of quality control for these targets is likely to be lower

https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-without-host-tools

ianthetechie:
> Something to consider is platform support by `rustc` and the Rust standard library. ARM64 iOS and ARM64 Android are classified as having 'Tier 2' support.

This is correct, but the Tier 2 label is a bit scarier than it sounds. For example, macOS on Apple Silicon is also somewhat infamously Tier 2 still 😉

It's worth noting that major projects are also using Rust in production on iOS and Android, including Firefox which uses Rust to share code across platforms.

louwers:
Can you try building with `#![no_std]`? https://docs.rust-embedded.org/book/intro/no-std.html

nyurik:
@louwers `no_std` would kill 90% of the Rust value -- it is mostly used for the embedded (no OS) cases (`writing firmware, kernel, or bootloader code`).  Are there any reasons for it?

nyurik:
Looking at `./build/bin/mbgl-render` output:
* original size is 124MB, shrinks to 7.8MB after running `strip` on it
* If I enable `lto = true` and `codegen-units = 1` in rustutils (PR is now updated), it shrinks to 122MB / 7.1MB --- 9% improvement on stripped
* That said, compiling original produces 120 / 6.8 MB, still 4% smaller on stripped.

I looked at the size increase text dump - looks really weird.  It shows a significant increase in all sorts of .cpp files, and I am really not sure why that might be the case.  The key changes are in these I think:

```
[NEW]    +171  [ = ]       0    /rust/deps/compiler_builtins-0.1.109/src/lib.rs/@/compiler_builtins.13ee9051c16629a4-cgu.016
[NEW]    +790  [ = ]       0    /rust/deps/compiler_builtins-0.1.109/src/lib.rs/@/compiler_builtins.13ee9051c16629a4-cgu.008
[NEW]  +130Ki  [NEW] +48.6Ki    /rust/deps/memchr-2.5.0/src/lib.rs/@/memchr.7066fcbbe06873ca-cgu.0
[NEW]  +180Ki  [NEW] +25.6Ki    library/alloc/src/lib.rs/@/alloc.bfbae7e348dce413-cgu.0
[NEW]  +829Ki  [NEW]  +207Ki    library/core/src/lib.rs/@/core.868bc93c3f2beb33-cgu.0
[NEW] +1.46Ki  [NEW]    +259    /rust/deps/compiler_builtins-0.1.109/src/lib.rs/@/compiler_builtins.13ee9051c16629a4-cgu.110
[NEW] +25.0Ki  [NEW] +13.2Ki    /home/bart/build-rust/corrosion_generated/cxxbridge/rustutils_headers/src/lib.cpp
[NEW] +3.14Mi  [NEW]  +619Ki    library/std/src/lib.rs/@/std.3c8ba8ebcf555201-cgu.0
[NEW] +5.40Ki  [NEW] +1.04Ki    library/panic_unwind/src/lib.rs/@/panic_unwind.79513d39ffd1496f-cgu.0
```

Most of these are 1-time cost, i.e. some core mem alloc, utf8, and panic handling, and a 4% increase might be a good trade in exchange for other benefits.  But clearly we should pay attention to that.

ianthetechie:
On `no_std`, it's worth noting that you only "pay" (in final binary size) for the parts of the std lib that you actually use, and these are almost all extremely helpful. It's possible to do `no_std`, but you don't really gain much except for an embedded target. And even there, you actually still do have access to a large part of what you initially think is in `std`. Most of that's actually in `core`, which is available, or `alloc`, which you can opt into (and of course it's available on all platforms we currently target).

nyurik:
Thx @ianthetechie! Why would we want to build a `cdylib`?  If i understand it correctly, that creates an `.so / .dll` - which has to be shipped together with the final binary as a dynamically loadable file.  Wouldn't it only make sense if we were building a shared lib distributed via .deb package? The current usecase is to create core component that gets linked with other c++ code into a single executable.

ianthetechie:
> Wouldn't it only make sense if we were building a shared lib distributed via .deb package? The current usecase is to create core component that gets linked with other c++ code into a single executable

Sorry I may have missed some of which platforms we're targeting / how it's built @nyurik ;) To distribute Rust code for Android via the NDK + Java bridges (JNI/JNA), you usually build a `cdylib`. Yes, that's dynamic (needs to be around somewhere), but the whole library (MapLibre Native) links in all of its native dependencies (statically). Just pointing that out since I assumed we'll eventually want that.

I regrettably don't have more details on why that's the case / what technical limitations there are, but that's what all the tooling bridging cargo and the NDK requires of crates. Maybe it doesn't apply to us since we're essentially linking up a library that will look like just a regular lib with headers and a C ABI already; I guess Bazel is driving a lot of this linking and by the time it gets to our NDK step, it's all indistinguishable anyways.

nyurik:
Ah, gotcha - yes, the resulting build target that wraps all Rust + C++ functionality could be dynamic - but that's up to the cmake/bazel/... to create and use in JNI. It would not affect how I build the low-level core components that get linked in.  Otherwise you end up with JNI -> C++ cdylib -> Rust cdylib.

As a separate project, I will try to wrap the whole ml-native as a rust lib. That target will need to support both rustlib and cdylib output.

ianthetechie:
@louwers To put the T2 target concern to rest, I pinged a few ppl on Mastodon to get an answer closer to the source, and Esteban Küber, a member of the compiler team responded: https://hachyderm.io/@ekuber/112841995275142925. TL;DR we can rely on stable channel Rust releases; just not nightly (which nobody is proposing here haha). It's more a reflection of CI resources than anything (and as such, perhaps unsurprisingly, eventually x86 macOS will eventually move to T2).

nyurik:
This PR has been rebased on the new docker implementation - so now it can be tried very easily without installing anything locally, while still not having to re-download anything on each `docker run` command

CMakeLists.txt
```diff
@@ -1355,6 +1354,35 @@ set_target_properties(
...
+set_target_properties(
+        rustutils
+        PROPERTIES
+        INTERFACE_MAPBOX_NAME "Rust Utils"
+        # INTERFACE_MAPBOX_URL "https://github.com/wolfpld/tracy.git"
+        INTERFACE_MAPBOX_AUTHOR "Yuri Astrakhan"
+        INTERFACE_MAPBOX_LICENSE ${CMAKE_CURRENT_LIST_DIR}/rustutils/LICENSE-MIT
+)
+# TODO: this is a dup of the above because license checker sees headers as a separate target
+set_target_properties(
```
alasram, line 1377: `set_target_properties(rustutils rustutils_headers ...)`

rustutils/Cargo.lock
```diff
@@ -0,0 +1,162 @@
+# This file is automatically @generated by Cargo.
```
alasram, line 1: Is this needed to be checked in

rustutils/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+name = "rustutils"
+version = "0.1.0"
+authors = ["Yuri Astrakhan <YuriAstrakhan@gmail.com>", "MapLibre contributors"]
+description = "Core Rust utilities for MapLibre Native"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/maplibre/maplibre-native"
+
+[lib]
+crate-type = ["staticlib"]
```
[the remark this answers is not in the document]
alasram: https://lifthrasiir.github.io/rustlog/why-is-a-rust-executable-large.html

rustutils/Cargo.lock
nyurik: @alasram yes, this file is usually checked in for the crates that produce a binary (i.e. an executable, a static or a dynamic library). This way everyone is using exactly the same dependency tree. This file is usually not checked in if its a simple rust library - i.e. something that is used as part of a bigger rust project.

CMakeLists.txt
nyurik: thanks, fixed

rustutils/Cargo.toml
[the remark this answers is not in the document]
maxammann: I think the Rust library is linked with the rest of maplibre to produce a final shared library. So the target of the Rust library should be staticlib for both platforms I believe.

bazel/BUILD.bazel
maxammann, line 1: Unusred files

.github/workflows/android-ci.yml
```diff
@@ -70,6 +70,9 @@ jobs:
           distribution: "temurin"
           java-version: "17"
 
+      - name: Add aarch64-linux-android for Rust toolchian
+        run: rustup target add --toolchain stable-x86_64-unknown-linux-gnu aarch64-linux-android armv7-linux-androideabi i686-linux-android
```
maxammann, line 74: The dependency to rustup is not hight right now, so this is fine. However, we should make sure not to depend too much on rustup. This is a mistake that I made in maplibre-rs. The mistake later shows up e.g. when integrating into other build systems.

Ideally, the only required dependency is `rustc`. Realstically there is no way around `cargo`. However, a strict requirement for `rustup` can be easily avoided.

Using rustup in the CI is totalyl fine! (always!) We should just avoid depending on rustup in e.g. cmake scripts.

nyurik: thx, wasn't aware that rustup is a problem in cmake/bazel, so will make sure to avoid it there

maxammann: Not neccassarily a problem, but rustup is just one way of installing Rust.

Note that the rust-toolchain.toml is for example a feature of rustup. It might just cause frictioin to depend on rustup early on :)

ntadej:
While this is meant to be a POC, let me express my opinion, that I am very hesitant to depend on Rust with Qt.
Could we pre-build such dependencies?

maxammann:
> While this is meant to be a POC, let me express my opinion, that I am very hesitant to depend on Rust with Qt. Could we pre-build such dependencies?

Why is that? Are QT projects somehow picky about how dependencies are built?

ntadej:
> > While this is meant to be a POC, let me express my opinion, that I am very hesitant to depend on Rust with Qt. Could we pre-build such dependencies?
> 
> Why is that? Are QT projects somehow picky about how dependencies are built?

There are a few users building for embedded devices where adding additional complexity may cause issues.

louwers:
I re-ran Bloaty manually Linux, looks a lot leaner now! https://gist.github.com/louwers/bb80f15df034061bb5ddc6022070aba0

This is a proof-of-concept to try to see how hard it is to integrate Rust into our build, because if we want to do that, that would be the first hurdle to overcome.

Whether we actually want to integrate Rust into our build, and thus allow certain parts of MapLibre Native be written in Rust, is something that warrants a discussion. I have my own thoughts about this, but in general any change that adds significant complexity needs to be well-justified in terms of what our users are interested in.

Adding a hard dependency on Rust will be a hard sell, especially to our users that care about binary size. If you want this effort to succeed, I would recommend an approach where existing modules can have opt-in Rust implementations. Then, when you have a foot in the door, you are in a good position to write new modules that add 'killer functionality' in just Rust (e.g. a MapLibre Tiles decoder). That way you would generate buy-in, trust and excitement in our community.

louwers:
@maxammann That is the file size, the VM size is what matters.

louwers:
Haha you were too quick. I guess both matter. @mwilsnd https://github.com/google/bloaty/blob/main/doc/using.md

.github/workflows/android-ci.yml
ianthetechie: heh I love that you mention `rust-toolchain.toml` @maxammann ;) I think that this is one of the most underrated features for cross-platform development Ind I do use it in my projects. That said, your concern is completely correct in the case of MapLibre and other projects which use Bazel for build orchestration.

nyurik (commented):
a few minor nits

louwers:
See https://github.com/maplibre/maplibre-native/pull/3137 I made some changes.

Issue #1693, opened 2023-09-26
Refactor `expression.test.cpp` to use std::filesystem


louwers:
This file is a bit ugly. Instead of using platform-specific APIs to instantiate the test names in `INSTANTIATE_TEST_SUITE_P` we can probably use std::filesystem instead.

https://github.com/maplibre/maplibre-native/blob/main/test/style/expression/expression.test.cpp

bilal614:
@louwers : I can give this Issue a go. Created PR https://github.com/maplibre/maplibre-native/pull/2649

PR #2649, opened 2024-07-24
Refactor expression.test.cpp to use std::filesystem

bilal614:
Refactor expression.test.cpp to use std::filesystem. Remove system specific file handling.
Addresses Outstanding Issue: https://github.com/maplibre/maplibre-native/issues/1693


louwers (changes requested):
Thanks, looks good once CI passes.

Android is not happy:

```
native/test/style/expression/expression.test.cpp
  /home/runner/work/maplibre-native/maplibre-native/test/style/expression/expression.test.cpp:93:17: error: use of undeclared identifier 'name'
              if (name.find("number-format") != std::string::npos) {
                  ^
  1 error generated.
```

bilal614:
Yep, just pushed a fix for Android build. Missed that one. thnx

bilal614:
@louwers Looking at some of the failing builds, some of the used toolchains rely on < c++17(don't like us using std::filesystem*). What do you think is the best course of action for those cases? Is it a possibility to disable the test for such cases or do you have another alternative in mind?

louwers:
@bilal614 Some APIs may be unavailable on some macOS and iOS versions, even though the compiler still supports C++17.

In the case of iOS, we can probably bump the iOS target to 13.0.


You could try taking out the [build (macOS, macos-13, RelWithDebInfo, 5.15.2, desktop, 10.13, x86_64)](https://github.com/maplibre/maplibre-native/actions/runs/10076090013/job/27857261184#logs) target from the matrix, but @ntadej would need to approve.

ntadej:
Hmm, there is no reason why Qt should not work though, but probably the build config needs to be updated.

Can you remind me where do we actually link with the filesystem library? This is separate to the standard library with some compilers.

bilal614:
@ntadej : I was able to get the qt-ci builds green by bumping the deployment_target to 10.15 qt-ci for macOS 13 build jobs(Ran a separate build job on forked repo https://github.com/bilal614/maplibre-native/actions/runs/10094629091/job/27912899722). I could not try out similar approach to ios-ci builds since those use "self-hosted" labels. Could you please help out with bumping the iOS target of the C++ unit test app to 13.0, since I'm not very familiar with Bazel builds?

louwers:
Yes let me try to fix iOS.

bilal614:
@louwers : I tried bumping up the minimum_ios version to 13.0 in the rules specified under platform/ios/BUILD.bazel. Could you please check if it looks good(makes sense) and trigger a build for any feedback. Thanks

louwers:
We can't bump the minimum iOS version across the board, just of the unit tests should be fine. I'll try it out locally now.

bilal614:
Yep, makes sense to keep the scope limited to the tests. Thanks, much appreciated 👍

bilal614:
@louwers : Seems like builds are green, thanks again for the help. Merge is still blocked on requested review change(already fixed).

louwers:
Yes I was doing one last sanity check to see if the tests still run, will merge.

louwers (approved):
Just realized  that `populateNames` could be renamed now that is not populating anything anymore, but let's fix that in the next one.

Thank you for this contribution! 💯

PR #2653, opened 2024-07-24
Add justfile to simplify docker usage

nyurik:
This adds a [justfile](https://just.systems/man/en/) to simplify docker usage. Eventually more recipes can be added to it to make executing any kind of long complex commands simpler.

image

Just is a very popular task runner, with 19k stars and 7.5 million downloads. It supports autocomplete, IDE integration, CI installation, and even github highlighting

This is an **optional** feature, and does not have to be used. I think we may want to use this method to keep all CI scripts in one place as well - in order to be able to run them both locally or in a workflow.

### DIFF
--- a/justfile
+++ b/justfile
@@ -6,6 +6,10 @@ set shell := ["bash", "-c"]
 @_default:
     {{just_executable()}} --list
 
+# interactively clean-up git repository, keeping IDE files
+git-clean:
+    git clean -dxfi -e .idea -e .clwb -e .vscode
+
 # (re-)build `maplibre-native-image` docker image for the current user
 init-docker:
     docker build -t maplibre-native-image --build-arg USER_UID=$(id -u) --build-arg USER_GID=$(id -g) -f docker/Dockerfile docker
--- a/justfile
+++ b/justfile
@@ -16,4 +16,4 @@ init-docker:
 
 # run a build command with docker, e.g. `just docker bazel build ...`, or open container shell with `just docker`
 docker *ARGS:
-    docker run --rm -it -v "$PWD:/app/" -v "$PWD/docker/.cache:/home/user/.cache" maplibre-native-image
+    docker run --rm -it -v "$PWD:/app/" -v "$PWD/docker/.cache:/home/user/.cache" maplibre-native-image {{ARGS}}
--- a/justfile
+++ b/justfile
@@ -14,6 +14,6 @@ git-clean:
 init-docker:
     docker build -t maplibre-native-image --build-arg USER_UID=$(id -u) --build-arg USER_GID=$(id -g) -f docker/Dockerfile docker
 
-# run a build command with docker, e.g. `just docker bazel build ...`, or open container shell with `just docker`
+# run command with docker, e.g. `just docker bazel build //:mbgl-core`, or open docker shell with `just docker`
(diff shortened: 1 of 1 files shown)


louwers (changes requested):
I don't want to introduce yet another tool that people should install. We already use a Makefile for Android. I also want to limit the amount of top-level files when possible, especially when those files are not relevant for all users (not everyone will use Docker).

Could you add a Makefile to the `docker` directory instead?

nyurik:
@louwers while ubiqutous, Makefile is something many people are trying to get rid of due to its syntax complexity and combersomness as a TASK manager.  Also, this is not specifically for docker, but rather for any recipe project-wide.

Justfile file is OPTIONAL - developers do not have to use it, but it offers great value that we already use in martin, tilespec, and maplibre-rs.  Also, there are [19k](https://github.com/search?q=path%3A**%2Fjustfile+NOT+is%3Afork&type=code) usages on github excluding forks -- each of which is likely a full blown, often complex project.

louwers:
You can use a Makefile as a simple task runner without dependencies too. That said, this is a monorepo, for better or worse. Many commands you might want to run are not even cross-platform. It does not make sense to have a project wide task runner.

It's also pretty embarrassing to have a task runner, but it cannot do the most important things developers want to do when they come to the project - build, run & compile for the platform they are interested in.

sjg-wdw:
Nothing the maintainers put in a repo is really optional is the thing.

nyurik:
As clearly observed from my prior work in numerous repos, I am pretty good at maintaining things, and also pretty good at getting them to a better state. I am willing to invest a lot of time in this repo, move it forward, introducing clear ways to build all targets one after another on all platforms.  Justfile would help me achieve this goal by keeping things cross-platform and simple to build, rather than deal with Makefiles or other build systems (i.e. the death by a thousand cuts argument)

sjg-wdw:
The Native repo had a whole lot of semi-broken dependences last year which the maintainers and team cleaned up.  We should be very careful when adding another level of build complexity.  It really needs to be compelling.

louwers:
> Justfile would help me achieve this goal by keeping things cross-platform and simple to build

It is simple for you because you have Docker installed and running, know Docker very well, have inspected and extended the Dockerfile, so you know what is happening when you run these commands and can debug any issues that occur.  But adding another level of indirection (or two in this case) is not helpful in general in my opinion.

I recommend keeping the justfile in the repo and create a personal `.gitignore` to exclude it from git.

nyurik:
@louwers lets not conflate two unrelated things - justfile and docker. They are not related, because next I will add things like  `cmake-core` and `bazel-core`, plus a `build-core` which will redirect to one of this targets.  And the main benefit of all this is that I can share these experiments with others while being on the shared repo - i.e. i do not want to fork the whole project because it will be a massive pain to merge it in later on.

sjg-wdw:
This really sounds like it should be worked out in a fork.

nyurik:
I have made a temporary wrapper repo for this functionality to avoid prolonging this discussion. Hopefully there will be some positive feedback in the future and we could bring back this topic. For those who want to use simpler way of building maplibre-native, see https://github.com/nyurik/just-maplibre-native

PR #2668, opened 2024-07-31
Reuse prefetched tiles to avoid empty screen

alexcristici:
When zooming out if needed tile is not loaded or not in cache it will try to use the prefetched tiles from a higher zoom level.

Before:

https://github.com/user-attachments/assets/22cdfb55-3472-4467-b9a2-3fad7a3f273f

After:

https://github.com/user-attachments/assets/bb6283b6-d2ac-4154-9858-3c3c8362fac0

Recordings are from Debug iPhone 6s so the behavior to be more visible, release build is much faster.

Before:

https://github.com/user-attachments/assets/b2511437-267c-4c2a-a55d-0bde157a921a

After:

https://github.com/user-attachments/assets/6f5856d5-fa65-4a09-9c26-d4f74e96a556


louwers:
This test is failing:

```
[ RUN      ] Map.PrefetchTiles
/home/runner/work/maplibre-native/maplibre-native/source/test/map/prefetch.test.cpp:63: Failure
Expected equality of these values:
  tiles.size()
    Which is: 19
  expected.size()
    Which is: 10
[  FAILED  ] Map.PrefetchTiles (318 ms)
```

alexcristici:
> This test is failing:
> 
> ```
> [ RUN      ] Map.PrefetchTiles
> /home/runner/work/maplibre-native/maplibre-native/source/test/map/prefetch.test.cpp:63: Failure
> Expected equality of these values:
>   tiles.size()
>     Which is: 19
>   expected.size()
>     Which is: 10
> [  FAILED  ] Map.PrefetchTiles (318 ms)
> ```

I will have to update the tests.

alexcristici:
Tests are passing, PR is ready for review.

src/mbgl/algorithm/update_renderables.hpp
```diff
@@ -14,12 +14,14 @@ template <typename GetTileFn,
           typename CreateTileFn,
           typename RetainTileFn,
           typename RenderTileFn,
-          typename IdealTileIDs>
+          typename IdealTileIDs,
+          typename PrefetchedTileMap>
```
louwers, line 18: I don't think `IdealTileIDs` nor `PrefetchedTileMap` need to be template parameters.

Can you change them to `const std::vector<OverscaledTileID>&` and `const std::map<OverscaledTileID, std::unique_ptr<Tile>> &`?

src/mbgl/renderer/tile_pyramid.cpp
```diff
@@ -23,6 +23,7 @@ namespace mbgl {
 using namespace style;
 
 static TileObserver nullObserver;
+static std::map<OverscaledTileID, std::unique_ptr<Tile>> emptyPrefetchedTiles;
```
louwers, line 26: Should be const.

src/mbgl/algorithm/update_renderables.hpp
alexcristici: Tests are using `MockTestData` instead of `Tile`. So they need to be template parameters for the tests.

alexcristici: Also `idealTiles` are defined as `set` in tests: `std::set<mbgl::OverscaledTileID>`

louwers: Thanks.

louwers:
In my testing the old tiles stick around too long, which has unexpected visual results, especially when the style at a higher zoom level very is different.

image

alexcristici:
@louwers Should be fine now. Please check again.

src/mbgl/algorithm/update_renderables.hpp
```diff
@@ -26,6 +26,7 @@ void updateRenderables(GetTileFn getTile,
                        const std::optional<uint8_t>& maxParentOverscaleFactor = std::nullopt) {
     std::unordered_set<OverscaledTileID> checked;
     bool covered;
+    bool parentOrChildTileFound;
```
louwers, line 29: You can just initialize it to false here.

louwers:
That looks better!

src/mbgl/algorithm/update_renderables.hpp
alexcristici: It is initialized with false in the right place. There is a for loop.

alexcristici:
pre-commit.ci autofix

alexcristici:
@louwers Not sure why precommit is failing.

src/mbgl/algorithm/update_renderables.hpp
louwers: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always

louwers (approved):

TimSylvester (approved):
Looks good.  Have you tried it in overscaled and wrapped cases?  It doesn't look like the new tests consider them.

mwilsnd (approved):

PR #2694, opened 2024-08-06
Add extra observer hooks for interesting internal processes

mwilsnd:
This PR adds a set of new observer events which can be used to instrument interesting events within the core library. These are primarily intended for use by external performance instrumentation tools and are thus split into `pre` and `post` events where relevant.

These events have also been passed up through the JNI layer to the Android platform.


include/mbgl/gfx/context.hpp
```diff
@@ -51,10 +52,13 @@ using UniqueDrawableBuilder = std::unique_ptr<DrawableBuilder>;
 using VertexAttributeArrayPtr = std::shared_ptr<VertexAttributeArray>;
 #endif
 
+static ContextObserver nullObserver;
```
louwers, line 55: This can be `const`. Consider using an anonymous namespace instead.

include/mbgl/tile/tile_operation.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include <cstdint>
+
+namespace mbgl {
+enum class TileOperation : uint8_t {
```
louwers, line 5: These could use some Doxygen comments.

```
    RequestedFromCache ///< like this
```

platform/android/MapLibreAndroidTestApp/src/main/AndroidManifest.xml
```diff
@@ -874,6 +874,18 @@
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value=".activity.FeatureOverviewActivity" />
+        </activity> <!-- Telemetry -->
```
louwers, line 877: Could you name this something else? I don't think it is telemetry (because the data is not sent anywhere) and because users often ask if we have any telemetry. I don't want there to be any confusion.

test/map/map.test.cpp
```diff
@@ -1684,3 +1687,166 @@ TEST(Map, InvalidUTF8InTile) {
...
+
+TEST(Map, ObserveTileLifecycle) {
+    util::RunLoop runLoop;
+
+    struct TileEntry {
+        OverscaledTileID id;
+        std::string sourceID;
+        TileOperation op;
+
+        bool operator==(const TileEntry& other) const noexcept {
```
louwers, line 1699: You should be able to default this `operator==` since we use C++20 now.

louwers (commented):
Left some comments but looks good otherwise.

We should probably expose this to iOS as well?

include/mbgl/gfx/context.hpp
mwilsnd: It can be moved to an anonymous namespace but can't be const as it's used to default-initialize a non-const pointer

platform/android/MapLibreAndroidTestApp/src/main/AndroidManifest.xml
mwilsnd: Renamed to "Events"

test/map/map.test.cpp
mwilsnd: Yup, looks like I don't even need to specify it with Clang 16.

louwers (approved):

include/mbgl/gfx/context.hpp
louwers: Alright!

include/mbgl/map/map_observer.hpp
```diff
@@ -66,6 +71,24 @@ class MapObserver {
     // Observe this event to easily mutate or observe shaders as soon
     // as the registry becomes available.
     virtual void onRegisterShaders(gfx::ShaderRegistry&) {};
```
TimSylvester, line 73: extra ';'

include/mbgl/renderer/renderer_observer.hpp
```diff
@@ -54,6 +61,17 @@ class RendererObserver {
 
     // Entry point for custom shader registration
     virtual void onRegisterShaders(gfx::ShaderRegistry&) {};
+    virtual void onPreCompileShader(shaders::BuiltIn, gfx::Backend::Type, const std::string&) {};
```
TimSylvester, line 64: more extra ';'

src/mbgl/tile/geometry_tile.cpp
```diff
@@ -312,12 +325,8 @@ void GeometryTile::onLayout(std::shared_ptr<LayoutResult> result, const uint64_t
 
     observer->onTileChanged(*this);
 
-    if (layoutResult) {
-        for (const auto& data : layoutResult->layerRenderData) {
-            if (data.second.bucket) {
-                data.second.bucket->check(SYM_GUARD_LOC);
-            }
-        }
+    if (!pending) {
```
TimSylvester, line 328: Did you mean to remove the guard checks?  Merge conflict?

mwilsnd: Think so, it was unintentional regardless and I've put it back

alexcristici (approved):

TimSylvester (approved):

Issue #2436, opened 2024-05-27
Replace usage of `android.os.AsyncTask` with Kotlin Coroutines


louwers:
As mentioned by @westnordost in [this comment](https://github.com/maplibre/maplibre-native/issues/2432#issuecomment-2132178993) [AsyncTask](https://developer.android.com/reference/android/os/AsyncTask) is deprecated and we should replace our usage of it with more modern APIs like Kotlin Coroutines.

A [developer](https://maplibre.org/jobs/developer-job-posting/) could claim a bounty under the [modernization](https://github.com/maplibre/maplibre/issues/270) bounty direction for this work.

The source files using `android.os.AsyncTask` should also be converted to Kotlin.

westnordost:
Five cents from a developer using Kotlin: Migrating to Kotlin coroutines is something that should not be done in isolation only to replace `AsyncTask`s but in concert with generally changing also the API to use suspending functions rather than callback-based functions. I am not sure how suspending APIs translate to Java code on the surface, though. _Maybe_, from the point of view of Java code, suspending functions look like functions with a callback parameter?

With Kotlin's suspending functions, the API to initialize the map could look like this (as if it were blocking calls, but they aren't), instead of juggling with callbacks:
```kotlin
suspend fun initMap() {
  val map = mapView.getMap()
  val style = map.setStyle("mapstyle.json")
  style.addImage("someImg", getDrawable(R.drawable.someImg))
  // ...
}
```

In fact, what I did was to write extension functions to wrap the calls for my use into a suspending API to profit of this convenience myself. E.g.
```kotlin

suspend fun MapView.awaitGetMap(): MapLibreMap = suspendCoroutine { cont ->
    getMapAsync { cont.resume(it) }
}
```
To make a blocking API that has no callback suspending, one would e.g.

```kotlin

suspend fun Style.awaitAddImage(name: String, bitmap: Bitmap) { 
    withContext(IO) { addImage(name, bitmap) }
}
```

tarkvara:
I believe all these _AsyncTask_ calls were removed as part of issue 1607.  I do see a couple of _AsyncTask_ calls, but only in Java code, not Kotlin… not sure where that Java code is actually used.

louwers:
This bounty can be paid out, PR is ready but we should hold off merging for a bit because it is a breaking change.

PR #2695, opened 2024-08-06
#2436 Replacing AsyncTask with coroutines

tarkvara:


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/LocalRequestTask.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.maplibre.android.MapLibre
+import org.maplibre.android.MapStrictMode
+import org.maplibre.android.log.Logger
+import java.io.IOException
+import java.io.InputStream
+
+internal class LocalRequestTask(private val scope: CoroutineScope, private val onCompletion: ((ByteArray?) -> Unit)?) {
+
+	fun execute(url: String) {
+		scope.launch(Dispatchers.IO) {
```
westnordost, line 17: (This comment applies to other Tasks, too)

This task has no state and is internal to MapLibre, i.e. potentially does not interface with Java code. The idiomatic Kotlin way for this would look like this, i.e. not a class but a suspending function)

```suggestion
suspend fun requestLocalUrl(url: String): ByteArray? = withContext(IO) {
   loadFile(...)
}
```

The current approach causes a leak of the coroutine scope because the `LocalRequestTask` is not garbage collected as long as the owner of `LocalRequestTask` is not. No idea how serious it is, my guess would be: not at all, but 🤷‍♂️ - it is quite similar to the problem with `AsyncTask`.

Now, it seems like (currently) it does interface with Java code (`NativeHttpRequest.java`). Actually, not sure how suspending functions look like from Java-code, but according to [stackoverflow answers](https://www.google.de/search?q=suspend+functions+in+java+code+site:stackoverflow.com&sca_esv=1c866aa5bb9857de&sa=X&ved=2ahUKEwjfmfm5-eKHAxVEwAIHHUW7F40QrQIoBHoECCMQBQ&biw=2048&bih=999&dpr=1.25), one can wrap it quite easily in a Java `CompletableFuture`.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/BitmapImageConversionTask.kt
```diff
@@ -0,0 +1,31 @@
...
+	private val job: Job = SupervisorJob(null)
+	private val scope = CoroutineScope(Dispatchers.IO + job)
+
+	fun execute(vararg params: ImageWrapper) {
+		scope.launch {
+			val images = params.map { Style.toImage(it) }
+			val nativeMap = nativeMap.get()
+			if (nativeMap != null && !nativeMap.isDestroyed) {
+				nativeMap.addImages(images.toTypedArray())
+			}
```
westnordost, line 24: Previously, this was done in the UI thread, now it is done in the IO thread. Is this fine?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/BitmapImageConversionTask.kt
```diff
@@ -0,0 +1,31 @@
...
+		}
+	}
+
+	fun cancel() {
+		job.cancel()
+	}
```
westnordost, line 30: `launch` returns a `Job`. You could also return it in `execute`.

In any case, (separate issue), here you cancel the `SupervisorJob`, i.e. the parent job. After the `SupervisorJob` is cancelled, IIRC no new jobs can be launched in that job. IIRC subsequent calls to `execute` would then just immediately cancel.
That would be fine if the `BitmapImageConversionTask` is a single-use object, but then, why does it need a `SupervisorJob`? In any case, needs documentation.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileDirsPathsTask.kt
```diff
@@ -0,0 +1,25 @@
...
+	fun execute(context: Context) {
+		CoroutineScope(Dispatchers.IO).launch {
+			try {
+				val path1 = FileSource.getCachePath(context)
+				val path2 = context.cacheDir.absolutePath
+				withContext(Dispatchers.Main) {
+					onCompletion(path1, path2)
+				}
+			} finally {
+				FileSource.unlockPathLoaders()
```
westnordost, line 21: Previously, this line was executed on the main thread. Now, it is executed on an IO thread (`AsyncTask::onCancelled` is called on the UI thread). Is this fine?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileDirsPathsTask.kt
```diff
@@ -0,0 +1,25 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class FileDirsPathsTask(
+	private val onCompletion: (resourcesCachePath: String, internalCachePath: String) -> Unit
+) {
+	fun execute(context: Context) {
+		CoroutineScope(Dispatchers.IO).launch {
```
westnordost, line 13: Hmm, this reads a bit odd. I have to look up what it does... this is a global coroutine scope, right?

I.e. it runs app-wide, independent of any e.g. view lifecycle. Is this what you want to do? Because, like other Tasks, this one has a callback. So, the callback is fired no matter what, even if the listener should have been removed already (out of lifecycle).

(As for the other tasks, the Kotlin-idiomatic way would be to have this as a suspend function, i.e.
```kotlin
suspend fun getResourcesCachePath(context: Context): String = 
    withContext(IO) { FileSource.getCachePath(context) }
```
(etc)

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+	private const val TAG = "Mbgl-FileUtils"
+
+	/**
+	 * Deletes a file asynchronously in a separate thread.
+	 *
+	 * @param path the path of the file that should be deleted
+	 */
+	fun deleteFile(path: String) {
+		// Delete the file in a separate thread to avoid affecting the UI
+		Thread {
```
westnordost, line 20: Inconsistency. You should use a coroutine with IO dispatcher.

westnordost (commented):
I read on Slack that extra eyes from Android (Kotlin) developers are appreciated, so here is my review. I think much of it could be written much shorter if written in a Kotlin-idiomatic way.

However, I don't know the big picture, so please take that into account.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/BitmapImageConversionTask.kt
artakka: Most probably should stay UI

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/LocalRequestTask.kt
tarkvara: If did look at the same StackOverflow article, and came to the conclusion that you could **not** easily do this.  The most promising approach required 1) an extra library and 2) SDK version 24+.  I didn't want to commit a change which did that.

In any event, this code is only called from _NativeHttpRequest.java_, which is never used anywhere (hence the _@Keep_ annotation).

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/BitmapImageConversionTask.kt
tarkvara: Resolved.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/BitmapImageConversionTask.kt
tarkvara: The `execute` call could potentially be launched several times.  The intent was that the `SupervisorJob` would cancel all of them.  Note that `BitmapImageConversionTask.cancel` is only called when the `Style` is cleared.

I'll add a comment to clarify that.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileDirsPathsTask.kt
tarkvara: I agree.  If we weren't trying to support SDK 21, the suspend function approach would be much more elegant.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileDirsPathsTask.kt
tarkvara: It shouldn't matter which thread this gets called on.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
tarkvara: Missed this because I was only dealing with `AsyncTask` calls.  I'll fix it to be consistent.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/LocalRequestTask.kt
westnordost: Try this. Doesn't it work? It is included in kotlinx-coroutines-core. 

https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.future/

I think that `org.jetbrains.kotlinx:kotlinx-coroutines-jdk8` library you probably meant is obsolete or something, at least the current version of it is completely empty.

tarkvara: Let me give that a try.  I'm not having any success getting that to compile.

Are you sure it's supported back to SDK 21?

westnordost: Hm no, on a second look, it appears that `CompletableFuture` in Android is only supported starting with SDK 24, after all:

https://developer.android.com/reference/java/util/concurrent/CompletableFuture

So it is not an issue with the coroutines library, but with Android.

westnordost: So, it looks like there is no easy way to interface suspending code to Android Java API 21+. 

Only when the code that uses this has also been migrated to Kotlin it could be made more Kotlin-idiomatic.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
```diff
@@ -105,7 +109,7 @@ public static synchronized FileSource getInstance(@NonNull Context context) {
    * @return the files directory path
    */
   @NonNull
-  private static String getCachePath(@NonNull Context context) {
```
louwers, line 108: Is this change on purpose?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
```diff
@@ -363,7 +327,7 @@ private static void lockPathLoaders() {
     resourcesCachePathLoaderLock.lock();
   }
 
-  private static void unlockPathLoaders() {
```
louwers, line 366: Does this need to be public?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
```diff
@@ -0,0 +1,63 @@
...
+					Logger.e(TAG, "Failed to delete file: $path")
+				}
+			}
+		} catch (exception: Exception) {
+			Logger.e(TAG, "Failed to delete file: ", exception)
+		}
+	}
+}
+
+fun checkFileReadPermissionAsync(file: File, scope: CoroutineScope, onCompletion: (Boolean, Throwable?) -> Unit) {
```
louwers, line 35: Nit: some documentation comments would be good on these.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
```diff
@@ -0,0 +1,63 @@
...
+	scope.launch(Dispatchers.IO) {
```
louwers, line 36: Can we remove this function since we don't seem to be using it?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/LocalRequest.kt
```diff
@@ -0,0 +1,49 @@
+package org.maplibre.android.http
+
+import android.content.res.AssetManager
+import org.maplibre.android.MapLibre
+import org.maplibre.android.MapStrictMode
+import org.maplibre.android.log.Logger
+import java.io.IOException
+import java.io.InputStream
+
+fun localRequest(url: String, onCompletion: ((ByteArray?) -> Unit)?) {
```
louwers, line 10: Does this need to be public?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileDirsPathsAsync.kt
```diff
@@ -0,0 +1,26 @@
+package org.maplibre.android.storage
+
+import android.content.Context
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+fun fileDirsPathsAsync(
```
louwers, line 9: Does this need to be public?

louwers (approved):
Looks good to me, but would be great if @artakka or @westnordost also could have another look.

I think we are changing public APIs as part of this change, so we'd need to bump a major version when releasing this.

If I understand it correctly, Kotlin is by default `public`, so we should be careful not to make internal helper functions part of our public API.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/LocalRequest.kt
tarkvara: No.  It's only ever used in the same package.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
tarkvara: No.  I'll restore it to the previous access.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.maplibre.android.log.Logger
+import java.io.File
+
+private const val TAG = "Mbgl-FileUtils"
+
```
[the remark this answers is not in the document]
tarkvara: Not that I can tell.  The `TAG` is only ever used for log statements within this file.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
tarkvara: Done.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileDirsPathsAsync.kt
tarkvara: Again, it's internal to the package.

louwers: I think it is public now (as that is the default) and should be marked `internal`?

tarkvara: Fixed.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/LocalRequest.kt
tarkvara: Sorry, I misunderstood your question.  Fixed.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/utils/FileUtils.kt
tarkvara: Added.

louwers:
One test is failing:

INSTRUMENTATION_STATUS: stack=java.lang.AssertionError: First attempt should succeed.
	at org.junit.Assert.fail(Assert.java:89)
	at org.maplibre.android.testapp.storage.FileSourceStandaloneTest$overridePathChangeCallTest$1$1.onError(FileSourceStandaloneTest.kt:81)
	at org.maplibre.android.storage.FileSource$1.onError(FileSource.java:313)
	at org.maplibre.android.storage.FileSource.setResourceCachePath(Native Method)
	at org.maplibre.android.storage.FileSource.internalSetResourcesCachePath(FileSource.java:296)
	at org.maplibre.android.storage.FileSource.lambda$setResourcesCachePath$1(FileSource.java:279)
	at org.maplibre.android.storage.FileSource$$ExternalSyntheticLambda1.invoke(D8$$SyntheticClass:0)
	at org.maplibre.android.utils.FileUtilsKt$checkFileWritePermissionAsync$1$1.invokeSuspend(FileUtils.kt:47)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
	at android.os.Handler.handleCallback(Handler.java:942)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7898)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@390829a, Dispatchers.IO]

INSTRUMENTATION_STATUS: stream=
Process crashed while executing overridePathChangeCallTest(org.maplibre.android.testapp.storage.FileSourceStandaloneTest):
java.lang.AssertionError: First attempt should succeed.
	at org.junit.Assert.fail(Assert.java:89)
	at org.maplibre.android.testapp.storage.FileSourceStandaloneTest$overridePathChangeCallTest$1$1.onError(FileSourceStandaloneTest.kt:81)
	at org.maplibre.android.storage.FileSource$1.onError(FileSource.java:313)
	at org.maplibre.android.storage.FileSource.setResourceCachePath(Native Method)
	at org.maplibre.android.storage.FileSource.internalSetResourcesCachePath(FileSource.java:296)
	at org.maplibre.android.storage.FileSource.lambda$setResourcesCachePath$1(FileSource.java:279)
	at org.maplibre.android.storage.FileSource$$ExternalSyntheticLambda1.invoke(D8$$SyntheticClass:0)
	at org.maplibre.android.utils.FileUtilsKt$checkFileWritePermissionAsync$1$1.invokeSuspend(FileUtils.kt:47)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
	at android.os.Handler.handleCallback(Handler.java:942)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7898)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@390829a, Dispatchers.IO]

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
```diff
@@ -197,31 +201,11 @@ public static void initializeFileDirsPaths(Context context) {
     ThreadUtils.checkThread(TAG);
     lockPathLoaders();
     if (resourcesCachePath == null || internalCachePath == null) {
```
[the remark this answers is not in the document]
westnordost: Anyway, I created a PR: #2759

louwers:
@tarkvara Do you still have this code? Could you open another PR?

PR #2700, opened 2024-08-07
Release MapLibre Native iOS 6.5.4

louwers:
Adds a script

```
scripts/generate-changelog.mjs ios
```

that makes my life a bit easier.

Merging this will release MapLibre Native iOS 6.5.4 which is a simple patch release with a bugfix.


cgalvan (approved):
Nice!

PR #2703, opened 2024-08-07
Move UBO updates from render layers to tweakers

TimSylvester:
This fits better with the long-term goal of having drawable updates only done by tweakers, centralizes the UBO updates rather than having them split across two areas, and will simplify the changes needed in the layer update methods for #2719.


louwers (changes requested):
Could you provide some more context?

src/mbgl/gfx/drawable.cpp
```diff
@@ -6,13 +6,20 @@
...
 
 namespace mbgl {
 namespace gfx {
 
 struct Drawable::Impl {
     gfx::ColorMode colorMode = gfx::ColorMode::disabled();
     gfx::CullFaceMode cullFaceMode = gfx::CullFaceMode::disabled();
+
+    std::shared_ptr<Bucket> bucket;
+    PaintPropertyBindersBase* binders = nullptr; // owned by `bucket`
```
louwers, line 19: Why not get it from the bucket when needed?

src/mbgl/gfx/drawable.cpp
```diff
@@ -6,13 +6,20 @@
...
+
+    Immutable<std::vector<RenderTile>> renderTiles = makeMutable<std::vector<RenderTile>>();
+    const RenderTile* renderTile = nullptr; // owned by `renderTiles`
```
louwers, line 22: Same here, why not get it from `renderTiles` when needed?

src/mbgl/gfx/drawable.cpp
TimSylvester: Just to avoid lookups that we've already done.  The paint property binder has to be found in a map by layer ID for each tile, and the render tile would have to be found in the collection by tile ID each time.

These lookups are pretty cheap individually, but get done a huge number of times, so I think they're worth eliminating.  Ideally, the tweakers execute many times for each render layer update.

src/mbgl/renderer/layers/render_line_layer.cpp
```diff
@@ -470,137 +473,32 @@ void RenderLineLayer::update(gfx::ShaderRegistry& shaders,
...
-            }
-
-            return *linePatternTilePropertiesUBO;
-        };
-
         auto updateExisting = [&](gfx::Drawable& drawable) {
             if (drawable.getLayerTweaker() != layerTweaker) {
                 // This drawable was produced on a previous style/bucket, and should not be updated.
                 return false;
             }
```
alexcristici, line 480: Probably we could remove the `updateExisting` lambda.

alexcristici (approved):
Nice!

sjg-wdw:
Are we done on this one?

PR #2705, opened 2024-08-07
Fix accidental regression conditional layer evaluation

TimSylvester:
Conditional layer evaluation was accidentally removed due to a merge conflict in [2202](https://github.com/maplibre/maplibre-native/commit/33a1e162bc9c3d2f5bfe8aad11cd39bc7ec19baa#diff-6555ec058a52c65059cbfca1b8dfaa09dfd87df52ca7010e2201fbf14dc6c8c1).


louwers:
Is it possible to add a regression test for this?

TimSylvester:
> Is it possible to add a regression test for this?

I don't see any straightforward way.  If we tracked the number of property evaluations in the stats, that could be made part of the probe check, but that's no simple matter.

mwilsnd (approved):

PR #2708, opened 2024-08-08
Do not use custom shell node-ci

louwers:


cgalvan (approved):

ntadej (approved):

acalcutt:
Fine with me also. I am not aware of a reason the shell was changed from the default and it seems to be working without it.

acalcutt (approved):

PR #2719, opened 2024-08-13
Tile diff

TimSylvester:
Only calculate which tiles were added and removed once per frame instead of separately in each layer update.


TimSylvester:
Well, after all that, it doesn't really help at all, and in fact seems to be just a bit slower.

In addition, I thought this was going to simplify things but, because layers can not be updated every frame, the complications involved in re-synchronizing offset the benefits.  It does potentially allow the large layer update methods to be split up into three parts, but they remain dependent (updates can find that a remove/add is necessary).

Finally, I think this would actually make it more difficult to transition the layer updates to a background thread.

So I'm leaning towards dropping this one.

image

image

PR #2720, opened 2024-08-14
Add `zoomLevel` options for `flyTo` camera

huyparody:
Add zoom-level options for `flyTo` function, for some use cases like flying from somewhere on the screen to the current user location, with a specific zoom level.

### DIFF
--- a/platform/ios/src/MLNMapView.h
+++ b/platform/ios/src/MLNMapView.h
@@ -1241,6 +1241,20 @@ MLN_EXPORT
 
 /**
  Moves the viewpoint to a different location using a transition animation that
+ evokes powered flight and a default duration based on the length of the flight
+ path.
+
+ The transition animation seamlessly incorporates zooming and panning to help
+ the user find his or her bearings even after traversing a great distance.
+
+ @param camera The new viewpoint.
+ @param zoomLevel The new zoom level for the map.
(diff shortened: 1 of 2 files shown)


louwers (changes requested):
The camera already includes an altitude. This conflicts with also allowing the user to provide a zoom level. You can use https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnaltitudeforzoomlevel to determine the altitude for the camera if you have a certain zoom level.

Please write a design proposal if the current API does not suit your needs. We should not make changes to existing public APIs.

huyparody:
> The camera already includes an altitude. This conflicts with also allowing the user to provide a zoom level. You can use https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnaltitudeforzoomlevel to determine the altitude for the camera if you have a certain zoom level.
> 
> Please write a design proposal if the current API does not suit your needs. We should not make changes to existing public APIs.

Then how do I `flyTo` a position and zoom simultaneously? I need to set the camera to follow the user's location at zoom level 15. I tried `setCamera` completion handler but the animation looks very weird, as it follows the user location to zoom level 21 (my max zoom level) and then it sets the zoom level to 15.

PR #2722, opened 2024-08-14
Scheduling optimizations and tracing

TimSylvester:
Don't hold the mailbox queue lock while calling `schedule`.
Eliminate a second key lookup when inserting a new item in the tagged queue map.

### DIFF
--- a/src/mbgl/actor/mailbox.cpp
+++ b/src/mbgl/actor/mailbox.cpp
@@ -98,7 +98,7 @@ void Mailbox::push(std::unique_ptr<Message> message) {
         return;
     }
 
-    bool wasEmpty;
+    bool wasEmpty = false;
     {
         MLN_TRACE_ZONE(queue lock)
         std::lock_guard<std::mutex> queueLock(queueMutex);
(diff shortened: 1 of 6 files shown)


mwilsnd (approved):

src/mbgl/actor/mailbox.cpp
```diff
@@ -88,19 +90,28 @@ void Mailbox::push(std::unique_ptr<Message> message) {
...
         return;
     }
 
-    std::lock_guard<std::mutex> queueLock(queueMutex);
-    bool wasEmpty = queue.empty();
-    queue.push(std::move(message));
-    auto guard = weakScheduler.lock();
-    if (wasEmpty && weakScheduler) {
-        weakScheduler->schedule(schedulerTag, makeClosure(shared_from_this()));
+    bool wasEmpty;
```
louwers, line 101: Should be initialized to something.

TimSylvester: Even though it's unconditionally set?  That will generate "value is never used" warnings on some compilers.

louwers: C++ Code Guidelines say yes. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es20-always-initialize-an-object

PR #2723, opened 2024-08-14
Toggle tile cache final API

alexcristici:


src/mbgl/renderer/render_orchestrator.hpp
```diff
@@ -93,6 +93,7 @@ class RenderOrchestrator final : public GlyphManagerObserver, public ImageManage
                             const std::optional<std::string>& stateKey);
 
     void setTileCacheEnabled(bool);
+    bool getTileCacheEnabled();
```
louwers, line 96: Can be const.

louwers (commented):
Looks good, but we should probably have some kind of test coverage for this.

louwers (changes requested):
Please add a simple test to the C++ unit tests (or modify an existing one) that just toggles it and renders something so we have some coverage. Ideally also for Android and iOS.

TimSylvester (approved):

louwers (approved):
Thanks for the smoke tests.

A test that tries to render without the cache enabled would be good at some point.

Issue #2715, opened 2024-08-12
Node macos metal implementation


acalcutt:
This has been a know issue for a while, but I wanted to make an Issue for it can be tracked.

**Is your feature request related to a problem? Please describe.**
Right now we are building node releases on the old opengl-2 code so we can continue to support macos. 

**Describe the solution you'd like**
I would like to see us be able to build the node lib files on macos, either by finding a way to build them in bazel, or restoring macos.cmake and getting it to work with the current release

**Describe alternatives you've considered**
I have been porting important changes to the 'opengl-2' branch were the code to build on macos still works, but porting things back and forth is a lot of work.

**Additional context**
Before the metal changes, there was a macos.cmake file that was used on macos
https://github.com/maplibre/maplibre-native/blob/opengl-2/CMakeLists.txt#L1136-L1137
https://github.com/maplibre/maplibre-native/blob/opengl-2/platform/macos/macos.cmake
This file and reference has been removed in the current release, so the cmake build just gets on error on macos
In macos.cmake is where the node lib files get built.
https://github.com/maplibre/maplibre-native/blob/opengl-2/platform/macos/macos.cmake#L105-L107
https://github.com/maplibre/maplibre-native/blob/opengl-2/platform/node/CMakeLists.txt

That code also uses
https://github.com/maplibre/maplibre-native/blob/opengl-2/platform/node/cmake/module.cmake
which is our version of https://github.com/mapbox/cmake-node-module , which actually builds the node lib files

I have tried a bit to get the cmake build working again by by putting back the missing cmake files and trying to get that build to work, but I am hitting errors now that I don't know how to fix. I am working on that here
https://github.com/maplibre/maplibre-native/pull/1793

Other than adding back the cmake build, I am wondering if something similar to glfw metal changes could be done here. I think the biggest challenge would be replacing the [module.cmake](https://github.com/maplibre/maplibre-native/blob/opengl-2/platform/node/cmake/module.cmake) file, which downloads the headers for and builds for a range of node versions.

louwers:
I built on your PR and I got a local Node.js Metal build with CMake on macOS that seems to work.

https://github.com/maplibre/maplibre-native/pull/2729

PR #2729, opened 2024-08-16
Add back CMake build config for macOS

louwers:
This builds on #2715 with some more fixes to the CMake config for macOS.

Reinstating these CMake build files is the easiest way to get Node.js working with Metal (and probably Qt too).


.github/workflows/node-ci.yml
```diff
@@ -78,12 +78,12 @@ jobs:
...
             arch: arm64
-          # Disabled until Metal backend is complete
-          # A release for macOS can be made from the opengl-2 branch
           # - runs-on: macos-12
           #   arch: x86_64
-          # - runs-on: macos-13-xlarge
+          # - runs-on: macos-13-large
           #   arch: arm64
+          - runs-on: macos-14
+            arch: arm64
```
acalcutt, line 86: Since this works can we also try 
```
          - runs-on: macos-14-large
            arch: x86_64
```

louwers: Oh neat. Yes let's try it.

acalcutt:
Seems like it compiles on x86_64, but the tests get this error

```
-[MTLTextureDescriptorInternal validateWithDevice:]:1357: failed assertion `Texture Descriptor Validation
MTLTextureDescriptor: Depth, Stencil, DepthStencil textures cannot be allocated with MTLStorageModeShared or MTLStorageModeManaged on this device.
```

Do you think this will only work with arm architecture?

louwers:
@acalcutt Should work, but GitHub might run some kind of emulation and not have a Metal enabled GPU.

louwers:
pre-commit.ci run

acalcutt (approved):

PR #2731, opened 2024-08-17
Improved compatiblity for NSPredicate predicateWithMLNJSONObject. (#2686)

### LINKED ISSUE #2686: iOS predicateWithMLNJSONObject more tolerant

Some filters that are tolerated by maplibre web and native drawing does not work with `NSPredicate+MLNPrivateAdditions predicateWithMLNJSONObject`. 

Here are some examples.
* `["!has", "attributeName"]` need to be re-written as `["!", ["has", "attributeName"]]` to work with `predicateWithMLNJSONObject`.
* `["!in", "attributeName", "A", "B"]` need to be re-written as `["!", ["in", ["get", "attributeName"], ["literal", ["A", "B"]]]]` to work with `predicateWithMLNJSONObject`.

It would be helpful if `NSPredicate+MLNPrivateAdditions predicateWithMLNJSONObject` would be just as tolarant as the filter handling used for drawing.

* One alternative is to re-write styles to the stricter filter syntax. 
* Another alternative is to create a more tolerante filter parser in the app outside of maplibre.

halset:
* Support `!has` and `!in`.
* Support `["in", "attributeName", "val"..]`.

Closes #2686

### DIFF
--- a/platform/darwin/src/NSPredicate+MLNAdditions.mm
+++ b/platform/darwin/src/NSPredicate+MLNAdditions.mm
@@ -177,7 +177,7 @@ static NSDictionary * const MLNPredicateOperatorTypesByJSONOperator = @{
         NSString *key = [objects objectAtIndex:1];
         return [NSPredicate predicateWithFormat:@"(%K!=nil)", key];
     }
-    // Handle ["!has", "attributeName"]. Also handled by expression using mgl_does:have, but this is simpler..
+    // Handle ["!has", "attributeName"].
     if ([op isEqualToString:@"!has"] && objects.count == 2 && [[objects objectAtIndex:1] isKindOfClass:[NSString class]]) {
         NSString *key = [objects objectAtIndex:1];
         return [NSPredicate predicateWithFormat:@"(%K==nil)", key];
--- a/platform/darwin/src/NSPredicate+MLNAdditions.mm
+++ b/platform/darwin/src/NSPredicate+MLNAdditions.mm
@@ -73,7 +73,7 @@ static NSDictionary * const MLNPredicateOperatorTypesByJSONOperator = @{
     if (operatorTypeNumber) {
         NSPredicateOperatorType operatorType = (NSPredicateOperatorType)[operatorTypeNumber unsignedIntegerValue];
         
-        // Handle ["in", "attributeName", "val"..] -> ["in", ["get", "attributeName"], ["literal", "val"..]]
+        // Handle ["in", "attributeName", "val"..] -> ["in", ["get", "attributeName"], ["literal", ["val"..]]]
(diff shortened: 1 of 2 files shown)


halset:
Need to work more on `!has` and add a test.

halset:
I have added code and tests for `has` and `!has` for `NSPredicate` without using `mgl_does:have`. I am unsure if this is the correct way to do it.

platform/darwin/src/NSPredicate+MLNAdditions.mm
```diff
@@ -73,6 +73,15 @@ + (instancetype)predicateWithMLNJSONObject:(id)object {
     if (operatorTypeNumber) {
         NSPredicateOperatorType operatorType = (NSPredicateOperatorType)[operatorTypeNumber unsignedIntegerValue];
         
+        // Handle ["in", "attributeName", "val"..] -> ["in", ["get", "attributeName"], ["literal", ["val"..]]]
```
louwers, line 76: This is not consistent with the MapLibre Style Spec. https://maplibre.org/maplibre-style-spec/expressions/#in

louwers (changes requested):
The JSON syntax should match that the [MapLibre Style Spec](https://maplibre.org/maplibre-style-spec/expressions/#expressions).

Only things that exist in the MapLibre Style Spec can be accepted.

platform/darwin/src/NSPredicate+MLNAdditions.mm
halset: Good point. Looks like there is a practice that I have been using and that are used in some examples that are according to the documentation.

Here are some examples from metrics/integration/styles/chicago.json of this practice. Not according to documentation, but works in .js and native, but not in `predicateWithMLNJSONObject`.
https://github.com/maplibre/maplibre-native/blob/c3a2fa247cd7dd15742919c537152d649fa9e31d/metrics/integration/styles/chicago.json#L68-L73
https://github.com/maplibre/maplibre-native/blob/c3a2fa247cd7dd15742919c537152d649fa9e31d/metrics/integration/styles/chicago.json#L128-L139

halset:
> Thanks for your contribution.
> 
> The JSON syntax should match that of the [MapLibre Style Spec](https://maplibre.org/maplibre-style-spec/expressions/#expressions). I think all the existing conversions do, but please correct me if I am wrong.
> 
> Only things that exist in the MapLibre Style Spec can be accepted.

Okay. Looks like the problem is that some of my styles use the old style filter expression. They still work when showing the map both in maplibre-js and maplibre-native, but not when parsing a filter as a predicate. 

Here are some examples:
https://github.com/maplibre/maplibre-native/blob/c3a2fa247cd7dd15742919c537152d649fa9e31d/metrics/integration/styles/chicago.json#L68-L73
https://github.com/maplibre/maplibre-native/blob/c3a2fa247cd7dd15742919c537152d649fa9e31d/metrics/integration/styles/chicago.json#L128-L139

It was confusing to me that filters that worked for drawing did not work with predicate. I will need to handle this locally in my apps. Either by updating the filters in my styles to the new expression syntax or create a more tolerant filter parser outside of maplibre-native.

PR #2739, opened 2024-08-20
Use ESM in Node.js code generation scripts

louwers:
- Updates Bazel dependency rules_nodejs and rules_js.
- Use Node 20 when Bazel runs Node.js (default is Node 16).
- Use ES modules with Node.js code generation scripts.


scripts/generate-style-code.js
```diff
@@ -98,7 +94,7 @@ global.evaluatedType = function (property) {
     } else {
       return `std::vector<${evaluatedType({type: property.value, name: property.name})}>`;
     }
-  default: throw new Error(`unknown type for ${property.name}`)
+  default: throw new Error(`unknown type for ${property.name} ${JSON.stringify(property)}`)
```
louwers, line 97:
```suggestion
  default: throw new Error(`unknown type for ${property.name}`)
```

cgalvan (approved):

PR #2743, opened 2024-08-21
Fix update time not being set when only drawable indexes are set

### LINKED ISSUE #2740: Failing assert iOS C++ Unit Tests
When running the iOS C++ Unit Tests from main, this assert fails in `Map.PrefetchTiles`.

https://github.com/maplibre/maplibre-native/blob/a4258ccbb8550b9dea9eac62680c4c83fc848502/src/mbgl/mtl/drawable.cpp#L194-L196

This line from the test is being executed.

```cpp
checkTilesForZoom(10, {11, 11, 11, 11, 11, 11, 11, 11, 11});
```

TimSylvester:
When only a drawable's indexes are updated but not the vertex attributes, the update time wasn't set, and they were seen as out-of-date later on.

For OpenGL, changes to indexes didn't trigger an update at all, so that may have been leading to an un-detected bug.  But the criteria for updates were also much too broad, so it was updating a lot when not really needed.  I split up the check for building the indexes from the check for building the attributes, and saved the attribute bindings with the drawable so that they don't need to be re-built (by rebuilding the attributes) every time.

Resolves #2740


louwers (approved):
Would be good if @alexcristici or @mwilsnd can have a quick look as well.

Do you know why the failing assert was not triggered on CI?

TimSylvester:
> Do you know why the failing assert was not triggered on CI?

No, I can't explain that.  Are you sure that asserts are enabled in the build that's running?

alexcristici (approved):

PR #2744, opened 2024-08-21
Add guard blocks and checks to `SymbolInstance`

TimSylvester:
Inserts 64-bit fixed-value guard blocks between each member in `SymbolInstance`, and checks to regularly validate that these values have not changed.

A certain customer is experiencing crashes in production which nobody has been able to reproduce.  The call stacks of these crashes have repeatedly pointed toward `SymbolInstance` values being corrupted, and the results of these checks confirm that that is occurring and reduces their measured crash rate noticeably, but has not led to any specific problems being discovered.

The failures don't occur at any specific point along the timeline of things being added to or queried in the index.  There are many examples of the checks passing just before `registerAtCrossTileIndex` and failing after, but other cases where they already fail before that.  The lack of a specific point hints that the changes are happening from another thread, but it's not clear how that's possible.

In `android-10.3.1-pre` every accessor validates the guard blocks.  Based on feedback from production, the failures of these checks are mostly around `CrossTileSymbolLayerIndex::addBucket` and `addLayer`, so I've backed that off and only implemented explicit checks in fewer places.

The checks can be disabled with a preprocessor setting, which should also eliminate any file size overhead of static strings from recording source locations.


include/mbgl/util/logging.hpp
```diff
@@ -8,6 +8,11 @@
 #include <string>
 #include <optional>
 
+#define STRINGIZE(x) STRINGIZE2(x)
+#define STRINGIZE2(x) #x
```
louwers, line 12: Can you try this? https://en.cppreference.com/w/cpp/utility/source_location

src/mbgl/layout/symbol_instance.hpp
```diff
@@ -90,43 +136,149 @@ class SymbolInstance {
...
+    /// Mark this item as failed (due to some external check) so that it cannot be used later
+    void forceFail() const;
+#else
+    bool check(std::string_view = {}) const { return true; }
+    bool checkIndex(const std::optional<std::size_t>&, std::size_t, std::string_view = {}) const { return true; }
+    bool checkIndexes(std::size_t, std::size_t, std::size_t, std::string_view = {}) const { return true; }
+    void forceFail() const {}
+#endif
+
+    const Anchor& getAnchor() const { return anchor; }
```
louwers, line 158: Why are the getters needed now?

src/mbgl/tile/geometry_tile.hpp
```diff
@@ -28,6 +28,8 @@ class TileAtlasTextures;
 
 class GeometryTile : public Tile, public GlyphRequestor, public ImageRequestor {
 public:
+    const std::thread::id renderThreadID = std::this_thread::get_id();
```
louwers, line 31: Nit: this variable name might be accurate now, but it won't be if this class is ever used off the render thread.

louwers (approved):
Looks 'good'! 🙂

src/mbgl/tile/geometry_tile.hpp
TimSylvester: If it's ever created off the render thread, yes.  But in that case I expect it would be removed entirely because the checks against it wouldn't make sense anymore.

sjg-wdw (approved):

src/mbgl/layout/symbol_instance.hpp
louwers: Can you remove the ones that are not used?

None of the getters that return references seem to be used.

TimSylvester: They're used in `symbol_layout.cpp`, see line 855.  I think they're all used somewhere, I removed the one field that wasn't referenced anywhere.

louwers (approved):

TimSylvester:
I was thinking about how to make the guard blocks fail in a test to get the coverage up, but the layout isn't accessible to the API, and I don't see a practical way to do it.  @mwilsnd's hooks might be helpful here, if we have a hook where `GeometryTileWorker` hands over the parsed tile or in placement updating, that passes the bucket through, that would be enough to get at the items and break them.

PR #2747, opened 2024-08-22
[node][main] Release node v6.0.0-pre.0

acalcutt:
This PR creates a node v6.0.0-pre.0 pre-release for testing purposes

- This is the first release with macos metal support, so that needs to be tested. 
- This is also the first release since we have changed the directory structure of the build scripts, so we will see how that goes. I had tested it in my own fork, but it is the first time here.

Does anybody have any more suggestions for the CHANGELOG?


platform/node/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 
 ## main
 
+## 6.0.0-pre.0
+* [Note] This is the first release that is back on the main branch.
+* This is the first release that uses Metal for rendering for MacOS. This is a graphics API from Apple that replaces OpenGL ES on Apple platforms.
+* This is the first release that uses OpenGL ES 3.0 for Windows and Linux.
+* MacOS binary is now built on MacOS-14.
```
louwers, line 8:
```suggestion
* macOS binary is now built on macOS 14.
```

louwers (approved):

PR #2753, opened 2024-08-23
Add semicolon after Tracy preprocessor macros

alexcristici:
Added semicolon inside the Tracy defines.

Tracy defines are called using semicolon and without semicolon, so I had a compilation error. I decided to go with no semicolon approach for calling and fixed redefinition errors by using `{ }`.


louwers (approved):

src/mbgl/renderer/image_manager.cpp
```diff
@@ -201,7 +201,9 @@ std::set<std::string> ImageManager::getAvailableImages() const {
     MLN_TRACE_FUNC()
     std::lock_guard<std::recursive_mutex> readWriteLock(rwLock);
 
-    MLN_TRACE_ZONE(copy)
+    {
+        MLN_TRACE_ZONE(copy)
```
louwers: Don't you need to include the return? Otherwise this zone does not measure anything, does it?

src/mbgl/gl/layer_group_gl.cpp
```diff
@@ -28,8 +28,10 @@ void TileLayerGroupGL::upload(gfx::UploadPass& uploadPass) {
         return;
     }
 
-    MLN_TRACE_FUNC()
-    MLN_ZONE_STR(name)
+    {
+        MLN_TRACE_FUNC()
+        MLN_ZONE_STR(name)
+    }
```
alexcristici: What about this? Seems to be duplicated?

src/mbgl/renderer/image_manager.cpp
alexcristici: I am not sure how Tracy zone works, so why the return needs to be included?

louwers: I think it uses RAII to check how long something is executing. So this will always report 0ms.

alexcristici: Then we cannot use multiple ZoneScoped in the same scope.

src/mbgl/gl/layer_group_gl.cpp
```diff
@@ -21,15 +21,15 @@ TileLayerGroupGL::TileLayerGroupGL(int32_t layerIndex_, std::size_t initialCapac
...
+    MLN_ZONE_STR(name)
 
     if (!enabled) {
         return;
     }
 
-    MLN_TRACE_FUNC()
-    MLN_ZONE_STR(name)
+    // MLN_TRACE_FUNC()
+    // MLN_ZONE_STR(name)
```
alexcristici, line 32: Commented because of this error:
`'___tracy_scoped_zone': redefinition; multiple initialization`

src/mbgl/renderer/image_manager.cpp
```diff
@@ -201,7 +201,7 @@ std::set<std::string> ImageManager::getAvailableImages() const {
     MLN_TRACE_FUNC()
     std::lock_guard<std::recursive_mutex> readWriteLock(rwLock);
 
-    MLN_TRACE_ZONE(copy)
+    // MLN_TRACE_ZONE(copy)
```
alexcristici, line 204: Commented because of this error:
`'___tracy_scoped_zone': redefinition; multiple initialization`

TimSylvester:
Ah, I did this before, but some of them were lost in conflict resolution for #2629.

Is the `___tracy_scoped_zone` compiler-specific?  If it's not going to work on Xcode, it would be nicer to just have the macros evaluate to nothing there.  But since it's a redefinition, they just need to be in a new scope block.

The cases with two `MLN_TRACE_FUNC` are an error, any secondary traces within a method should be `MLN_TRACE_ZONE`.

src/mbgl/actor/mailbox.cpp
```diff
@@ -90,7 +90,7 @@ void Mailbox::push(std::unique_ptr<Message> message) {
         }
     }};
 
-    MLN_TRACE_ZONE(push lock)
+    // MLN_TRACE_ZONE(push lock)
```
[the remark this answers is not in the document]
alasram: Thanks for removing the semicolon from the Macro.
I have a PR in review which fixes the errors: [here](https://github.com/alasram/maplibre-native-experimental/blob/alasram/texture_2d_pool/src/mbgl/actor/mailbox.cpp#L94) 
I also changed `push lock` to `push_lock` so in the future we fix Tracy and make the label part of the variable name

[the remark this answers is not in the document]
alexcristici: Thanks @alasram. I did the changes.

louwers:
@TimSylvester Works fine on macOS too.

src/mbgl/actor/mailbox.cpp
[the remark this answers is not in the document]
alasram: Really thanks for adding the `;`. Not only it's uncommon to have macro calls without the `;` but it also won't play nice with clang-format

alasram:
> Ah, I did this before, but some of them were lost in conflict resolution for #2629.
> 
> Is the `___tracy_scoped_zone` compiler-specific? If it's not going to work on Xcode, it would be nicer to just have the macros evaluate to nothing there. But since it's a redefinition, they just need to be in a new scope block.
> 
> The cases with two `MLN_TRACE_FUNC` are an error, any secondary traces within a method should be `MLN_TRACE_ZONE`.

Not compiler specific. Tracy creates the same variable name for `ZoneScoped` `ZoneScopedN` We need a Tracy-side change to use the label in the variable name generator

src/mbgl/gl/drawable_gl.cpp
```diff
@@ -23,7 +23,7 @@ DrawableGL::~DrawableGL() {
 }
 
 void DrawableGL::draw(PaintParameters& parameters) const {
-    MLN_TRACE_FUNC();
+    MLN_TRACE_FUNC()
```
alasram, line 26: I thought we were doing the opposite. It doesn't feel right to have instructions without the `;` at the end. What if we change the macros to actual functions? Think of the `assert` macro in `assert.h`

alasram: Also clang-format doesn't interpret macros as it's tidying the actual code not pre-processed one

TimSylvester:
> Not compiler specific. Tracy creates the same variable name for `ZoneScoped` `ZoneScopedN` We need a Tracy-side change to use the label in the variable name generator

Why does it work in Android, though?

src/mbgl/gl/drawable_gl.cpp
TimSylvester: I'm not sure if it's coming from clang, but if the semicolons are outside the macro, I get a ton of these in AndroidStudio:

image

alasram: we need to use `(void)(0)` when Tracy is disabled to avoid the empty statement spam

src/mbgl/gl/layer_group_gl.cpp
alasram: `MLN_ZONE_STR` adds a string text to the Zone. From Tracy manual:
```
Use the ZoneText(text, size) macro to add a custom text string that the profiler will display along
with the zone information (for example, name of the file you are opening).
```

src/mbgl/gl/drawable_gl.cpp
alexcristici: Already used that.

alexcristici: @TimSylvester What platform?

TimSylvester: That was Android Studio on MacOS, but while it was evaluating to nothing, so I expect the above will resolve it.

louwers: Seems to work OK now.

PR #2759, opened 2024-08-23
ensure ReentrantLock is unlocked after being locked, and on same thread

### LINKED ISSUE #2432: Crash on Android while in background
**Describe the bug**
I used an app with MapLibre 11.0.0, then did something else, and on next time opening the app about an hour later I noticed it had crashed, probably while not using the phone.
Stack Trace (update: replaced with the strack trace from 11.1.0 as this one shows line numbers):
```
FATAL EXCEPTION: main
Process: de.westnordost.streetcomplete.expert, PID: 677
java.lang.IllegalMonitorStateException
at java.util.concurrent.locks.ReentrantLock$Sync.tryRelease(ReentrantLock.java:156)
at java.util.concurrent.locks.AbstractQueuedSynchronizer.release(AbstractQueuedSynchronizer.java:1291)
at java.util.concurrent.locks.ReentrantLock.unlock(ReentrantLock.java:466)
at org.maplibre.android.storage.FileSource.unlockPathLoaders(FileSource.java:367)
at org.maplibre.android.storage.FileSource.access$100(FileSource.java:31)
at org.maplibre.android.storage.FileSource$FileDirsPathsTask.onPostExecute(FileSource.java:224)
at org.maplibre.android.storage.FileSource$FileDirsPathsTask.onPostExecute(FileSource.java:204)
at android.os.AsyncTask.finish(AsyncTask.java:755)
at android.os.AsyncTask.access$900(AsyncTask.java:192)
at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:772)
at android.os.Handler.dispatchMessage(Handler.java:107)
at android.os.Looper.loop(Looper.java:214)
at android.app.ActivityThread.main(ActivityThread.java:7356)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:491)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:940)
```
There is no app code involved, so I don't know whether the crash might be related to mis-handling MapLibre somehow.

**To Reproduce**
Unclear how to reproduce it. Possibly triggered by some stuff Android does with background apps.
The stack trace might not be particularly useful, it doesn't even contain line numbers. Is there anything I can do to help finding out what is wrong?

 - OS: Android 10 / LineageOS 17.1
 - Platform Android
 - Version 11.0.0

westnordost:
fixes #2432

The reentrant locks were engaged unconditionally on the then-main thread but the async task was only started conditionally. The unlocking of the locks was managed by the async task.
This lead to two issues:

- if the condition checked for in `initializeFileDirsPaths` was already met, the locks were engaged but never unlocked. Any subsequent attempt to lock would just block execution forever.

- (hypothetical, not tested:) the thread id of the app's main thread when the async task finished and thus calls `onPostExecute` or `onCancel` on that thread may differ from the thread id of the app's main thread when `initializeFileDirsPaths` was invoked: E.g. if the app was sent to background and effectively re-started later by the Android system. However, reentrant locks require to be locked/unlocked on the same thread, otherwise they will throw an `IllegalMonitorStateException`. The previous mechanism could not ensure this.

In any case, I am pretty sure either of these two issues is the cause for #2432.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
@@ -203,7 +203,7 @@ public class FileSource {
     @Override
     protected Void doInBackground(Context... contexts) {
       getResourcesCachePath(contexts[0]);
-      getInternalCachePath(context[0]);
+      getInternalCachePath(contexts[0]);
     }
   }
 
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
@@ -204,6 +204,7 @@ public class FileSource {
     protected Void doInBackground(Context... contexts) {
       getResourcesCachePath(contexts[0]);
       getInternalCachePath(contexts[0]);
+      return null;
     }
   }
 
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
@@ -195,33 +195,15 @@ public class FileSource {
   @UiThread
   public static void initializeFileDirsPaths(Context context) {
     ThreadUtils.checkThread(TAG);
-    lockPathLoaders();
-    if (resourcesCachePath == null || internalCachePath == null) {
-      new FileDirsPathsTask().execute(context);
-    }
+    new FileDirsPathsTask().execute(context);
   }
 
-  private static class FileDirsPathsTask extends AsyncTask<Context, Void, String[]> {
+  private static class FileDirsPathsTask extends AsyncTask<Context, Void, Void> {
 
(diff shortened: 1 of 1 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/storage/FileSource.java
```diff
@@ -358,16 +340,6 @@ private static boolean isPathWritable(String path) {
...
-  private static void lockPathLoaders() {
-    internalCachePathLoaderLock.lock();
-    resourcesCachePathLoaderLock.lock();
-  }
-
-  private static void unlockPathLoaders() {
-    resourcesCachePathLoaderLock.unlock();
-    internalCachePathLoaderLock.unlock();
-  }
-
```
westnordost, line 370: If you wonder where the locking/unlocking is done, it's done in `getResourcesCachePath` and `getInternalCachePath`, respectively. No need to duplicate the code from these methods.

louwers (approved):
Looks good! Thanks for digging into this.

We'll have a release out with this soon.

Issue #2760, opened 2024-08-24
Unused index buffers


alasram:
**Describe the bug**
I noticed that many index buffers are created when tiles are loaded but they are never used in an actual draw which means creating and uploading these buffers is a waste.
The reason for the creation of these buffers needs to be investigated and once the cause is understood we need to avoid creating these buffers.

sjg-wdw:
Can you take a look at this one @alexcristici
First question would be how we tell these buffers are being created and then not used.

alasram:
[This commit](https://github.com/alasram/maplibre-native-experimental/commit/f53653418933847f2031f906627121c7a7a2577d) prints to the log index buffers that aren't used in any draw

Issue #2499, opened 2024-06-12
Proposal for pipelining the rendering


alasram:
The [Tracy profiler](https://github.com/wolfpld/tracy) shows MapLibre is mostly GPU bound when rendering a complex style on a high end laptop. An example on i9-11950H and its iGPU:
![i9_11950H_iGPU](https://github.com/maplibre/maplibre-native/assets/5223673/97ec950d-f2ce-4a40-b0d3-550136196e14)

When running the same content on a lower power system, i.e a Qualcomm Adas chip, the profile significantly changes:
![SA8155](https://github.com/maplibre/maplibre-native/assets/5223673/f0ed5926-3e9d-4cc0-ad37-4a39d5585ca4)

MapLibre becomes a bit CPU bound with `GeometryTile` methods such as `setData` (simply copying and passing image data to worker threads) becoming very slow and `createRenderTree` taking milliseconds therefore starving the GPU.

Optimizing geometry and render tree creation would be very beneficial for lower-end hardware but a first simple method to alleviate performance is to pipeline the renderer. Instead of:
![render_tread](https://github.com/maplibre/maplibre-native/assets/5223673/b6f86912-d0e2-4c56-8473-1a8bacf12de6)

The Render thread will be split into multiple threads and pipelined:
![pipeline](https://github.com/maplibre/maplibre-native/assets/5223673/14af8aff-9cd6-49ba-a4bc-9c60e2c81458)

A simple implementation is to have persistent threads `GeometryTile`, `RenderTree` and `Render`. A thread safe producer-consumer FIFO sits between each pipeline stage (the arrow in the above figure). The size of the FIFO controls the depth of the pipeline: compromise between how far behind the rendering is and the memory overhead of a deep queue (Not sure if the existing mailboxing method used in MapLibre can be leveraged for this pipelining purpose).

alasram:
@louwers @sjg-wdw 
Let me know if there are concerns and who should be added to the discussion.
Also I cannot change the assignee so feel free to assign it to me

sjg-wdw:
It's an interesting idea and thanks for putting together a sketch of it.

I'm curious what's taking all the time in createRenderTree().  I'd expect the upload to be a big part of it, but is it just copying memory and such?  We've tried to cut that down a bit in newer versions, though the old buckets were pretty good about sharing it.

[...]

PR #2761, opened 2024-08-24
Free threaded resources on Android

alasram:
When tiles are streamed, GPU resources are uploaded to the GPU synchronously in the render thread. A quick change in Zoom or location can block the render thread due to very slow uploads: Many uploads are serialized delaying actual rendering and affecting the user experience. The following Tracy screenshot shows a slow frame. Most of the frame is processing uploads of vertex buffers to the GPU:
slow_frame

Zoomed version of the same screenshot where we see actual rendering at the very end of the frame:
slow_frame_zoom

[Pipelining the renderer](https://github.com/maplibre/maplibre-native/issues/2499) would alleviate the problem but it won't be enough when many uploads are processed in the same frame. Also, pipelining the renderer requires significant efforts. An easier solution that may scale better is to distribute uploads to as many threads as possible and only wait for them to finish when they are needed in a draw call. This can scale better when many free CPU cores exist.

This PR adds shared EGL contexts to the Android backend to issue buffer uploads in worker threads.
The existing thread pool has been updated to enable persistent EGL contexts (I noticed very poor performance on Qualcomm drivers when migrating contexts between threads hence the thread pool with persistent contexts: no calls to `eglMakeCurrent` every frame and in different threads). This allows concurrent uploads to the GPU:
slow_frame_with_async_upload

The uploads are synchronized with the draws: the render thread only waits for a buffer when a draw needs it. This allow drawing while uploading data:
parallel_draw_and_upload
The uploads are currently randomly scheduled. Parallelizing drawing and uploading can be improved by first scheduling uploads for resources that get used late in the frame, e.g. uploading resources used by translucent objects before opaque objects since  opaques are rendered first.

This PR is specific to Android and is made such as other backends are unaffected and have minimal code change. Once Vulkan is the official backend on Android similar and better handling of resources can be done.

Texture uploads are also slow but less frequent and will be handled in a separate PR:
upload_texture_future_work

There are unused index buffer that we currently wait on before we destroy them and this is slowing down the render thread #2760
unused_index_buffers
We can destroy them in a future frame but we should understand why they are there in the first place.

There are slowdowns in the the render thread that will be investigated separately (mailbox related):
slow_mail_box_push_lock


alasram:
There's a lot Tracy related noise so it might be easier to review the first commit which has the relevant code

src/mbgl/gl/fence.cpp
```diff
@@ -56,14 +59,75 @@ bool Fence::isSignaled() const {
...
+        GLenum fenceStatus = glClientWaitSync(fence, GL_SYNC_FLUSH_COMMANDS_BIT, timeout);
+        switch (fenceStatus) {
+            case GL_ALREADY_SIGNALED:
+                [[fallthrough]];
+            case GL_CONDITION_SATISFIED:
+                return;
+            case GL_TIMEOUT_EXPIRED:
+                Log::Error(
+                    Event::OpenGL,
+                    "glClientWaitSync timeout after " + std::to_string(i * timeout / oneSecInNanoSec) + " seconds");
```
louwers, line 97: CI complains about fallthrough here

sjg-wdw:
Just tagging @mwilsnd and @alexcristici to take a look at this.

alasram:
I will be making this opt-in because the performance is variable depending on vendors:
- Emulator. Shared EGL contexts not well supported (known issue)
- Mali drivers (Google pixel 6 and 7) on Android: uploads are like memcpy so no benefit of using shared contexts
- Adreno drivers on Android: buffer updates are slow compared to a memcpy so this is useful
- Adreno drivers on QNX: buffer upload perf is much lower than a regular Android driver and shared contexts scale the perf (screenshots are on QNX)

louwers (commented):
Is the added complexity worth it if it is opt-in?

Probably a tiny fraction of users has enough OpenGL knowledge and control over their distribution for this to be useful as a feature.

If the performance gains are significant, can we make an informed choice for the user automatically?

alasram:
I changed this to always on by default on emulator and any device if GLES 3 is supported otherwise a GLES 2 context is requested and multi threading is disabled.

PR #2780, opened 2024-08-29
Add textFitWidth and textFitHeight Properties to Sprites

qqz003:
This PR migrates the changes from ["Add textFitWidth and textFitHeight Properties to Sprites"
](https://github.com/maplibre/maplibre-gl-js/pull/4019) to MapLibre Native, adding textFitWidth and textFitHeight properties to sprites. These properties help manage text fitting within shields, addressing issues with small text sizes without needing multiple layers.


metrics/integration/render-tests/icon-text-fit/textFit-grid-long/expected_Alt.png
louwers, line 1: This file can be deleted I think.

metrics/integration/render-tests/icon-text-fit/textFit-grid-long/expected-mac.png
louwers, line 1: This file can be deleted.

louwers:
On iOS there is a failing render test.

[textFit-grid-long.zip](https://github.com/user-attachments/files/16817530/textFit-grid-long.zip)

metrics/integration/render-tests/icon-text-fit/textFit-grid-long/expected-mac.png
qqz003: These two files were migrated from JS. I tried to delete the file "expected-mac.png" and run the render test locally on my MacBook, but it failed, it seems that the render result on Mac of this test is a little different. Therefore, I believe both of them are necessary.
@DavidBuerer could you please help to explain the usage of these two file?

qqz003:
> On iOS there is a failing render test.
> 
> [textFit-grid-long.zip](https://github.com/user-attachments/files/16817530/textFit-grid-long.zip)

qqz003:
Sorry @louwers I closed this PR unintentionally and could you please help to approve the workflow again? 
And as for the failed test "[textFit-grid-long](https://github.com/user-attachments/files/16817530/textFit-grid-long.zip)", it could pass on my Mac, and the online iOS render test has also passed, have you deleted the file "expected-mac.png" on your side?

qqz003:
Hi, @louwers 

Thank you for the comment!

As for the two baseline files the "text-grid-long," I ran and debugged the render test on my Mac and found that they will be used for testing. The test runner uses a regex to load all the expected images under the test case folder and compare them with the "actual.png." If one of their differences is less than the threshold, the test will be regarded as "pass." The logic can be found in the file "render-test/runner.cpp," in the method `checkRenderTestResults`. And I have sent the e-mail to @DavidBuerer, who is the origin author of these two files, and maybe he could help to answer this question.

As for the failed workflow, honestly, I didn't find the root cause yet because it was caused by the failure of the render tests running on Linux, but I tried to run the render tests locally, and they passed. All the failed test cases have a common point: they all use the sprite "stretch," and this PR updated the "stretch.json" to add a new sprite. This may cause the issue, but what is peculiar is that all the newly added test cases that used the sprite in "stretch" in this PR did not fail. To fix this issue, I need more information:
1. How do the online tests load the sprite? From the "cache-style.db" or some remote server? I tried to update the "cache-style.db" but it still failed.
2. Is there a way to obtain the test result file from the online workflow? I'd like to know why they failed.

metrics/integration/render-tests/icon-text-fit/textFit-grid-long/expected-mac.png
DavidBuerer: With the original change, I was surprised by the results being off by one pixel on the various platforms.  I was assured that that sometimes happens and that having variants for each platform was supported and sometimes required.  It's been a while, but as I recall I let the build fail and copied the variants from the visual comparison.

louwers:
Thanks for pointing me to that! I didn't know about `readExpectedImageEntries`.

We only run the render tests under iOS, Linux and Android currently. So we might need some variations of the expected images for these platforms.

qqz003:
@louwers,

We have identified the cause of the render test failure in linux-ci. It aligns with your comment that all the failed test cases are using the "stretch" sprite. Since the size of the "stretch.json" changed, we need to update their network probes as well.

However, I have a question. I noticed that you only added the network probe for Linux. Does this mean that only the Linux build will check the network probe?

Additionally, the device test has failed, and it appears to be running on a remote device. How can I obtain the detailed test report? We would like to understand the reason for the failure.

Thank you!

louwers:
> However, I have a question. I noticed that you only added the network probe for Linux. Does this mean that only the Linux build will check the network probe?

I think so.

> Additionally, the device test has failed, and it appears to be running on a remote device. How can I obtain the detailed test report? We would like to understand the reason for the failure.

You can check the logs on AWS Device Farm: https://github.com/maplibre/maplibre-native/wiki/AWS-Device-Farm-Logs

It's probably easier to run the render tests on a local device if you have one available.

qqz003:
@louwers thanks, but I have run the render test on both my Android and iOS devices, and it passed. So, I'm wondering why it fails remotely. As for the failed case "textFit-grid-long" you shared, honestly, I can't repro it.
Anyway, I'll first check the online render test results to see if there's any clue for these failures. Maybe it's caused by some configuration, I guess. Thank you again for the information!
Thank you again for the information!

louwers:
@qqz003 Are you sure the new render tests were included in the runs you did locally?

qqz003:
@louwers yes, I'm sure, here's my result [textFit-grid-long.zip](https://github.com/user-attachments/files/16882874/textFit-grid-long.zip). 
I noticed that in the result you provided, there is only a single "expected.png". Did you delete the other two files when you run the render test? 
Additionally, could you clarify how you executed the render test - was it in a simulator or on an iPhone? Is this test case the only one failed? And what's the parameter you set when run the bazel? Using drawable or some other flag? (like `bazel run //platform/ios:xcodeproj --@rules_xcodeproj//xcodeproj:extra_common_flags="--//:renderer=drawable"`)

louwers:
You should generate an Xcode project and run the render test on a real device. https://github.com/maplibre/maplibre-native/blob/main/platform/ios/CONTRIBUTING.md#bazel

On CI it also runs on a real device, not in a simulator.

Yes I only kept one of the expected images when sharing it, but they were all slightly different.

qqz003:
@louwers, I haven't found the reason for the failure in the render test yet. I suspect it may be caused by the renderer because the iPhone should use "Metal" while the simulator may use "OpenGL." However, I don't understand why there is a one-pixel difference between different renders. It seems that this test case is not stable enough, which is why @DavidBuerer created more than one baseline for it.
I have debugged this issue locally and I'm sure that the output of the newly added method `applyTextFit` works well for both the simulator and the device. Could we first ignore this test case in this PR? Since it's a newly added one, and it may be caused by the renderer, it shouldn't be a blocking issue for this change.
On the other hand, since I can't obtain the detailed report from AWS devices nor repro it locally, I don't know if the failure of the render test in Android is caused by the same issue. Therefore, I ignored this test case in my latest change to see if the render test will pass. Please help approve the workflow to check the result. Thank you!

random3940:
> @louwers, I haven't found the reason for the failure in the render test yet. I suspect it may be caused by the renderer because the iPhone should use "Metal" while the simulator may use "OpenGL." However, I don't understand why there is a one-pixel difference between different renders. It seems that this test case is not stable enough, which is why @DavidBuerer created more than one baseline for it. I have debugged this issue locally and I'm sure that the output of the newly added method `applyTextFit` works well for both the simulator and the device. Could we first ignore this test case in this PR? Since it's a newly added one, and it may be caused by the renderer, it shouldn't be a blocking issue for this change. On the other hand, since I can't obtain the detailed report from AWS devices nor repro it locally, I don't know if the failure of the render test in Android is caused by the same issue. Therefore, I ignored this test case in my latest change to see if the render test will pass. Please help approve the workflow to check the result. Thank you!

@louwers, could you please approve the workflow? And what do you think about allowing 1px difference here? I was looking at render tests and there are few where JS and native produce different results.

louwers:
The Vulkan render test failing is not a blocker.

random3940:
> The Vulkan render test failing is not a blocker.

Awesome, thank you for the review! Please tell us if there is something else you want to see to get this merged in.

louwers:
@random3940 I will have a closer look today.

For future reference, here is the diff of the binary SQLite database (I used https://www.sqlite.org/sqldiff.html):

```
UPDATE resources SET data=x'1f8b08087ee0c6660003737472657463682e6a736f6e00d593b16ac3301040777f85d06c812da769e8d4a9740b64698bc96012a91218d928077608f9f75e12a5b51aa19a261dba9ddf9df0bb936e9710428d3682b59505fa40760810f51866e929de0ee24eaf41e1f7e40c94d0ef0a86a4d5bda81715e80669eee006ac80957a455496799692225b2efddc5b30b7968865556f04827deae9b2556340983fd0e6b7d376e44bf5587baa3f9e09f405a207a981758ab9bfcc4dbd65ad6ddac6a2a1a9eaffddb32b38f4f9a4e1c599d241b3d42f793ebb536f089fb3935a821026f08aef62b3b9bfc1334e498987c8f4f70f1a9415a105cca3eed30bf77cbc3b47abc90f565d1370e251a7e21aa7c3a4f8d8d57fe4bd2f566431b1d985d86cfc12f0c8f57ecff9d2c93ef9005811ddfc62050000' WHERE id=164;
INSERT INTO resources(id,url,kind,expires,modified,etag,data,compressed,accessed,must_revalidate) VALUES(176,'http://sprites/textFit.json',6,NULL,NULL,NULL,x'1f8b0808f24fbc660003746578744669742e6a736f6e00e593bf0ac23010c6f73e45e9dc822e8a3e80b8093aa8148762a30996a4c4035b4adfdd8b26d0d43fade814b7cb77974bee977c95e7fb019002660ca20b8dce2009ece942aea864fcd45e0753bfc21db8a7c07010dee3b2115f580a14d7c381512861470a288d8d92b38264cb049850855ad4276d508ae349e84f86bb9d9dda9ad4a8914a0fa81e92ec4cb4b2171c0857e7ddbaa846b71d3aad675deb6b06ed01edb2b9b9fb431d96d5e10b783c2bff071c0efb15b45c8a5c481c892799fbd49ad3fee4afb58daa9ec3396a9f9a5441e86b50a781f531e72b584f8de924ad8f4cd9f5b78c21ada6ce21ebeb488b429725ff03d93b4f76e2b24ce936af5eae7c00e6d5de1539fc6f47d40a0000',1,1724416726,0);
INSERT INTO resources(id,url,kind,expires,modified,etag,data,compressed,accessed,must_revalidate) VALUES(177,'http://sprites/textFit.png',5,NULL,NULL,NULL,x'1f8b08086c4ebc660003746578744669742e706e6700eb0cf073e7e592e2626060e0f5f4700902d22940ecc6c10624adda0c338114637190bb13c3ba73322f811c9674475f4706868dfddc7f1259817cce028fc862060631551066f40c52f900d460e5e9e2185211f7f6e246cec30c22472ef84c6f345afcbde896089b76c211d15f1f66c5cf8b7b7fbe30cdddf198488a0f327187ede5bcbdab7beb172a84ffffd39f13292078739382d2a654479794d34d1d27a6b10a4ed4e556528a167371d9a6d3d131f414788b2ca97eb7ceec5fa9b4cabaa61ecd133390081fcb73fa4b6f16a4307a4fa804062d83a7ab9fcb3aa78426009ec7fbb8a5010000',1,1724416726,0);
INSERT INTO resources(id,url,kind,expires,modified,etag,data,compressed,accessed,must_revalidate) VALUES(178,'http://geojson/grid.json',2,NULL,NULL,NULL,x'1f8b08087990c4660003677269642e6a736f6e00bd93310b83301085f7fc8a90594128a5d0b5d0b97b71b07ada40f4427a1d44fcefb589ad421d9bcb9877ef2e1ff73208391d45bd057594ea0c053d1d9cd018284963a792505007e131155d077fb5e59babbd641d5a70a4bd67b178ad0463de36429b1aa8497de571d5a1016c815cffebff8cbda0ee68353434477495ee0a0aaf95e9214be46e2ff3658808a3e47f495a5d5506a2b2309138dddce32e856327611f4c01cb58587822c6c3c213b2c82c3724c296296369ec1f33d3f0a48c8b8627679b34b918c50bd84fe5a462070000',1,1724416726,0);
INSERT INTO resources(id,url,kind,expires,modified,etag,data,compressed,accessed,must_revalidate) VALUES(179,'http://geojson/grid-tall.json',2,NULL,NULL,NULL,x'1f8b08087990c4660003677269642d74616c6c2e6a736f6e00bd93310f82301085f7fe8aa633244c9ab89a38bb1b0684039b14aea9c74008ff5d6c514864b4d7b1efde5dbfdceb28e47c140d16d449aa0b14d43b38a3315092c64e25a1a00ec2732eba8dfe6acfb7547bc93ab4e0487bcf6af15a09c6bc6d8436355093facad3a64303d802b9e1d7ff197b45ddd16668688ee82add15145e2bd36396c84326f3758808a3e47f495a5d5506a2b2309138dd3ce22e856327611f4c01e361e189180f0b4fc822b3dc91085ba68ca5b17fcc42c393322e1a9e9cedd2e462122f6859c7b462070000',1,1724416726,0);
UPDATE sqlite_sequence SET seq=179 WHERE rowid=1;
```

src/mbgl/sprite/sprite_parser.cpp
```diff
@@ -150,6 +159,33 @@ std::optional<style::ImageContent> getContent(const JSValue& value, const char*
     return std::nullopt;
 }
 
+std::optional<style::TextFit> parseTextFit(const std::string_view& value) {
```
louwers, line 162: Tests covering these two functions should be added to `test/sprite/sprite_parser.test.cpp`.

Make sure to include all branches, including the else branch of the log statement.

src/mbgl/text/shaping.cpp
```diff
@@ -92,6 +92,45 @@ PositionedIcon PositionedIcon::shapeIcon(const ImagePosition& image,
...
 }
 
+PositionedIcon PositionedIcon::applyTextFit() const {
+    if (!_image.textFitWidth && !_image.textFitHeight) {
+        return *this;
+    }
+    auto width = _right - _left;
+    auto height = _bottom - _top;
+
+    auto contentWidth = _image.content->right - _image.content->left;
```
louwers, line 102: Please have a look over all definitions in the PR and make sure that all immutable values are marked `const`.

include/mbgl/style/image.hpp
```diff
@@ -14,6 +14,12 @@ namespace style {
 using ImageStretch = std::pair<float, float>;
 using ImageStretches = std::vector<ImageStretch>;
 
+enum class TextFit : uint8_t {
```
louwers, line 17: We follow the C++ Core Guidelines, which state

> [Specify the underlying type of an enumeration only when necessary](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-underlying)

Especially in this case I think it makes sense to just use the default.

metrics/ignores/platform-all.json
```diff
@@ -85,6 +85,7 @@
     "render-tests/fill-pattern/update-feature-state": "https://github.com/mapbox/mapbox-gl-native/issues/15895",
     "render-tests/geojson/inline-linestring-fill": "current behavior is arbitrary",
     "render-tests/icon-text-fit/text-variable-anchor-overlap": "https://github.com/mapbox/mapbox-gl-native/issues/15809",
+    "render-tests/icon-text-fit/textFit-grid-long": "Needs to be investigated and fixed.",
```
louwers, line 88: Maybe we can only ignore it on the platforms where it is failing?

For me it is running fine locally on Android.

If you would prefer to merge this and look into this later, let's create an issue so we don't forget.

louwers (changes requested):
This is an excellent PR! I have a few nits, please have a look at them.

I will make a release for iOS and Android once merged.

src/mbgl/sprite/sprite_parser.cpp
qqz003: I see, will add a new test case that could cover parsing the `TextFit`

metrics/ignores/platform-all.json
qqz003: Oh, it's because that I'm not sure if the online Android render test also failed with this render test, (it also running fine locally on my side), but it seems not, so I'll move this to the `platform-iOS`.

include/mbgl/style/image.hpp
qqz003: Thanks, actually I wrote it this way by referring to the existing code (like the enums in the file: src/mbgl/style/types.cpp), I thought this could save some space because the default type of the enum in C++ is `int`.

louwers: There is something to be said for it, but since the space saving is minimal sticking to the default might be better.

I will leave it up to your judgement!

src/mbgl/sprite/sprite_parser.cpp
louwers: Thanks!

qqz003: Have added the test "SpriteParsingTextFit" to cover the TextFit parsing.

src/mbgl/text/shaping.cpp
qqz003: Done.

include/mbgl/style/image.hpp
qqz003: yeah, you're right, this enum won't be used very frequently, so it won't save too much space, it looks like using default to keep the code clear is better.

src/mbgl/sprite/sprite_parser.cpp
louwers: OK. Do you also want to test the error logging like the other tests?

metrics/ignores/platform-all.json
louwers: Well, looks like it failed on the Pixel 7 Pro on CI.

Might be device-dependent. Maybe best to disable it after all so we can merge this.

louwers (approved):
Can be merged once CI passes (again) and the conversations are resolved.

src/mbgl/sprite/sprite_parser.cpp
qqz003: Oh, you're right, the value of the TextFit may be invalid, I will add another test case to test it.

metrics/ignores/platform-all.json
qqz003: Yeah, but I don't know why is it failed since we couldn't obtain the result file now, I think your PR will be helpful.

src/mbgl/sprite/sprite_parser.cpp
qqz003: Have added two new test cases to test the invalid and null `TextFit` values.

louwers:
We should ignore or fix that one failing render test on all platforms again, because on CI it was failing on Android too.

qqz003:
@louwers The checks have just finished, with only the "Vulkan" test failing. Could you please help merge it? Thank you!

louwers:
Yes!

louwers:
I have created a follow-up issue for the render test. https://github.com/maplibre/maplibre-native/issues/2829

PR #2782, opened 2024-08-29
Android auto support

thanhdang198:
Added zoomBy(), getZoom() and setZoom() to MapLibreMap


louwers:
Please provide some more context. Why is the title of the pull request Android Auto support?

thanhdang198:
I'll provide more context before open another PR. Thanks @louwers

louwers:
I don't think this PR can be re-opened, so perhaps create a new one? Thanks!

Issue #2779, opened 2024-08-28
Crash on startup AndroidRendererFrontend


westnordost:
**Describe the bug**
On app startup, a user reported that the following crash occurs:

```
#00 pc 00000000008a971c  libmaplibre.so
#01 pc 00000000008a6484  libmaplibre.so
#02 pc 00000000004a3740  libmaplibre.so (mbgl::android::AndroidRendererFrontend::AndroidRendererFrontend(mbgl::android::MapRenderer&)+96)
#03 pc 00000000004d7b1c  libmaplibre.so (mbgl::android::NativeMapView::NativeMapView(_JNIEnv&, jni::Object<mbgl::android::NativeMapView> const&, jni::Object<mbgl::android::FileSource> const&, jni::Object<mbgl::android::MapRenderer> const&, float, unsigned char)+204)
#04 pc 00000000004f6610  libmaplibre.so (std::__ndk1::unique_ptr<mbgl::android::NativeMapView, std::__ndk1::default_delete<mbgl::android::NativeMapView> > jni::MakePeer<mbgl::android::NativeMapView, jni::Object<mbgl::android::NativeMapView> const&, jni::Object<mbgl::android::FileSource> const&, jni::Object<mbgl::android::MapRenderer> const&, float, unsigned char>(_JNIEnv&, jni::Object<mbgl::android::NativeMapView> const&, jni::Object<mbgl::android::FileSource> const&, jni::Object<mbgl::android::MapRenderer> const&, float, unsigned char)+92)
[...]

PR #2786, opened 2024-08-30
Add RelWithDebInfo to Android CMake build

louwers:
We are having an issue tracking down a crash (#2779). I think this will include debug symbols in the release.


cgalvan (approved):

louwers:
This did not have the desired result.

PR #2799, opened 2024-09-04
Downgrade to Java 11

louwers:
Make sure the library can be used with Java 11.

Note that Google recommends always setting the latest `targetSdkVersion`, which is 34 right now. However, when using SDK 34, Java 17 needs to be used. So we should drop support for Java 11 soonish.


alasram (approved):

cgalvan (approved):

louwers:
Workflows were failing due to some GitHub issue, retrying.

PR #2801, opened 2024-09-04
Expose RENDERMODE_CONTINUOUSLY and RENDERMODE_WHEN_DIRTY

alasram:
To save power, MapLibre only renders the map scene when it becomes dirty (an Android event causes the map to change). This behavior is desired in production. However, when testing performance and particularly trying to automatically detect stuttering, it is better to use continuous rendering.
`GLSurfaceView` already has the ability to switch between `RENDERMODE_CONTINUOUSLY` and `RENDERMODE_WHEN_DIRTY`. This PR simply exposes these options to MapLibre client apps.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapView.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapView.java
@@ -140,10 +140,10 @@ public class MapView extends FrameLayout implements NativeMapView.ViewCallback {
     // add accessibility support
     setContentDescription(context.getString(R.string.maplibre_mapActionDescription));
     setWillNotDraw(false);
-    initialiseDrawingSurface(options);
+    initializeDrawingSurface(options);
   }
 
-  private void initialiseMap() {
+  private void initializeMap() {
     Context context = getContext();
 
     // callback for focal point invalidation
@@ -305,7 +305,7 @@ public class MapView extends FrameLayout implements NativeMapView.ViewCallback {
     }
   }
 
-  private void initialiseDrawingSurface(MapLibreMapOptions options) {
+  private void initializeDrawingSurface(MapLibreMapOptions options) {
     mapRenderer = MapRenderer.create(options, getContext(), () -> MapView.this.onSurfaceCreated());
     renderView = mapRenderer.getView();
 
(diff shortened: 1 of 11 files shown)


alasram:
@louwers CI is failing. Can you tell me what log I should look at to understand the failure?

louwers:
Hmm the logs should show up if you click through but

```
Test spec output for run 'arn:aws:devicefarm:us-west-2:373521797162:run:20687d72-0e46-403e-8f03-0941850665bc/5146eb99-3855-474d-aa1c-8f132baf221e' not found
```

AWS Device Farm logs can also be found via this guide: https://github.com/maplibre/maplibre-native/wiki/AWS-Device-Farm-Logs

Did you try running the (Android) C++ Unit Tests locally? They are in `test/android`.

alasram:
@louwers All tests are passing on my phone. The runner hangs at the end but this is true on main as well. It shows this log without returning:
```
2024-09-05 14:36:42.267 11218-11255 GTest                   org.maplibre.cpp_test_runner         I  Test program ended with 955 tests run.
2024-09-05 14:36:42.267 11218-11255 mbgl                    org.maplibre.cpp_test_runner         I  {Thread-139}[General]: TestRunner finished with status: '0'
2024-09-05 14:36:42.268 11218-11247 mbgl                    org.maplibre.cpp_test_runner         I  {Thread-2}[General]: TestRunner done
```

mwilsnd (approved):

louwers (changes requested):
How do users know when the view is ready for this method to be called?

I don't really like this API. If I understand it correctly, it is OpenGL specific. Taking an `int` (instead of some kind of `enum class`) is acceptable for internal API but not so much for a public API. And if we want people to configure low level rendering, we should expose that on some low level object, otherwise the MapView will turn into a kitchen sink.

alasram:
@louwers You're right that exposing it as-is to client apps without a proper enum is wrong. I fixed this. It should work in both Vulkan and GL. It won't work for TextureView which seems to be a deprecated feature.

louwers (changes requested):
Thanks for adding the enum.

I have one more request, we should add some test coverage. At the minimum just toggling the rendering refresh mode similar to: https://github.com/alexcristici/maplibre-native/blob/f74f4ff3cf3983a07cbd25391103d8f2950846d5/platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/maps/MapLibreMapTest.kt#L471-L481

louwers (approved):

PR #2802, opened 2024-09-05
Add android auto support

thanhdang198:
- Added zoomBy(), getZoom() and setZoom to MapLibreMap
- Added androidautoapp module to MapLibreAndroidTestApp


louwers (changes requested):
Thanks for sharing you working implementation of Android Auto!

Since hardly any changes to the base library, I think this would be better as a separate project. I will ask around if there is more interest from the community for such a project.

frankkienl:
Very interested in Android Auto 👍

louwers:
I also scheduled Android Auto as a discussion item for the next TSC Meeting. 

7 PM to 8 PM CEST, MapLibre Native, https://zoom.us/j/97047175665?pwd=8C1EUbqXoNdUBJdmkfuhrb45PuxjzJ.1

(That time likely won't work for @thanhdang198).

On Slack I discussed with @frankkienl why I believe a separate repo for an Android Auto example project is a better idea:

- The maplibre/maplibre-native repo focuses on the core renderer in C++ plus some platform bindings that wrap that renderer and essential functionality for a map toolkit on those platforms.  We also have apps part of this repo, but they exist primarily for the purpose of testing the platform bindings and/or core renderer. Platforms that do not directly integrate with the C++ renderer should in my opinion not be added to maplibre/maplibre-native. Android Auto is arguably such a platform.
- Since end users using Android Auto will use MapLibre Android directly, having it in a separate project is a better demonstration of how they would intergrate it.
- Maybe the project will grow out into a utility library to make intergrating MapLibre Android with Android Auto easier to use. In this case a separate project is definitely the better option.

@frankkienl said he will discuss with his colleagues if @flitsmeister can kick-off such a repo until we onboard it in the MapLibre organization.

rubenmx:
I see some of the code I wrote for Flitsmeister, including my exact comments 😂 Curious to know how you got it.
I feel that if there would be official Android Auto support from MapLibre, that there should be a more efficient way instead of drawing a bitmap of the MapView to the Android Auto surface. If something like that were possible, I'd be very interested.

louwers:
@rubenmx Oof, if that is really true and not some crazy coincidence, that is something we take very seriously.

@thanhdang198 Did you write this code yourself?

rubenmx:
We did share it privately with another developer, I don't think its that big a deal, just curious if it ended up somewhere online or something.

louwers:
@rubenmx Including code that infringes on someone copyright is a huge liability both for MapLibre as a project and our users, so we take great care to avoid that and have [a policy](https://github.com/maplibre/maplibre/blob/main/INTELLECTUAL_PROPERTY_VIOLATION_POLICY.md) in place to handle situations where this happened.

We rely on contributions following GitHub TOS which state:

> Whenever you add Content to a repository containing notice of a license, you license that Content under the same terms, and **you agree that you have the right to license that Content under those terms**.

https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license

thanhdang198:
> @rubenmx Oof, if that is really true and not some crazy coincidence, that is something we take very seriously.
> 
> @thanhdang198 Did you write this code yourself?

I saw it in Slack chat and I mentioned it before. You can see this in the Slack thread. 
I added more additional code for the gesture handles, like double-click zoom, two-finger zoom, and move the map.
With the gesture handle, I confirmed that I do not copy anywhere OUTSIDE the maplibre-native project. And of course, it not against the Mapbox license too.

thanhdang198:
> We did share it privately with another developer, I don't think its that big a deal, just curious if it ended up somewhere online or something. Edit: I see now that Sjoerd shared in on the MapLibre slack.



Sorry if this PR is against any license of your code. I found they were shared in Slack without any license included.
If you need any additional action before it can be shared with the public or don't allow it, please let me know.
Thank you.

louwers:
@thanhdang198 Ah so Flitsmeister shared some of this code on Slack? Thanks for clarifying.

Let's see if we hear back from @frankkienl and @rubenmx if @flitsmeister can release this code with a proper license. 👍 

Closing this PR for now, but feel free to re-open one with the needed changes to MapLibreMap (still needs some tests).

PR #2805, opened 2024-09-05
Removing aftermarket expression installer

tarkvara:
- Removed `MLNAftermarketExpressionInstaller`, since it doesn't work anyways.
- The reimplementation of missing functions from the MapLibre Style Spec is out of scope for this PR.


louwers (commented):
When I run MultipleImagesExample from MapLibreApp, it now crashes. I think what is happening is that the patching still works, but Apple blocks calling actually calling the Objective-C implementations. 

So I expect all the functions with

> Assignment expressions lack underlying Objective-C implementations.

to still work.

Say what, maybe we can keep the `MLNAftermarketExpressionInstaller` in for now, because removing it will turn a warning into a hard crash, and instead just remove the broken tests?

tarkvara:
I'm withdrawin

tarkvara:
I'm withdrawing this PR.  It doesn't accomplish what it's supposed to, and it causes more trouble than it's worth.

PR #2811, opened 2024-09-06
Ensure that all depth values are rendered before any color values

### LINKED ISSUE #2591: Incorrect rendering of extruded geometry
Android release android-v11.0.1 (also 11.0) shows artifacts when rendering extruded geometry (such as buildings).
The issue is only seen in the drawable GL renderer. Teh legacy GL renderer does not have the issue:
angle3
angle2
angle1

The bad geometry is camera dependent

**To Reproduce**
- Run on Android devices with a map that has 2.5D geometry data
- Make sure MapLibreAndroid-drawable library is used
- Zoom into the geometry and navigate the map to see incorrect culled geometry along a line
-  Issue not present when switching to MapLibreAndroid-legacy library

TimSylvester:
Resolves #2591 ... ?

### DIFF
--- a/src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
+++ b/src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
@@ -433,6 +433,7 @@ void RenderFillExtrusionLayer::update(gfx::ShaderRegistry& shaders,
                 builder->setEnableColor(false);
                 builder->setRenderPass(drawPass);
                 builder->setCullFaceMode(gfx::CullFaceMode::backCCW());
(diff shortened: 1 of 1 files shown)


alasram:
Thanks @TimSylvester I'll test this change

alasram:
@TimSylvester brilliant! This works great on my side.

alasram (approved):

louwers (approved):

louwers:
Thanks for reporting @alasram and for the fix @TimSylvester. Would be good to create a regression test for this. I created an issue. https://github.com/maplibre/maplibre-native/issues/2818

Issue #2807, opened 2024-09-06
`Map.ObserveTileLifeCycle` test failing


louwers:
This new test is failing on Android (on CI).

Can't reproduce it locally.

```
09-05 16:47:30.191  1946  2433 I GTest   : Starting: Map.ObserveTileLifecycle
09-05 16:47:30.192  1946  2433 I mbgl    : {Thread-66}[Style]: Style 'Water' has 3 layers:
09-05 16:47:30.192  1946  2433 I mbgl    :   None: 3
09-05 16:47:30.263  2280  3277 W ModuleInitIntentOp: Dropping unexpected action com.google.android.gms.phenotype.COMMITTED
09-05 16:47:30.264  2280  3276 W ModuleInitIntentOp: Dropping unexpected action com.google.android.gms.phenotype.COMMITTED
09-05 16:47:30.265  2280  3276 W ModuleInitIntentOp: Dropping unexpected action com.google.android.gms.phenotype.COMMITTED
09-05 16:47:30.266  2280  3289 I SyncIntentOperation: (REDACTED) Handling the intent: %s.
09-05 16:47:30.266  1946  2433 I GTest   : FAILED: Value of: stage
09-05 16:47:30.266  1946  2433 I GTest   : Expected: is equal to 1-byte object <01>
09-05 16:47:30.266  1946  2433 I GTest   :   Actual: 1-byte object <05>
09-05 16:47:30.270  1946  2433 D mbgl    : {Thread-66}[General]: Texture2DPool: total allocations 9
09-05 16:47:30.270  1946  2433 D mbgl    : {Thread-66}[General]: Texture2DPool: total reused allocations 0
09-05 16:47:30.270  1946  2433 D mbgl    : {Thread-66}[General]: Texture2DPool: total textures released 9
09-05 16:47:30.270  1946  2433 D mbgl    : {Thread-66}[General]: Texture2DPool: total textures released and deallocated 9
09-05 16:47:30.270  1946  2433 D mbgl    : {Thread-66}[General]: Rendering Stats:
09-05 16:47:30.270  1946  2433 D mbgl    : numDrawCalls = 8
09-05 16:47:30.270  1946  2433 D mbgl    : stencilUpdates = 1
09-05 16:47:30.272  2280  3287 I MdiSyncModule: flag handling is skipping irrelevant intent.
09-05 16:47:30.272  2280  3287 I MlBenchmarkInstaller: (REDACTED) Intent: %s
[...]

PR #2819, opened 2024-09-10
Improve tile lifecycle determinism 

mwilsnd:
Addresses some observed behavior in the lifetime of tiles that wasn't initially expected. This PR makes some minor changes to how GeometryTiles update their pending status, to better reflect what the tile is actually doing. One of the more interesting observations is how a tile pyramid's call to `setLayers()` can dispatch a task on the worker thread before the tile has even loaded.

This PR should address #2807


louwers:
Looks like it will also fix #2826!

src/mbgl/tile/geometry_tile.cpp
```diff
@@ -245,11 +245,13 @@ void GeometryTile::setData(std::unique_ptr<const GeometryTileData> data_) {
 void GeometryTile::reset() {
     MLN_TRACE_FUNC();
 
+    if (pending) {
+        observer->onTileAction(id, sourceID, TileOperation::Cancelled);
+        pending = false;
+    }
+
     // Mark the tile as pending again if it was complete before to prevent
     // signaling a complete state despite pending parse operations.
```
louwers, line 254: Is this comment still accurate?

test/map/map.test.cpp
```diff
@@ -1762,17 +1768,13 @@ TEST(Map, ObserveTileLifecycle) {
...
-        // Lower zooms can also be seen, but not always, so we
-        // ignore them.
-    };
-
     for (const auto& tile : expectedTiles) {
         TileOperation stage = TileOperation::NullOp;
+        bool parsing = false;
 
-        for (const auto& op : tileOps) {
+        // NOLINTNEXTLINE
```
louwers, line 1775: Maybe include the specific lint rule you are disabling?

louwers (approved):

src/mbgl/tile/geometry_tile.cpp
mwilsnd: Not really, but I don't think it ever truly was in the context of `reset()`. I've revised the comments in this method and cleared the `loaded` state which more accurately expresses what this method does.

test/map/map.test.cpp
```diff
@@ -1772,6 +1772,7 @@ TEST(Map, ObserveTileLifecycle) {
         TileOperation stage = TileOperation::NullOp;
         bool parsing = false;
 
+        // Suppressing warning-as-error to use range-based for. Index is desired for debugging context.
         // NOLINTNEXTLINE
```
louwers, line 1776:
```suggestion
        // NOLINTNEXTLINE(modernize-loop-convert)
```

Like this I meant.

TimSylvester (approved):

alexcristici (approved):

PR #2827, opened 2024-09-11
Make sure swapBehavior is NoFlush

alasram:
There's a performance regression on Android with OpenGL due to calling `glFinish` every frame. This is because the new variable  `MapRenderer::swapBehaviorFlush` is set to `true` by default. This PR simply sets it to `false` to get the previous behavior which doesn't call `glFinish` unless the client explicitly change the behavior by calling `setSwapBehavior`

### DIFF
--- a/platform/android/MapLibreAndroid/src/cpp/map_renderer.hpp
+++ b/platform/android/MapLibreAndroid/src/cpp/map_renderer.hpp
@@ -143,12 +143,13 @@ private:
     std::shared_ptr<UpdateParameters> updateParameters;
     std::mutex updateMutex;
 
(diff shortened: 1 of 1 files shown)


alasram:
@adrian-cojocaru I didn't check Vulkan but I hope it supports both `Flush` and `NoFlush`. If not then it's better to force the working behavior inside the VK backed then fix swapchain sync ticket

adrian-cojocaru:
Vulkan supports both. The variable ended up as true due to some refactoring and always testing with flush enabled (it should have always been false).

adrian-cojocaru (approved):

Issue #1654, opened 2023-09-18
Location Indicator Layer


sjg-wdw:
Reimplement the location indicator using the new Drawable/Builder/LayerGroup structure.

sjg-wdw:
The thinking was this would be a great test case for a new custom layer.
So when you get a chance to figure out how we should reuse the builders/buckets.  See if you can get this going!

PR #2836, opened 2024-09-16
Use drawables for location indicator layer

adrian-cojocaru:
This PR adds guards for OpenGL code in `src/mbgl/renderer/layers/render_location_indicator_layer.cpp` and adds an alternative implementation using drawables for the location indicator elements.

- added a set of new simple shaders: `CommonShader`, `CommonTexturedShader`
- added mipmapping and runtime mip generation
- added texture sampler anisotropy value/check


TimSylvester (approved):

include/mbgl/shaders/shader_defines.hpp
```diff
@@ -62,6 +64,11 @@ enum {
     collisionTextureCount
 };
 
+enum {
```
louwers, line 67: Prefer `enum class`.

louwers:
Closes #1654

src/mbgl/renderer/layers/render_location_indicator_layer.cpp
```diff
@@ -894,12 +1032,271 @@ void RenderLocationIndicatorLayer::render(PaintParameters& paintParameters) {
...
+        auto& drawableUniforms = circleDrawable.mutableUniformBuffers();
+        drawableUniforms.createOrUpdate(shaders::idCommonUBO, &commonUBO, context);
+
+        commonUBO.color = renderImpl->parameters.errorRadiusBorderColor;
+        auto& outlineDrawableUniforms = circleOutlineDrawable.mutableUniformBuffers();
+        outlineDrawableUniforms.createOrUpdate(shaders::idCommonUBO, &commonUBO, context);
+    };
+
+    const shaders::CommonUBO ubo = {/* .matrix */ util::cast<float>(renderImpl->getProjectionPuck()),
+                                    /* .color */ Color::black()};
```
alexcristici, line 1250: This part that updates the UBO should be in the layer tweaker files: `location_indicator_layer_tweaker.hpp/cpp`.

alexcristici (changes requested):

include/mbgl/shaders/shader_defines.hpp
adrian-cojocaru: Everything else in the file uses `enum`. I would prefer keeping them the same so it's either updating all of them to `enum class` now or later.

louwers: Created an issue.

alexcristici (approved):

Issue #2791, opened 2024-08-31
Android x86_64 emulator crashes (11.2.0)


westnordost:
**Describe the bug**
On displaying the map, the Android emulator now crashes. This is new in MapLibre native for Android 11.2.0.

Log:
```
2024-08-31 18:53:33.742  4383-4383  DEBUG                   crash_dump64                         A        #01 pc 00000000008b276e  /data/app/~~Bf5r8m0AUQgO9I1Un4iM_g==/de.westnordost.streetcomplete.debug-_PZ5lVN0NkAKor_jPnrn-Q==/lib/x86_64/libmaplibre.so (BuildId: 5ee7125836a38fac58c05a97ac4b3e9ca8aec3a0)
2024-08-31 18:53:33.742  4383-4383  DEBUG                   crash_dump64                         A        #02 pc 00000000008e04ef  /data/app/~~Bf5r8m0AUQgO9I1Un4iM_g==/de.westnordost.streetcomplete.debug-_PZ5lVN0NkAKor_jPnrn-Q==/lib/x86_64/libmaplibre.so (BuildId: 5ee7125836a38fac58c05a97ac4b3e9ca8aec3a0)
2024-08-31 18:53:33.742  4383-4383  DEBUG                   crash_dump64                         A        #03 pc 00000000008e1863  /data/app/~~Bf5r8m0AUQgO9I1Un4iM_g==/de.westnordost.streetcomplete.debug-_PZ5lVN0NkAKor_jPnrn-Q==/lib/x86_64/libmaplibre.so (BuildId: 5ee7125836a38fac58c05a97ac4b3e9ca8aec3a0)
2024-08-31 18:53:33.743  4383-4383  DEBUG                   crash_dump64                         A        #04 pc 00000000006c31ca  /data/app/~~Bf5r8m0AUQgO9I1Un4iM_g==/de.westnordost.streetcomplete.debug-_PZ5lVN0NkAKor_jPnrn-Q==/lib/x86_64/libmaplibre.so (BuildId: 5ee7125836a38fac58c05a97ac4b3e9ca8aec3a0)
[...]

PR #2841, opened 2024-09-16
Allow building the legacy renderer for Android

alasram:
Suppress the build error when legacy renderer is selected on Android. The legacy renderer is still useful because it is stable when running on an Emulator. The legacy renderer can be fully removed when [2791](https://github.com/maplibre/maplibre-native/issues/2791) is fixed.


louwers (approved):
It's on it's way out, but sure.

PR #2842, opened 2024-09-16
Hillshade geometry update fixes

### LINKED ISSUE #2838: hillshade rendering artifacts, regression in iOS 6.6.0
**Describe the bug**
The iOS 6.6.0 release seems to have introduced a bug related to hillshade rendering, while the map is moving there are sometimes lots of flashing, and line artifacts. But once all data is loaded, it renders correctly. There are several examples of this happening in the attached video. This was not happening in 6.5.4

**Expected behavior**

**Screenshots**

https://github.com/user-attachments/assets/99ebc3e8-5f53-43ee-a72c-d7addb229ce1

 - OS: iOS 17.x
 - Platform iOS
 - Version 6.6.0

TimSylvester:
Fixes a problem where masking changes in hillshade tile geometry are not recognized as updated.

Fixes a bug where vertex attributes on a drawable replaced by ones created earlier are not applied to the underlying buffers.  This would have been an issue previously, if the newly-applied attributes were not marked dirty, as with shared static tile geometry.

Fixes a bug where fill updates would apply the fill vertices to the triangulated outline drawables (if enabled), which would lead to errors if they were applied, but they weren't because of the previous item.

Resolves #2838 ?


louwers:
Better, but now some tiles don't get loaded at all until I change the zoom a bit:

591EFB3E-87D3-44A3-B26E-7A978891AECD_1_102_o

alasram:
@louwers the behavior in your screenshots is already present in main on Android

louwers (approved):

JesseCrocker (approved):
I have not looked at the code, bust i tested this in my application, and it fixes the issue.

TimSylvester:
Looks like we have a longstanding issue with updating the vertices without updating the segments.  That's fine in Metal, but leads to problems in GL because the segments retain Vertex Array identifiers, so we get GL errors if we actually replace the vertices alone.

louwers (changes requested):
@TimSylvester Render tests are failing unfortunately.

louwers (approved):

alexcristici (approved):

PR #2845, opened 2024-09-17
Add array support for icon-padding

### LINKED ISSUE #2368: Accept CSS style padding different on different sides for `icon-padding` property
For the [`icon-padding`](https://maplibre.org/maplibre-style-spec/layers/#icon-padding) property MapLibre GL JS support providing an array of floats as [padding](https://maplibre.org/maplibre-style-spec/types/#padding).

Right now we only support a single float (which applies padding to all sides).

random3940:
Port [feature from MapLibre/JS](https://github.com/maplibre/maplibre-gl-js/pull/1289) to restore parity and close #2368. Implementation follows same idea as original PR by adding new Padding type to MapLibre/Native DSL implementation.


include/mbgl/util/padding.hpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    float top = 0;
+    float right = 0;
+    float bottom = 0;
+    float left = 0;
+
+    explicit operator bool() const { return top != 0 || right != 0 || bottom != 0 || left != 0; }
+
+    std::array<float, 4> toArray() const;
```
louwers, line 61: You should import array here, CI is complaining about this.

include/mbgl/util/padding.hpp
```diff
@@ -0,0 +1,80 @@
...
+                right = values[1];
+                bottom = values[2];
+                left = values[3];
+                break;
+            default:
+                throw std::invalid_argument("Padding must have between 1 and 4 elements");
+        }
+    }
+
+    float top = 0;
```
louwers, line 54: We should make this immutable probably.

scripts/style-spec-reference/v8.json
```diff
@@ -1529,7 +1529,8 @@
         "basic functionality": {
           "js": "0.10.0",
           "android": "2.0.1",
-          "ios": "2.0.0"
+          "ios": "2.0.0",
```
louwers, line 1532: We maintain this in https://maplibre.org/maplibre-style-spec/

We don't make releases for macOS anymore.

louwers (commented):
Exciting!

include/mbgl/util/padding.hpp
random3940: This breaks constructor from std::span. I don't want to copy this logic in several places where we call it. I

I can make it immutable and add factory method instead of ctor, but it feels like an overkill for such a simple struct with 4 fields. I'd prefer to keep it mutable. Please ping me if you if you feel strongly about this and want me to change this.

include/mbgl/util/padding.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    explicit operator bool() const { return top != 0 || right != 0 || bottom != 0 || left != 0; }
+
+    std::array<float, 4> toArray() const;
+
+    // Used by ValueFactory<Padding>::make()
+    mbgl::Value serialize() const;
+};
+
+inline bool operator==(const Padding& paddingA, const Padding& paddingB) {
```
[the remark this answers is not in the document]
random3940: Indeed, thank you for pointing that out!

scripts/style-spec-reference/v8.json
random3940: Removed the change.

louwers:
Error: Failed to find metric expectations for: metrics/integration/render-tests/icon-padding/default/style.json

Error: Failed to find metric expectations for: metrics/integration/render-tests/icon-padding/databind/style.json

include/mbgl/util/padding.hpp
louwers: I tried this:

```cpp
struct Padding {
    // Empty/zero padding on all sides.
    // Padding() = default;

    // Same padding on all sides.
    Padding(float value)
        : top(value),
          right(value),
          bottom(value),
          left(value) {}

    Padding(float t_, float r_, float b_, float l_)
        : top(t_),
          right(r_),
          bottom(b_),
          left(l_) {}

    /// Padding values are in CSS order: top, right, bottom, left.
    Padding(const std::span<const float>& values)
        : Padding(Padding::paddingTupleFromSpan(values)) {}

    const float top = 0;
    const float right = 0;
    const float bottom = 0;
    const float left = 0;

    explicit operator bool() const { return top != 0 || right != 0 || bottom != 0 || left != 0; }

    bool operator==(const Padding&) const = default;

    std::array<float, 4> toArray() const;

    // Used by ValueFactory<Padding>::make()
    mbgl::Value serialize() const;

private:
    // Helper function to compute padding values from a span
    static std::tuple<float, float, float, float> paddingTupleFromSpan(const std::span<const float>& values) {
        switch (values.size()) {
            case 1:
                return {values[0], values[0], values[0], values[0]};
            case 2:
                return {values[0], values[1], values[0], values[1]};
            case 3:
                return {values[0], values[1], values[2], values[1]};
            case 4:
                return {values[0], values[1], values[2], values[3]};
            default:
                throw std::invalid_argument("Padding must have between 1 and 4 elements");
        }
    }

    // Private constructor used by the std::span constructor
    Padding(const std::tuple<float, float, float, float>& padding)
        : top(std::get<0>(padding)),
          right(std::get<1>(padding)),
          bottom(std::get<2>(padding)),
          left(std::get<3>(padding)) {}
};
```

But I am getting some weird compile errors.

random3940: I was mostly averse to complicating this simple struct, but I love immutable things as the next guy, so I'll look into using this approach! :)

louwers (approved):

include/mbgl/util/padding.hpp
random3940: Bart, the tuple approach works wonderfully. I don't think I've changed anything, except moving impl to cpp. 

But... immutable Padding causes more changes downstream. There are several places where code either uses assignment operator directly or via autogenerated operators in classes which include Padding as a member (e.g. PositionedIcon). Immutability breaks those.

So far I've updated 6 places, mostly by either adding std::optional wrapper or by replacing `x = foo;` with `x.emplace(foo)` if the object was already an optional.  

Then I got to expressions code and found out that generic code around `defaultValue` expects underlying types to be mutable too. It is probably fixable too, but at this point I doubt that the whole caboodle is worth it and wanted to double-check with you. 

Most (all?) other simple structs I've checked are declared mutable and immutability is added by slapping const on them as necessary. Maybe we should follow the same idea with Padding?

louwers: Yes exactly I also went down that rabbit hole for a bit. 🙈 

Let's keep it the way it is.

PR #2846, opened 2024-09-17
Try using MoltenVK on macOS

louwers:
```
cmake -D MLN_WITH_VULKAN=ON -DMLN_DRAWABLE_RENDERER=ON -DCMAKE_BUILD_TYPE=Debug -B build -S . -GNinja -DVULKAN_SDK=/Users/bart/VulkanSDK/1.3.290.0/macOS
cmake --build build --target mbgl-glfw
```

Does not quite work yet:

```
build/platform/glfw/mbgl-glfw         
2024-09-17 16:16:16.142 mbgl-glfw[35844:887089] [ERROR] {}[Render]: Vulkan extensions not found
libc++abi: terminating due to uncaught exception of type vk::IncompatibleDriverError
zsh: abort      build/platform/glfw/mbgl-glfw
```

### DIFF
--- a/include/mbgl/vulkan/renderer_backend.hpp
+++ b/include/mbgl/vulkan/renderer_backend.hpp
@@ -3,13 +3,16 @@
 #include <mbgl/gfx/renderer_backend.hpp>
 #include <mbgl/gfx/context.hpp>
 
+#define VULKAN_HPP_DISPATCH_LOADER_DYNAMIC 1
+
 #ifdef _WIN32
 #define VK_USE_PLATFORM_WIN32_KHR
 #elif defined(__ANDROID__)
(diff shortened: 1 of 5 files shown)


birkskyum:
This is quite interesting. I imagine the initial question is how close the perf is to Metal, right? I'm not experienced in Native mobile dev (built more with Ionic/Capacitor/Tauri/Unity etc.), but if the perf is sufficient, it it possible to imagine that some Vulkan-specialized teams would actually prefer to use the Vulkan backend throughout for cross-platform distribution? Or would it simply be a non-starter compared to the dx delivered by the native Metal backend on iOS?

adrian-cojocaru (approved):

TimSylvester (approved):

PR #2859, opened 2024-09-21
Run Node.js ARM workflow on GitHub runner

louwers:
Work around for gcc bug. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -76,7 +76,7 @@ jobs:
         include:
           - runs-on: ubuntu-22.04
             arch: x86_64
-          - runs-on: [self-hosted, linux, ARM64]
+          - runs-on: MapLibre_Native_Linux_ARM_8_core
             arch: arm64
           - runs-on: macos-14
             arch: arm64
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -76,7 +76,7 @@ jobs:
         include:
           - runs-on: ubuntu-22.04
             arch: x86_64
-          - runs-on: MapLibre_Native_Linux_ARM_8_core
+          - runs-on: MapLibre_Native_Ubuntu_22_04_ARM_8_core
             arch: arm64
           - runs-on: macos-14
             arch: arm64
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -46,7 +46,7 @@ jobs:
         include:
           - runs-on: ubuntu-22.04
             arch: x86_64
-          - runs-on: [self-hosted, linux, ARM64]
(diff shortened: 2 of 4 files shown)


.github/workflows/node-ci.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         include:
           - runs-on: ubuntu-22.04
             arch: x86_64
-          - runs-on: [self-hosted, linux, ARM64]
+          - runs-on: MapLibre_Native_Linux_ARM_8_core
```
acalcutt, line 79: Don't forget this runner change also needs to be done in node-release

acalcutt:
Is 'MapLibre_Native_Linux_ARM_8_core' a ubuntu 22.04 runner?

louwers:
@acalcutt Ubuntu 24.04 actually https://github.com/maplibre/maplibre/issues/413

acalcutt:
Ubuntu 24.04 would change the compatibility, this would no longer work on ubuntu 22.04 we say we support in the README. I am not sure I am ready to move tot hat but at some point we should.

The linux version is picky about the libs of the system it was build on.

birkskyum:
> Ubuntu 24.04 would change the compatibility, this would no longer work on ubuntu 22.04 we say we support in the README. I am not sure I am ready to move to that but at some point we should.
> 
> The linux version is picky about the libs of the system it was build on.


There is a 22.04 option, so let me know if it would be better to recreate the runner with that instead.
Screenshot 2024-09-22 at 09 09 06

acalcutt:
I think it would be good to do 22.04 for now, at least until we are ready for a new breaking release.

louwers:
@birkskyum Could you make two and include the Ubuntu version in the name?

cgalvan (approved):

acalcutt:
Any update on this @birkskyum and @louwers ? Can we make a Ubuntu 20.04 for this build for now? It looks like right now the existing runner is already down, so moving it would be good.

.github/workflows/node-release.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         include:
           - runs-on: ubuntu-22.04
             arch: x86_64
-          - runs-on: [self-hosted, linux, ARM64]
+          - runs-on: MapLibre_Native_Linux_ARM_8_core
```
acalcutt, line 49: Don't forget to change this to match ci

acalcutt:
I guess I was wrong....we were running on 22.04 before. sorry about that!

.github/workflows/node-release.yml
louwers: CI already complained! :)

acalcutt (approved):

PR #2880, opened 2024-09-28
Add PropertyFactory.iconPadding(Float) overload for better backcompat

random3940:
See #2873 for details.


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/layers/property_factory.java.ejs
```diff
@@ -90,6 +90,18 @@ public class PropertyFactory {
...
+<% } -%>
+<% if (property.type === 'padding') { -%>
+  /**
+   * <%- propertyFactoryMethodDoc(property) %>
+   *
+   * @param value a Float value
+   * @return property wrapper around Float
+   */
+  public static PropertyValue<Float> <%- camelizeWithLeadingLowercase(property.name) %>(Float value) {
+    return new LayoutPropertyValue<>("<%-  property.name %>", value);
```
louwers, line 102: Shouldn't this be wrapped in `new Float[]{value}`?

random3940: Converter<Padding> supports [both numbers and arrays](https://github.com/maplibre/maplibre-native/blob/63d67079e0ab6cf73d37223a50b81c3cf7128e40/src/mbgl/style/conversion/constant.cpp#L118), so it will work both ways. With number there is a miniscule gain because we don't need to get Convertible into std::vector :)

louwers (commented):
Should we restore the old test for the single Float version as well?

random3940:
> Should we restore the old test for the single Float version as well?

Added test that single number works for padding too.

louwers (approved):

louwers:
Thank you!

PR #2887, opened 2024-10-01
Make sure `Scheduler::GetCurrent()` cannot return a nullptr

louwers:
I think when the call to `JNI_OnLoad` and thus `registerNatives` on the main thread gets delayed, it is possible that  `Scheduler::GetCurrent()` returns a `nullptr`, which crashes the program. This change prevents a `nullptr` from being returned, falling back to a thread-local Scheduler.

Another attempt to fix https://github.com/maplibre/maplibre-native/issues/2779


src/mbgl/actor/scheduler.cpp
```diff
@@ -23,6 +24,10 @@ void Scheduler::SetCurrent(Scheduler* scheduler) {
 }
 
 Scheduler* Scheduler::GetCurrent() {
+    if (!localScheduler) {
+        thread_local util::RunLoop runLoop;
+        SetCurrent(&runLoop);
```
mwilsnd, line 29: Does it make sense to do this here or on the caller side? I see RunLoop destructors explicitly setting the current scheduler to nullptr so I'm not sure if it makes sense to have GetCurrent always return a valid instance.

louwers: I think it makes sense, none of the current call sites assume that `GetCurrent()` ever return `nullptr`.

louwers:
@tdcosta100 Do you know why we get linker errors on Windows?

adrian-cojocaru (approved):

PR #2889, opened 2024-10-01
Metal macOS render tests updated

alexcristici:
Metal macOS render tests updated from Metal iOS expectations and ignores.

Screenshot 2024-10-01 at 23 58 33


cgalvan (approved):

.github/workflows/node-ci.yml
```diff
@@ -265,15 +265,13 @@ jobs:
           cmake --build build
 
       - name: Run render tests on macOS
+        id: render_tests
         if: runner.os == 'macOS'
-        # A few are failing
-        # https://github.com/maplibre/maplibre-native/issues/2741
-        # Almost all are failing on Intel GitHub runners (but only on CI)
-        continue-on-error: true
```
acalcutt, line 272: removing this will probably still fail in macos x64, since almost all test were failing in the runner (unrelated to the actual test failers this PR fixes)

louwers: Yup, that is why I allowed it to fail on x84 again now.

PR #2892, opened 2024-10-02
Group descriptor sets by binding frequency

adrian-cojocaru:
Group uniform and textures in global/layer/drawable descriptor sets to minimize the number of bindings


louwers (commented):
Can you provide more context for this PR in the PR description?

louwers:
Thanks!

TimSylvester:
Looks like the only effect in Metal/GL is to change around the indexes where things are bound?

adrian-cojocaru:
@TimSylvester In general yes. It reorders some enum values so that all layer specific ids are contiguous.

TimSylvester (approved):

louwers (approved):

mwilsnd (approved):

alexcristici (approved):
Nice!

PR #2908, opened 2024-10-07
Line atlas fix

### LINKED ISSUE #2735: Metal: Incorrect reuse of sprites for line-pattern
**Describe the bug**

In our product we use a style with `line-pattern`. In versions that use MetalKit for map rendering we noticed a problem with filling lines with pattern. 

I have prepared a test project demonstrating the problem (attached). 

The project style is available in style.json file or at https://api.npoint.io/7dac9318bfb95b18a9dc.

The example uses a simple geojson with two lines, one should be filled with a black dashed line and the other with yellow icons. 

On a small zoom everything shows up as it should, but if you zoom in on the map you can see that the styles get mixed up. 
The yellow line starts to appear as a black dashed line and vice versa. 

The problem seems to be the reuse of sprites.

We have only noticed this problem for the `line-pattern` property.

Before using the pre-metalKit version, we had no problems. The problem persists on the latest version 6.5.4 , and is also definitely present on 6.5.0

[MlnPatterBug.zip](https://github.com/user-attachments/files/16662818/MlnPatterBug.zip)

**Screenshots**
photo_2024-08-19 19 10 45
photo_2024-08-19 19 10 47
photo_2024-08-19 19 10 50

 - OS: iOS
 - Platform: iOS
 - Version: 6.5.4

TimSylvester:
I think this came from the fill layer or something, but doesn't make sense here.  We can potentially share some of these, because they are generated in the worker and assigned to multiple buckets, but we would have to use a map and look them up not just have a single cache item, and that's probably not worth it.

Resolves #2735

### DIFF
--- a/platform/darwin/.gitignore
+++ b/platform/darwin/.gitignore
@@ -1,3 +1,5 @@
 # Generated list files from code generation
 /scripts/generate-style-code.list
 /scripts/update-examples.list
+vulkan.xcworkspace
+
--- a/src/mbgl/renderer/layers/render_line_layer.cpp
+++ b/src/mbgl/renderer/layers/render_line_layer.cpp
@@ -558,19 +558,17 @@ void RenderLineLayer::update(gfx::ShaderRegistry& shaders,
 
             // texture
             if (const auto& atlases = tile.getAtlasTextures(); atlases && atlases->icon) {
-                if (!iconTweaker) {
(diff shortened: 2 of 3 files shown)


sjg-wdw (approved):
Cool!

alexcristici (approved):
Nice!

louwers (commented):
Can we add a regression test?

PR #2921, opened 2024-10-09
Add support for `text-variable-anchor-offset` property

### LINKED ISSUE #2358: Support `text-variable-anchor-offset` property
This is a new style property that is especially suited for label placement of asymmetrical icons. I have included the description below as well as some links for reference. The discussion thread has some useful illustrations for the motivation behind this property.

A PR that implements this property for MapLibre Native would be highly welcomed.

> ### text-variable-anchor-offset
> 
> _Layout property. Optional array, alternating enum and point values._
> 
> To increase the chance of placing high-priority labels on the map, you can provide an array of text-anchor locations, each paired with an offset value. The renderer will attempt to place the label at each location, in order, before moving on to the next location+offset. Use text-justify: auto to choose justification based on anchor position.
> 
> The length of the array must be even, and must alternate between enum and point entries. i.e., each anchor location must be accompanied by a point, and that point defines the offset when the corresponding anchor location is used. Positive offset values indicate right and down, while negative values indicate left and up. Anchor locations may repeat, allowing the renderer to try multiple offsets to try and place a label using the same anchor.
> 
> When present, this property takes precedence over text-anchor, text-variable-anchor, text-offset, and text-radial-offset.
> 
> Example
> 
> ```json
> {
>   "text-variable-anchor-offset": ["top", [0, 4], "left", [3, 0], "bottom", [1, 1]]
> }
> ```
> 
> When the renderer chooses the `top` anchor, [0, 4] will be used for `text-offset`; the text will be shifted down by 4 ems. When the renderer chooses the `left` anchor, [3, 0] will be used for `text-offset`; the text will be shifted right by 3 ems.

Discussion MapLibre Style Spec: https://github.com/maplibre/maplibre-style-spec/issues/112

MapLibre Style Spec Definition: https://maplibre.org/maplibre-style-spec/layers/?h=text+offset#text-variable-anchor-offset

Implementation MapLibre GL JS: https://github.com/maplibre/maplibre-gl-js/pull/2914

wangyingfang:
Closes #2358. Support text-variable-anchor-offset property


include/mbgl/style/layers/symbol_layer.hpp
```diff
@@ -9,6 +9,7 @@
 #include <mbgl/style/property_value.hpp>
 #include <mbgl/style/expression/formatted.hpp>
 #include <mbgl/util/color.hpp>
+#include <mbgl/util/variable_anchor_offset_collection.hpp>
```
random3940, line 12: @wangyingfang, shouldn't there also be change in layer.hpp.ejs template? Otherwise those changes will be overwritten on next re-gen, right?

scripts/generate-style-code.mjs
```diff
@@ -26,21 +26,21 @@ function parseCSSColor(/** @type {string} **/ str) {
...
- * @param {any} property 
+ * @param {any} property
  */
 function isLightProperty(property) {
   return property['light-property'] === true;
 };
 
 /**
- * @param {any} property 
+ * @param {any} property
```
random3940, line 36: Could you please drop those writespace-only changes from PR?

src/mbgl/layout/symbol_instance.cpp
```diff
@@ -206,6 +206,17 @@ bool SymbolInstance::hasSdfIcon() const {
     return symbolContent & SymbolContent::IconSDF;
 }
 
+std::vector<style::SymbolAnchorType> SymbolInstance::getTextAnchors() const {
+    std::vector<style::SymbolAnchorType> result;
+    if (textVariableAnchorOffset) {
+        for (const auto& anchorOffset : *textVariableAnchorOffset) {
```
random3940, line 212: It would be good to call vector.reserve() upfront.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -358,6 +359,67 @@ std::array<float, 2> SymbolLayout::evaluateVariableOffset(style::SymbolAnchorTyp
...
+std::optional<VariableAnchorOffsetCollection> SymbolLayout::getTextVariableAnchorOffset(const SymbolFeature& feature) {
+    std::optional<VariableAnchorOffsetCollection> result;
+
+    // If style specifies text-variable-anchor-offset, just return it
+    if (!textVariableAnchorOffset.isUndefined()) {
+        std::vector<AnchorOffsetPair> anchorOffsets;
+        auto variableAnchorOffset = layout->evaluate<TextVariableAnchorOffset>(zoom, feature, canonicalID);
+        if (!variableAnchorOffset.empty()) {
+            // Convert offsets from EM to PX, and apply baseline shift
+            for (const auto& anchorOffset : variableAnchorOffset) {
```
random3940, line 371: And here too vector::reserve() might help. Check other places where we know size ahead of time? I expect those arrays to be small, but it's still a good practice to do where we can.

src/mbgl/renderer/buckets/symbol_bucket.cpp
```diff
@@ -356,6 +356,27 @@ bool SymbolBucket::hasFormatSectionOverrides() const {
     return *hasFormatSectionOverrides_;
 }
 
+bool SymbolBucket::hasVariableTextAnchors() const {
+    auto hasTextVariableAnchorOffset = [&]() -> bool {
+        auto tvao = layout->get<TextVariableAnchorOffset>();
+        const bool hasAnchorOffset = tvao.match([](const auto&) { return true; });
```
random3940, line 362: Curious, what case this condition covers? When property is not set at all?

src/mbgl/style/expression/value.cpp
```diff
@@ -8,37 +8,40 @@ namespace style {
...
+
+            return type::Array(itemType.value_or(type::Value), arr.size());
+        });
 }
 
 std::string toString(const Value& value) {
     return value.match([](const NullValue&) { return std::string(); },
-                       [](const Color& c) { return c.stringify(); }, // avoid quoting
+                       [](const Color& c) { return c.stringify(); },                         // avoid quoting
+                       [](const VariableAnchorOffsetCollection& v) { return v.toString(); }, // avoid quoting
```
random3940, line 44: nit: duplicated comment?

src/mbgl/layout/symbol_instance.hpp
```diff
@@ -183,7 +184,7 @@ class SymbolInstance {
     std::optional<size_t> getPlacedIconIndex() const { return placedIconIndex; }
     std::optional<size_t> getPlacedVerticalIconIndex() const { return placedVerticalIconIndex; }
     float getTextBoxScale() const { return textBoxScale; }
-    std::array<float, 2> getVariableTextOffset() const { return variableTextOffset; }
+    std::optional<VariableAnchorOffsetCollection> getTextVariableTextOffset() const { return textVariableAnchorOffset; }
```
random3940, line 187: typo - should be getTextVariableAnchorOffset()?

include/mbgl/util/variable_anchor_offset_collection.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include <mbgl/style/types.hpp>
+#include <mbgl/util/feature.hpp>
+#include <mbgl/util/geometry.hpp>
+
+#include <string>
+#include <unordered_map>
+#include <vector>
```
random3940, line 9: It looks like we need `#include <array>` here to fix node.js CI.

random3940: I think it can also fix qt-ci build.

src/mbgl/renderer/buckets/symbol_bucket.cpp
wangyingfang: add isExpression in PossiblyEvaluatedPropertyValue class to check TextVariableAnchorOffset is expression.

src/mbgl/style/expression/value.cpp
wangyingfang: no, will revert the change in Color line next commit.

wangyingfang: It was change when do clang format

wangyingfang:
@random3940 please help review again.
@louwers could you help run the workflows?

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -358,6 +359,69 @@ std::array<float, 2> SymbolLayout::evaluateVariableOffset(style::SymbolAnchorTyp
     return result;
 }
 
+std::optional<VariableAnchorOffsetCollection> SymbolLayout::getTextVariableAnchorOffset(const SymbolFeature& feature) {
```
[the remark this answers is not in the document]
wangyingfang: I will keep use optional. I think it is more clear for me.

src/mbgl/renderer/buckets/symbol_bucket.cpp
```diff
@@ -356,6 +356,22 @@ bool SymbolBucket::hasFormatSectionOverrides() const {
     return *hasFormatSectionOverrides_;
 }
 
+bool SymbolBucket::hasVariableTextAnchors() const {
+    auto hasTextVariableAnchorOffset = [&]() -> bool {
```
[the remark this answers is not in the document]
wangyingfang: I think use lambda is more clear, so keep it.

include/mbgl/util/variable_anchor_offset_collection.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
```
louwers, line 1: This should not be in `util`. This directory is for utilities that are useful across the codebase, this code is too specific.

platform/darwin/src/MLNStyleValue_Private.h
```diff
@@ -239,6 +240,31 @@ class MLNStyleValueTransformer {
   // Color
   void getMBGLValue(MLNColor *rawValue, mbgl::Color &mbglValue) { mbglValue = rawValue.mgl_color; }
 
+  // VariableAnchorOffsetCollection
+  void getMBGLValue(id rawValue, mbgl::VariableAnchorOffsetCollection &mbglValue) {
+    if ([rawValue isKindOfClass:[NSArray class]]) {
+      NSArray *array = (NSArray *)rawValue;
+      if (array.count % 2 != 0) {
+        [NSException raise:NSInvalidArgumentException
+                    format:@"VariableTextAnchorOffset array should have even number of elements."];
```
louwers, line 249: Grammar: should be `VariableTextAnchorOffset array should have an even number of elements.`

src/mbgl/style/conversion/constant.cpp
```diff
@@ -136,6 +136,55 @@ std::optional<Padding> Converter<Padding>::operator()(const Convertible& value,
...
+std::optional<VariableAnchorOffsetCollection> Converter<VariableAnchorOffsetCollection>::operator()(
+    const Convertible& value, Error& error) const {
+    if (!isArray(value)) {
+        error.message = "value must be an array";
+        return std::nullopt;
+    }
+
+    const auto arraySize = arrayLength(value);
+    if (arraySize < 1 || arraySize % 2 != 0) {
+        error.message = "array must containing an even number";
```
louwers, line 148:
```suggestion
        error.message = "array must containing an even number of elements";
```

wangyingfang:
@louwers @random3940 please help review again.

random3940:
LGTM!

louwers:
Thank you, having another look over it.

include/mbgl/util/interpolate.hpp
```diff
@@ -143,6 +147,27 @@ struct Interpolator<Padding> {
...
+                                              const float t) const noexcept {
+        assert(a.size() == b.size());
+        std::vector<AnchorOffsetPair> offsetMap;
+        for (size_t index = 0; index < a.size(); index++) {
+            const auto& aPair = a[index];
+            const auto& bPair = b[index];
+            assert(aPair.anchorType == bPair.anchorType);
+            auto offset = std::array<float, 2>{interpolate(aPair.offset[0], bPair.offset[0], t),
+                                               interpolate(aPair.offset[1], bPair.offset[1], t)};
+            offsetMap.emplace_back(aPair.anchorType, offset);
```
louwers, line 164: This can throw which would crash the program (because of `noexcept`). Is it possible to avoid allocation?

random3940: Uhm, I also didn't like the idea of vector in this type, but I don't know how to avoid it. It's a collection without static size limit, so we have to allocate it on heap. 

I guess it's possible to do some shenanigans with inline storage for cases when there are only few elements. But this wouldn't fix the issue, it will just make it less probable. We are open to try ideas if you have some!

louwers: If it is unavoidable we should remove `noexcept` here (and from the base class), otherwise the program will crash when an exception is thrown here.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -358,6 +359,69 @@ std::array<float, 2> SymbolLayout::evaluateVariableOffset(style::SymbolAnchorTyp
...
+                variableTextOffset = {{layout->evaluate<TextOffset>(zoom, feature, canonicalID)[0] * util::ONE_EM,
+                                       layout->evaluate<TextOffset>(zoom, feature, canonicalID)[1] * util::ONE_EM}};
+            }
+
+            std::vector<AnchorOffsetPair> anchorOffsets;
+            anchorOffsets.reserve(variableTextAnchor.size());
+            for (auto anchor : variableTextAnchor) {
+                auto offset = variableTextOffset;
+                offset = SymbolLayout::evaluateVariableOffset(anchor, offset);
+                anchorOffsets.emplace_back(AnchorOffsetPair{anchor, offset});
```
louwers, line 415: This will construct an `AnchorOffsetPair` and then use the move constructor. Exactly the same as wen using `push_back`. So I would use `push_back` here.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -358,6 +359,69 @@ std::array<float, 2> SymbolLayout::evaluateVariableOffset(style::SymbolAnchorTyp
...
+            } else {
+                variableTextOffset = {{layout->evaluate<TextOffset>(zoom, feature, canonicalID)[0] * util::ONE_EM,
+                                       layout->evaluate<TextOffset>(zoom, feature, canonicalID)[1] * util::ONE_EM}};
+            }
+
+            std::vector<AnchorOffsetPair> anchorOffsets;
+            anchorOffsets.reserve(variableTextAnchor.size());
+            for (auto anchor : variableTextAnchor) {
+                auto offset = variableTextOffset;
+                offset = SymbolLayout::evaluateVariableOffset(anchor, offset);
```
louwers, line 414: Maybe use two different variables here?

src/mbgl/style/conversion/constant.cpp
```diff
@@ -136,6 +136,55 @@ std::optional<Padding> Converter<Padding>::operator()(const Convertible& value,
...
+        std::optional<float> xOffset = toNumber(arrayMember(offsetValue, 0));
+        std::optional<float> yOffset = toNumber(arrayMember(offsetValue, 1));
+        if (!xOffset || !yOffset) {
+            error.message = "anchor offset must have two numbers";
+            return std::nullopt;
+        }
+
+        auto anchorType = anchor.value();
+        auto offset = std::array<float, 2>{xOffset.value(), yOffset.value()};
+        collection.emplace_back(AnchorOffsetPair{anchorType, offset});
```
louwers, line 182:
```suggestion
        collection.push_back(AnchorOffsetPair{anchorType, offset});
```

louwers (approved):
Thank you for this excellent contribution!

I left a few more small comments, let's merge after those are addressed.

@random3940 Feel free to also formally approve it using GitHub's UI.

include/mbgl/util/interpolate.hpp
wangyingfang: Add runtime error as JS

wangyingfang:
@louwers @random3940 please help review again.

random3940:
@louwers, iOS C++ test fails with this error: 
```
<Code>SignatureDoesNotMatch</Code>
<Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
```
It looks like some CI glitch to me. Just in case I've re-ran tests from this branch locally and they pass.

random3940 (approved):

louwers:
I will take a look at it.

louwers (approved):

PR #2858, opened 2024-09-20
Enable crash mitigation when running OpenGL in the Android emulator

mwilsnd:
In the Android Open Source Project, there exists a translation layer that sits between an OpenGL client application and the driver, called Goldfish. For whatever reason, MapLibre's OpenGL backend crashes Goldfish inside `GL2Encoder::s_glDrawElements`. The following mitigation addresses this crash by first detecting if we're in an emulated environment and then deleting all drawables every frame. The crash appears to be related to buffer re-use, but I've yet to locate a minimum reproduction sample that triggers this crash.

Just deleting all drawables every frame and rebuilding them works around the reuse problem, at the cost of emulator performance. It sure beats crashing though.
 
Closes #2791

PR #2927, opened 2024-10-11
Index buffers reverted to dirty flag instead of timestamps

alexcristici:
The index buffers were recreated for each drawable even though 2 or more drawables used the same index buffer. 

We need to recreate index buffers only when the buffer data is dirty.
New index buffer will be shared through the `indexes` variable for all drawables that are using it.

This change also fixes the Android emulator crash. 

I also removed some unused variables: `indexBuffer`, `attributeBuffers` and `instanceBuffers`.


louwers:
Could you provide some more context in the PR description?

alexcristici:
> Could you provide some more context in the PR description?

Sure, not ready yet, I thought I created as a draft. I will change now.

alexcristici:
Ready for review.

platform/android/MapLibreAndroid/src/cpp/map_renderer.cpp
```diff
@@ -257,7 +257,7 @@ void MapRenderer::onSurfaceCreated(JNIEnv& env, const jni::Object<AndroidSurface
     // If we're running the emulator with the OpenGL backend, we're going to crash eventually,
     // unless we enable this mitigation.
     if (inEmulator()) {
-        renderer->enableAndroidEmulatorGoldfishMitigation(true);
```
alexcristici, line 260: @mwilsnd For now I just disabled old workaround for Android crash. I would remove the code after you confirm is not crashing on your side too.

platform/android/MapLibreAndroid/src/cpp/map_renderer.cpp
```diff
@@ -257,7 +257,7 @@ void MapRenderer::onSurfaceCreated(JNIEnv& env, const jni::Object<AndroidSurface
...
+        renderer->enableAndroidEmulatorGoldfishMitigation(false);
```
alexcristici, line 260: @mwilsnd For now I just disabled old workaround for Android crash. I would remove the code after you confirm is not crashing on your side too.

sjg-wdw:
@TimSylvester cool with you?

TimSylvester (approved):

mwilsnd (approved):
Looks to be resolved, nice find!

alexcristici:
Removed workaround for Android emulator crash: https://github.com/maplibre/maplibre-native/pull/2858.

PR #2932, opened 2024-10-14
Enable CMake Unity build for Android

louwers:
> When this property is set to true, the target source files will be combined into batches for faster compilation. This is done by creating a (set of) unity sources which #include the original sources, then compiling these unity sources instead of the originals. This is known as a Unity or Jumbo build.

https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html

Experiment to allow for making [unity builds](https://github.com/j-jorge/unity-build) with CMake. This has the potential to significantly speed up build times.

Edit: looks like for Android `libmaplibre.so` compilation time goes from `2m 6s` to `53s`.


ntadej:
Can you maybe provide a bit of a description what is that and why is it useful? 😊

louwers:
@ntadej Yeah I just opened it to see if it builds. 🙈

PR #2949, opened 2024-10-18
Fix descriptor pool fragmentation

adrian-cojocaru:
Added a descriptor pool per descriptor type to avoid fragmentation.


TimSylvester (approved):

src/mbgl/vulkan/context.cpp
```diff
@@ -27,8 +27,20 @@
 #include <algorithm>
 #include <cstring>
 
-#ifndef MLN_VULKAN_DESCRIPTOR_POOL_SIZE
-#define MLN_VULKAN_DESCRIPTOR_POOL_SIZE 30000
+#ifndef MLN_VULKAN_GLOBAL_DESCRIPTOR_POOL_SIZE
+#define MLN_VULKAN_GLOBAL_DESCRIPTOR_POOL_SIZE 3 * 4
```
louwers, line 31: Could you use `constexpr` here? Ideally without all caps.

adrian-cojocaru: They're used as defines so they can be overridden with build time configs if needed. I don't have a strong preference either way since they're just initial values in resizable containers.

louwers: I suggest we use `constexpr` until someone comes along that actually wants to set these with a build config.

louwers (approved):

alexcristici (approved):

PR #2956, opened 2024-10-22
Change `parameters.currentLayer` to index instead of layerIndex

alexcristici:
In order to compare the `parameters.currentLayer` correctly with `parameters.opaquePassCutoff` we need to compute the `parameters.currentLayer` every frame instead of using the `layerGroup.getLayerIndex()`.

I also removed `drawable.getLayerIndex()`.


src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -368,37 +371,36 @@ void Renderer::Impl::render(const RenderTree& renderTree,
...
-        parameters.depthRangeSize = 1 -
-                                    (maxLayerIndex + 3) * PaintParameters::numSublayers * PaintParameters::depthEpsilon;
+        parameters.depthRangeSize = 1 - (orchestrator.numLayerGroups() + 2) * PaintParameters::numSublayers *
+                                            PaintParameters::depthEpsilon;
 
         // draw layer groups, opaque pass
+        int32_t i = static_cast<int32_t>(orchestrator.numLayerGroups()) - 1;
         orchestrator.visitLayerGroups([&](LayerGroupBase& layerGroup) {
-            parameters.currentLayer = layerGroup.getLayerIndex();
+            parameters.currentLayer = i--;
```
TimSylvester, line 381: It looks like the counting direction is swapped between 3D and opaque passes.

alexcristici: Yes, the opaque pass is in reversed order in legacy, but on drawable is not so I had to adjust the index. I will explore drawing opaque layers in reverse order as on legacy in different PR. I think they did it like this to optimize the rendering based on depth test.

TimSylvester: Interesting that it never caused any problems.

alexcristici: It affects the `opaqueCutOff` logic, so the fill-extrusion depth logic. Until now I saw the `currentLayer` it is used only there.

src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -368,37 +371,36 @@ void Renderer::Impl::render(const RenderTree& renderTree,
...
         const auto debugGroup(parameters.renderPass->createDebugGroup("drawables-translucent"));
-        const auto maxLayerIndex = orchestrator.maxLayerIndex();
         parameters.pass = RenderPass::Translucent;
-        parameters.depthRangeSize = 1 -
-                                    (maxLayerIndex + 3) * PaintParameters::numSublayers * PaintParameters::depthEpsilon;
+        parameters.depthRangeSize = 1 - (orchestrator.numLayerGroups() + 2) * PaintParameters::numSublayers *
+                                            PaintParameters::depthEpsilon;
 
         // draw layer groups, translucent pass
+        int32_t i = static_cast<int32_t>(orchestrator.numLayerGroups()) - 1;
```
louwers, line 393: Can't setting the current layer index be set by `visitLayerGroups`, or maybe have it pass the correct layer index to the lambda? Capturing a local variable by reference in this way looks noisy and feels error-prone.

TimSylvester (approved):

src/mbgl/renderer/renderer_impl.cpp
alexcristici: Should be fixed now.

louwers (approved):

Issue #2879, opened 2024-09-27
Android: Location puck switch icon on rotating camera


Fabi755:
**Describe the bug**
We have a really strange behavior of the location icon. And I am at the end of my ideas.

I'm currently working on upgrading the MapLibre navigation to v11. After upgrade (from v11.0.2-pre0) to v11.1.0 (and above) the location puck gets flickering by rotating. After more research we found out that the icon goes to the `maplibre_user_stroke_icon` drawable. The puck is going back to the right drawable, when zoom in/out, or update the location.

The `maplibre_user_stroke_icon` is used by two styles (`maplibre_backgroundDrawable` & `maplibre_backgroundDrawableStale`). Changing the styles has no effect to displayed icon in my tests.

I can't see this issue on the sample app (build from source). But I see this issue also in a new project from scratch. I can't see this issue on my Emulator running on a Mac.

**To Reproduce**

1. Clone [my sample repository](https://github.com/Fabi755/maplibre-puck-issue/tree/main)
2. Run app on a real device
3. Allow using location, to make things work
4. Rotate map
5. Puck is flickering between GPS puck and red dot (`maplibre_user_stroke_icon`)

(Use the enable/disable) button to see the stroke-icon state, without location updates. If disabled, only zoom in/out goes back to the GPS puck.

**Expected behavior**
Icon should not changed to the stroke icon, and the GPS puck should always showed.

**Screenshots**

https://github.com/user-attachments/assets/0c5f889e-e8fd-4f68-87b1-6079ec1fb8b3

**Platform information (please complete the following information):**
- OS: Android
- Platform: Android
- Version: MapLibre 11.1.0 - 11.5.0

**Additional context**

alasram:
I'm having the exact same problem but it's only reported in 11.5. (11.1 seems to not have this issue)

alasram:
The VAO of the Drawable is changing which means we're running vertex input commands outside the drawable that are affecting the bound VAO.
A solution is to unbind the VAO after drawing but a correct solution would ensure we never run commands that affect VAO outside a Drawable object.
This code works for me:
https://github.com/alasram/maplibre-native-experimental/commit/473fdfa784f73e337c9b7edee93bc95eb3815237

louwers:
@alasram Could you make a PR?

alasram:
@boldtrn @Fabi755 Can you confirm the issue is fixed in Android v11.6.1? We can then close the issue

boldtrn:
[...]

PR #2957, opened 2024-10-22
Fix puck rendering behavior Android

alasram:
The puck is typically rendered last and while its VAO is bound command that change the VAO are executed causing glitches. I didn't debug what commands are affecting the VAO but as a workaround the VAO is unbound after drawing which ensures the state of a Drawable object remains clean without adding overhead. This works around #2879

### DIFF
--- a/src/mbgl/gl/drawable_gl.cpp
+++ b/src/mbgl/gl/drawable_gl.cpp
@@ -71,6 +71,8 @@ void DrawableGL::draw(PaintParameters& parameters) const {
             context.draw(glSeg.getMode(), mlSeg.indexOffset, mlSeg.indexLength);
         }
     }
+    // Unbind the VAO so that future buffer commands outside Drawable do not change the current VAO state
+    glBindVertexArray(0);
 
 #ifndef NDEBUG
     context.bindVertexArray = value::BindVertexArray::Default;
--- a/src/mbgl/gl/drawable_gl.cpp
+++ b/src/mbgl/gl/drawable_gl.cpp
@@ -72,11 +72,9 @@ void DrawableGL::draw(PaintParameters& parameters) const {
         }
     }
     // Unbind the VAO so that future buffer commands outside Drawable do not change the current VAO state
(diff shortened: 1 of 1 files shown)


TimSylvester:
Curious, I don't think that should be necessary.  Doesn't it mean that either those commands shouldn't be happening at all, or that they are happening with the wrong bindings?

louwers:
Android render test timing out is just because we were running benchmarks, should restart...

louwers (approved):

alasram:
@TimSylvester I haven't really debug this. I only noticed in renderdoc that the buffers bound to the VAO change when the glitch happens. This means that while the puck VAO is bound (It's the last drawable in the frame), and before the next frame's drawable, there are calls that bind buffers such as `glBindBuffer(GL_ELEMENT_ARRAY_BUFFER)`

adrian-cojocaru:
I haven't been able to reproduce the glitch, but from what @alasram described a single `glBindVertexArray(0)` at the end of the frame (`context::endFrame`) should block any further updates to the puck's VAO.

alasram:
This is a legit fix.
After puck rendering, the upload pass might create or update index buffers. This can lead to issues with VAO bindings, as the newly created/updated buffers are bound to `GL_ELEMENT_ARRAY_BUFFER` without considering the active VAO: calls to `UploadPass::createIndexBufferResource` or `UploadPass::updateIndexBufferResource`
This PR is enough to address the issue. An alternative approach, though less conservative, is to unbind the VAO at the start of the upload pass (reducing GL calls). Unless there's evidence of performance benefits with VAOs in a specific implementation, the simpler approach is bind a dummy VAO when a context is created and bind vertex buffers directly for each drawable.
I recommend merging this PR as it provides a straightforward solution to a real bug.

TimSylvester:
Sorry, I was out since 10/30.  Something must be making GL calls without binding its own buffer first but if it's not leading to an observed bug, I guess it's not a big deal.

Issue #2445, opened 2024-05-29
Tracking Issue MapLibre GL JS Render Test Parity


louwers:
*Will be filled in*.

PR #2958, opened 2024-10-22
Add Tile LOD controls

alasram:
Expose tile level of details (LOD) parameters to Android. Controlling the LOD gives MapLibre clients the ability to control performance scalability. This is already exposed to some level with the parameter `pixelRatio` but the option `pixelRatio` changes the view so it's typically set once when the application starts and the styling and camera need to be setup for the chosen `pixelRatio`.
This PR exposes controlling an offset (shift) to the zoom level. If performance drops below a certain FPS, an application can decide to lower the zoom level to reduce the number of processed and displayed tiles.
Similarly when the screen is tilted (Often the case during active Navigation on large display infotainment devices), tiles away from the camera could use lower zoom levels to maintain a certain level of performance. This PR exposes controlling the pitch angle threshold and the distance to camera to lower zoom levels.

Example:

What we currently get on a 4K display with 45 degrees pitch
default

Example where zoom is offset by +0.75 (usually a negative value would be used to improve performance):
zomm_add_0 75

Example where the pitch and distance to camera parameters allow for a lower zoom levels away from the camera view target:
reduced_lod_at_low_pitch


louwers (changes requested):
Very useful and MapLibre Native should definitely include this functionality in my opinion.

A few things:

- MapLibre GL JS has a similar mechanism. We should aim to be aligned here. See debug tiles: https://demotiles.maplibre.org/debug-tiles/#9/0/0/0/70 @HarelM can probably give more info.
- We should expose this functionality to iOS as well or at least create a tracking issue.
- The Android sample app should be updated with a simple demo for functionality.
- Needs better test coverage, both C++ Unit tests, render tests(?) as well as for the Android platform bindings. Maybe there are some [render tests from MapLibre GL JS](https://github.com/maplibre/maplibre-native/issues/2445) that we can port?
- Tests seem to be failing.

Let me know what you think.

louwers:
We're going to have a meeting about this PR and try to get it over the line together with @alexcristici @TimSylvester @sjg-wdw @yousifd do you want to join as well?

HarelM:
@NathanMOlson is probably the right address for questions about the relevant PR in gl-js.
Note that we have added a way to override the method to control the level of details, allowing a fine grained control in this flow.

NathanMOlson:
> We're going to have a meeting about this PR and try to get it over the line together with @alexcristici @TimSylvester @sjg-wdw @yousifd do you want to join as well?


I'm happy to help you understand how gl-js does it.

The native algorithm is currently much different from gl-js. I see a problem in one of the examples that would be resolved by using the gl-js algorithm. The tiles labeled "1" and "2" are closer to the camera than the tile labeled "3", but are rendered at a lower zoom level. This should not happen (except due to zoom quantization, which is not in play in this example).

image

sjg-wdw:
The way I've worked around this problem in the past is to calculate an adjusted screen space value based on how much of the polygon is on screen.  So if takes up 1000 pixels and only a third of it is on screen, you weight it at 3000 pixels.  Otherwise you get strange results toward the edges.

I don't think that's what's happening here, as MapLibre isn't really doing a screen real estate calculation.  But I'd like for us to dig into what this adjustment would do.

yousifd:
@louwers This line in `tile_pyramid.cpp` is causing the tests to fail. I'm not sure what @alasram was trying to avoid here exactly and I'm wondering if clamping it to `parameters.transformState.getMaxZoom()` and `parameters.transformState.getMinZoom()` would make more sense?
```
    // Optionally shift the zoom level
    double zoom = util::clamp<double>(
        parameters.transformState.getZoom() + parameters.tileLodZoomShift, zoomRange.min, zoomRange.max);
```

sjg-wdw:
Still need to figure out the zoom level clamping.
Let's mark this new API as experimental.

We'll try to get this more compatible with the gl-js version in the future, but we don't have the resources right now.

adrian-cojocaru (approved):

test/tile/tile_lod.test.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <mbgl/style/sources/custom_geometry_source.hpp>
+#include <mbgl/test/util.hpp>
+
+#include <filesystem>
+
+using namespace mbgl;
+using namespace mbgl::util;
+using namespace ::testing;
+
+//#define UPDATE_EXPECTED
```
louwers: Could you remove this comment?

louwers (approved):
Thanks for adding the tests. Looks good. :)

louwers:

NathanMOlson:
I would like to use the tile selection algorithm from MapLibre GL JS in Native (see https://github.com/maplibre/maplibre-gl-js/pull/4779 and https://github.com/maplibre/maplibre-gl-js/pull/5719).

Would this change be accepted as a PR in Native, and if so should it replace the existing algorithm or be included alongside it?

sjg-wdw:
> I would like to use the tile selection algorithm from MapLibre GL JS in Native (see [maplibre/maplibre-gl-js#4779](https://github.com/maplibre/maplibre-gl-js/pull/4779) and [maplibre/maplibre-gl-js#5719](https://github.com/maplibre/maplibre-gl-js/pull/5719)).
> 
> Would this change be accepted as a PR in Native, and if so should it replace the existing algorithm or be included alongside it?

We looked at it as we were evaluating this PR.  The big problem at the time was that the JS one was pretty janky.  A number of bad corner cases easily triggered.  Is it sufficiently fixed now?  Is there something this one doesn't do that you need?

NathanMOlson:
> Is it sufficiently fixed now?

I'm not aware of any bad corner cases, but that doesn't mean there aren't any. GL JS Behaves how I want it to in all the cases I've used.

> Is there something this one doesn't do that you need?

Yes. I want points closer to the camera to be rendered in greater detail than points further from the camera. The current native algorithm renders the center ("look at" point) in the highest detail, and decreasing detail away from the center.

sjg-wdw:
I'm not sure if your demo site is out of date, so this may no longer hold true.  I poked around a little bit and got it to do this.

Screenshot 2026-01-15 at 4 26 19 PM

Tile loading in 3D is a profoundly hard problem and it really needs a much more complex solution.  The one being used on mobile is just a minor tweak, which keeps the complexity at bay.  It's also not a very satisfying solution, so I can understand the urge for more.

The JS solution struck me as attempting to be comprehensive, but falling short in these sorts of corner cases.  That works better for JS as they're not held to the sort of frame rate and resource constraints mobile is.  If we start thrashing tile loading, for example, I'm going to have to explain it to our sponsor, explain how it got like that and what the plan is for reverting it.  Not a lot of fun.

So if the JS solution comes to mobile, I'd ask that it be opt-in and off by default.  The Hudhud Maps folks might also have some thoughts as they're using the existing tilt in production.

NathanMOlson:
> I'm not sure if your demo site is out of date, so this may no longer hold true.

The [demo site](https://nathanmolson.github.io/lod_control) is out of date, although it shows a version that I personally prefer to the one that is currently in GL JS. :)

> I poked around a little bit and got it to do this

That's an interesting one that certainly looks undesireable! The only way I've been able to reproduce anything like that is to decrease the "Tile Zoom Deadband" parameter below 0. While the demo site lets you play with a wide range of parameters, Zoom Rate would be limited to the range [0,2] and Tile Zoom Deadband would be non-negative in practical use.

There are also other parts of the code involved in displaying the tile, which sometimes cause a different set of tiles to be displayed than what is returned by the tile cover algorithm. 

Putting aside the details of the GL JS implementation, the result I want is

`tileScale` = `k` / `distanceToTile` * `cos(tileOffNadirAngle)^N`.

Adjusting `N` from 0 to 1/2 to 1 causes tiles to be displayed at equal screen width, equal screen area, or equal screen height. ("Equal" meaning "within a factor of 2" due to the quantization of tile zoom).

sjg-wdw:
> Putting aside the details of the GL JS implementation, the result I want is
> 
> `tileScale` = `k` / `distanceToTile` * `cos(tileOffNadirAngle)^N`.
> 
> Adjusting `N` from 0 to 1/2 to 1 causes tiles to be displayed at equal screen width, equal screen area, or equal screen height. ("Equal" meaning "within a factor of 2" due to the quantization of tile zoom).

That's sounds nicely constrained.  You could add that as a new center mode, for example, leaving the default in place. Then it's opt-in for anyone who wants to use it.  If it's got some wonky corner cases, not a big deal.

In the end everything like this is going to have problems.  The proper way to do this work is to measure the screen area taken up by a tile and have a hard limit on both screen area and number of tiles.  But the logic for that gets complex once you mix in projections and a globe.

PR #2959, opened 2024-10-23
Fix LineBucket::addGeometry() empty coordinates.

clinthidinger:
This fix avoids using a negative number to index the coordinates inside the "len" lambda in LineBucket::addGeometry().  It does an early return similar to the one on line ~90 for invalid geometry.

### DIFF
--- a/src/mbgl/renderer/buckets/line_bucket.cpp
+++ b/src/mbgl/renderer/buckets/line_bucket.cpp
@@ -54,6 +54,10 @@ void LineBucket::addFeature(const GeometryTileFeature& feature,
 void LineBucket::addGeometry(const GeometryCoordinates& coordinates,
                              const GeometryTileFeature& feature,
                              const CanonicalTileID& canonical) {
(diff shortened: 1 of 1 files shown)


cgalvan (approved):
Can we add a unit test for this?

clinthidinger:
> Can we add a unit test for this?

Are there docs for how to run the unit tests?  When I run the mbgl-test-runner project, it says: 

[==========] Running 0 tests from 0 test suites.
[==========] 0 tests from 0 test suites ran. (1 ms total)
[  PASSED  ] 0 tests.

I tried filtering for Buckets.LineBucket, but that didn't help.

louwers:
That should work. What platform are you on? Could be that we don't have it set up properly for Windows.

clinthidinger:
> That should work. What platform are you on? Could be that we don't have it set up properly for Windows.

I'm on Windows 11 with GLFW as the platform.

louwers (approved):

PR #2965, opened 2024-10-25
Encode the style URL in iOS

qqz003:
For iOS versions under 17.0, `[NSURL URLWithString:]` cannot create an `NSURL` object with the url string that containing any characters that need to be encoded. Consequently, it cannot load the custom style correctly if the style file contains such a URL.

### DIFF
--- a/platform/darwin/src/http_file_source.mm
+++ b/platform/darwin/src/http_file_source.mm
@@ -226,8 +226,9 @@ BOOL isValidMapboxEndpoint(NSURL *url) {
 MLN_APPLE_EXPORT
 NSURL *resourceURL(const Resource& resource) {
     
-    NSURL *url = [NSURL URLWithString:@(resource.url.c_str())];
-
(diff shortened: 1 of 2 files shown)


louwers:
Please add a unit test to `http_file_source.test.cpp`.

qqz003:
> Please add a unit test to `http_file_source.test.cpp`.

The method `resourceURL` is not defined in `http_file_source.hpp`; it's only used for Apple platforms. So, maybe it's better to add the test to the class `MLNResourceTests`? However, I didn't find how to run the Darwin tests. They don't appear in the test navigator of Xcode, and I didn't find the related workflow (maybe included in some other workflow?).

louwers:
@qqz003 Yes that sounds good.

The `ios_test` test target should include this test.

image

louwers (approved):

qqz003:
@louwers Looks that I don't have the write access to merge the PR, could you please help to merge it? Thank you!

louwers:
@qqz003 Thanks for the reminder!

louwers:
@qqz003 @random3940 I had to revert this change due to regressions.

Happy to work with you to address this in another way...

Issue #917, opened 2023-03-20
Run Windows build on CI with MBGL_EGL = ON


louwers:
We have a regression because some files were not compiled on CI.

If someone could look into making sure we compile these files on CI, that would prevent regressions in the future.

This is the file in question (there might be more).

```
platform/windows/src/headless_backend_egl.cpp
```

louwers:
Note CI is also not running: https://github.com/maplibre/maplibre-gl-native/pull/918

acalcutt:
How would you want to do this? It would need to be built with different options since the default node build has MBGL_EGL=OFF . We would have to add the -DMBGL_EGL=OFF option to the configure step ( https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/node-ci.yml#L163-L169 ) , and if we did that it wouldn't test the non-EGL mode.

I could maybe add a variable to the matirix and add a few more os runs.... or maybe maybe a copy of this ci that runs an EGL build...

acalcutt:
I think ci don't run on that PR because the paths are restricted and don't include the windows/linux/macos playform.
https://github.com/maplibre/maplibre-gl-native/blob/main/.github/workflows/node-ci.yml#L30-L45

etnav:
Flag is called `MBGL_WITH_EGL`, was a typo I made :smile:

PR #2968, opened 2024-10-25
Add Windows to CI tests

tdcosta100:
Until now, except for Node-CI, Windows didn't have tests yet. So now I'm adding them, based on Linux-CI. I created a matrix with 4 renderers (OpenGL, EGL, Vulkan and OSMesa), and 2 rendering modes (Legacy and Drawable). Two combinations are disabled:

* EGL/Drawable (the tests are throwing exceptions, I don't know what causes it or how to fix. [More details](https://github.com/maplibre/maplibre-native/actions/runs/11513187647/job/32066096165))
* Vulkan/Legacy - since this combination does not exist in Linux, I'm doing the same in Windows

I left some considerations on ignored tests, so others can investigate and fix them in the future. Also, some unit and render tests appear to have specific Windows issues (like font rendering, for example, which will always have some differences, and some number precision issues), so I tweaked these cases, but this is not a definitive solution and suggestions are welcome.

Closes https://github.com/maplibre/maplibre-native/issues/917


test/src/mbgl/test/sqlite3_test_fs.cpp
```diff
@@ -234,16 +234,21 @@ static int sqlite3_test_fs_access(sqlite3_vfs* vfs, const char* zPath, int flags
...
 }
 
 namespace mbgl {
 namespace test {
 
 SQLite3TestFS::SQLite3TestFS() {
-    sqlite3_vfs* unix_fs = sqlite3_vfs_find("unix");
-    if (unix_fs == nullptr) {
+#ifdef WIN32
+    sqlite3_vfs* os_fs = sqlite3_vfs_find("win32");
```
louwers, line 246: Maybe try passing `nullptr` here?

> If zVfsName is NULL then the default VFS is returned.

https://www.sqlite.org/c3ref/vfs_find.html

test/CMakeLists.txt
```diff
@@ -185,7 +185,7 @@ else()
         ${PROJECT_SOURCE_DIR}/test/src/mbgl/test/http_server.cpp
         PROPERTIES
         COMPILE_FLAGS
-        -Wno-shadow
+        $<$<NOT:$<BOOL:${WIN32}>>:-Wno-shadow>
```
louwers, line 188: Would 

```
$<$<OR:$<CXX_COMPILER_ID:GNU>,$<CXX_COMPILER_ID:Clang>>:-Wno-shadow>
```

be more appropriate here?

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,263 @@
...
+env:
+  SCCACHE_GHA_ENABLED: "true"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # INFO: We are cancelling the concurrency group if the change is on PR. For workflow dispatch, this will not work.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+permissions:
+  id-token: write         # needed for AWS
```
louwers, line 25: Not used by this workflow.

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,263 @@
...
+            ${{ env.renderer_flag_cmake }} `
+            ${{ env.rendering_mode_flag_cmake }}
+
+      - name: Build MapLibre Native Core
+        run: |
+          cmake --build build --target mbgl-core mbgl-test-runner mbgl-render-test-runner mbgl-expression-test mbgl-render mbgl-benchmark-runner
+
+      # mbgl-render (used for size test) & mbgl-benchmark-runner
+
+      - name: Upload mbgl-render as artifact
```
louwers, line 145: These two steps can be removed, since we don't do a size test or benchmark in a subsequent workflow like on Linux.

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,263 @@
...
+          path: |
+            build/mbgl-benchmark-runner.exe
+      - name: Configure AWS Credentials
+        if: matrix.renderer == 'opengl' && matrix.rendering_mode == 'drawable' && github.ref == 'refs/heads/main' && vars.OIDC_AWS_ROLE_TO_ASSUME 
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-west-2
+          role-to-assume: ${{ vars.OIDC_AWS_ROLE_TO_ASSUME }}
+          role-session-name: ${{ github.run_id }}
+      - name: Upload mbgl-render & mbgl-benchmark-runner to S3
```
louwers, line 167: This one should also be removed.

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,263 @@
...
+          path: |
+            metrics/windows-${{ matrix.renderer == 'osmesa' && 'opengl' || matrix.renderer }}.html
+      
+      # expression tests
+
+      - name: Run expression test
+        shell: bash
+        run: build/expression-test/mbgl-expression-test.exe
+
+      - if: github.event_name == 'pull_request'
```
louwers, line 249: Not needed

test/storage/online_file_source.test.cpp
```diff
@@ -13,6 +13,13 @@
 
 using namespace mbgl;
 
+#ifdef WIN32
+// Windows doesn't fail immediately like other OS
```
louwers, line 17: Can that be fixed?

test/src/mbgl/test/sqlite3_test_fs.cpp
tdcosta100: Yes, much better this way! Correcting now.

test/CMakeLists.txt
tdcosta100: I agree, this is better. Maybe even better `$<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wno-shadow>`, since only Visual Studio doen't support it?

.github/workflows/windows-ci.yml
tdcosta100: Removed!

.github/workflows/windows-ci.yml
tdcosta100: Removed!

.github/workflows/windows-ci.yml
tdcosta100: Removed!

.github/workflows/windows-ci.yml
tdcosta100: Removed!

test/storage/online_file_source.test.cpp
tdcosta100: I don't know if there is a better approach. Unix-like systems fail immediately, and Windows fails after waiting 2 seconds. How to deal with it? A separate file, maybe? A constant defined elsewhere? I'm open to suggestions.

louwers: It's fine this way.

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,230 @@
...
+
+      - name: Configure MapLibre Native Core
+        env:
+          CI: 1
+          VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
+          VCPKG_KEEP_ENV_VARS: CMAKE_CXX_COMPILER_LAUNCHER;CMAKE_C_COMPILER_LAUNCHER
+          CMAKE_C_COMPILER_LAUNCHER: sccache
+          CMAKE_CXX_COMPILER_LAUNCHER: sccache
+        shell: bash
+        run: |
```
louwers, line 106: Shellcheck detects an issue with this script:

```
Argument mixes string and array. Use * or separate argument
```

louwers: Since this is quite a large script, maybe it can be extracted in a file so it can be easily ran.

tdcosta100: You are right. Done!

tdcosta100: Now Shellcheck tests passed =D!

test/CMakeLists.txt
louwers: What other compilers support it? I would just stick to the ones that support it.

tdcosta100: I searched about it and I agree with you, better stick with the explicit compiler list to enable this flag.

.github/changed-files.yml
```diff
@@ -12,6 +12,23 @@ linux:
   - 'BUILD.bazel'
   - '.bazelrc'
   - '.bazelversion'
+windows:
```
louwers, line 15: By the way, the only reason I use this complicated setup is so that I can mark the Android / Linux / iOS workflows as required. You can also use https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

This means you only need one job.

louwers (approved):
Maybe we should disable the metrics for Windows. It's going to be a bit of a pain to have to update them for yet another platform when changing anything. Actually I think ideally we should have some workflow to update the metrics automatically on a PR.

Anyway, that is a concern for later. LGTM!

tdcosta100:
> Maybe we should disable the metrics for Windows. It's going to be a bit of a pain to have to update them for yet another platform when changing anything. Actually I think ideally we should have some workflow to update the metrics automatically on a PR.
> 
> Anyway, that is a concern for later. LGTM!

Well, maybe. But redoing metrics it's kind of easy to do. My biggest concern here is testing on Qt platform, so I still need @ntadej to update the Docker image to use CMake 3.24.

louwers:
@tdcosta100 I think the plan is to drop support for Qt5. Would that be sufficient?

.github/changed-files.yml
tdcosta100: I found this a little bit confusing because I'm not used to it. Are you saying I don't need to add a `windows` entry here? Because the jobs in `windows-ci.yml` are meant to test every possible build flag combination, so this setup is useful to me at least. Please let me know what parts are unnecessary, because I'm learning about it yet.

tdcosta100:
> @tdcosta100 I think the plan is to drop support for Qt5. Would that be sufficient?

I think so, but until then, how can we pass this PR in the tests?

louwers:
Either ignore the Qt workflows (they are not required for merging) or remove all matrix blocks in qt-ci that mention Qt 5. https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/qt-ci.yml#L80-L87

tdcosta100:
I will see what's going wrong and also update it to match the main branch.

tdcosta100:
Found the reason. With the release of CMake 3.31, `vcpkg` stopped working properly in CI. I will PR an update for it.

tdcosta100:
I fixed the Qt5 configuration in this branch, CMake version is bigger than 3.24, so everything works as indended. Now tests are passing flawlessly (at least for while).

ntadej:
Looks OK to me!

Sorry for silence, I was very busy lately.

louwers:
@ntadej No worries, thanks for having a look!

PR #2975, opened 2024-10-27
Use the recommended actionlint hook in pre-commit hooks

tdcosta100:
Currently the pre-commit hooks uses actionlint hook, but it does not work well with shellcheck, specially PowerShell scripts in workflow flies. Looking at the [README](https://github.com/Mateusz-Grzelinski/actionlint-py?tab=readme-ov-file), it recommends another way to use this hook, so I changed it and then the pre-commit hooks are not complaining about incorrect PowerShell script parsing anymore. So I'm bringing this change to the project.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,9 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
-    hooks:
-    -   id: check-yaml
-        args: [--allow-multiple-documents]
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v5.0.0
+  hooks:
+  - id: check-yaml
+    args: [--allow-multiple-documents]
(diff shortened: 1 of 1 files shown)


louwers (changes requested):
This does not work with pre-commit.ci https://results.pre-commit.ci/run/github/314553426/1730008081.gbyXcM-MSOyDlV3inPbm7g

As you can see, it's not running.

Also I think shellcheck is not enabled anymore after this change.

Maybe we can ignore the windows-ci file.

tdcosta100:
> Shellcheck would not be enabled anymore after this change. I didn't find a way to enable both shellcheck and actionlint other than the current setup.
> 
> Maybe we can ignore the windows-ci file.
> 
> By the way, shellcheck should not run for PowerShell: https://github.com/rhysd/actionlint/blob/main/docs/checks.md

You are right, using the configuration I suggest in this PR disables the spellcheck, so I'm dropping it. My problem was resolved by explicitly setting the shell configuration where the shellcheck error comes in, and then the errors disappeared.

Issue #2854, opened 2024-09-19
Move Android Examples Documentation out of mdBook Documentation


louwers:
The [mdBook Documentation](https://maplibre.org/maplibre-native/docs/book/) should be documentation for contributors of MapLibre Native.

A separate documentation site should be generated for Android Examples.

I plan on using MkDocs for this.

PR #2978, opened 2024-10-28
Split out Android examples to separate documentation site

louwers:
Create a separate documentation site for Android Examples. I wish it was possible to add examples to the API documentation, similar to the current Apple docs. However, that is not possible with Dokka at the moment.

I ported the existing example documentation and added some new content. I think it's best to merge now and then gradually add more examples.

image

Include code blocks from the sample app with [Pymdown Snippets](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/). This way we can ensure that the examples are up to date.

I fixed a small bug with the Snapshotter where it would try to render attribution even if it was empty (in the case of Demotiles). If needed I can split this out in a separate PR.

See also https://github.com/maplibre/maplibre-native/issues/2854


louwers:
@cgalvan Could you try building it locally? You'll need Docker. Instructions are part of the PR.

platform/android/docs/lat-lng-bounds.md
```diff
@@ -0,0 +1,8 @@
+# Using the LatLngBounds camera API
+
+!!! info
+    This guide is a work in progress.
```
cgalvan, line 4: Do we already have a backlog item to flesh this out?

cgalvan (approved):

platform/android/docs/lat-lng-bounds.md
louwers: I'll include it in this PR.

Issue #2829, opened 2024-09-13
Investigate `render-tests/icon-text-fit/textFit-grid-long` failing


louwers:
This render test was added as part of #2780.

There are some pixel differences for iOS (Metal) and Android (OpenGL and Vulkan). We should add some alternate expected images and enable this render test.

random3940:
@louwers, closing this per @qqz003 PR.

PR #2983, opened 2024-10-31
Enable and update baseline for "icon-text-fit/textFit-grid-long"

qqz003:
Enable the render test "icon-text-fit/textFit-grid-long" and add an alternative to let it pass when running in iPhone, since an alternative has already been added for Vulkan. #2829


louwers (approved):
Thanks, I'll merge this soon.

PR #2987, opened 2024-11-02
Allow configuring Call.Factory instead of OkHttpClient (#3137)

### LINKED ISSUE #2986: Change HttpRequestImpl client to Call.Factory
OkHttpClient is a concrete instance of Call.Factory (https://square.github.io/okhttp/3.x/okhttp/okhttp3/Call.Factory.html)

This is typically the right API for SDKs to require as it allows more flexibility in call creation such as

- Switching to a non OkHttp implementation (https://github.com/google/cronet-transport-for-okhttp)
- Switching between different clients
- adding logging tracing before call creation

As an example, see the coil image loader https://github.com/coil-kt/coil/blob/0582d7587e946c1d19340aee15227bf6aefc7770/coil-network-okhttp/src/commonMain/kotlin/coil3/network/okhttp/OkHttpNetworkFetcher.kt#L21

But in maplibre it's specifically OkHttpClient at

https://github.com/maplibre/maplibre-native/blob/b58c0c24cc311225770618158a8f6ab16edd7b4d/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/module/http/HttpRequestImpl.java#L108

yschimke:
This change allows users to configure a `Call.Factory` instead of just an `OkHttpClient` for making HTTP requests. It changes the `setOkHttpClient` method in a source compatible, but binary incompatible way.

This change provides more flexibility and allows users to use different HTTP clients, like those using Cronet.

Closes #2986

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/module/http/HttpRequestImpl.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/module/http/HttpRequestImpl.java
@@ -22,6 +22,7 @@ import java.net.NoRouteToHostException;
 import java.net.ProtocolException;
 import java.net.SocketException;
 import java.net.UnknownHostException;
+import java.util.Objects;
 
 import javax.net.ssl.SSLException;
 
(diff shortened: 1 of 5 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/module/http/HttpRequestImpl.java
```diff
@@ -105,9 +105,14 @@ public static void enableLog(boolean enabled) {
     HttpLogger.logEnabled = enabled;
   }
 
+  @Deprecated
   public static void setOkHttpClient(@Nullable OkHttpClient okHttpClient) {
```
louwers, line 109: Would it be a breaking change to take a `Call.Factory` here?

yschimke: Binary incompatible, but source compatible

louwers: If it's source compatible I'd say go ahead and just change the signature, if you don't mind. 🙂

louwers (approved):

louwers:
Thank you, will be part of the next release.

PR #2989, opened 2024-11-05
Fix crash when user updates icons with images of different sizes

alasram:
We run into several crashes related to users calling `MapLibreMap.getStyle.addImage(imageId, imageBitmap)` with the same `imageId` but `imageBitmap` with different sizes. In response to this call, MapLibre finds the texture atlas rectangle corresponding to `imageId` and updates the rectangle with `imageBitmap` without checking if `imageBitmap` fits the rectangle. This PR is a quick workaround to avoid the crash and unblock users. If `imageBitmap` is lager than the atlas rectangle for `imageId` then `imageBitmap` gets resized to fit the rectangle.
A probably better approach is to detect these cases and repack the atlas such as the rectangle allocated to `imageId` is the maximum size for all previously updated images.


louwers (commented):
Could you add a regression tests? Perhaps to the Android Instrumentation tests?

PR #2990, opened 2024-11-05
Clarify padding order in the Java API

ianthetechie:
Pretty simple PR hopefully. When working on a bounding box camera for a Jetpack Compose wrapper, I realized that the padding order was nowhere to be found in the docs. So I drilled down to the C++, used that padding order, and was very confused 🙃

This should save others from such a fate.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMap.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMap.java
@@ -55,6 +55,7 @@ interface NativeMap {
   @NonNull
   CameraPosition getCameraPosition();
 
+  // Note for implementors: the ordering of the padding is left, top, right, bottom
   CameraPosition getCameraForLatLngBounds(@NonNull LatLngBounds bounds, int[] padding, double bearing, double pitch);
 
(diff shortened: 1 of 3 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMap.java
```diff
@@ -55,6 +55,7 @@ void easeTo(@NonNull LatLng center, double zoom, double bearing, double pitch, d
   @NonNull
   CameraPosition getCameraPosition();
 
+  // Note for implementors: the ordering of the padding is left, top, right, bottom
   CameraPosition getCameraForLatLngBounds(@NonNull LatLngBounds bounds, int[] padding, double bearing, double pitch);
```
ianthetechie, line 59: I have attempted to keep this PR as minimal as possible, but if there's appetite for a deeper API change, I'd suggest that this API should be deprecated and replaced with something unambiguous. In our compose wrapper, we have something with this general shape:

```
data class CameraPadding(
    val left: Int,
    val top: Int,
    val right Int,
    val bottom: Int
)
```

louwers (approved):

louwers (approved):

Issue #2941, opened 2024-10-15
Support camera roll


louwers:
MapLibre GL JS issue: https://github.com/maplibre/maplibre-gl-js/pull/4780

MapLibre Style Spec issue: https://github.com/maplibre/maplibre-style-spec/pull/858

PR #3011, opened 2024-11-12
Camera Roll

NathanMOlson:
https://github.com/maplibre/maplibre-native/issues/2941


NathanMOlson:
@louwers can you approve the github workflow? I want to examine the wreckage.

louwers:
I have sent an invite so the workflows run automatically.

platform/android/MapLibreAndroid/src/cpp/map/camera_position.cpp
```diff
@@ -27,6 +34,8 @@ jni::Local<jni::Object<CameraPosition>> CameraPosition::New(jni::JNIEnv& env,
 
     // convert tilt, core ranges from  [0 rad, 1,0472 rad], android ranges from 0 to 60
     double tilt_degrees = options.pitch.value_or(0);
+    double roll_degrees = options.roll.value_or(0);
+    double fov_degrees = options.fov.value_or(36.86989764584402);
```
NathanMOlson: Is there a good way to use `mbgl::util::rad2deg(mbgl::util::DEFAULT_FOV)` from here?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/camera/CameraPosition.kt
```diff
@@ -227,7 +288,10 @@ class CameraPosition
                 val lat = typedArray.getFloat(R.styleable.maplibre_MapView_maplibre_cameraTargetLat, 0.0f).toDouble()
                 val lng = typedArray.getFloat(R.styleable.maplibre_MapView_maplibre_cameraTargetLng, 0.0f).toDouble()
                 target = LatLng(lat, lng)
+                centerAltitude = typedArray.getFloat(R.styleable.maplibre_MapView_maplibre_cameraCenterAltitude, 0.0f).toDouble()
                 tilt = typedArray.getFloat(R.styleable.maplibre_MapView_maplibre_cameraTilt, 0.0f).toDouble()
+                roll = typedArray.getFloat(R.styleable.maplibre_MapView_maplibre_cameraRoll, 0.0f).toDouble()
+                fov = typedArray.getFloat(R.styleable.maplibre_MapView_maplibre_cameraFov, 36.86989764584402f).toDouble()
```
NathanMOlson, line 296: Is there a good way to use `mbgl::util::rad2deg(mbgl::util::DEFAULT_FOV)` from here?

NathanMOlson:
@louwers At long last this PR is ready for review.

test/util/tile_cover.test.cpp
```diff
@@ -65,24 +65,6 @@ TEST(TileCover, PitchIssue15442) {
               util::tileCover({transform.getState()}, 2));
 }
 
-TEST(TileCover, PitchOverAllowedByContentInsets) {
```
louwers, line 68: Why was this test removed?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/camera/CameraPosition.kt
louwers: Not really.

platform/android/MapLibreAndroid/src/cpp/map/camera_position.cpp
louwers: I think you can just include the right header here.

src/mbgl/map/transform.cpp
```diff
@@ -679,19 +716,12 @@ double Transform::getMaxPitchForEdgeInsets(const EdgeInsets& insets) const {
```
NathanMOlson, line 720: The behavior of `getMaxPitchForEdgeInsets()` was previously "Return a pitch value that puts the top of the screen just below the horizon". The new behavior is "Return the pitch value that puts the top of the screen at the same angle it would be at maxPitch with zero padding." This change is needed because seeing above the horizon is now possible and desired.

platform/android/MapLibreAndroid/src/cpp/map/camera_position.cpp
NathanMOlson: Done.

test/util/tile_cover.test.cpp
NathanMOlson: This was removed to support a change that didn't make the final cut (involving `getMaxPitchForEdgeInsets()`). I've restored it and updated the expected values to match the new behavior. See my note below on the change in behavior.

NathanMOlson:
> Thanks for implementing this! Looks good. Just curious, what is your use case?

My goal is to render map info (primarily roads) on top of video from an airborne camera.

> We should consider adding an API for iOS as well before merging. I am guessing you only care about Android, correct?

I do not care about Android or iOS, I'm using Linux. Can you provide any hints on what those iOS changes would look like? I don't use or develop for iOS.

> I think we should also update the GLFW app even though it is just for development.

I'll take a look at that.

louwers:
> I do not care about Android or iOS, I'm using Linux. Can you provide any hints on what those iOS changes would look like? I don't use or develop for iOS.

No worries, I'll take care of that then.

> My goal is to render map info (primarily roads) on top of video from an airborne camera.

Cool!

NathanMOlson:
I've added support for FOV, roll, and high pitch to the GLFW app. FOV and roll are only settable from the command line. Pitch is changeable with the mouse, and there is a new `maxPitch` command-line arg to allow higher pitch angles.

When the horizon is below the top of the screen, I sometimes see labels briefly flying in the sky. They disappear once the map fully loads.
image

It seems to be displaying data that is behind the camera, from tiles that are no longer appropriate for the view.

For my use case, I plan to only use fully-loaded maps, so I can live with this issue. Will this need to be resolved before merging?

louwers:
@NathanMOlson As long as it is not an issue with the default max pitch it should be fine.

src/mbgl/map/transform_state.cpp
```diff
@@ -230,12 +249,12 @@ void TransformState::updateStateFromCamera() {
...
 
     // Compute bearing and pitch
     double newBearing;
     double newPitch;
-    camera.getOrientation(newPitch, newBearing);
+    double newRoll;
+    camera.getOrientation(newRoll, newPitch, newBearing);
     newPitch = util::clamp(newPitch, minPitch, maxPitch);
 
     // Compute zoom level from the camera altitude
```
NathanMOlson, line 262: This function always puts the center point at sea level, which does not work for pitch >=90. The equation for zoom is also poorly behaved near 90 degrees. The center point can be anywhere along the boresight of the camera, so there are infinitely many choices for zoom that would work. I like the idea of this function setting the zoom level to a fixed number regardless of pitch. This would require some unit tests to change, and I'm not sure if anyone relies on the existing behavior (or uses the FreeCamera options). To make this work for all pitch values, I would have to remove z-limiting in `TransformState::setCameraPosition()`.

The lowest impact approach would be to use the existing behavior at low pitch, and use a fixed zoom for high pitch angles or when the camera is underground.

Another idea is to use the GL JS approach, which is to get rid of the `FreeCameraOptions` methods and instead provide a helper function like GL JS's `calculateCameraOptionsFromCameraLngLatAltRotation(cameraLngLat: LngLatLike, cameraAlt: number, bearing: number, pitch: number, roll?: number)`, which would return a `CameraOptions` struct that could be used with `map.jumpTo()`.

I can also implement my desired functionality outside MapLibre if we don't want to mess with this stuff here.

NathanMOlson: OK, I updated using the low-impact approach. Behavior stays the same for pitch below the horizon, but `FreeCameraOptions` now work with pitch angles at/above the horizon as well.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/camera/CameraPosition.kt
```diff
@@ -75,6 +100,21 @@ class CameraPosition
...
+     * @param zoom    Zoom level at target. See zoom(float) for details of restrictions.
+     * @param tilt    The camera angle, in degrees, from the nadir (directly down). See tilt(float)
+     * for details of restrictions.
+     * @param bearing Direction that the camera is pointing in, in degrees clockwise from north.
+     * This value will be normalized to be within 0 degrees inclusive and 360 degrees
+     * exclusive.
+     * @param padding Padding in pixels. Specified in left, top, right, bottom order.
+     * @throws NullPointerException     if target is null
+     * @throws IllegalArgumentException if tilt is outside the range of 0 to 90 degrees inclusive.
+     */ @Keep internal constructor(target: LatLng?, zoom: Double, tilt: Double, bearing: Double, padding: DoubleArray?) : this(target, 0.0, zoom, tilt, bearing, 0.0, 36.86989764584402, padding)
```
louwers: Nit: I think this should be on its own line after the `/*`.

Also, could you make a variable for `36.86989764584402`?

src/mbgl/util/camera.cpp
```diff
@@ -15,7 +15,7 @@ using namespace std::numbers;
 namespace mbgl {
 
 namespace {
-double vec2Len(const vec2& v) noexcept {
+/*double vec2Len(const vec2& v) noexcept {
```
louwers: Stray comment.

NathanMOlson: Maplibre does not compile on my system with this unused code present. Should I remove it entirely, or restore it?

louwers: Since CI passes you can remove it completely.

NathanMOlson: Done.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/camera/CameraPosition.kt
NathanMOlson: Done.

PR #3012, opened 2024-11-12
Attempt to fix instrumentation tests

louwers:
Try to fix instrumentation tests after #2991.


HarelM (approved):
Did it work?

louwers:
@HarelM I can only tell after merging, as the device test workflow uses what is in `main` for security reasons.

cgalvan (approved):

Issue #3023, opened 2024-11-16
6.8.0 fails to load/render tiles, 6.7.1 succeeds


hactar:
**Describe the bug**
I have a map style I've been using for many maplibre versions thats always worked so far. But, in 6.8.0, when this map style is clean loaded (ie not upgraded from from an app that previously downloaded as 6.7.1), MapLibre does not display the map. I have included a sample project. If you change the dependency to use 6.7.1 in the example project, all works and you see Vienna. If you change the dependency back to 6.8.0, the map continues to appear. Delete the app and build and run again and the map no longer appears.

**To Reproduce**
1. Download the attached sample project.
2. Build and run for iPhone 16 pro 18.1 simulator (I also tried with my 18.1 device).
3. **You will see a "View Loaded" text but nothing else.**
4. In the package dependencies, change from 6.8.0 to 6.7.1.
5. Build and run again.
6. **Now you will see the map.**
7. Change the package dependency back to 6.8.0.
8.  build and run again.
9. **Map continues working.**

**Expected result**
The map should load fine on a clean install under 6.8.0.

**Screenshots**
<table>
  <tr>
    <td style="text-align: center;">
      <img src="https://github.com/user-attachments/assets/b57950d9-d1f4-4afb-87bb-db91665d7fd8" alt="6.8.0" style="width: 50%;">
      <p>6.8.0</p>
    </td>
    <td style="text-align: center;">
      <img src="https://github.com/user-attachments/assets/2c56e908-9985-4fcb-a059-702ccd71f4e7" alt="6.7.1" style="width: 50%;">
      <p>6.7.1</p>
    </td>
  </tr>
</table>

**Platform information (please complete the following information):**
 - OS: iOS 18.1, Xcode 16.1
 - Platform: iOS Simulator and device, though in the actual app it sometimes does work on device.
 - Version: 6.8.0

**Additional context**
Example project: 
[MapSixEightIssue.zip](https://github.com/user-attachments/files/17787012/MapSixEightIssue.zip)

louwers:
Thanks, I will have a look at it.

I see this in the console:

> fopen failed for data file: errno = 2 (No such file or directory)

louwers:
Also doesn't load for me on older versions...

louwers:
I did git bisect and I think d140d5e1659b3804f90f2e2f7f1b8e304a66decc caused a regression.

#2965

sargunv:
[...]

Issue #2993, opened 2024-11-05
MLNMapViewDelegate on 6.8.0 doesn't work


seraph3b:
**Describe the bug**

on 6.8.0 mapViewWillStartLoadingMap, mapViewDidFinishRenderingMap and mapView(_ mapView: MLNMapView, didFinishLoading style: MLNStyle) did not be fired even map appears (guess also mapViewDidFailLoadingMap and so on don't work)

**To Reproduce**
After loaded a style, map appears but app get stucked waiting for didfinishrenderingmap to continue 

**Expected behavior**
delegate fired

**Platform information (please complete the following information):**
 - OS: ios 18.1
 - iOS
 - map libre 6.8.0

louwers:
Please provide a minimal reproduction.

louwers:
I tried to reproduce this, but I was unable.

seraph3b:
we have 5 different maps and this issue happens with every maps.
only "func mapView(_ mapView: MLNMapView, didFinishLoading style: MLNStyle) {" is fired (tryed on simulator and phisical devices with ios 18+ or ios 15+)

currently using xcode 16.1, swift package manager -> maplibre 6.8.0 

with 6.7.1 it works as expected after setting style (self.mapView.styleURL = URL(string:"our map url")), map is loaded and fired:
- func mapView(_ mapView: MLNMapView, didFinishLoading style: MLNStyle) {
- func mapViewDidFinishRenderingMap(_ mapView: MLNMapView, fullyRendered: Bool) {

louwers:
Thanks.

I'm investigating the issue.

louwers:
I added these delegate methods to the AnimatedLineExample.swift and used AWS Open Data Standard Light and they fired as expected.

Could you share your style (can be private on Slack) so I can give it a try?

seraph3b:
here you are @louwers  https://tiler.3bmeteo.com/api/maps/bright_v2/style.json

seraph3b:
on viewdidload:
self.mapView.styleURL = URL(string:"https://tiler.3bmeteo.com/api/maps/bright_v2/style.json")
 self.mapView.clearsContextBeforeDrawing = true
            self.mapView.maximumZoomLevel = maximumZoomLevel
            self.mapView.minimumZoomLevel = minimumZoomLevel
            
            // se CurrentRadar == swiss_radar limita al nord!
[...]

PR #3024, opened 2024-11-16
Revert "Encode the style URL in iOS (#2965)"

louwers:
This reverts commit d140d5e1659b3804f90f2e2f7f1b8e304a66decc.

This commit has caused several regressions https://github.com/maplibre/maplibre-native/issues/3023 

Potentially also https://github.com/maplibre/maplibre-native/issues/2993

I want to revert it, put out a release, and then look for a fix.


HarelM (approved):

HarelM:
I would advise to split a revert PR with a version release PR, but that might be an overkill.
I think it helps to see in the git history a commit for "bump version" (besides the tagging of course) since then you can easily see which commit was added in which version without a lot of hassle.
I use an action to create a bump version PR, which also takes care of the change log modification. I think it's a nice way to solve this, also it allows me to release a version without any extra approval (which is nice by its own).

louwers:
@HarelM I agree that is a cleaner way to do it. I am guilty of lumping unrelated things into one PR, because CI for this repo is quite heavy and takes a long time to run. Not really a good reason though.

I usually update the changelog as part of the version bump PR. I also think that releasing a version should require an approval.

HarelM:
I all you do is "push the release button" I don't see a lot of merit in approving this. But I guess it also depends on how fast you get a PR approval for something so trivial, which I find to long in my case...

PR #3030, opened 2024-11-19
Batch up scheduling of deferred deletions

TimSylvester:
In each tile cache (one per tile source), released tiles are kept in a vector and scheduled for release all at once, at the end of `update`, which is called once per frame.

The capture below is from ~150 seconds of the Android benchmark, 5300 frames.  Of 384 total, it shows 138 instances of just one tile being scheduled, up to 3 instances of 13 being released at once.  440 total scheduled tasks were averted in this case, roughly half.

Manually zooming in and out on the same area for ~600 frames produces similar results.

If we only did the release after accumulating several tiles, we could reduce the number of tasks significantly more, but we might end up holding onto unused tiles for many frames in some cases.

~~Additionally, this avoids allocating a `shared_ptr` for each `unique_ptr` to make it copyable along with the lambda capture, using a more generic move-only function wrapper.~~


src/mbgl/tile/tile_cache.cpp
```diff
@@ -26,43 +41,54 @@ void TileCache::setSize(size_t size_) {
...
+    }
+
+    // Block destruction until the cleanup task is complete
+    {
+        std::lock_guard counterLock{deferredSignalLock};
+        deferredDeletionsPending++;
+    }
+
+    threadPool.schedule(
+        util::unique_function<void()>{[this, wrap_{CaptureWrapper{std::move(pendingReleases)}}]() mutable {
```
louwers, line 80: You are moving the data member `pendingReleases` here.

include/mbgl/util/unique_function.hpp
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include <functional>
+
+namespace mbgl {
+namespace util {
+
+// C++20 non-copying lambda capture, pending C++23 `move_only_function`
```
louwers, line 8: Maybe we can use something like `https://github.com/zhihaoy/nontype_functional/blob/main/include/std23/move_only_function.h` instead so we can more easily migrate to `std::move_only_function` when we can use C++23.

It's also better tested.

TimSylvester: I'll give it a try.  Would we add that whole thing as a dependency, or just pull in the one file?

louwers: If it's not too much hassle I would add it to `vendor/nontype_functional`. Could just be the two files we need.

src/mbgl/tile/tile_cache.cpp
TimSylvester: Interesting, there used to be language on cppreference implying that was valid, but it's gone now.  Seems like it was removed since 2022, based on [this](https://news.ycombinator.com/item?id=32907069).

louwers: image
image

TimSylvester: I changed it to move the elements to a temporary yesterday, but it sounds like it's sufficient to use `clear()` to return it to a well-defined valid state, right?

louwers (approved):

louwers:
For future reference: once C++23 is out we can use std::move_only_function

PR #3035, opened 2024-11-21
Eliminate copies in deferred cleanup

TimSylvester:
Pulls in an implementation of `move_only_function`.

This simplifies deferred deletions, since there's no temporary copy to potentially outlast the schedule call, and eliminates the need to transfer ownership from each `unique_ptr` to a `shared_ptr` so that it can be copied.

Unfortunately, the change of function types for the scheduler spreads to the map callbacks due to usage like:

`pathChangeCallback = Scheduler::GetCurrent()->bindOnce(...`

This is further complicated by the fact that of the callbacks are explicitly copied:

```
void FileSourceRequest::setResponse(const Response& response) {
    // Copy, because calling the callback will sometimes self
    // destroy this object. We cannot move because this method
    // can be called more than once.
```

So some but not all callbacks are changed to the new function type, and typedefs are added to clarify.

This will make it possible to eliminate the extra copy steps in deferred tile destruction.


include/mbgl/storage/file_source.hpp
```diff
@@ -35,24 +36,28 @@ enum FileSourceType : uint8_t {
 // GeoJSONSource, RasterSource, VectorSource, CustomGeometrySource and other *Sources.
 class FileSource {
 public:
+    template <typename T = void()>
+    using Callback = std23::move_only_function<T>;
+    template <typename T = void()>
+    using CopyableCallback = std::function<T>;
```
louwers, line 42: I personally don't like these at all. I think it's much cleaner to have APIs don't obfuscate the fact that they use simple STL templates.

include/mbgl/util/unique_function.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <functional>
+
+namespace mbgl {
+namespace util {
+
+// C++20 non-copying lambda capture, pending C++23 `move_only_function`
+// Adapted from https://stackoverflow.com/a/52358928/135138
+template <typename T>
+class unique_function : public std::function<T> {
```
louwers, line 11: This one can be removed now I think?

louwers (changes requested):
Please see comments.

louwers (approved):

PR #3040, opened 2024-11-22
Expose Android NativeMapView and move creation to the ModuleProvider

mwilsnd:
This PR exposes the NativeMapView implementation and delegates construction to the ModuleProvider via a new factory method. This allows an external library to implement a custom NativeMapView and replace it at runtime by implementing the ModuleProvider interface. This enables a much cleaner method of working with the native core layer of MapLibre in Android projects.

For existing users, nothing will change. Advanced users which already implement their own ModuleProvider will need to implement the new `createNativeMapView` method. The implementation of this method can be copied from the included `ModuleProviderImpl.java` file.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMapView.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMapView.java
@@ -44,7 +44,7 @@ import java.util.Arrays;
 import java.util.List;
 
 // Class that wraps the native methods for convenience
-final class NativeMapView implements NativeMap {
+public final class NativeMapView implements NativeMap {
 
   private static final String TAG = "Mbgl-NativeMapView";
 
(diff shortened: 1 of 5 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMapView.java
```diff
@@ -44,7 +44,7 @@
 import java.util.List;
 
 // Class that wraps the native methods for convenience
-final class NativeMapView implements NativeMap {
+public final class NativeMapView implements NativeMap {
```
louwers, line 47: Should we drop the `final` so it's easier to create a custom NativeMap with inheritance?

mwilsnd: Let me do some testing, I'm currently uncertain if the JNI bindings will work as expected with a derived class.

louwers:
Closing for now, please open a new PR if someone wants to pick this up.

PR #3046, opened 2024-11-27
Fix the issue that the empty polyline cannot be updated

grab-liyanjin:
An example code to reproduce the issue is 
```
Polyline line = mapLibreMap.addPolyline(new PolylineOptions());
// Anything to update the polyline will fail
line.setPoints(...);
line.setColor(...);
```
The reason is that PolylineContainer will ignore the empty polyline the first time.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/PolygonContainer.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/PolygonContainer.java
@@ -27,12 +27,10 @@ class PolygonContainer implements Polygons {
   @Override
   public Polygon addBy(@NonNull PolygonOptions polygonOptions, @NonNull MapLibreMap maplibreMap) {
     Polygon polygon = polygonOptions.getPolygon();
-    if (!polygon.getPoints().isEmpty()) {
-      long id = nativeMap != null ? nativeMap.addPolygon(polygon) : 0;
-      polygon.setId(id);
-      polygon.setMapLibreMap(maplibreMap);
-      annotations.put(id, polygon);
-    }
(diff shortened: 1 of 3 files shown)


hactar:
Hmm, I wonder why that check was there in the first place...

Did you test what happens if you add an empty polyline and don't add any points/colors etc later in the code?

louwers:
Thanks for the PR!!

We should probably make this behavior consistent across other types of containers. For example:

https://github.com/maplibre/maplibre-native/blob/06fb1816a6c400af051a0ec2605d0eff25eb1f12/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/PolygonContainer.java#L30-L35

Also there is a `AnnotationManagerTest.kt` where a test can be added.

grab-liyanjin:
> Hmm, I wonder why that check was there in the first place...
> 
> Did you test what happens if you add an empty polyline and don't add any points/colors etc later in the code?

This issue was found by our developer during map integration, nothing will happen with empty point, but cannot update the polyline later since it's not really added into the map

grab-liyanjin:
> Thanks for the PR!!
> 
> We should probably make this behavior consistent across other types of containers. For example:
> 
> https://github.com/maplibre/maplibre-native/blob/06fb1816a6c400af051a0ec2605d0eff25eb1f12/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/PolygonContainer.java#L30-L35
> 
> Also there is a `AnnotationManagerTest.kt` where a test can be added.

Hi louwers, I checked all the containers including **PolygonContainer**, **PolylineContainer**, **MarkerContainer**, **AnnotationContainer**, and **ShapeAnnotationContainer**, only **PolygonContainer** and **PolylineContainer** have this kind of logic, so made changes to these two classes and UT as well

louwers (approved):

PR #3051, opened 2024-11-27
Add thermal status to benchmark results

adrian-cojocaru:


platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/benchmark/BenchmarkActivity.kt
```diff
@@ -254,7 +263,13 @@ class BenchmarkActivity : AppCompatActivity() {
...
         mapView.removeOnDidFinishRenderingFrameListener(listener)
 
-        return BenchmarkRunResult(fps, encodingTimeStore, renderingTimeStore)
+        var thermalStatus = -1;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            val powerManager = getSystemService(Context.POWER_SERVICE) as PowerManager
+            thermalStatus = powerManager.currentThermalStatus
+        }
+
+        return BenchmarkRunResult(fps, encodingTimeStore, renderingTimeStore, thermalStatus)
```
louwers, line 272: A bit cleaner would be a getter since this function is already getting pretty big, but it's fine this way as well.

louwers (approved):

Issue #2787, opened 2024-08-30
Vulkan Early Access Testing (Android)


louwers:
<img src="https://github.com/user-attachments/assets/109b43b0-8b61-4980-9cd2-1b91d8dfad1e" width="50%">

---

We are looking for users interested in the upcoming Vulkan support for Android.

We will be releasing an 'early access' MapLibre Android version (a.k.a. pre-release) that will _just_ support Vulkan in **September 2024**. This will give people an opportunity to test the new Vulkan backend on Android. Joining this effort will help the rendering team to ensure there are no regressions as well as with optimizing performance.

Edit: the first [pre-release has been released](https://github.com/maplibre/maplibre-native/releases/tag/android-v11.3.0-vulkan-pre0)!

## Reporting Issues

When [reporting an issue](https://github.com/maplibre/maplibre-native/issues/new) (here or in a separate issue), the following are required:

- Style: If your style is public or can be made available, that will help a lot.
- Lat / Long / Zoom: Where exactly does the issue occur?
- Screenshot
- Android version
- MapLibre Native version
- Device

When available:

- Stack Trace / Vulkan Error / Core Dump
- Minimal style: Can you provide a minimal style that reproduces the issue?

westnordost:
I could create some alpha releases alongside the stable ones for StreetComplete (once the MapLibre migration has been stabilized in the first place) for people to try out. 

Beyond detecting simple rendering issues, this only makes sense if there's a straightforward way _for these users_ to make useful reports for crashes in native / Vulkan code. I understand that the logged crash trace is not that useful for you and digging out the tombstone seems to require root privileges on the phone and/or usage of developer tools (adb), plus some special build configuration (we haven't fully figured out yet) beforehand.

Also, how would you sell it to users so they want to be the first people to try it out? Is there anything to gain from that?

louwers:
> I could create some alpha releases alongside the stable ones for StreetComplete (once the MapLibre migration has been stabilized in the first place) for people to try out.

That would be amazing!

> Also, how would you sell it to users so they want to be the first people to try it out? Is there anything to gain from that?

Being an early tester and testing it with your style and app means that any bugs you report will have the attention of the rendering team, and will probably be fixed by the time we make an official release.

mtpdog:
I can take part too to test in complex environment with many raster, vector and extruded layers.
[...]

PR #3056, opened 2024-11-28
Add latency to surface updates

adrian-cojocaru:
Surface rotation events (and updated info) come a few frames later after 90 degree rotations. Adding a delay between the event and surface update fixes https://github.com/maplibre/maplibre-native/issues/2787#issuecomment-2488533005


louwers (approved):
Is this also how games and the like solve this problem?

adrian-cojocaru:
I'm not sure if it's the best solution, but I've seen this type of fix a few times.

TimSylvester (approved):
Seems reasonable.  It might be a good idea to include an explicit recursion limit just in case the SDK doesn't behave as expected.

PR #3066, opened 2024-12-04
Fix crash on unsupported attribute type conversion

### LINKED ISSUE #3042: It should not crash on style which works with maplibre 10.2.0
EDIT: The cause I thought in the initial description of the issue is incorrect. After further investigation we pinpointed the actual cause of the experienced crash which I added as a comment to the issue: https://github.com/maplibre/maplibre-native/issues/3042#issuecomment-2510799281

When I use a style.json with map libre version 11.5.1 which does not have;

"maputnik:renderer": "mbgljs"

in its metadata, the maplibre sdk causes a crash with the log attached.
[maplibre-renderer-crash.txt](https://github.com/user-attachments/files/17919481/maplibre-renderer-crash.txt)

1. Use a style.json that does not have the "maputnik:renderer": "mbgljs" in its metadata
2. Open a map using this style.json
3. Observe crash upon drawing certain tiles

Expected behavior:
The Maplibre SDK should not crash, or mention that this metadata is mandatory.

 - OS: Android - 14
 - Platform: Android
 - Version: 11.5.1
 - Device: Samsung S23

This issue does not happen on version maplibre 10.2.0. It only started happening after a switch to version 11.5.1.

TimSylvester:
Check matched attribute before trying to clear its dirty flag, preventing a crash when there is no match.

Remove assertion to avoid killing debug builds while attempting to set up the attribute.

Resolves #3042

### DIFF
--- a/src/mbgl/gl/upload_pass.cpp
+++ b/src/mbgl/gl/upload_pass.cpp
@@ -271,7 +271,9 @@ gfx::AttributeBindingArray UploadPass::buildAttributeBindings(
             return;
         }
 
-        overrideAttr->setDirty(false);
+        if (overrideAttr) {
+            overrideAttr->setDirty(false);
+        }
 
         bindings[index] = {
             /*.attribute = */ {defaultAttr.getDataType(), offset},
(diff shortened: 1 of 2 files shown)


louwers (approved):

louwers:
Thank you for looking into this and fixing it.

I created an issue to make regression test (I assigned myself, but if you want you can also add it to this PR). https://github.com/maplibre/maplibre-native/issues/3067

Issue #3058, opened 2024-11-29
Cannot exclude compassView from VoiceOver (accessibility) when map is not north


krupupakku:
### MapLibre iOS Version

6.8.0

### iOS Version

18.0

### Device

iPhone 13

### What happened?

Hey! as I'm not using the compassView I need to exclude it from being read from the iOS accessibility VoiceOver but I noticed even if I'm setting
```
        map.compassView.isAccessibilityElement = false
        map.compassView.accessibilityElementsHidden = true
```

is the map is not facing north, it's being read by screen reader. It's also visible in accessibility list elements (two fingers, three taps on screen).
If the map is not rotated there's no problem, screen reader is ignoring the compass view

extra info: rendering or hiding compass view is not changing things.

Some screens:
| / | Map | List |
|--|--|--|
| Not North | <img src="https://github.com/user-attachments/assets/0f2bdf43-61b0-4360-84fb-28043282208e" width="200" /> | <img src="https://github.com/user-attachments/assets/cdccecc0-b2c8-4b49-8b3d-3917ab1d1e38" width="200" /> |
| North | <img src="https://github.com/user-attachments/assets/a06a9031-933c-431c-829b-8971b2379c59" width="200" /> | <img src="https://github.com/user-attachments/assets/68bd8365-370b-48b1-9738-d7396cf7197d" width="200" /> |

### Steps to reproduce

- just show a simple map, 
- try disable compassView from accessibility (I'm doing it in updateUIView)
- rotate the map to not face north

### Relevant log output

```shell

```

louwers:
Thanks for the report. I am not super aware how Voice Over works.

We have a [developing guide](https://github.com/maplibre/maplibre-native/blob/main/platform/ios/CONTRIBUTING.md#create-the-xcode-project) if you want to try looking into this yourself.

https://github.com/maplibre/maplibre-native/blob/8b32c9a78309af3df8ffadf5fa4276a135212262/platform/ios/src/MLNCompassButton.mm#L4

PR #3068, opened 2024-12-04
fix: Disable accessibility on hidden compass button

ocarrascoe:
Checking the [issue 3058](https://github.com/maplibre/maplibre-native/issues/3058), it might be solved by disabling the accessibility on the `hideCompass` function because if the compass is hidden there is no point in showing it for the accessibility.

Testing it fixes the problem, let me know if it's useful!

Also, you can check if on the other hand, the function `showCompass`, you can set the opposite

```
self.isAccessibilityElement = YES;
self.accessibilityElementsHidden = No;
```

but as they are the default values, maybe it is not needed, just mind a case of being forced to hide and show the compass button in the same instance. What do you think?

### DIFF
--- a/platform/ios/src/MLNCompassButton.mm
+++ b/platform/ios/src/MLNCompassButton.mm
@@ -124,6 +124,8 @@
 
 - (void)hideCompass:(BOOL)animated {
     animated ? [self animateToAlpha:0] : [self setAlpha:0];
+    self.isAccessibilityElement = NO;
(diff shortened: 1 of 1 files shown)


louwers (approved):
I agree that when the compass is hidden, there's probably no expectation that the direction is included for users using Voice Over.

Thanks for your contribution, will be included in the next iOS release!

PR #3069, opened 2024-12-05
Add bold support for CJK characters

wangyingfang:
Use system font family to draw characters of CJK, but don't support bold font on iOS. Add similar logic as Android side.
image
image


louwers (commented):
Can't you use a bold font?

platform/darwin/src/local_glyph_rasterizer.mm
```diff
@@ -266,6 +279,11 @@ CGContextHandle context(CGBitmapContextCreate(
     
     CTLineDraw(*line, *context);
     
+    // Release the bold font if it was created
+    if (boldFont) {
+        CFRelease(boldFont);
```
louwers, line 284: This is never called if the function throws. Should use `CFHandle`.

louwers:
I wasn't aware this is exactly what MapLibre Android and MapLibre GL JS do...

platform/darwin/src/local_glyph_rasterizer.mm
```diff
@@ -191,16 +191,29 @@ CFDictionaryRefHandle attributes(
...
 
+    // Can't use CTFontRefHandle
+    // because the boldFont will be released after it is out of the isBold condition
+    CTFontRef boldFont = NULL;
+    if (isBold)
+    {
+        // Create a bold variant of the font
+        boldFont = CTFontCreateCopyWithSymbolicTraits(font, 0.0, NULL, kCTFontBoldTrait, kCTFontBoldTrait);
+        if (!boldFont) {
+            throw std::runtime_error("Unable to create bold font");
```
louwers, line 211: Maybe instead of throwing you can log an error and set `boldFont` to `font`.

louwers (approved):
I added a test. Please take a look if it looks OK to you.

Also, please have a look at my comments. Thank you.

random3940:
I find it peculiar that we need to figure out bold flag from the name, but change seems to reflect what Android does and fixes the bug. LTGM.

random3940 (approved):

platform/darwin/src/local_glyph_rasterizer.mm
```diff
@@ -191,16 +193,26 @@ CFDictionaryRefHandle attributes(
...
     CFStringRefHandle string(CFStringCreateWithCharacters(NULL, reinterpret_cast<UniChar*>(&glyphID), 1));
     if (!string) {
         throw std::runtime_error("Unable to create string from codepoint");
     }
 
+    // Create a bold variant of the font
+    CTFontRefHandle boldFont(CTFontCreateCopyWithSymbolicTraits(font, 0.0, NULL, kCTFontBoldTrait, kCTFontBoldTrait));
+    if (!boldFont) {
+        CFStringRefHandle familyNameHandle(CTFontCopyFamilyName(font));
+        [MLNNativeNetworkManager.sharedManager errorLog:@"Unable to create bold font for %@", *familyNameHandle];
```
louwers, line 209: I think using `Log::Error` is more appropriate here. Otherwise LGTM.

PR #3077, opened 2024-12-10
Remove `Pass3D` 

### LINKED ISSUE #3039: Filters applied to fill extrusion layers are not rendered unless a manual zoom is applied to the map
I am setting filters to layers using the layer.setFilter(<expression>) method dynamically from code. The filters take effect for all layers immediately except the fill and fill-extrusion layers(from what I observe, maybe other layers are not rendered properly that I missed). For the fill layers, the filter is applied only if a manual zoom is applied to the map. Otherwise, the filter takes effect after another filter is applied. 

To explain further;

- The user changes to level 1 on the map, a filter is applied on all layers for level 1 using expression programmatically.
- All of the layers filter the features correctly except the fill and fill-extrusion layers.
- User changes to level 2, filter is applied to all layers, all show correct features except the fill and fill-extrusion layers. The fill and fill-extrusion layers still show level 1 features.
- User changes to level 3, the filtering on the fill layer shows the previous filter. So, on the fill and fill-extrusion layers, features with level 2 are drawn.
- User changes to level 1, the features on level 3 are drawn.
- The only way to refresh this layer filter and show the correct features is to manually change the zoom of the map via gestures.

1. Set a filter to all layers programmatically. 
2. Observe filtering takes effect immediately for only certain layers. Not fill and fill-extrusion layers to be precise.
3. Zoom in/out manually, observe features are correctly rendered for all layers.

The filtering on the layers should take effect immediately.

 - Platform: Android
 - Version: Maplibre(11.5.1)
 - Devices: Samsung Galaxy S23(Android 14) - Pixel 8 Pro Emulator (Android 35, 16KB page size)

This issue started happening after I upgraded from 10.2.0 to 11.5.1 for the 16kb device support on the new version. Using the same style, and the same filtering logic, I get the expected behavior on version 10.2.0. 

There are no logs indicating any issue with rendering or filtering.

TimSylvester:
... to prevent a bug where `removeTile(Translucent|Pass3d)` doesn't remove drawables added with just `Translucent`.

Resolves #3039

### DIFF
--- a/src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
+++ b/src/mbgl/renderer/layers/render_fill_extrusion_layer.cpp
@@ -63,9 +63,8 @@ void RenderFillExtrusionLayer::evaluate(const PropertyEvaluationParameters& para
         parameters.getCrossfadeParameters(),
         unevaluated.evaluate(parameters, previousProperties->evaluated));
 
-    passes = (properties->evaluated.get<style::FillExtrusionOpacity>() > 0)
-                 ? (RenderPass::Translucent | RenderPass::Pass3D)
-                 : RenderPass::None;
+    passes = (properties->evaluated.get<style::FillExtrusionOpacity>() > 0) ? RenderPass::Translucent
+                                                                            : RenderPass::None;
     properties->renderPasses = mbgl::underlying_type(passes);
     evaluatedProperties = std::move(properties);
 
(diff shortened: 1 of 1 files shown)


sjg-wdw:
How do we feel about merging this before Tim goes on vacation?

louwers (approved):
Sounds like a good idea to get a final bugfix in before the end of the year. We can always revert it if it causes any problems.

Seems like the 3D pass is even more dead code now. But I am not too worried about it with the revival on the roadmap.

Would be good to have a regression test for this, I tried (#3073) but it passes even before this.

TimSylvester:
Yeah, I never figured out why it doesn't fail in that test.

PR #3089, opened 2024-12-12
UBO consolidation

alexcristici:
UBOs was reorganized for all shaders. As a convention: `drawableUBO` contains the matrix, interpolations and other parameters that are used only in vertex shader.
UBOs that were bound per drawable (like `drawableUBO` or `tilePropsUBO`) are now consolidated in a vector per layer instance and bound per layer. **The UBOs are consolidated for Metal and Vulkan backends.**
Uniform blocks were removed, they are not used anymore because UBOs are bound in the layer where there is no access to shader informations. So now there are some reserved UBO slots for vertex or fragment binding only.


TimSylvester (approved):

include/mbgl/shaders/mtl/collision.hpp
```diff
@@ -0,0 +1,202 @@
+#pragma once
+
+#include <mbgl/shaders/collision_layer_ubo.hpp>
+#include <mbgl/shaders/shader_source.hpp>
+#include <mbgl/shaders/mtl/shader_program.hpp>
+
+namespace mbgl {
+namespace shaders {
+
+#define COLLISION_SHADER_COMMON \
```
louwers, line 10: Could you get rid of the macro? Maybe use a constexpr `std::string` instead?

alexcristici: I tried something but it's not working to merge strings in initialization like this.

alexcristici: I mean here: `static constexpr auto source = COLLISION_SHADER_COMMON R"(`

louwers: Hmm seems like std::string can actually be constexpr, but only short strings.

Trying to figure out how we can concat constexpr `std::string_view`s

louwers: @alexcristici https://github.com/alexcristici/maplibre-native/pull/29

alexcristici: Looks like a good solution, but I would say let's merge this PR as it is, and make this change in the follow up PR.

louwers: @alexcristici OK sounds good to me.

louwers (approved):

PR #3094, opened 2024-12-16
Release MapLibre iOS 6.9.0

louwers:


HarelM (approved):
I would suggest to create an action that does that so you could release without waiting for someone to approve as this is a very technical operation, this is what I do for GL-JS.
But that might just be me, and I probably already said that, so sorry for repeating myself. :-)

louwers:
@HarelM Thanks! I think the four eyes principle should apply to releases. Also I update the changelog for every release (need to weed out irrelevant changes per platform) so a PR is needed anyway...

PR #3103, opened 2024-12-23
cxx17 compatability fixes

jwinarske:
-fixes to build with c++17

https://en.cppreference.com/w/cpp/feature_test#cpp_lib_source_location


louwers:
MapLibre Native is a C++20 project now and we don't aim to be backwards compatible with C++17. Although we don't rely on C++20 in a lot of places yet, I think it will be helpful for the project to embrace new functionality such as concepts and ranges.

Feel free to bring up this topic during the TSC Meeting or in a GitHub Discussion if this is important to you, so we can have a discussion about it.

Issue #644, opened 2023-01-02
Labels getting clipped with some sources but not others


acalcutt:
Labels are getting clipped with some sources in tile mode, while others are not getting clipped

For an example, 

This map is using western-zone-latest.mbtiles from the example https://github.com/maptiler/tileserver-gl/issues/618#issuecomment-1328116302 which was made with  tilemaker. you can see the label is clipped.
https://tiles.wifidb.net/styles/western-zone/?raster#15/22.8167/72.7381
![image](https://user-images.githubusercontent.com/3792408/210261828-730b998c-b58b-47c3-9485-60967bf30dbe.png)

However this is the same style using a planetiler generated mbtiles file
https://tiles.wifidb.net/styles/bright/?raster#15/22.8167/72.7381
![image](https://user-images.githubusercontent.com/3792408/210262543-518540d2-3e50-40d6-83fb-4038180cec35.png)

What's the difference? Both these are rendered in tile mode. 

If I set tileserver to run in static mode, the second example clips text like the first example. The first example does not improve in static mode.

_Originally posted by @acalcutt in https://github.com/maplibre/maplibre-gl-native/issues/632#issuecomment-1369099484_

tdcosta100:
Nice we can have a conversation not being offtopic hahaha. So, the real problem is the way the tiles are generated with TileMaker. It keeps only the features that are inside one particular tile. That means it doesn't include point features outside that tile. Planetiler seems to be smarter and include surrounding point features that go outside the tile, hence the non-clipping situation.

When I mentioned static, I mean the possibility of rendering the surrounding tiles, that is, with a margin that could include the point where the labels are rendered. This is my way to solve this my problem because I had my tiles generated with TileMaker. If you use static mode without a margin, labels will be clipped in some situations, probably worse than rendering with tile mode, like you demonstrated.

The Maplibre-GL-Native code already deals with this situation by rendering elements that go outside the tile borders. The only thing it needs is to have the tile data including points outside the tile. This is why TileMaker tiles are more prone to label clipping than Planetiler.

acalcutt:
That makes sense. 

One of my idea to work around this in tileserver was to set the mode to static when the TileMargin option was set. similar to how I made it switch when using static tile mode  https://github.com/maptiler/tileserver-gl/pull/608 . That kind of makes it hard to have different sourced mbtiles though.

tdcosta100:
[...]

Issue #284, opened 2022-05-19
Labels getting clipped in tile mode (node version)


acalcutt:
I mentioned this issue in https://github.com/maplibre/maplibre-gl-native/issues/279 , but did not want to pollute that thread with an unrelated problem.

In my testing of the node version of maplibre-gl-native in my branch of [tileserver-gl](https://github.com/acalcutt/tileserver-gl/) I have noticed labels getting clipped more than in the mapbox-gl-native version.

For example, in the maplibre-gl-native version it is consistently cutting off labels all over the place like this
![maplibre](https://user-images.githubusercontent.com/3792408/169196629-feada2a1-0579-49a1-82bf-cdf879a0334f.png)

Where in the maptiler version using mapbox-gl-native v5.02, does not have this issue 
![mapbox](https://user-images.githubusercontent.com/3792408/169196781-45d50301-de5c-409f-ae87-c1d48200e3be.png)

Looking at several similar threads I found these related mapbox issues
https://github.com/mapbox/mapbox-gl-native/issues/15805
https://github.com/mapbox/mapbox-gl-native/issues/14011
https://github.com/mapbox/mapbox-gl-native/pull/15880
The first article in this list mentions this was improved in v5.0.2, which I did find the change they mentioned here https://github.com/mapbox/mapbox-gl-native/commit/d500b67d4a73ccbfbc59b78a3277538b01d1cb5e. the padding was increased when in tile mode to help the labels to help with the trimming (a change which still exists, slightly rewritten). Also, in tileserver-gl they added the code to use tile mode here https://github.com/maptiler/tileserver-gl/commit/5048388d1fbfce3f983e6ee0338bafaa90e947ce

Looking though the related code in https://github.com/maplibre/maplibre-gl-native/blob/main/src/mbgl/text/collision_index.cpp#L26-L50 it seems like the changes that added padding in tile mode are still there and looking in my tileserver-gl code the tile mode option is still set ( https://github.com/acalcutt/tileserver-gl/blob/main/src/serve_rendered.js#L574 )

From this, to me, it seems like it should be showing the labels correct like in the second screenshot, but for some reason it is not. 

One idea I had is maybe it is ignoring the tile mode switch? I'm guessing that would be [this code](https://github.com/maplibre/maplibre-gl-native/blob/main/platform/node/src/node_map.cpp#L1403-L1409) in the node binding code?

[...]

PR #3109, opened 2024-12-26
Add optional map mode argument to render

kkirov:
# Label Clipping Prevention in mbgl-render

This PR adds the ability to use `MapMode::Tile` with `mbgl-render` to prevent label clipping at tile boundaries, which is particularly important for map tile generation.

## Changes
- Adds support for `MapMode::Tile` while maintaining `MapMode::Static` as the default
- No breaking changes to existing behavior
- Enables proper label placement across tile boundaries when explicitly using `MapMode::Tile`

## Technical Background
Label clipping occurs because `MapMode::Static` treats each tile in isolation, causing labels near tile edges to be truncated. `MapMode::Tile` mode considers neighboring tiles during label placement, ensuring continuous and properly placed labels across tile boundaries.

## Related Issues
- maplibre/maplibre-native#644 - Label clipping discussion
- maplibre/maplibre-native#284 (comment) - Technical explanation of tile modes

### DIFF
--- a/bin/render.cpp
+++ b/bin/render.cpp
@@ -38,7 +38,8 @@ int main(int argc, char* argv[]) {
     args::ValueFlag<uint32_t> widthValue(argumentParser, "pixels", "Image width", {'w', "width"});
     args::ValueFlag<uint32_t> heightValue(argumentParser, "pixels", "Image height", {'h', "height"});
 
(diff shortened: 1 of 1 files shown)


louwers (approved):
LGTM but you need to run clang-format.

```
pre-commit install
pre-commit run --all-files
```

PR #3111, opened 2025-01-02
Fix broken .geojson url

justinclift:
The path for MapLibre GL JS documentation assets appears to have changed, so the path in the test app needs updating to match.

Without this change, that url 404's.


louwers (approved):
Thanks!

Issue #3084, opened 2024-12-11
Labels jittery on some styles


louwers:
### MapLibre Android Version

11.7.0

### Android SDK Version

Android 11

### Device

Samsung S10

### What happened?

Jittery labels.

https://github.com/maplibre/maplibre-native/issues/2787#issuecomment-2513318976

https://github.com/maplibre/maplibre-native/issues/2787#issuecomment-2334921372

### Steps to reproduce

Use Versatiles Colorful and zoom in: https://tiles.versatiles.org/assets/styles/colorful.json

### Renderer

Vulkan

### Relevant log output

```shell

```

Issue #3116, opened 2025-01-06
Marker Flashing with 11.7.1 Vulkan


ekamp:
### MapLibre Android Version

11.7.1 Vulkan

### Android SDK Version

Android 12

### Device

Samsung Galaxy S20

### What happened?

SymbolLayer / Markers flash when using the latest 11.7.1-Vulkan version.https://github.com/user-attachments/assets/0c58d11f-7198-4f49-9558-fd4c8bee192b
 

### Steps to reproduce

Add a SymbolLayer with filter and pan or move the map along the same zoom level.

### Renderer

Vulkan

### Relevant log output

```shell

```

louwers:
Thanks for the report!

> Add a SymbolLayer with filter and pan or move the map along the same zoom level.

Could you add some sample code?

cc @adrian-cojocaru

Issue #3085, opened 2024-12-11
Vector elements floating


louwers:
### MapLibre Android Version

1.6.1-vulkan-pre1

### Android SDK Version

?

### Device

?

### What happened?

Reported by @mtpdog

> The only small issue I've noticed so far with 11.6.1-vulkan-pre1 is that all vector elements (vector maps, annotations) do not "stick" firmly to the raster background on quick camera movements. They kind of floating around their normal positions. Probably the vector and raster parts of rendering are not synchronized perfectly. Would be great if this is fixed as well.

### Steps to reproduce

Unclear but @adrian-cojocaru seems to be on it.

@mtpdog if you could, please clarify

### Renderer

Vulkan

### Relevant log output

```shell

```

mtpdog:
This issue seems to be fixed in 11.7.0.

louwers:
@mtpdog Also with Vulkan?

mtpdog:
> [@mtpdog](https://github.com/mtpdog) Also with Vulkan?

Yes. I've tested it with Vulkan and OpenGL. Both work fine! (there was no issue with OpenGL anyway).

PR #3122, opened 2025-01-07
Fix in-flight frame update

adrian-cojocaru:
Select and update the correct frame uniform/vertex buffers. Currently the same set of buffers were being used (each frame) generating race conditions after the GPU submission.

Fixes https://github.com/maplibre/maplibre-native/issues/3084, https://github.com/maplibre/maplibre-native/issues/3085 and should address similar issues (https://github.com/maplibre/maplibre-native/issues/3116, https://github.com/maplibre/maplibre-native/issues/2787#issuecomment-2513318976)


louwers:
!benchmark android

louwers (changes requested):
Pretty sure there is some kind of performance regression.

Still benchmarking locally.

louwers (commented):
My bad, was using a wrong (debug) build...

louwers (approved):

PR #3124, opened 2025-01-08
Replace macro with compile-time string

alexcristici:


cgalvan (approved):

louwers:
Nice, thanks! I contacted the author of the code snippet. I'll update the PR with the license.

adrian-cojocaru (approved):

louwers (changes requested):
Bloaty results came back. Seems like it's not optimizing out the unused strings. So this is probably not worth it after all unless we find another approach. What do you think?

louwers:
Interesting, so now it's a size reduction.

include/mbgl/shaders/mtl/shader_group.hpp
```diff
@@ -63,7 +64,8 @@ class ShaderGroup final : public ShaderGroupBase {
             addAdditionalDefines(propertiesAsUniforms, additionalDefines);
 
             auto& context = static_cast<Context&>(gfxContext);
-            const auto shaderSource = std::string(shaders::prelude) + source;
+            // C++26 will allow operator+ with std::string and std::string_view
```
louwers, line 67: I think this comment can be removed now

louwers (commented):
Now the concatenation happens at runtime. Probably minimal performance impact?

alexcristici:
> Now the concatenation happens at runtime. Probably minimal performance impact?

Yes, no noticeable differences.

PR #3126, opened 2025-01-08
Replace deprecated Node Buffer constructor

kudlav:
Node.js `new Buffer(size)` is deprecated since: v6.0.0.

Replace it with `Buffer.from(string)` and `Buffer.alloc(size)`.

The `Buffer.alloc(size)` returns zero-filled memory which was default for `new Buffer(size)` starting from v8.0.0.

https://nodejs.org/docs/latest-v22.x/api/buffer.html#new-bufferbuffer


acalcutt (approved):
This change looks good to me, based on the documentation and tests passing. I think I have seen that warning before, thouh I don't see it in the ci.

PR #3138, opened 2025-01-11
[qt] Drop support for Qt5 and update to Qt 6.8.1

ntadej:
I think it's time to drop support for Qt5 to allow more modern toolsets for Qt and start implementing Metal and Vulkan support.

Also update to Qt 6.8.1 and MSVC2022 on Windows.


ntadej:
It builds, I say we merge it 😊

louwers (approved):

Issue #3099, opened 2024-12-19
Vulkan Native Crash when using TextureMode


ekamp:
### MapLibre Android Version

11.7.1 - Vulkan

### Android SDK Version

Android 12

### Device

Samsung Galaxy S20

### What happened?

When updating to the latest 11.7.1-Vulkan , when placing the app in the background or leaving the map open for a min or two the app crashes with a native trace : 
When disabling `textureMode` within the `MapLibreMapOptions` the issue no longer occurs. For our use-case we would like to continue to benefit from using TextureMode within our application.

### Steps to reproduce

1. Launch application and inflate a `MapView`
2. Navigate away from the MapView placing it in the background
3. Application crashes with a native stack trace linked below

```


### Renderer

Vulkan

### Relevant log output

```shell
[1219/161542.394465:ERROR:elf_dynamic_array_reader.h(64)] tag not found
2024-12-19 13:15:42.459 26747-27093 libc                    com....          A  Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 27093 (TextureViewRend), pid 26747 (roid.driver.dev)
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  Build fingerprint: 'samsung/x1quex/x1q:12/SP1A.210812.016/G981U1UES3FVI6:user/release-keys'
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  Revision: '14'
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  ABI: 'arm64'
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  Processor: '7'
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  Timestamp: 2024-12-19 16:15:42.560716234-0500
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  Process uptime: 129s
2024-12-19 13:15:43.491 27697-27697 DEBUG                   crash_dump64                         A  Cmdline: com....
[...]

PR #3144, opened 2025-01-15
Fix TextureMode crash

adrian-cojocaru:
Reordered some surface lifecycle events to avoid rendering with no valid surface available. Fixes https://github.com/maplibre/maplibre-native/issues/3099


louwers (commented):
Any chance you can add an instrumentation test that reproduces this?

adrian-cojocaru:
I've used Android Studio's Layout Inspector to trigger the crash since I haven't been able to consistently reproduce it with just the device events.

alexcristici (approved):

louwers (approved):

moodstubos:
@adrian-cojocaru 
thank you very much for this fix. It resolves some `Cause: null pointer dereference` issues. 
Question: Could you do the same for non-vulkan version?

adrian-cojocaru:
@moodstubos The same scenario was tested with OpenGL, but everything was ok there. 
If you have any more info or any repro steps open a new issue and I'll take a look.

PR #3145, opened 2025-01-15
Fix scheduler test failure

TimSylvester:
Add a failing unit test to catch a rare case where `Scheduler::GetSequenced` cycles through the available items faster than expected, causing the test to fail.  Then simplify that method to fix the problem.

### DIFF
--- a/src/mbgl/actor/scheduler.cpp
+++ b/src/mbgl/actor/scheduler.cpp
@@ -48,29 +48,22 @@ std::shared_ptr<Scheduler> Scheduler::GetBackground() {
 
 // static
 std::shared_ptr<Scheduler> Scheduler::GetSequenced() {
-    const std::size_t kSchedulersCount = 10;
+    constexpr std::size_t kSchedulersCount = 10;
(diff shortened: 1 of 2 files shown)


src/mbgl/actor/scheduler.cpp
```diff
@@ -48,29 +48,22 @@ std::shared_ptr<Scheduler> Scheduler::GetBackground() {
...
-            if (lastUsedIndex == i) result = scheduler;
-            continue;
-        }
-        result = std::make_shared<SequencedScheduler>();
-        weak = result;
-        lastUsedIndex = i;
-        break;
+    if (auto scheduler = weaks[lastUsedIndex].lock()) {
+        return scheduler;
+    } else {
```
louwers, line 62: nit: else branch can be removed

louwers (approved):

src/mbgl/actor/scheduler.cpp
TimSylvester: I'll clean it up in #3035

PR #3147, opened 2025-01-16
Fix broken links in node readme

kudlav:
Fix broken links in node platform readme.


louwers (changes requested):
I think these should use permalinks otherwise they will be broken again. :)

louwers (approved):

kudlav:
Severity values are actually `ERROR`, `WARNING`, ...
The correct source is [src/mbgl/util/event.cpp](https://github.com/maplibre/maplibre-native/blob/0dea8bad2ff68a95772a6e3b3a788946c0596f15/src/mbgl/util/event.cpp).
Updated link to the status codes RFC.

acalcutt (approved):
Looks good to me now, thanks

PR #3149, opened 2025-01-16
Add timeout-minutes to windows-ci workflow

louwers:
This workflow step gets stuck for 6 hours right now. https://github.com/maplibre/maplibre-native/actions/runs/12811603515/job/35721558246

Add a timeout so it will time out quicker.


cgalvan (approved):
Nice! Do we know what makes it timeout intermittently?

tdcosta100:
> Nice! Do we know what makes it timeout intermittently?

The timeout usually comes from the `vcpkg` step, unfortunately it has a bunch of problems in the last updates, so we have to wait some fixes from this component until this problem is gone. This workaround is a good way to deal with this problem meanwhile.

tdcosta100:
Hmmm 5 minutes is not enough to `vcpkg` if it's going to build the packages. Maybe increasing it to 45 minutes is better, maybe 60 if it's not enough for EGL build (building ANGLE is time-consuming).

PR #3154, opened 2025-01-18
Support defining a custom style layer from iOS Swift

christian-boks:
I wanted to use a custom style layer in a Swift app on iOS, but discovered that the `typedef struct MLNBackendResource` structure used to expose the backend wasn't visible in Swift, so I converted it to an Objective-c interface.

And I've also converted the Objective-c example to Swift and added it to the Swift example app, to show how to use it.

### DIFF
--- a/platform/darwin/bazel/files.bzl
+++ b/platform/darwin/bazel/files.bzl
@@ -161,6 +161,7 @@ MLN_DARWIN_PRIVATE_HEADERS = [
 
 MLN_DARWIN_PUBLIC_OBJCPP_SOURCE = [
     "src/MLNAttributionInfo.mm",
+    "src/MLNBackendResource.mm",
     "src/MLNComputedShapeSource.mm",
     "src/MLNCustomStyleLayer.mm",
     "src/MLNDefaultStyle.mm",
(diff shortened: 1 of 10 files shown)


stefankarschti (approved):

louwers (approved):
Thanks this is awesome!

> Got it to compile with the OpenGL backend too.

We're not really supporting that anymore, but I guess it doesn't hurt.

This example doesn't work when I rotate the map. But that can be addressed in a follow-up PR.

I think this is technically a breaking change, but since custom style layers are a somewhat experimental feature I think it is acceptable.

christian-boks:
> Thanks this is awesome!
> 
> > Got it to compile with the OpenGL backend too.
> 
> We're not really supporting that anymore, but I guess it doesn't hurt.

It looks like there is a CI step that still compiles the code without the `--//:renderer=metal` option tough.

> This example doesn't work when I rotate the map. But that can be addressed in a follow-up PR.

I see what you mean, I had just tested it on the simulator before. I guess I need to rotate the matrix too, not just scale it. 

> I think this is technically a breaking change, but since custom style layers are a somewhat experimental feature I think it is acceptable.

👍

louwers:
> It looks like there is a CI step that still compiles the code without the --//:renderer=metal option tough.

Oh wow, that is an oversight. I will fix that.

stefankarschti:
> This example doesn't work when I rotate the map. But that can be addressed in a follow-up PR

I think that's fine, the original example didn't implement the projection matrix either.

PR #3158, opened 2025-01-19
Update vcpkg to 2025.01.13

tdcosta100:
Update `vcpkg` to 2025.01.13 commit tag. Hopefully this will reduce the timeouts in `Windows CI` workflow.

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -51,9 +51,40 @@ jobs:
         run: |
           Write-Host "Changed file(s): ${{ steps.changed-files.outputs.windows_all_changed_files }}"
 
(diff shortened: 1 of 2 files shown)


louwers (approved):
Ah it needs to download a bunch of dependencies as part of configuring? Makes sense to increase it then.

tdcosta100:
> Ah it needs to download a bunch of dependencies as part of configuring? Makes sense to increase it then.

Download AND build dependencies, if they are not in cache. This is the reason it takes so long sometimes. I'm considering make just one `vcpkg` pass, so all build commands benefit from not building the dependencies simultaneously.

cgalvan (approved):

PR #3164, opened 2025-01-21
Small fixes ios-ci.yml

louwers:
- Add permissions to deploy to GitHub pages to `ios-ci.yml`.
- Make sure Metal backend is used for C++ Unit Test app.

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -29,6 +29,7 @@ on:
 permissions:
   id-token: write         # needed for AWS
   contents: write         # allow making a release
+  pages: write
 
 jobs:
   pre_job:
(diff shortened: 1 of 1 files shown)


.github/workflows/ios-ci.yml
```diff
@@ -29,6 +29,7 @@ on:
 permissions:
   id-token: write         # needed for AWS
   contents: write         # allow making a release
+  pages: write
```
cgalvan, line 32: Nit: Could add a comment like the other permissions so we know what this permission is needed for

cgalvan (approved):

.github/workflows/ios-ci.yml
louwers: Will add it in a follow-up.

PR #3174, opened 2025-01-24
 chore: remove multiple empty trailing newlines

nyurik:
Some IDEs auto-fix these, and ensure there is at least one, so easier to keep them clean-ish in the repo with a single commit

To find it, i used these two commands with ripgrep:

```
rg -l -U '[^\n]\z' --type txt -g '!**vendor**'
rg -l -U '\n\n\z' --type txt -g '!**vendor**'
```


louwers:
Let's fix all then https://github.com/maplibre/maplibre-native/pull/3176

PR #3177, opened 2025-01-24
Fix incorrect Object/function checking in node index.js wrapper

kudlav:
Node
Fix incorrect Object/function checking in node index.js wrapper.

## Changes
- replace `options instanceof Object` by `typeof options === "object"`
- replace `options.request instanceof Function` by `typeof options.request === "function"`

Using `instanceof Object` can be problematic because it won't work correctly for all objects.
E.g. the following object `{ ratio: 1, request: [Function: request] }` returns `options instanceof Object` `false` and at the same time `typeof options.request === "function"` `true`. Simillar to the function check.

### Other changes

- Use modern `let` and `const` instead of `var` to prevent hoisting issues
- Unify single and double quote to double quote - use just one type in a single file
- do not repeat `if (options...` every time, use nested if


louwers (commented):
Looks good, the Node.js codebase can be modernized a bit. I would recommend setting up an automated formatter, maybe with pre-commit. Could you add some regression tests for the scenarios you describe?

Maybe we can create a Node.js channel on Slack to coordinate the continued development of the Node.js platform?

louwers:
@acalcutt I am going through old PRs. This looks good, so I'm merging it.

PR #3181, opened 2025-01-27
Rust wrapper

nyurik:


platform/rust/core-src/ARCHITECTURE.md
```diff
@@ -0,0 +1 @@
+../../../ARCHITECTURE.md
```
louwers, line 1: This file should not exist anymore.

platform/rust/src/wrapper.cc
```diff
@@ -0,0 +1,48 @@
...
+
+uint32_t get_42() {
+    return 42;
+}
+
+std::unique_ptr<TileServerOptions> TileServerOptions_new() {
+  return std::make_unique<TileServerOptions>();
+}
+std::unique_ptr<TileServerOptions> TileServerOptions_mapbox() {
+  return std::make_unique<TileServerOptions>(TileServerOptions::MapLibreConfiguration());
```
louwers, line 15: These are pending deprecation and removal.

platform/rust/src/wrapper.cpp
```diff
@@ -0,0 +1,47 @@
+#include "maplibre-native/include/tile_server_options.h"
+#include "maplibre-native/include/map_renderer.h"
+#include "maplibre-native/src/lib.rs.h"
+
+namespace ml {
```
louwers, line 5: I suggest using `mln::rustbridge` as namespace name.

`ml` is Machine Learning for me

.github/workflows/linux-ci.yml
```diff
@@ -111,6 +111,14 @@ jobs:
           cmake -B build -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMLN_WITH_CLANG_TIDY=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DMLN_WITH_COVERAGE=ON ${{ env.renderer_flag_cmake }}
           cmake --build build --target mbgl-core mbgl-test-runner mbgl-render-test-runner mbgl-expression-test mbgl-render mbgl-benchmark-runner
 
+      # test rust platform
+      - if: matrix.renderer == 'drawable-rust'
```
louwers, line 115: We should move this to rust-ci.

platform/rust/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// FIXME: Remove this before merging
```
louwers, line 1: .

CommanderStorm:
since we moved things to its own wrapper, I think this PR is moot, right?

nyurik:
tbh, i would much rather move this here to avoid the unneeded extra complexity of the two repos - esp because now you see many other projects are using it, and it clearly shows value?

birkskyum:
Taking a step back, it's not just two repos. There is the option to keep the wrapper in maplibre-native-rs, and just add a CI check here in Native that clones the wrapper repo to test it with upcoming native versions. It's a bit of an "ecosystem ci" check, which is an approach that will soon be used for Qt SDK (which prev lived in maplibre-native) and I think strikes a good balance.

It's inspired by the [vite-ecosystem-ci](https://github.com/vitejs/vite-ecosystem-ci) project, which keep the vite core repo clean by making integration tests, and it's protecting as well against regressions as a monorepo would.

Without an ecosystem check like that, everything that depends on native has to be part of native core repo to be sure... maplibre-compose, maplibre-react-native, maplibre-native-qt, flutter-maplibre-gl etc., the rs bindings aren't really different in that regard, so i think it's best we try to establish a solution that'll work for other dependants too.

CommanderStorm:
I see that the issues that we currently have are not very MLN specific.

Given that otherwise I would have to subscribe to all MLN issues or Bart would need to tag me (both horrible) I think keeping us separate and adding something like ecosystem CI would be nice.

The changes that I see as TODO would unnecessarily spam people, making maintaining something as high traffic as MLN more difficult.

nyurik:
oops, i totally misread - i thought this was the PR that added the `justfile` to the core MLN, not the rust wrapper.  I'm fine with the wrapper being separate too

PR #3184, opened 2025-01-31
Change runtime error to the nullptr returning

TatyanaPolunina:
- We are attempting to create shaders during the update of layers. It seems more appropriate to return a nullptr for a shader that fails to create, rather than stopping the app if one shader is not created (as exceptions are not caught).
- Proper handling is implemented in all callers when a shader is nullptr, so we should not experience inconsistent behavior in such cases.
- Most shader creation errors are related to low memory on iOS. Since we have lazy initialization for shaders, the behavior may be restored in the next level update if iOS frees the necessary memory.

### DIFF
--- a/include/mbgl/shaders/mtl/shader_group.hpp
+++ b/include/mbgl/shaders/mtl/shader_group.hpp
@@ -70,7 +70,8 @@ public:
             assert(shader);
             if (!shader || !registerShader(shader, shaderName)) {
                 assert(false);
-                throw std::runtime_error("Failed to register " + shaderName + " with shader group!");
+                Log::Error(Event::Shader, "Failed to register " + shaderName + " with shader group!");
+                return nullptr;
             }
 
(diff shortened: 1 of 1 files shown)


adrian-cojocaru (approved):
Returning null works, but on each subsequent call of `getOrCreateShader` the same shader creation will be retried and it can end up compiling the same code multiple times per frame. 
An empty/debug "Null" shader should fix this, but I don't think the current checks can handle it. 
Either way a performance hit is better than an unhandled exception.

TatyanaPolunina:
> Returning null works, but on each subsequent call of `getOrCreateShader` the same shader creation will be retried and it can end up compiling the same code multiple times per frame. An empty/debug "Null" shader should fix this, but I don't think the current checks can handle it. Either way a performance hit is better than an unhandled exception.

Potentially it could be even better. As shader is not created, that's why we have Null. Also we usually have "compile error" together with "memory warning" on IOS, so I expect that on next "updateLayers" IOS could clear the memory and we would correctly create shaders so app would work. But need to be investigated for sure

PR #3230, opened 2025-02-16
reset depth stencil state for render pass

yousifd:
Because the depth stencil state of the render pass
is no reset correctly after the custom layer is drawn
we need to manually do this to ensure that we don't
write to the depth or stencil buffers when we shouldn't

### DIFF
--- a/src/mbgl/mtl/render_pass.cpp
+++ b/src/mbgl/mtl/render_pass.cpp
@@ -72,6 +72,17 @@ void RenderPass::endEncoding() {
     }
 }
 
+void RenderPass::resetState() {
+    currentDepthStencilState.reset();
+    currentStencilReferenceValue = 0;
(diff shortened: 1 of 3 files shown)


src/mbgl/mtl/render_pass.cpp
```diff
@@ -72,6 +72,22 @@ void RenderPass::endEncoding() {
     }
 }
 
+// Reset the states and bindings of the render pass to deal
```
louwers, line 75:
```suggestion
// Resets the states and bindings of the render pass to deal
```

Maybe move this to the header and use a triple-slash comment?

yousifd: Done

cgalvan (approved):

stefankarschti (approved):

TimSylvester (approved):

PR #3234, opened 2025-02-17
Allow configuring iOS build with CMake

louwers:
- Allows configuring Xcode with CMake for iOS builds.
- Add new job to `ios-ci` that builds the `mbgl-core`, `ios-sdk-static` and `app` CMake targets for iOS.
- Move some files that are implementations for the core to `platform/darwin/core` to better separate out the iOS SDK.
- Fix some build issues (unused/uninitialized variables).
- Add license for SMCalloutView.
- Fix typo `MLN_PUBLIC_IOS_APP_SOPURCE`.
- Update MapLibre Developer Documentation to reflect changes in this PR.
- Remove some unused files.
- Add `CMakePresets.json`. Other platforms will be added in a follow-up PR. Info here: https://martin-fieber.de/blog/cmake-presets/


.github/workflows/ios-ci.yml
```diff
@@ -332,13 +332,30 @@ jobs:
         run: |
           VERSION=${{ env.version }} COCOAPODS_TRUNK_TOKEN=${{ secrets.COCOAPODS_PASSWORD }} pod trunk push MapLibre.podspec
 
+  ios-build-cmake:
+    needs: pre_job
+    runs-on: macos-14
```
cgalvan, line 337: Should we matrix this to run across all our supported major version? Do we support 13, 14, and 15?

docs/mdbook/src/ios/README.md
```diff
@@ -76,3 +76,12 @@ You can also build targets from the command line. For example, if you want to bu
 ## Swift App
 
 There is also an example app built with Swift instead of Objective-C. The target is called `MapLibreApp` and the source code lives in `platform/ios/app-swift`.
+
+## CMake
+
+It is also possible to generate an Xcode project using Xcode. As of February 2025 only the `mbgl-core` target is supported.
```
cgalvan, line 82:
```suggestion
It is also possible to generate an Xcode project using CMake. As of February 2025 only the `mbgl-core` target is supported.
```

cgalvan (approved):

.github/workflows/ios-ci.yml
louwers: No we only support building with a recent version of Xcode. I guess we could use `macos-latest` instead here.

src/mbgl/mtl/drawable.cpp
```diff
@@ -252,6 +252,7 @@ void Drawable::draw(PaintParameters& parameters) const {
             const auto stencilMode = enableStencil ? parameters.stencilModeForClipping(tileID->toUnwrapped())
                                                    : gfx::StencilMode::disabled();
             impl->depthStencilState = context.makeDepthStencilState(depthMode, stencilMode, renderable);
+            assert(newStencilMode);
```
louwers, line 255: Issue: https://github.com/maplibre/maplibre-native/issues/3248

platform/darwin/darwin.cmake
```diff
@@ -0,0 +1,164 @@
...
+    "${MLN_GENERATED_DARWIN_CODE_DIR}/MLNFillStyleLayer_Private.h"
+    "${MLN_GENERATED_DARWIN_CODE_DIR}/MLNHillshadeStyleLayer_Private.h"
+    ${MLN_GENERATED_DARWIN_STYLE_PUBLIC_HEADERS}
+)
+
+find_program(BAZEL bazel REQUIRED)
+
+add_custom_command(
+    OUTPUT ${MLN_GENERATED_DARWIN_STYLE_SOURCE} ${MLN_GENERATED_DARWIN_STYLE_HEADERS}
+    COMMAND ${BAZEL} build //platform/darwin:generated_code
```
louwers, line 127: I use Bazel to generate the code so I don't have to redo all that build logic in CMake...

platform/darwin/darwin.cmake
```diff
@@ -0,0 +1,164 @@
...
+        mbgl-core
+        PRIVATE
+            ${PROJECT_SOURCE_DIR}/platform/default/src/mbgl/mtl/headless_backend.cpp
+    )
+endif()
+
+include(${PROJECT_SOURCE_DIR}/vendor/icu.cmake)
+
+set(CMAKE_OBJC_FLAGS "-fobjc-arc")
+set(CMAKE_OBJCXX_FLAGS "-fobjc-arc")
```
louwers, line 78: Not sure if this is a good idea. We don't seem to use `-fobjc-arc` for the Bazel builds? 🤔

platform/darwin/darwin.cmake
```diff
@@ -0,0 +1,164 @@
+enable_language(OBJC OBJCXX Swift)
```
louwers, line 1: Common code `ios.cmake` and `macos.cmake`.

CMakePresets.json
```diff
@@ -0,0 +1,51 @@
...
+      "generator": "Xcode",
+      "description": "Create Xcode project for iOS",
+      "binaryDir": "${sourceDir}/build-ios",
+      "cacheVariables": {
+        "CMAKE_SYSTEM_NAME": "iOS",
+        "CMAKE_CXX_COMPILER_LAUNCHER": "ccache",
+        "MLN_WITH_METAL": "ON",
+        "MLN_WITH_OPENGL": "OFF",
+        "MLN_DRAWABLE_RENDERER": "ON",
+        "MLN_WITH_RTTI": "ON"
```
louwers, line 21: I get link errors when I don't enable RTTI...

Screenshot 2025-02-23 at 00 46 22

cgalvan: I think we'd lose access to `dynamic_cast` if we disable RTTI

louwers: Fixed by making sure RTTI is disabled for all translation units.

TimSylvester:
I had to upgrade to the latest CMake, maybe the `cmake_minimum_required` should be increased?

The Team ID `QMC7N4BDVZ` is embedded in `platform/ios/app/CMakeLists.txt`.  I guess that needs to come from a config file like `config.bzl` or environment variable.

louwers:
@TimSylvester Yes if you use the CMakePresets.json you need CMake 3.31.

@TimSylvester That's the one on CI. Yes I'll fix that.

Issue #3052, opened 2024-11-27
Support initializing map with string JSON style on iOS


michalgwo:
**Is your feature request related to a problem? Please describe.**
On Android, we can set style to a map either as a URI or JSON using Style.Builder i.e. `Style.Builder().fromJson()` or `Style.Builder().fromUri()`, but on iOS the only option (at least it's the only one I found) is to set `styleURL`. Setting a style with JSON as a String is useful in my use case when I edit the style in code before initializing a map. The same applies to MapSnapshotter.

**Describe the solution you'd like**
- Create a new instance property of MLNMapView - styleJSON and/or create the function setStyleJSON. 
- Create a new instance property of MLNMapSnapshotOptions - styleJSON 

**Describe alternatives you've considered**
I was thinking about saving a style to a file every time it's edited and then reading it from the file, but it's more a workaround than a real solution. I didn't really explore this option, so I don't even know if it's possible.

louwers:
This is implemented now. You can use this constructor

https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapview/initwithframe:stylejson:

Or initialize with MLNMapOptions: https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre/mlnmapoptions

michalgwo:
It's implemented only for `MLNMapView`, but not for `MLNMapSnapshotter`

PR #3240, opened 2025-02-20
Allow initializing MLNMapView with style json

wangyingfang:
Address [issues 3052 - Support initializing map with string JSON style on iOS](https://github.com/maplibre/maplibre-native/issues/3052)


random3940 (approved):
Assuming that it's the same API as we have on Android - LGTM.

platform/darwin/test/MLNMapViewTests.m
```diff
@@ -133,32 +149,84 @@ - (void)testVisibleCoordinatesCompletion {
...
+    XCTAssertNil(error, @"Style JSON should be valid JSON");
+    XCTAssertNotNil(jsonObject, @"Style JSON should parse to a valid object");
+    XCTAssertTrue([jsonObject isKindOfClass:[NSDictionary class]], @"Style JSON should represent a dictionary");
+
+    // Reset style loading expectation before setting new style
+    _styleLoadingExpectation = nil;
+
+    // Test setting style JSON
+    NSString *newStyleJSON = @"{\"version\": 8, \"sources\": {}, \"layers\": []}";
+    self.mapView = [[MLNMapView alloc] initWithFrame:CGRectMake(0, 0, 64, 64) styleJSON:newStyleJSON];
```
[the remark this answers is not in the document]
louwers: Maybe also modify the map in some way (add a layer programmatically) and check that it is reflected in the JSON?

platform/darwin/test/MLNMapViewTests.m
```diff
@@ -133,32 +149,84 @@ - (void)testVisibleCoordinatesCompletion {
...
+    self.mapView.tileCacheEnabled = NO;
+    XCTAssertEqual(self.mapView.tileCacheEnabled, NO);
 
-    mapView.tileCacheEnabled = YES;
-    XCTAssertEqual(mapView.tileCacheEnabled, YES);
+    self.mapView.tileCacheEnabled = YES;
+    XCTAssertEqual(self.mapView.tileCacheEnabled, YES);
 }
 
+- (void)testStyleJSON {
```
louwers, line 179: Maybe split these out in multipel tests?

louwers, line 179: What happens when you set it to `nil`?

wangyingfang: Will throw exception if set it to nil.

wangyingfang:
@louwers please help review and re-run the pipelines.

platform/ios/src/MLNMapView.mm
```diff
@@ -522,6 +522,19 @@ - (instancetype)initWithFrame:(CGRect)frame styleURL:(nullable NSURL *)styleURL
...
 }
 
+- (instancetype)initWithFrame:(CGRect)frame styleJSON:(NSString *)styleJSON
+{
+    if (self = [super initWithFrame:frame])
+    {
+        MLNLogInfo(@"Starting %@ initialization.", NSStringFromClass([self class]));
+        MLNLogDebug(@"Initializing frame: %@ styleJSON: %@", NSStringFromCGRect(frame), styleJSON);
+        [self commonInit];
+        self.mbglMap.getStyle().loadJSON([styleJSON UTF8String]);
```
louwers, line 532: If approach will first load the default style. It is possible that the user will first see the default style flash before this style is finally loaded.

platform/ios/src/MLNMapView.mm
```diff
@@ -569,6 +582,33 @@ - (void)setStyleURL:(nullable NSURL *)styleURL
...
+}
+
+- (void)setStyleJSON:(NSString *)styleJSON {
+    if (!styleJSON) {
+        [NSException raise:NSInvalidArgumentException
+                    format:@"Style JSON cannot be nil."];
+        return;
+    }
+
+    // Verify JSON is valid
```
louwers, line 596: We're already parsing the style when loading the style, so we parse the JSON another time.

This does not check if the style is really valid, it only checks if it is valid JSON.

louwers:
@wangyingfang @random3940 Any ideas how we can address the concerns I have raised?

Can we (also) allow passing a JSON at initialization?

wangyingfang:
@louwers I'm so sorry for not addressing your comments in a timely manner. It's too busy this month. I will resolve these comments next week.

louwers:
@wangyingfang No worries! Thanks for confirming you still want to work on it.

wangyingfang:
@louwers I have resolved the comments, please help review again.

platform/ios/src/MLNMapView.mm
louwers: I think it is wasteful to parse JSON only to check if what the user passes is valid JSON. On Android we also do not do this check. It is the responsibility of the user to pass valid JSON. 

Another thing is: the JSON parser part of iOS might be slightly different than the JSON parser in the core. For example, if we ever swap out the JSON parser in the core (e.g. to enable JSON5, which allows comments) now all of the sudden iOS will reject JSON that would be perfectly valid as far as the core is concerned.

Also, again, this does not check if the style is valid, it only checks if the JSON is valid.

If you want to synchronously validate the style before handing it off to be loaded asynchronously, you should add this logic to the core. Otherwise, please remove the validation from this method for the reasons listed above.

louwers (changes requested):
Thanks! I have one more comment, see above.

platform/ios/src/MLNMapView.mm
wangyingfang: @louwers  Have updated

platform/ios/src/MLNMapView.mm
```diff
@@ -552,6 +568,12 @@ - (nonnull NSURL *)styleURL
         return self.residualStyleURL;
     }
 
+    // If style was initialized with JSON, return nil
+    if (self.mbglMap.getStyle().getJSON().length() > 0 &&
+        self.mbglMap.getStyle().getURL().empty()) {
+        return nil;
```
louwers, line 574: The return type is marked with `nonnull` and the interface with `null_resettable`, which means this getter should not return null if I understand it correctly. We should not modify this, because I think it would change the Swift API.

Maybe you can return `[NSURL URLWithString:@"local://style.json"]` instead?

platform/ios/src/MLNMapView.mm
```diff
@@ -522,6 +523,21 @@ - (instancetype)initWithFrame:(CGRect)frame styleURL:(nullable NSURL *)styleURL
...
 }
 
+- (instancetype)initWithFrame:(CGRect)frame styleJSON:(NSString *)styleJSON
+{
+    if (self = [super initWithFrame:frame])
+    {
+        MLNLogInfo(@"Starting %@ initialization.", NSStringFromClass([self class]));
+        MLNLogDebug(@"Initializing frame: %@ styleJSON: %@", NSStringFromCGRect(frame), styleJSON);
+        [self commonInit];
+        // Please don't call self.styleURL = nil, which load the default style
```
louwers, line 533: We can remove this comment I think.

Let's mark `styleJSON` as `nonnullable` and `nonnull` in the header.

louwers (approved):
Thanks, I have two more comments.

platform/ios/src/MLNMapView.mm
wangyingfang: Will remove the comment. But I think don't need mark `styleJSON` as `nonnullable` and `nonnull`, because the header has done nonnullable by default with `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`

wangyingfang:
@louwers have resolved comments, please review again.

platform/ios/src/MLNMapView.mm
louwers: Good point thanks.

louwers (approved):

Issue #2928, opened 2024-10-11
Node 6.0.0/6.0.1.pre.0 freezing in macos/metal after about 32 renders


acalcutt:
I have gotten a report in slack and the tileserver-gl repo that the node 6.0.0 for macos is freezing. It will work at first, but after browsing the map the application freezes and stops returning rendered images.
https://github.com/maptiler/tileserver-gl/issues/1402
https://osmus.slack.com/archives/C02B2CBSNBU/p1728481702513059

The same code used works fine with the last opengl-2 release, 5.4.1

**To Reproduce**
I have created a simplified node application that just generates the same image 50 times. On Windows this generates all 50 images, but when I test on macos it seems to generate 32 images then freeze, similar to what is happening in tileserver-gl

The test app can be used like this 
```
https://github.com/acalcutt/maplibre-node-test.git
git checkout macos_test
npm i
node .
```

**Expected behavior**
This test application should generate 50 images in the root folder, but right now on macos it freezes after 32 images are generated and I have to kill the node process to get it to end

**Screenshots**
Not working on macos
<img width="1440" alt="Screenshot 2024-10-11 at 9 21 06 AM" src="https://github.com/user-attachments/assets/b9f94b97-8c70-458f-b026-9c7b30794d4b">

Working on windows
![image](https://github.com/user-attachments/assets/c8c9c9e9-d9cc-4a7d-a40b-090d5c2d5704)

**Platform information (please complete the following information):**
 - OS: macOS Sonoma 14.3 / Apple M2
 - Platform: Node.js 20/22
 - Version: 6.0.0/6.0.1-pre.0

**Additional context**
IThis issue does not seem to occur in Windows or Linux, Just macOS. I am not completly sure how to troubleshoot this.

acalcutt:
@louwers do you have any idea how I might troubleshoot this? I tried to make a "debug" release but it didn't really give me any more information. I also tried to enable the "tracy" flag but it didn't seem to work on metal

Seems strange it works 32 times then just freezes

louwers:
If I have to guess it's something CFRunLoop related. I can try replacing it with the libuv implementation to see if that works. It's a bit a shot in the dark but that worked for GLFW.

louwers:
32 images... That sounds like a suspicious number. Maybe something is not cleaned up properly and filled up.

I'll put it on the agenda for Wednesday and see if anyone has any ideas.

lseelenbinder:
[...]

PR #3245, opened 2025-02-21
Add iOS/macOS observer hooks

adrian-cojocaru:
Follow-up to https://github.com/maplibre/maplibre-native/pull/2694. Adds `MapObserver` events to iOS/macOS delegates.


louwers (commented):
Could you add some example usage to the Objective-C and/or Swift based apps?

acalcutt:
Do you think this could help troubleshoot https://github.com/maplibre/maplibre-native/issues/2928 . I was trying to figure out how to troubleshoot the node process crash in xcode but wasn't really sure what hooks to watch for since I have never done it before. Does this add hooks I could watch for?

adrian-cojocaru:
@acalcutt Using `onWillStartRenderingFrame`/`onDidFinishRenderingFrame` might help identify if the problem is inside a frame, but considering your test case has a static map I don't think the existing events will help that much.

platform/darwin/src/MLNTileOperation.h
```diff
@@ -0,0 +1,19 @@
+#import <Foundation/Foundation.h>
+
+#ifndef NS_ENUM
```
louwers, line 3: Should be already defined.

louwers (commented):
image
Some documentation strings should be added.

platform/macos/src/MLNMapViewDelegate.h
```diff
@@ -227,6 +228,50 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (BOOL)mapView:(MLNMapView *)mapView shouldRemoveStyleImage:(NSString *)imageName;
 
+// MARK: Shader Compilation
+
+- (void)mapView:(MLNMapView *)mapView
+    shaderWillCompile:(NSInteger)id
+              backend:(NSInteger)backend
```
louwers, line 235: Last nit: does it even make sense to expose this? Since it will be always Metal for iOS?

louwers (approved):

platform/macos/src/MLNMapViewDelegate.h
adrian-cojocaru: I had the same question when adding it. Decided to leave it in for parity with Android and C++ core.

adrian-cojocaru:
Thanks for adding the documentation!

PR #3246, opened 2025-02-21
Add support for parsing MLT-format vector tile sources

### LINKED ISSUE #3725: Raster tile source change is not detected
If you switch to a new style which has a raster source of the same ID with different `encoding` field, the source would keep the cached `Tileset` value, due to that field not being part of the equality test for `Tileset`.

This is speculative, from reading the code, not observed directly.

Now that we're on C++20, we can just use the default operators, though they must be requested explicitly.

TimSylvester:
Loading MLT tiles with embedded metadata from a custom MBTiles file built from the Bing example tiles.  An HTTP source should work as well, but I haven't tried it.

Geometry is loading and rendering correctly in most cases. Here, roads from Bing tiles are loaded from MVT and drawn in red, and the same layer from the MLT tiles and drawn in blue.  Water features are drawn from both with opacity, so everything seems to be lining up well.

image

This is Americana rendering from MLT, generally looking good:

The changes to `DebugBucket` and `SourcePrepareParameters` make the tile labels more useful when multiple sources are in use:

image

Also resolves #3725

### DIFF
--- a/test/style/conversion/source_options.test.cpp
+++ b/test/style/conversion/source_options.test.cpp
@@ -13,9 +13,7 @@ using namespace mbgl::style::conversion;
 TEST(SourceOptions, Basic) {
     Error error;
     auto converted = convertJSON<SourceOptions>("{}", error);
-    ASSERT_TRUE(converted);
-    ASSERT_FALSE(converted->rasterEncoding);
-    ASSERT_FALSE(converted->vectorEncoding);
+    ASSERT_FALSE(converted);
 }
 
 TEST(SourceOptions, ErrorHandling) {


louwers:
I can help with the build tooling. We will discuss it during the standup today.

acalcutt:
I've been thinking about how this would be done in TileServer-GL (maplibre-native node) since we are using custom request handlers. I think there I would have to use the js version of MLT decoding since it would not be able to use the built in ones unless i get rid of that custom request handler.

I will keep an eye on this to see how it gets done here.

include/mbgl/util/tileset.hpp
```diff
@@ -18,24 +18,25 @@ class Tileset {
         XYZ,
         TMS
     };
-    enum class DEMEncoding : bool {
-        Mapbox,
-        Terrarium
+    enum class Encoding : std::uint8_t {
```
louwers: Probably cleaner to split this.

TimSylvester: done

louwers (commented):
A few more comments:

We should probably make it possible to construct a `TileSet` and `VectorSource` on Android. If you want we can merge it first.

The names `VectorMLTTile` and `VectorMVTTile` may not be ideal since if you read out the abbriviation it's "vector MapLibre Tile Tile" or "Vector mapbox vector tile tile".

Is `MLT_UNPACK_MULTI_GEOMETRY` still under development? Or can we decide either to enable or disable it and remove the define?

TimSylvester:
I'd prefer to do JNI stuff separately.  Does it even depend on this?

I considered `MLTile` and `MVTile` but it also sounds weird.  But happy to change it if you want.

I never really resolved which was the better way to do things, but it's working well enough this way, so I'll remove the option.

louwers (approved):

louwers:
Qt CI is not happy with this:

/home/runner/work/maplibre-native/maplibre-native/source/src/mbgl/style/conversion/source_options.cpp:28:15: error: converting to ‘std::optional<mbgl::style::SourceOptions>’ from initializer list would use explicit constructor ‘constexpr std::optional<_Tp>::optional(std::in_place_t, _Args&& ...) [with _Args = {}; typename std::enable_if<__and_v<std::is_constructible<_Tp, _Args ...> >, bool>::type <anonymous> = false; _Tp = mbgl::style::SourceOptions]’
   28 |     return {{}};

src/mbgl/style/conversion/source_options.cpp
```diff
@@ -0,0 +1,33 @@
...
+        } else if (encoding && *encoding == "mlt") {
+            return {{.vectorEncoding = Tileset::VectorEncoding::MLT}};
+        } else {
+            error.message =
+                "invalid encoding - valid types are 'mapbox' and 'terrarium' for raster sources, 'mvt' and 'mlt' for "
+                "vector sources";
+            return std::nullopt;
+        }
+    }
+    return {{}};
```
louwers:
```suggestion
    return {};
```

PR #3255, opened 2025-02-25
Add .git to .bazelignore

louwers:
@alexcristici Could you try if this fixes the indexing problem with Xcode?


TimSylvester:
It probably also makes sense to ignore `.cache`, `.vscode`, `CMakeScripts`, `build`, `node_modules`, and perhaps more.

alexcristici (approved):
Works great now!

PR #3283, opened 2025-03-06
Allow Rust to build core as a static lib

nyurik:


louwers:
Should live in https://github.com/nyurik/maplibre-native-rs

louwers (approved):

Issue #3304, opened 2025-03-12
Android - PMTiles - Asset/Local Source - Crash on Invalid Header / Decompress


zach-snell:
### MapLibre Android Version

11.8.2

### Android SDK Version

35

### Device

Google Pixel Tablet Simulator API 35

### What happened?

I have already diagnosed the issue and will look to open a PR that works but thought I would open an issue for discussion around it.

This may occur on other platforms but I have only tested with Android for now.

In particular, the logic added around PM Tiles expects the providers to respect requested dataRange.
While this is successful with most web sources, not all (any?) of the local sources respect the range or have the ability to handle that concept. As such, when the library tries to operate on the data you have a variety of errors such as invalid compression, invalid header, unexpected data, etc. (Basically it expects a few bytes and gets the whole pmtiles archive every time).

As a stop-gap fix, I created a pmtiles_files_source.cpp modification that wraps requests and ensures the response is the appropriate offset/range. This is very likely not efficient and an improvement could be made by making the android resource provider actually handle pulling specific range of binary data.

However, it is working currently as I am working on a game for my son who likes Geography.
Will link to this issue with my variant for review.

### Steps to reproduce

Launch the MapLibreTestApp
Modify one of the variants to use a pmtiles sourced as pmtiles://asset://world.pmtiles (or any pmtiles archive).

I did this by just over-riding the style.json in the simple map example. 

### Relevant log output

```shell

```

louwers:
Thanks for the report! This is a relatively new feature so I think it is quite likely we overlooked some things. Ideally we should not only fix any problems, but also add some test coverage for loading local PMTiles files.

Thanks for sharing your fix. If you can share some more details about the problem that would be helpful as well.

zach-snell:
I agree. 

However, I also acknowledge I am not as in depth on this library and so tried to make the change as small as possible to show the issue and a resolution.

It may be prudent to modify this to a more comprehensive solutions.
Opened PR here to look at / discuss :  https://github.com/maplibre/maplibre-native/pull/3305

tdcosta100:
@zach-snell, could you please test the changes in #3404?

louwers:
Fix released: https://github.com/maplibre/maplibre-native/releases/tag/android-v11.8.8

PR #3305, opened 2025-03-13
#3304 - Initial resolution for local pmtiles asset, likely better resolution would be upstream in providers.

zach-snell:


louwers:
Not sure if this is something that should be implemented in the PMTiles library itself? (cc @bdon)

@zach-snell Can you provide some more details in what cases the current implementation is problematic?

tdcosta100:
> This was an issue when loading PMTiles directly as an asset on Android.
> 
> i.e.
> 
> pmtiles://asset://world.pmtiles
> 
> In this use case the Android Asset Provider would return the full file without respecting the dataRange concept and break the current implementation.

Shouldn't it be handled in the asset URL handler instead of directly doing it in the PMTiles handler? Asking it because PMTiles puts a range data within the file request, but it's up to each handler read this data. As far as I know, there isn't handling of ranges in asset yet. I did it only for `file://` and `http(s)://` protocols.

louwers:
Yes I think that's the way to go. 👍

zach-snell:
> > This was an issue when loading PMTiles directly as an asset on Android.
> > i.e.
> > pmtiles://asset://world.pmtiles
> > In this use case the Android Asset Provider would return the full file without respecting the dataRange concept and break the current implementation.
> 
> Shouldn't it be handled in the asset URL handler instead of directly doing it in the PMTiles handler? Asking it because PMTiles puts a range data within the file request, but it's up to each handler read this data. As far as I know, there isn't handling of ranges in asset yet. I did it only for `file://` and `http(s)://` protocols.

I agree with you, which is what I was noting before. Though I have less knowledge of the overall code base as to if changing that would impact other areas.

I was mostly working on resolving this for a game for my kid who likes geography ha.
That does feel like the correct solution - if we are using dataRange, every provider should support it.

tdcosta100:
If it works for you, it's a solution, despite not an optimal one. I will dive in the asset handler. The PMTiles component relies in forwarding the original request into a general request so other handlers can provide the data. I initially thought only local files and network files, but not asset URLs, but this can be provided, of course. It's a work in progress component, so as people use and ask for fixed and features, we will improve it.

zach-snell:
Absolutely, if my day job ever gives me some free time I will consider addressing it in the asset manager.
This may apply to other sources on other platforms - so while the fix I have created is not the proper solution - it would likely work as a fix across all providers while each provider is updated / reviewed.

tdcosta100:
Hi, @zach-snell. Could you please tell if #3404 fullfills your needs? If so, we can close this PR.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100, line 1: Since #3404 was merged, I think this solution is not needed anymore.

PR #3310, opened 2025-03-14
Add workaround for android emulator as crash is still presented

TatyanaPolunina:
- There is still crash on Android emulator, inside Android Studio 2023.3.1
- Such workaround is solving the issues
- Here is blocker to move to the latest Android release for Lyft


platform/android/MapLibreAndroid/src/cpp/map_renderer.cpp
```diff
@@ -232,6 +253,14 @@ void MapRenderer::onSurfaceCreated(JNIEnv& env, const jni::Object<AndroidSurface
     renderer = std::make_unique<Renderer>(backend->getImpl(), pixelRatio, localIdeographFontFamily);
     rendererRef = std::make_unique<ActorRef<Renderer>>(*renderer, mailboxData.getMailbox());
 
+#if MLN_RENDER_BACKEND_OPENGL
+    // If we're running the emulator with the OpenGL backend, we're going to crash eventually,
+    // unless we enable this mitigation.
+    if (inEmulator()) {
+        renderer->enableAndroidEmulatorGoldfishMitigation(true);
```
ekamp, line 260: Wondering here if in the future we can point to Vulkan. A few folks mentioned that Vulkan should work better with Google emulation

TatyanaPolunina: I still beleive that fix will be found ealier and we can remove this workaround again :)

louwers (approved):

PR #3313, opened 2025-03-17
fix glfw platform feature state demo

cjbd:
issue:

feature state demo for glfw platform doesn't work now

fix:

the `us_states.geojson` url in `glfw_view.cpp` is not available anymore, the [new link](https://maplibre.org/maplibre-gl-js/docs/assets/us_states.geojson) is from https://maplibre.org/maplibre-gl-js/docs/examples/hover-styles/, which works


louwers (approved):
Thank you!

PR #3315, opened 2025-03-17
Revert #3035 which was causing memory growth issue

louwers:


TimSylvester (approved):
Bummer.  I guess we'll wait for C++23 after all.

louwers:
@TimSylvester Yeah bummer! Do you think it is an issue with the library?

PR #3322, opened 2025-03-19
Rendering statistics view

adrian-cojocaru:
Rendering statistics overlay using a modified symbol layer that displays formatted info from `gfx::RenderingStats`. Enabled by default in `ObserverActivity.kt` / `ObserverExample.swift` and can be toggeled on/off in the iOS and glfw samples.

<p align="center"></p>


alexcristici (approved):

TimSylvester (approved):

src/mbgl/gfx/rendering_stats.cpp
```diff
@@ -99,5 +111,191 @@ std::string RenderingStats::toString(std::string_view sep) const {
...
+    } else if (value >= kb) {
+        ss << value / kb << "KB";
+    } else {
+        ss << value << "B";
+    }
+
+    ss << "\n";
+};
+
+void RenderingStatsView::update(const std::unique_ptr<style::Style>& style, const gfx::RenderingStats& stats) {
```
louwers, line 227: Instead of taking a `const std::unique_ptr<style::Style>&` as a parameter here, I think it is cleaner to take a `style:Style&`. You would need to pass the check the pointer is not null to the caller.

There are a few other places in this PR where a reference to a unique pointer is used as a parameter, those can maybe also be changed although I only checked this one.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapView.java
```diff
@@ -1188,8 +1188,9 @@ public interface OnDidFinishRenderingFrameListener {
      * Called when the map has finished rendering a frame
      *
      * @param fully true if all frames have been rendered, false if partially rendered
+     * @param stats rendering statistics
      */
-    void onDidFinishRenderingFrame(boolean fully, double frameEncodingTime, double frameRenderingTime);
+    void onDidFinishRenderingFrame(boolean fully, RenderingStats stats);
```
louwers, line 1193: This is technically a breaking change right?

louwers: Maybe we can deprecate this API and add a new one that has the full rendering stats?

platform/macos/src/MLNMapViewDelegate.h
```diff
@@ -168,11 +169,31 @@ NS_ASSUME_NONNULL_BEGIN
...
  @param mapView The map view that has just redrawn.
- @param frameTimeNanos The time taken to render the frame, in nanoseconds
+ @param fullyRendered A Boolean value indicating whether the map is fully rendered or not.
+ @param frameEncodingTime The time taken to encode the frame, in milliseconds.
+ @param frameRenderingTime The time taken to render the frame, in milliseconds.
  */
 - (void)mapViewDidFinishRenderingFrame:(MLNMapView *)mapView
                          fullyRendered:(BOOL)fullyRendered
-                             frameTime:(double)frameTime;
+                     frameEncodingTime:(double)frameEncodingTime
```
louwers, line 178: This is also a breaking change right?

louwers (changes requested):
Still needs some documentation, we should just make a brief mention of the rendering statistics HUD and how to enable it.

I think the rendering statistics callbacks are also interesting to point out in the documentation. The HUD is quite a niche use case (I think), but I expect many users will be interested in getting more performance data out of the library.

We should also avoid making a breaking change. Maybe we can deprecate the existing APIs and add a new one with the full rendering statistics? What do you think?

platform/macos/src/MLNMapViewDelegate.h
adrian-cojocaru: This is the macOS delegate. iOS was updated with encoding/rendering times as parameters and mac lagged behind with the changes.

src/mbgl/gfx/rendering_stats.cpp
adrian-cojocaru: Initially there was a branching path if the was no style, but it didn't make sense to show stats for an empty view.

platform/macos/src/MLNMapViewDelegate.h
louwers: Oh yes.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapView.java
adrian-cojocaru: I don't remember exactly why I added new methods for iOS and core, but decided to update the existing one on Android.
Added a new listener with rendering stats.

adrian-cojocaru:
Updated the Android listeners to mirror the iOS delegates (no more breaking changes on Android).
Both the existing and new `onDidFinishRenderingFrame` hooks should be mentioned in the observability documentation started in https://github.com/maplibre/maplibre-native/pull/3409.

louwers:

louwers:

PR #3326, opened 2025-03-20
Add regression test for #3323, bug in layer dependency tracking

### LINKED ISSUE #3323: Bug in layer dependency tracking vs style layer diff mechanism
**Describe the bug**

We have found a bug in MapLibre that was introduced in https://github.com/maplibre/maplibre-native/pull/2113 @TimSylvester 

It is in interaction between the layer dependency tracking and how the style diff works when changing a map style.

What we believe is the issue:
- When a style is first loaded and used, the render instance is created (e.g. `RenderBackgroundLayer`) and the `styleDependencies` are evaluated: https://github.com/maplibre/maplibre-native/pull/2113/files#diff-ff88e05e953ff23945e34b9cb5623cd2c2a6368aa4ac5a0016f49404aed7dfffL43
- When subsequently the style/layer is changed (and if the layer id and type is the same), the same `RenderBackgroundLayer` instance is reused
  - Because `diffLayers` considers the layer to be the same
  - `RenderOrchestrator::createRenderTree` just "updates" the existing `renderLayer` instance at https://github.com/maplibre/maplibre-native/blob/92e9a46c1dc27957cc8462baf805742d03db6176/src/mbgl/renderer/render_orchestrator.cpp#L274-L285
  - But the `styleDependency` update happens only in the render layer constructors

**To Reproduce**

1. Load a style with a single-color `background` layer
3. View the map
4. Change the `background` layer to have a color based on `zoom` expression:
```
background-color: ["interpolate", ["linear"], ["zoom"], 0, "#ff0000", 14, "#00ff00"]
```
5. The zoom-based styling will not work correctly -- the same color will always be used (depending on the initial zoom first used) (the render layer instance does not know that the layer now newly depends on zoom when it previously wasn't and does not reevalute the expressions on zoom change

**Expected behavior**

The map rendering should work correctly independent on the previous state of the style.

**Solution**

Not sure what is the best way to fix this is. We tried extending the `diffLayers` to take into account layer dependencies, but that is probably not the correct solution? Ideally the `styleDependencies` should update in `RenderLayer::layerChanged` but that might require some refactoring?

TimSylvester:
Update the layer dependencies kept in the render layer when the associated style layer is updated or replaced, so that the layer updates correctly in response to zoom, etc., actions.

Adds and fixes a test where the zoom change doesn't trigger an update after a style layer modification.

Hopefully resolves #3323


TimSylvester:
The new test failed as expected before commit `e91f95a`:

Image

TimSylvester:
@petrsloup See if this resolves your problem

louwers (approved):
Happy you got the test to work.

@petrsloup Please also have a look if you can. Thanks for reporting the issue.

cgalvan (approved):

PR #3332, opened 2025-03-24
Change Java Transfrom class from final to normal

grab-liyanjin:
We need to hack/hook several classes to implement the Android Auto outside, full implementation is here https://github.com/maplibre/maplibre-native/pull/3333/files


louwers (approved):
Please edit the "Internal use" comment.

> Note: this class is in principle for internal use. Override and use at your own risk.

PR #3333, opened 2025-03-24
[Draft] Grab full Android auto implementation

grab-liyanjin:


louwers (commented):
We have a separate repo for a MapLibre Android Auto app:

https://github.com/maplibre/MapLibre-Android-Auto-Sample

Do you think we should have the Android Auto app as part of this repo? Or can you upstream it there?

platform/android/MapLibreAuto/src/main/java/org/maplibre/android/maps/ProjectionAuto.java
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (c) 2012-2025 Grab Taxi Holdings PTE LTD (GRAB), All Rights Reserved. NOTICE: All information contained herein is, and remains the property of GRAB. The intellectual and technical concepts contained herein are confidential, proprietary and controlled by GRAB and may be covered by patents, patents in process, and are protected by trade secret or copyright law.
```
louwers, line 2: This is still a draft PR, so I think this is an oversight this is still here and Grab intends to open source this code.

Code contributed to MapLibre repositories should not contain any licenses, but should be licensed under the same license as the repository (also as per GitHub TOS):

> Whenever you add Content to a repository containing notice of a license, you license that Content under the same terms, and you agree that you have the right to license that Content under those terms.

https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license

grab-liyanjin: oh, this code is generated automatically, we will definitely remove these things when it's an official commit

grab-liyanjin:
> We have a separate repo for a MapLibre Android Auto app:
> 
> https://github.com/maplibre/MapLibre-Android-Auto-Sample
> 
> Do you think we should have the Android Auto app as part of this repo? Or can you upstream it there?

 It depends. In Grab, we just add new modules(auto-demo and auto) in the project, as it's easier to share the common parts, such as publish maven scripts, version declaration and sign configurations. But since MapLibre already has separate project for auto, I can follow the current structure, commit auto things to the project https://github.com/maplibre/MapLibre-Android-Auto-Sample

boldtrn:
Thanks for opening this PR @grab-liyanjin - this looks very promising. I was just wondering what the current state of this PR is?

platform/android/MapLibreAuto/src/main/java/org/maplibre/android/maps/TransformAuto.java
```diff
@@ -0,0 +1,197 @@
...
+    //
+
+    @Nullable
+    @UiThread
+    public CameraPosition getCameraPosition() {
+        if (cameraPosition == null) {
+            cameraPosition = invalidateCameraPosition();
+        }
+        return cameraPosition;
+    }
```
ahmedre, line 78: from testing this pr, this method causes the camera position to _always_ be 0,0, even when the underlying native map has the correct position. i believe that this is because `cameraPosition` is never set here beyond the first `invalidateCameraPosition`, and the parent `Transform` class has its own `cameraPosition`.

what if we delegate this method to the parent and remove the `cameraPosition` variable from this class altogether? i tested this and it fixed the camera position for me.

sjg-wdw:
Do we want to revisit this?

louwers:
@sjg-wdw That's really up to Grab.

I think since this is just using public MapLibre Android APIs Android Auto integration doesn't have to live in this repo.

ahmedre:
I spent a lot of time thinking about this and testing this out for HudHud (tested this one within our app and shipped to production with another patch I built based on this one) - I did a detailed analysis and posted some findings in my blog post [here](https://helw.net/2025/11/16/maplibre-on-android-auto/). In summary, there's a performance equivalent solution that is a lot less code using Android's `Presentation` and `VirtualDisplay` apis that we're going with instead.

PR #3334, opened 2025-03-24
Fix Typo

kangyul:
I've noticed a small typo in src/mbgl/gfx/upload_pass.hpp. I think the typo doesn't affect the compilation of the project since it's not being used. However, I thought it would be nice to be corrected.


kangyul:
If so, would it be nice to remove it or leave it as it is?

louwers:
Will piggy bank this on some other PR, thanks! :)

PR #3364, opened 2025-04-05
Retry with curl on CI

louwers:
Make CI more reliable by using `--retry 5` with curl when getting the status of parent workflows on CI.


louwers:
@HarelM Not really, sometimes, very rarely, GitHubs APIs are unreliable.

PR #3373, opened 2025-04-10
Support for routes and vanishing routes rendering

siddharthcmdz:
This PR adds native support for routes and route vanishing. One or more routes can be added/updated/disposed and several properties like color, thickness, etc can be set problematically . This implementation makes use of gradient color expressions to color the traffic zones and hence we only have one draw call to render the route with different traffic zones. There is one more draw call / layer for the casing behind the route.

The key entry points are in the class RouteManager and these APIs are exposed via native map view for android. This changeset is also implemented on opengl and tested as well.

Here is a list of functionalities with this PR:
* Add/remove/dispose routes
* set casing and inner color of the routes
* set geometry of the route without java's geojson source (this was identified as a bottleneck)
* set thickness of the route either as a static width or dynamically change with specified different zoom level
* set location of the route with respect to layers
* create multiple route segments aka traffic zones and modify them anytime
* set custom colors for route segments
* set the priority of render of route segments if route segments overlap
* support for picking the route given a screen space point
* capture a snapshot of the routes as a json file (useful for testing/debugging if client code is complex and heavy and you want to test with glfw view) 
* capture stats that helps how complex the routes are
* glfwview example of using routes

**Demo**
A non comprehensive demo of route as simple circles, static in width, and vanishes route across all (you can vanish route on whatever route you want)
[Screencast from 04-09-2025 05:16:24 PM.webm](https://github.com/user-attachments/assets/d049fc4b-099a-436f-a6cb-725481389d5d)


louwers (commented):
Thanks for working on this awesome feature.

I will take a close look later this week. Some comments from taking a quick first look:

- We don't use `malloc` or even `new` anywhere in the codebase, but instead use [smart pointers](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r20-use-unique_ptr-or-shared_ptr-to-represent-ownership) for improved memory safety. 
- Other guidelines from the C++ Core Guidelines also apply, such as avoiding using macros for program text manipulation.
- Everything in `include` is part of the public API of the library. If it doesn't need to be there, add it to `src` (although we have been moving a lot to `include` lately).
- A feature like this needs C++ Unit Tests as well as Android Tests.
- We should make sure this functionality is available for Vulkan (and probably Metal) as well, I hope someone is available to assist with this.
- We should add iOS bindings as well, I think @ianthetechie offered some help with that.
- I recommend adding a new Activity to the MapLibreAndroidTestApp to demo this functionality. That will also make it easy to reference that code in the Android Examples Documentation.

PR #3381, opened 2025-04-11
Update node-release workflow - bring over changes from node-ci

acalcutt:
This PR includes several updates to the node release workflow to prepare for a new release

1.) Bring over changes from the node-ci that did not make it over to node-release to prepare for a new node pre-release.

2.) I have updated @acalcutt/node-pre-gyp-github to fix the vulnerable packages. This has been included in this release and tested in https://github.com/WifiDB/maplibre-native/actions/runs/14563311631 and https://github.com/WifiDB/maplibre-native/releases/tag/node-v6.0.1-pre.9

3.) I have switched back to @mapbox/node-pre-gyp which now making new releases. I have been working with them to make releases there, so I think it makes sense to move away from my fork. Their latest release no longer supports node 16, so support for node 16 has been removed.

4.) I have removed a few dev dependencies from the node package.json to fix some vulnerabilities (there are now none). I don't think they are actually used anywhere and were just copied over from the main package.json. removing them didn't seem to cause any issues and there are likely other packages that could be removed from dev dependencies that are not actaully used in the node platform.


louwers (approved):
You might want to consider consolidating these two workflows.

FF-98:
Hi all,
can you give a rough estimation on when the new pre release for the node version of maplibre-native will be available?

acalcutt:
I am going to try and put in a PR for release with the next week.

Is there anything in particular you are waiting for? I need to look at updating the changelog, so if there is anything i should note let me know. I am mainly trying to fix the vulnerable packages right now, but i know PMTiles support was also added since the last release.

FF-98:
I am mainly interested on the vulnerability fixes. Maybe this could also be released in a final 6.0.1 version?

PR #3198, opened 2025-02-04
Implement dynamic texture atlas

alexcristici:

PR #3384, opened 2025-04-11
Remove legacy renderer from source, build files and CI

louwers:
@alexcristici is working on https://github.com/maplibre/maplibre-native/pull/3198 which once it lands will offer greatly reduced peak memory usage. We have decided this is probably a good moment to get rid of the legacy renderer, because it is not worth it to backport the required changes to get that to compile as well.

We wanted to get rid of it for some time now, but thus far it hasn't really affected anyone's work and it was sometimes still useful to investigate regressions and performance. 

We are very dependent on the new rendering architecture to support different backends, the last round of testing indicated similar or better performance of the new drawable renderer and we want to move forward refactoring the codebase to be more aligned with the new rendering architecture.

### DIFF
--- a/.github/scripts/windows-ci_configure_wrapper.ps1
+++ b/.github/scripts/windows-ci_configure_wrapper.ps1
@@ -12,12 +12,6 @@ switch ($env:RENDERER)
 	'osmesa' { $compile_flags += '-DMLN_WITH_OSMESA=ON'; break; }
 }
 
-switch ($env:RENDERING_MODE)
-{
-	'legacy'   { $compile_flags += '-DMLN_LEGACY_RENDERER=ON'  ; break; }
-	'drawable' { $compile_flags += '-DMLN_DRAWABLE_RENDERER=ON'; break; }
-}
-
 Write-Host 'Compile flags: ' $compile_flags
 
 & cmake -B build -G Ninja $($compile_flags)
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -187,9 +187,9 @@ jobs:
 
       - name: Build Instrumentation Tests, copy to platform/android
         run: |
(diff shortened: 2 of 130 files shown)


alasram:
Can we make this part of a bigger release that is well tested on Emulator. e.g. Android v12.0.0? We currently run legacy on emulator as it is more stable and faster (There are workarounds in drawable that make emulation slow)

louwers:
@alasram Is using Vulkan in the emulator an option for you?

alasram:
Using Vulkan is generally fine but some devices we support are still on OpenGL. Ideally emulator CI runs OpenGL but I guess we can live with Vulkan only emulation since device running drawable and emulator running legacy are different anyway

birkskyum:
The conversation above confuse me a bit - Does this remove support for OpenGL v3+?

alasram:
No just the legacy GL renderer. The newer drawable GL renderer will remain

birkskyum:
if vulkan isn't an option, can the emulators also not use the opengl v3+ backend?

louwers:
@birkskyum I think we use some new features in the Drawable renderer that are not emulated very well yet. There's some regressions in terms of stability/performance (again, in the emulator).

alexcristici (approved):
Looks good! 
Probably there are some files that needs to be removed too, but we can do that as a follow up PR.

louwers:
@tdcosta100 It seems that windows-ci egl + legacy works, but egl + drawable doesn't. It was disabled.

Merging this now, but do you think it is possible to get drawable + egl to work? Right now it just times out.

tdcosta100:
> @tdcosta100 It seems that windows-ci egl + legacy works, but egl + drawable doesn't. It was disabled.
> 
> Merging this now, but do you think it is possible to get drawable + egl to work? Right now it just times out.

Thanks for pointing out! I will take a look at this issue and PMTiles as well in the next days.

PR #3397, opened 2025-04-15
TextureViewRenderThread#eventQueue should be LinkedList

Tagakov:
Every time `eventQueue.remove(0)` is called the whole array is being shifted left. Since only operations that are performed on the queue are `remove(0)` `add()` and `isEmpty()` LinkedList seems to be a perfect fit.


louwers (approved):

louwers:
Makes sense.

Issue #3112, opened 2025-01-02
map user location annotation not updating location in followWithHeading user tracking mode


matthiaskmt:
### MapLibre iOS Version

6.8.1

### iOS Version

any

### Device

iPhone 13

### What happened?

The map view is not updating the user location while using followWithHeading user tracking mode while the user heading is changing. The location annotation moves again when the user heading remains stable. This is not ideal for use cases like cycling.  

### Steps to reproduce

- Great a map and set user tracking mode to followWithHeading
- Start moving with small changes to the compass.

### Relevant log output

```shell

```

matthiaskmt:
I also checked version mapLibre 6.9.0

Issue #2134, opened 2024-02-23
[iOS] MLNUserTrackingModeFollowWithHeading causing issue with MLNUserLocationAnnotationView location update


JuLink:
**Describe the bug**
Setting the userTrackingMode of the MapView to MLNUserTrackingModeFollowWithHeading leads to bumpy update of the MLNUserLocationAnnotationView. 

**To Reproduce**
1. Use the MapLibre Sample app, but change the style to be able to see more clearly the updates.
2. Set the userTrackingMode to `MLNUserTrackingModeFollowWithHeading` by clicking on the top right navigation bar item (until it represent an arrow with a line in front of it).
3. Start walking around the neighborhood (I did not figure out how to reproduce it on the simulator unfortunately).
4. Notice how the location jump from position to position instead of smoothly following the coordinates of the user.

**Expected behavior**
The update to the `MLNUserLocationAnnotationView` should be smooth exactly like when you use the `MLNUserTrackingModeFollow` userTrackingMode.

**Screenshots**
In the video you can see the difference of update between the `MLNUserTrackingModeFollow` and the `MLNUserTrackingModeFollowWithHeading` update to the user annotation. In the first part of the video, I'm using the Follow mode and in the second part I'm using the  FollowWithHeading. You can see that in the second part of the video the update is not as good as in the first part.
https://github.com/maplibre/maplibre-native/assets/1457215/d2ec55ad-9679-44ae-9fcd-3d520ba497d7

Here is the cut from the video that show the issue:
|Smooth update (follow)| Jumping update (followWithHeading)|
|------|------|
|![ok_480](https://github.com/maplibre/maplibre-native/assets/1457215/c1a32ec4-00a8-4705-95ef-bdf86a1f628a)|![not_ok_480](https://github.com/maplibre/maplibre-native/assets/1457215/55f0ca69-fc52-4e96-8555-0a9fbf5ee7a6)|

As you can see in the second case, the rotation of the map is smooth but the update to the location is not.

**Platform information (please complete the following information):**
 - OS: iOS
 - Platform: iOS
 - Version: This is present on the main branch of the repo and also on latest versions of the SDK (it was there in the 5.x version and it's still there in the 6.1.0 version)

**Additional context**
By doing a few tests I came to the conclusion that the issue is some kind of animation issues between the annotation of the user and the rotation of the map.
[...]

PR #3401, opened 2025-04-17
Fix: follow with heading tracking mode cancels user location indicator update

GenChe:
### Issue Link :link:
#2134 and #3112

### Goals :soccer:
* Fix: follow with heading tracking mode cancels user location indicator update

### Implementation Details :construction:

As far as I understand, the fact that `_setDirection` doesn’t preserve the center coordinate when `UserTrackingMode` is `followWithHeading` seems to be a bug. I added the fix without trying to rework the animation flow. To me, it seems reasonable that if heading updates occur very frequently, they will interrupt the interpolated animation for location changes. However, this shouldn’t cause the blue dot to remain stuck at an old position until the heading updates stop.

### Testing Details :mag:

Tested using a slightly modified version of the sample app (View Route Simulation). I added support for heading updates along with simulated location updates.

In real-world scenarios, this should look even better, since location updates will generally be less abrupt.

https://github.com/user-attachments/assets/a0911cae-54cb-4c42-bbca-06113684d4b1

https://github.com/user-attachments/assets/7c615f9a-600d-4027-8d69-0760bb44b9b1


louwers (changes requested):
The documentation of the public `setDirection:direction:animated` API notes:

>  Changing the heading rotates the map without changing the current center
 coordinate or zoom level.

And this calls `_setDirection:direction:animated` in its implementation.

So you need to find another fix for this, because this is a breaks the `setDirection:direction:animated` method.

GenChe:
> The documentation of the public `setDirection:direction:animated` API notes:
> 
> > Changing the heading rotates the map without changing the current center
> > coordinate or zoom level.
> 
> And this calls `_setDirection:direction:animated` in its implementation.
> 
> So you need to find another fix for this, because this is a breaks the `setDirection:direction:animated` method.

Thanks!

I’ve updated the PR to separate the two possible use cases and apply the center coordinate only when the call is triggered by a new heading update. I also added an animation timing function, similar to [this example](https://github.com/maplibre/maplibre-native/blob/0257d53dfb5a07994a8d1db3f2d39058d47e2e57/platform/ios/src/MLNMapView.mm#L6233-L6242), as used for incremental location changes. The animation looks less choppy to me now.

I tried to dig deeper to figure out what exactly causes the user location annotation to get stuck, but I wasn’t able to pinpoint the issue. My guess is that it’s due to a desync between the `userLocation.coordinate` and the value used to calculate `userLocationAnnotationViewCenter`.

That said, I think applying the last `userLocation.coordinate` on a heading update is reasonable when `userTrackingMode` is set to `follow` or `followWithHeading` — which is the case here.

GenChe:
Hello @louwers 

Please tell me if you want me to change anything else in this PR.

Thanks!

platform/ios/src/MLNMapView.mm
```diff
@@ -6428,7 +6445,7 @@ - (void)locationManager:(__unused id<MLNLocationManager>)manager didUpdateHeadin
         if (headingDirection >= 0 && self.userTrackingMode == MLNUserTrackingModeFollowWithHeading
             && self.userTrackingState != MLNUserTrackingStateBegan)
         {
-            [self _setDirection:headingDirection animated:YES];
+            [self _setDirection:headingDirection shouldCenter:YES animated:YES];
```
louwers, line 6448: I understand `_setDirection` needs the current location so that it can do the animation in one go, but maybe consider passing the center instead of a boolean. I don't think the `_setDirection` method should have anything to do with the location manager.

GenChe: I changed parameter from `shouldCenter` to `center`. And so public `setDirection` method passes `kCLLocationCoordinate2DInvalid` to private `_setDirection` to avoid centering.

GenChe: @louwers is there anything else that needs to be changed or you are fine with it? 

This is a quite important feature for us, and we want to have it merged (and released) as soon as possible.

louwers: Looks good now, thanks!

GenChe:
Hello @louwers

Thanks for enabling auto-merge, but PR should still be reviewed and accepted before merging.

louwers (approved):

GenChe:
thanks @louwers!

seems like pod publish step failed for release 6.13.0 and we can't get this version (because we are using CocoaPods). Can you restart pod publish step, please?

Issue #3366, opened 2025-04-07
Sparse PMTiles source not rendering (all platforms)


louwers:
We have an issue with a particular PMTiles source not rendering. I tried Android, iOS and macOS.

> It looks like this in Maputnik (visualising the lakes of Switzerland):
<img width="1512" alt="Bildschirmfoto 2025-04-07 um 09 48 39" src="https://github.com/user-attachments/assets/dd0b6680-b9ee-4627-bdbd-6202b46cf446" />

_Originally posted by @manuelroth in https://github.com/maplibre/maplibre-native/discussions/3360#discussioncomment-12747469_

Here is Demotiles style with a PMTiles layer added so it is a bit easier to find Switzerland. 

`
https://gist.githubusercontent.com/louwers/b35f629c593b391cd373f76e31adddee/raw/9da03582361065cc4f79a5a5e188ec584c253181/style.json
`

<img width="1938" alt="Image" src="https://github.com/user-attachments/assets/86a34ba0-250b-4ed4-9bef-eb501aef3377" />

manuelroth:
The PMTiles source created by the following steps:
1. Download [Swisstopo's Light Base Map tileset](https://api3.geo.admin.ch/services/sdiservices.html#gettilesets) in MBTiles format: https://vectortiles.geo.admin.ch/tiles/ch.swisstopo.base.vt/v1.0.0/ch.swisstopo.base.vt.mbtiles
2. Convert it to PMTiles format using pmtiles CLI: `pmtiles convert ch.swisstopo.base.vt.mbtiles ch.swisstopo.base.vt.pmtiles`

louwers:
FYI @tdcosta100 I am looking into this, let me know if you have any ideas...

louwers:
Somehow at z=6 it is loading 6/33/41 and not 6/33/22

tdcosta100:
Hi @louwers. I have been reading this issue but did not have time investigate until now. I looked at the [PMTiles](https://github.com/protomaps/PMTiles) repository and they updated the header we use in Maplibre. So could you please try to build with the last commit of the `vendor/pmtiles` submodule and see if it has any impact on this issue?

acalcutt:
I notice in the metadata the mbtiles file uses a tms scheme. isn't the defautl xyz? could the coordinates be inverted?

![Image](https://github.com/user-attachments/assets/86f95b50-29a8-480b-9ec2-7c16da439811)

Does the file display correctly with mbtiles?

[...]

PR #3403, opened 2025-04-17
Force PMTiles metadata to always have XYZ tile scheme

tdcosta100:
This PR reflects the discussion #3366, where it was demonstrated that other PMTiles libraries always assume XYZ tile addressing, so it will be done in here as well.


tdcosta100:
@louwers, could you please take a look here?

tdcosta100:
Exactly! In order to not mess up with existing data, fist I check if there is a scheme key in the JSON object, and create one if not, then hardcoded it to XYZ. It's a little adaptation from the previous code, which also created a scheme key if not existed, but preserved the existing value.

acalcutt (approved):

louwers (approved):

PR #3404, opened 2025-04-17
Add support to range requests in AssetFileSource

tdcosta100:
As pointed out in #3304, PMTiles support was not extended to AssetFileSource, so this PR corrects this.

### DIFF
--- a/platform/default/src/mbgl/storage/asset_file_source.cpp
+++ b/platform/default/src/mbgl/storage/asset_file_source.cpp
@@ -26,8 +26,8 @@ public:
           resourceOptions(resourceOptions_.clone()),
           clientOptions(clientOptions_.clone()) {}
 
-    void request(const std::string& url, const ActorRef<FileSourceRequest>& req) {
-        if (!acceptsURL(url)) {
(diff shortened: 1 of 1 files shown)


tdcosta100:
`node-ci` is failing. Investigating the issue, it seems NAN needs to be updated to 2.22.2. Also, it doesn't work with the ABI version 137 (`v24.0.0`), there is a issue about it: https://github.com/nodejs/nan/issues/996. @acalcutt, can I update the NAN version and exclude v24 while NAN is incompatible with it in another PR?

acalcutt:
Please do submit a PR if you have the time. I think it makes sense to update NaN, and also if that isn't enough for node 24 to work then disabling it for now makes sense.

 if we do disable a release an issue should be made so it isn't forgotten that it was disabled and needs to be fixed.

tdcosta100:
`node-ci` is working fine now, thanks @acalcutt!

louwers:
Do you have time to add a test?

If not happy to merge this as is.

tdcosta100:
You are right! I will add a test before the merge.

louwers:
Cool, thanks!

louwers (approved):

PR #3406, opened 2025-04-18
Fix incomplete feature state updates in GeometryTile and SourceFeatureState [Vector Tile Layer].

NithinAashikPS:
This PR fixes a bug where feature state updates were not applied correctly in `GeometryTile::setFeatureState` due to a mismatch between `layerID` and `sourceLayerID`. It also addresses an issue in `SourceFeatureState::coalesceChanges` where only diffed state changes were passed, potentially causing incomplete state rendering.

**Summary of Changes:**

- Use `renderData.layerProperties->baseImpl->sourceLayer` as the correct `sourceLayerId` for state lookup.
- Ensure `currentStates` are merged into `changes` to reflect the complete state at the time of dispatch.

**Impact:**

- Fixes missing or incorrect application of feature states during rendering **vector tiles**.
- Ensures consistent and predictable behavior when setting or clearing feature states across tiles.


louwers (commented):
We should probably add a regression test for this.

Doing so would involve writing a render test that uses the [`setFeatureState`](https://github.com/maplibre/maplibre-native/blob/b03d4b348e1f74f09fb3ec3a535aecfef09da8ff/render-test/parser.cpp#L637-L638) test operation. Would that be something you are comfortable tackling? I wrote some [documentation](https://maplibre.org/maplibre-native/docs/book/render-tests.html) on the render tests.

NithinAashikPS:
> We should probably add a regression test for this.
> 
> Doing so would involve writing a render test that uses the [`setFeatureState`](https://github.com/maplibre/maplibre-native/blob/b03d4b348e1f74f09fb3ec3a535aecfef09da8ff/render-test/parser.cpp#L637-L638) test operation. Would that be something you are comfortable tackling? I wrote some [documentation](https://maplibre.org/maplibre-native/docs/book/render-tests.html) on the render tests.

Thanks for the suggestion! I took a look and it seems like the existing render test ([render-tests/feature-state/vector-source/style.json](https://github.com/maplibre/maplibre-native/blob/main/metrics/integration/render-tests/feature-state/vector-source/style.json)) already covers this case, including the use of setFeatureState. 

However, after digging deeper, I noticed that the test isn't actually effective as a regression check — the `layer id` and `source-laye`r are both set to `"poi_label"`, which unintentionally masks the bug. This naming overlap causes the test to pass even if the fix isn’t applied, so the bug escapes detection.

I’d be happy to correct the test by renaming the `layer id` to something like `"poi_layer"` to ensure it properly verifies the fix. Let me know if you’d prefer that I update the existing test or create a new one specifically for this.

bitmap

louwers:
@NithinAashikPS Great! Did you by any chance check if it fails before this PR, and now passes?

NithinAashikPS:
> @NithinAashikPS Great! Did you by any chance check if it fails before this PR, and now passes?

After changing the `layer id` the old code fails. New code is working fine.

louwers:
Nice, that's a good test then.

NithinAashikPS:
> Nice, that's a good test then.

Thank You.
What should I do next?

louwers:
I will have another look over it and merge it!

NithinAashikPS:
> I will have another look over it and merge it!

Got it 👍

NithinAashikPS:
@louwers @mnutt @mdaines Can anyone review this bug fix?

MarkusWendorf:
@NithinAashikPS is this change relevant to this feature being available https://maplibre.org/maplibre-style-spec/expressions/#feature-state ? I'm looking to implement the `Create a hover effect` [example](https://maplibre.org/maplibre-gl-js/docs/examples/hover-styles/) with maplibre native but I'm confused about the state of the feature state feature.

There's another feature tracking ticket, but there's already a lot of code and even tests mentioning the feature state property.

NithinAashikPS:
Yes both are same feature.

PR #3409, opened 2025-04-18
Add action journal

adrian-cojocaru:


sjg-wdw (approved):
Looks good!

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/events/ObserverActivity.kt
```diff
@@ -50,9 +51,18 @@ class ObserverActivity : AppCompatActivity(),
         mapView.addOnSpriteRequestedListener(this)
         mapView.getMapAsync {
             it.setStyle(
-                Style.Builder().fromUri(TestStyles.getPredefinedStyleWithFallback("Streets"))
+                Style.Builder().fromUri("https://demotiles.maplibre.org/style.json")
```
louwers, line 54: Can use `TestStyles.Demotiles`

platform/ios/BUILD.bazel
```diff
@@ -145,7 +146,7 @@ apple_xcframework(
         ],
         "device": ["arm64"],
     },
-    minimum_os_versions = {"ios": "12.0"},
```
louwers, line 148: This is a breaking change. Is there a way to avoid that?

src/mbgl/util/action_journal_impl.cpp
```diff
@@ -0,0 +1,522 @@
+#include <mbgl/util/action_journal_impl.hpp>
+#include <mbgl/util/logging.hpp>
+#include <mbgl/style/style.hpp>
+#include <mbgl/map/map.hpp>
+
+#include <rapidjson/document.h>
+#include <rapidjson/stringbuffer.h>
+#include <rapidjson/writer.h>
+
+#include <regex>
```
louwers, line 10: std::regex is hopelessly slow. Can we avoid it?

src/mbgl/util/action_journal_impl.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+        clientName = clientOptions.name();
+        clientVersion = clientOptions.version();
+
+        styleName = style.getName();
+        styleURL = style.getURL();
+    }
+
+public:
+    std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds> time;
```
louwers, line 36: Does this need to be mutable?

src/mbgl/util/action_journal_impl.cpp
```diff
@@ -0,0 +1,522 @@
...
+    : map(map_),
+      options(options_),
+      scheduler(Scheduler::GetSequenced()) {
+    assert(!options.path().empty());
+    assert(options.logFileSize() > 0);
+    assert(options.logFileCount() > 1);
+
+    options.withPath(options.path() + "/" + ACTION_JOURNAL_FOLDER_NAME);
+
+    openFile(detectFiles(), false);
```
louwers, line 123: Returns a bool, presumable to indicate success. But failure is silently ignored?

src/mbgl/util/action_journal_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+class ActionJournalEvent;
+
+class ActionJournal::Impl : public MapObserver {
+public:
+    Impl(const Map& map, const ActionJournalOptions& options);
+    ~Impl() = default;
+
+    const Map& getMap() const { return map; }
+
+    std::vector<std::string> getLog();
```
louwers, line 25: Can be const if you mark the mutex as mutable.

include/mbgl/util/action_journal.hpp
```diff
@@ -0,0 +1,28 @@
...
+namespace util {
+
+class ActionJournalOptions;
+
+class ActionJournal {
+public:
+    ActionJournal(const Map& map, const ActionJournalOptions& options);
+    ~ActionJournal() = default;
+
+    std::vector<std::string> getLog();
```
louwers, line 20: Can probably be const.

include/mbgl/util/action_journal_options.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace mbgl {
+
+namespace util {
+
+class ActionJournalOptions {
+public:
+    ActionJournalOptions() = default;
+    ~ActionJournalOptions() = default;
+
+    ActionJournalOptions& enable(bool value) {
```
louwers, line 14: Honestly I would make `ActionJournalOptions` objects immutable if possible. You could return a new `ActionJournalOptions` if you think that is a good API.

include/mbgl/util/action_journal.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <memory>
+#include <filesystem>
+
+namespace mbgl {
+
+namespace util {
+
+class ActionJournalOptions;
+
+class ActionJournal {
```
louwers, line 15: Public APIs could use some docstrings. Adding them to private APIs also does not hurt.

louwers (changes requested):
What is the rationale behind this functionality? It has been discussed during the TSC Meetings, but it would be good to include it in the PR description as well.

What is the relation between the 'action journal' and the logging framework already built into the library (`logging.hpp`)? I feel like there is some overlap and we can possibly consolidate the two better. Right now for any particular event happening it is not clear to me if it should be logged or added to the action journal.

It looks like we just have an API to read the action journal in an in-memory array. Wasn't the whole point of the action journal system to be able to read it in case things went wrong? Are users supposed to read it, and then clear it, periodically? If so, won't events get lost between reading it in memory and clearing it? We should add some guidance in the form of documentation how users are supposed to use these APIs.

We discussed how we can give the user some flexibility in terms of format, since many people use different logging systems. Right now the JSON structure is baked into the C++ core. I suspect it will be relatively hard for people to parse and process it this way.

Could you explain the reasoning why you decided just exposing events to the platform is not enough? I think we should avoid adding stuff to the core if it is at all possible to let users build on top of the library.

In retrospect I think it would have been better to have this type of discussions on a design proposal PR.

Definitely needs tests, but I would like to see some clarification on the overal design first,

src/mbgl/util/action_journal_impl.cpp
adrian-cojocaru: It's used only on init on a few strings. It can be done without the regex if needed, but the cost is minimal in this case.

platform/ios/BUILD.bazel
adrian-cojocaru: I've removed/worked around all references of `std::filesystem::path`, but there is one remaining `std::filesystem::directory_iterator` used to detect the existing files. One workaround is using the `ghc::filesystem` that was being slowly removed from the codebase or provide platform specific implementations.

src/mbgl/util/action_journal_impl.hpp
adrian-cojocaru: It touches most of the member variables.

include/mbgl/util/action_journal_options.hpp
adrian-cojocaru: I've been mirroring `MapOptions` and `ClientOptions` since all of them are passed in the same place and to be consistent with the different types of options.

src/mbgl/util/action_journal_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+    void onMapCreate();
+    void onMapDestroy();
+
+protected:
+    void log(ActionJournalEvent& value);
+    void log(ActionJournalEvent&& value);
+
+    // file operations
+    std::string getFilepath(uint32_t fileIndex);
+    uint32_t detectFiles();
```
louwers, line 64: What does this method do?

src/mbgl/util/action_journal_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+    void onMapCreate();
+    void onMapDestroy();
+
+protected:
+    void log(ActionJournalEvent& value);
+    void log(ActionJournalEvent&& value);
+
+    // file operations
+    std::string getFilepath(uint32_t fileIndex);
```
louwers, line 63: Can be marked const.

src/mbgl/util/action_journal_impl.cpp
adrian-cojocaru: The open is retried on each subsequent, but indeed the initial error should be logged.

src/mbgl/util/action_journal_impl.hpp
adrian-cojocaru: If the current file is in use (it is in most cases) it will seek and read it's contents. It's intended to reuse the file pointer.

src/mbgl/util/action_journal_impl.cpp
adrian-cojocaru: Fixed.

include/mbgl/util/action_journal.hpp
adrian-cojocaru: Mirrors `ActionJournal::Impl::getLog`

src/mbgl/util/action_journal_impl.hpp
adrian-cojocaru: Fixed.

adrian-cojocaru:
> What is the relation between the 'action journal' and the logging framework already built into the library (logging.hpp)? I feel like there is some overlap and we can possibly consolidate the two better. Right now for any particular event happening it is not clear to me if it should be logged or added to the action journal.

The main difference is that the action journal is persistent and can be used to view previous sessions. `logging.hpp` uses either `std::cerr`, `NSLog` or `__android_log_print` to print the messages to a console (debugging tools required).

> What is the purpose of the APIs to read the action journal in an in-memory array? Isn't the whole point of the action journal system to be able to read it in case things went wrong? 

It's mainly a convenience function. I've used it to display the log in in a separate view.

> Are users supposed to read it, and then clear it, periodically? If so, won't events get lost between reading it in memory and clearing it? Does clearing the log means files get deleted? 

Clearing does indeed delete the files. The example pulls the log, prints it and clears it to keep the next trigger small and with latest entries. The clear function is not intended to be used periodically, it provides a safe way for the user to delete the files (since the current log file is in use/locked) if it's ever needed (regaining space, new app version, etc).

> We discussed how we can give the user some flexibility in terms of format, since many people use different logging systems. Right now the JSON structure is baked into the C++ core. I suspect it will be relatively hard for people to parse and process it this way.

The proposed format was a filter for the event list and additional elements to be added to each json entry.

> Could you explain the reasoning why you decided just exposing events to the platform is not enough? I think we should avoid adding stuff to the core if it is at all possible to let users build on top of the library.

The exposed `MapObserver` will always be more flexible since it gives the user the opportunity to collect any data it needs from the map/app state and store it and any format. The journal is mostly a best effort to collect the most relevant top level events that can help debug a previous crash or problematic session and it can serve as an implementation example.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -897,6 +897,49 @@ public void cycleDebugOptions() {
...
+   *   "event" : {
+   *     "action" : "RequestedFromNetwork",
+   *     "tileX" : 0,
+   *     "tileY" : 0,
+   *     "tileZ" : 0,
+   *     "overscaledZ" : 0,
+   *     "sourceID" : "openmaptiles"
+   *   }
+   * }`
+   */
```
westnordost, line 929: What is the consideration / use case for returning an array of JSON strings? On first glance, it seems unusual to not return a `List<JournalEvent>`, `JournalEvent` being some structured data class, here.

If this is *just* for logging, then why is it JSON?

louwers:
> To that last question, we're going for a 'black box' flight recorder sort of approach. This is really aimed at reconstructing what might have happened in long tail crashes that don't give us coherent stack traces.

Is this the reason why we want this to live in the C++ core?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
adrian-cojocaru: The event data is semi structured. A dictionary/json is human readable, can be extended easily (without having to bind and pass different classes through jni) and it covers the scenario where the logs are uploaded to a cloud service/database and serve as a loose schema for some analytics table.

westnordost: Is this normal? To put structured data into the log output to later parse it back into structured data sounds to me like shoehorning functionality into the log system that doesn't really belong there.

After all, one does also not send off (user behavior) analytics data by first logging it and then later parsing those logs. So, why not send off the journal events in the same way as you would send off analytics data?

westnordost: (And what's the data format of sending off data to some analytics server should be data format agnostic. Maybe some company rather prefers sending data in some protobuf or whatevever)

sjg-wdw: > Is this normal? To put structured data into the log output to later parse it back into structured data sounds to me like shoehorning functionality into the log system that doesn't really belong there.

The Action Journal has its own output destination, so it's not getting mixed in with the log output.

sjg-wdw:
> > To that last question, we're going for a 'black box' flight recorder sort of approach. This is really aimed at reconstructing what might have happened in long tail crashes that don't give us coherent stack traces.
> 
> Is this the reason why we want this to live in the C++ core?

I think it's more that we want the same thing on both platforms.

src/mbgl/map/map.cpp
```diff
@@ -20,6 +20,8 @@
 #include <mbgl/util/mapbox.hpp>
 #include <mbgl/util/math.hpp>
 #include <mbgl/util/tile_coordinate.hpp>
+#include <mbgl/util/action_journal.hpp>
+#include <mbgl/util/action_journal_impl.hpp>
```
stefankarschti, line 24: can we avoid dependence on the journal implementation? creating a higher level interface of ActionJournal may help.

src/mbgl/util/chrono.cpp
```diff
@@ -49,6 +49,22 @@ std::string iso8601(Timestamp timestamp) {
...
+std::string iso8601(std::chrono::time_point<std::chrono::system_clock, Milliseconds> timestamp) {
+    std::time_t time = std::chrono::system_clock::to_time_t(timestamp);
+    std::tm info;
+    _gmtime(&time, &info);
+
+    long long ms =
+        std::chrono::duration_cast<Milliseconds>(timestamp - std::chrono::system_clock::from_time_t(time)).count() %
+        1000;
+
+    char buffer[sizeof("yyyy-mm-ddThh:mm::ss.000Z")];
```
stefankarschti, line 61: _(minor)_ duplicate `:` after `mm`

src/mbgl/map/map.cpp
adrian-cojocaru: Avoided. The `Impl` calls were tied to the `ActionJournal` lifetime and were redundant.

src/mbgl/util/chrono.cpp
adrian-cojocaru: Nice catch!

hy9be:
I don't seem to see any documentations for this feature in the PR. Should we add that?

louwers:
@hy9be Definitely! We need offer some guidance how people can best use this functionality on both iOS and Android. Beyond just showing the APIs to enable it we need to document the event format and write some words in what circumstances you might want to use the action journal, where files get written and how you could process them.

Happy to assist there if we have an example in the Swift-based example app (for iOS) and the Android example app.

stefankarschti (approved):

src/mbgl/shaders/shader_source.cpp
```diff
@@ -0,0 +1,69 @@
...
+                  {BuiltIn::HillshadeProgram, "HillshadeProgram"},
+                  {BuiltIn::LineGradientProgram, "LineGradientProgram"},
+                  {BuiltIn::LinePatternProgram, "LinePatternProgram"},
+                  {BuiltIn::LineSDFProgram, "LineSDFProgram"},
+                  {BuiltIn::LineProgram, "LineProgram"},
+                  {BuiltIn::RasterProgram, "RasterProgram"},
+                  {BuiltIn::SymbolIconProgram, "SymbolIconProgram"},
+                  {BuiltIn::SymbolSDFTextProgram, "SymbolSDFTextProgram"},
+                  {BuiltIn::SymbolSDFIconProgram, "SymbolSDFIconProgram"},
+                  {BuiltIn::SymbolTextAndIconProgram, "SymbolTextAndIconProgram"}});
```
alexcristici, line 67: Most of the legacy programs got removed, except for `ClippingMaskProgram`.
List order got updated, you can guide by the `shader_source.hpp`.

alexcristici (changes requested):

src/mbgl/shaders/shader_source.cpp
adrian-cojocaru: The file is generated by `shaders/generate_shader_code.mjs`. Re-running the script updated the strings.

platform/ios/MapLibre.docc/ActionJournalExample.md
```diff
@@ -0,0 +1,77 @@
...
+Events are stored as JSON objects with the following format:
+
+| Field | Type | Required | Description |
+| :---- | :--: | :------: | :---------- |
+| name | string | true | event name |
+| time | string | true | event time ([ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) with milliseconds) |
+| styleName | string | false | currently loaded style name |
+| styleURL | string | false | currently loaded style URL |
+| clientName | string | false | `ClientOptions::name()` passed on map initialization |
+| clientVersion | string | false | `ClientOptions::version()` passed on map initialization |
```
louwers, line 35: This seems to be an Android API. Is there something similar for iOS? Can you update it?

adrian-cojocaru: `ClientOptions`? I never found if they're exposed on the platform level. `ClientOptions::name()` and `ClientOptions::version()` can be passed when creating the core map.

src/mbgl/util/chrono.cpp
```diff
@@ -49,6 +49,22 @@ std::string iso8601(Timestamp timestamp) {
     return buffer;
 }
 
+std::string iso8601(std::chrono::time_point<std::chrono::system_clock, Milliseconds> timestamp) {
+    std::time_t time = std::chrono::system_clock::to_time_t(timestamp);
+    std::tm info;
+    _gmtime(&time, &info);
```
louwers, line 55: `_gmtime` is not thread-safe. `gmtime_r` is I think?

src/mbgl/util/chrono.cpp
```diff
@@ -49,6 +49,22 @@ std::string iso8601(Timestamp timestamp) {
...
+
+    long long ms =
+        std::chrono::duration_cast<Milliseconds>(timestamp - std::chrono::system_clock::from_time_t(time)).count() %
+        1000;
+
+    char buffer[sizeof("yyyy-mm-ddThh:mm:ss.000Z")];
+    const std::size_t offset = std::strftime(buffer, sizeof(buffer), "%FT%T", &info);
+    snprintf(buffer + offset, sizeof(buffer) - offset, ".%03lldZ", ms);
```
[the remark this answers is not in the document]
adrian-cojocaru: iOS 16.3 adds support for it. It was using `std::format` initially.

platform/ios/MapLibre.docc/ActionJournalExample.md
louwers: We should remove this comment from the platform-documentation then I think.

louwers (commented):
Thanks for working on the Android / iOS documentation!

I am still not sure how the platform APIs are supposed to be used. When calling `getLogFiles`, the log files could have changed by the time we get the chance to read them.

Reading log files in memory with `getActionJournalLog` is also really inefficient I think. So I am unsure why someone would want to use that.

Why don't we just let people know where the log files live, and let people read/process them themselves in any way they please?

I have another question: when people use multiple (map) instances, I think it is very useful to know which action journal events come from which map. Is there any way people can differentiate that right now?

src/mbgl/util/chrono.cpp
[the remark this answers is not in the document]
louwers: Oof OK, need to hold off a bit then.

adrian-cojocaru:
Files should be processed either at application startup or shutdown (usual time for crash detection). 
`getLogFiles` and `getActionJournalLog` are just quick and dirty methods for runtime viewing of the logs.

> Why don't we just let people know where the log files live, and let people read/process them themselves in any way they please?

The location is configured by the user (with a default value for each platform).

> I have another question: when people use multiple (map) instances, I think it is very useful to know which action journal events come from which map. Is there any way people can differentiate that right now?

Each map has it's own action journal. Initially I tried generating unique paths for each instance based on the `ClientOptions` (that never got exposed to the platform level?), but a simpler solution is manually setting different paths using `ActionJournalOptions`.

src/mbgl/util/chrono.cpp
adrian-cojocaru: There's a define at the top of the file that remaps `_gmtime` to `gmtime_r`/`gmtime_s` depending of the platform.

louwers:
> simpler solution is manually setting different paths using `ActionJournalOptions`

Good point, but on iOS this is not possible since `MLNSettings` is global.

platform/ios/app-swift/Sources/ObserverExample.swift
```diff
@@ -10,10 +10,12 @@ class ObserverExampleView: UIViewController, MLNMapViewDelegate {
...
 
         // #-example-code(actionJournalOptions)
-        MLNSettings.actionJournalOptions.enabled = true
+        let options = MLNMapOptions()
+        options.actionJournalOptions.enabled = true
+        mapView = MLNMapView(frame: view.bounds, options: options)
         // #-end-example-code
 
-        mapView = MLNMapView(frame: view.bounds, styleURL: AMERICANA_STYLE)
+        mapView.styleURL = AMERICANA_STYLE
```
louwers, line 18:
```suggestion
        mapView.styleURL = AMERICANA_STYLE
```

Note: this will load this style asynchronously and will cause it to first load the default style.

Maybe we can add the style to MLNMapView as well?

adrian-cojocaru: To `MLNMapOptions`? That works. I can also disable the default style loading for this specific constructor.

louwers: Yes I think it makes sense to move towards a paradigm where we use a configuration object. This will avoid further explosion of different initializer permutations.

louwers (approved):

louwers (approved):

include/mbgl/util/action_journal.hpp
adrian-cojocaru: Added string for the public api (+ some comments for private methods)

louwers:

louwers:

PR #3411, opened 2025-04-18
Update Rust cxx version

nyurik:
Also updates all other rustutils dependencies

According to CXX author, as long as the version of cxx everywhere matches, it should work ok.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -71,10 +71,10 @@ darwin_config(
 )
 
 bazel_dep(name = "rules_rust", version = "0.59.2")
-bazel_dep(name = "cxx.rs", version = "1.0.154")
+bazel_dep(name = "cxx.rs", version = "1.0.157")
 git_override(
     module_name = "cxx.rs",
-    commit = "204d76f53ab2380dd9d532795781beb51c953520",  # 1.0.154
+    commit = "f9d547b60324bc02d9983622159973a75d06ea10",  # 1.0.157
     remote = "https://github.com/dtolnay/cxx.git",
 )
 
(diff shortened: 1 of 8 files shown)


MODULE.bazel
```diff
@@ -71,10 +71,10 @@ darwin_config(
 )
 
 bazel_dep(name = "rules_rust", version = "0.59.2")
-bazel_dep(name = "cxx.rs", version = "1.0.154")
+bazel_dep(name = "cxx.rs", version = "1.0.157")
 git_override(
     module_name = "cxx.rs",
-    commit = "204d76f53ab2380dd9d532795781beb51c953520",  # 1.0.154
+    commit = "f9d547b60324bc02d9983622159973a75d06ea10",  # 1.0.157
```
louwers, line 77:
```suggestion
    tag = "1.0.157",
```

MODULE.bazel
```diff
@@ -71,10 +71,10 @@ darwin_config(
 )
 
 bazel_dep(name = "rules_rust", version = "0.59.2")
-bazel_dep(name = "cxx.rs", version = "1.0.154")
+bazel_dep(name = "cxx.rs", version = "1.0.157")
```
louwers, line 74: We should ignore cxx.rs in `.github/rennovate.json` with `ignoredeps` to make sure it does not get auto-updated `https://docs.renovatebot.com/configuration-options/#ignoredeps

louwers (commented):
Please see comments.

MODULE.bazel
louwers: Actually we can get rid of the `git_override` because cxx.rs is in the Bazel Central Registry now. https://registry.bazel.build/modules/cxx.rs

nyurik: @louwers fixed (i think), please take a look

nyurik: nope, the registry is several versions behind

louwers: @nyurik let's use the version that is in the registry?

louwers (approved):

PR #3412, opened 2025-04-18
Allow including Mermaid diagrams in Developer Docs

louwers:
Use a package [mdbook-mermaid](https://github.com/badboy/mdbook-mermaid) to allow including [Mermaid](https://mermaid.live/) graphs in the Developer Documentation. This way it is easier to update and fix diagrams (several of them are out of date now). They are also look nicer and adapt to the current theme.

mdbook-mermaid instructs commiting a JS blob into your repo, which I think is not a good idea, so I use Mermaid from CDN instead. 

@sjg-wdw FYI

Before

image

After

image

### DIFF
--- a/.github/workflows/gh-pages-mdbook.yml
+++ b/.github/workflows/gh-pages-mdbook.yml
@@ -1,7 +1,10 @@
-name: Build & Deploy mdBook documentation
+name: gh-pages-mdbook
 
 on:
   workflow_dispatch:
+  pull_request:
+    paths:
+      - 'docs/mdbook/**'
   push:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


sjg-wdw:
Neat!

.github/workflows/gh-pages-mdbook.yml
```diff
@@ -20,8 +23,8 @@ jobs:
       - name: Install Dependencies
         shell: bash
         run: sudo apt-get install -y libwayland-dev libxkbcommon-dev # Required for winit
-      - name: Install mdbook-alerts
-        run: cargo install mdbook-alerts
+      - name: Install mdbook-alerts, mdbook-mermaid
```
nyurik, line 26: you might want to use a cache-optimized https://github.com/marketplace/actions/cargo-install

nyurik (approved):
nice!  one small nit to skip cargo builds, but that could be done separate

.github/workflows/gh-pages-mdbook.yml
louwers: Our cache is pretty much non-functional, because it is way too small.

Also I am bit more held back including separate GitHub Actions after the last security incident.

PR #3426, opened 2025-04-24
Make sure Android Instrumentation Tests pass again

louwers:
Had to ignore two tests that need to be investigated later. https://github.com/maplibre/maplibre-native/issues/3425

### DIFF
--- a/platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/styles/TestStyles.kt
+++ b/platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/styles/TestStyles.kt
@@ -5,8 +5,6 @@ import org.maplibre.android.maps.Style
 object TestStyles {
     const val DEMOTILES = "https://demotiles.maplibre.org/style.json"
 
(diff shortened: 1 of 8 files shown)


alexcristici (approved):

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/styles/TestStyles.kt
```diff
@@ -33,7 +31,7 @@ object TestStyles {
             val style = Style.getPredefinedStyle(name)
             return style
         } catch (e: Exception) {
-            return VERSATILES
+            return OPENFREEMAP_LIBERY
```
cgalvan, line 34: Nit: LIBERTY

cgalvan (approved):

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/styles/TestStyles.kt
louwers: Whoops.

PR #3431, opened 2025-04-26
Fix build with GCC 15

ntadej:
Reported by @Ferenc- in https://github.com/maplibre/maplibre-native-qt/issues/203

None of my test environments have GCC 15 at this point but it is a sensible fix.


louwers (approved):

Ferenc-:
Should we also add GCC 15 compilation tests to CI?

PR #3435, opened 2025-04-29
add bounds to rendering cli

Bl4d3s:
I noticed that the CLI did not support a bounding box (north/west/south/east coordinates) yet.

This PR adds support for this feature and ensures a user either passes those or the latitude/longitude/zoom pair.

Not sure, whether this should be added to the Changelog?

if so, maybe like this:
`[bin] add renderer support for LatLngBound via CLI`

### DIFF
--- a/bin/render.cpp
+++ b/bin/render.cpp
@@ -29,10 +29,20 @@ int main(int argc, char* argv[]) {
 
     args::ValueFlag<double> pixelRatioValue(argumentParser, "number", "Image scale factor", {'r', "ratio"});
 
-    args::ValueFlag<double> zoomValue(argumentParser, "number", "Zoom level", {'z', "zoom"});
+    // grouping ensures either bounds or center based position is used
+    args::Group boundsOrCenterZoom(argumentParser, "Position (either one):", args::Group::Validators::AtMostOne);
 
-    args::ValueFlag<double> lonValue(argumentParser, "degrees", "Longitude", {'x', "lon"});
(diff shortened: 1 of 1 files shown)


louwers (approved):
Sure!

Thanks.

Bl4d3s:
Thanks for the review 😄 

Example invocation would be:
```
xvfb-run -a /bin/mbgl-render --style style.json --output image.png \
--north <northWestLatitude> --west <northWestLongitude> --south <southEastLatitude> --east <southEastLongitude> \
--width 512 --height 512
```

louwers:
I feel like it would make more sense to pack this in one argument, since all 4 need to be specified. What do you think?

Bl4d3s:
I see your point, but isn’t this what the argument groups are for? 

It’s the same for the latitude/longitude center, which I based it on.
Those are also separate arguments despite both being required in unison.

but I don’t have a strong opinion :)

louwers:
OK good point, but `--north` is not very descriptive, and `--north-bound` might get too long.

I think `--latlng x x x x` might be the cleanest. But... I don't have a strong opinion either.

Bl4d3s:
If we were to combine the option, should it not be  `--bounds x x x x`?

Otherwise the destinction between the current `--latitude x --longitude x` might get lost?

louwers:
Yes!

Bl4d3s:
Will update the PR this or next week accordingly.

louwers (approved):

PR #3436, opened 2025-04-29
Release Node v6.1.0

acalcutt:
This PR releases Maplibre Node v6.1.0

I tested this windows and linux (x64/arm64) and it seem to work fine there. 

I also tested in macos, but like the existing release it works at first but the locks up the application after a certain number of renders, like mentioned in https://github.com/maplibre/maplibre-native/issues/2928 . I do wonder if there are any ideas to fix this (I have tried working with AI to look at the code and troubleshot but came up with nothing).  @louwers would vulkan be an alternative on macos? I am wondering if it would behave differently.

### DIFF
--- a/platform/node/CHANGELOG.md
+++ b/platform/node/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## main
 
-## 6.1.0-pre.0
+## 6.1.0
 * Add `textFitWidth` and `textFitHeight` properties to sprites ([#2780](https://github.com/maplibre/maplibre-native/pull/2780)).
   More information can be found in the [MapLibre Style Spec](https://maplibre.org/maplibre-style-spec/sprite/#text-fit-properties).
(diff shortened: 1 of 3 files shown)


louwers:
> would vulkan be an alternative on macos

Yes it would! It does have a runtime dependency though (MoltenVK). I can try it out. Do you want to wait with the release?

acalcutt:
I think we can release this since it inproves several vulnerabilities and the macos issue is not new. however maybe after this gets released we could give vulkan a try in macos and see if it can work. 

Would vulkan on macos need new requirements installed on the OS in is running on?

acalcutt:
It looks like MoltenVK installable throgh brew. I think that is how we have people install dependencies on macos right now in workflow (and tileserver). looks like it is https://formulae.brew.sh/formula/molten-vk

acalcutt:
I tested just adding it to brew and changing the macos-node preset to use it in https://github.com/maplibre/maplibre-native/pull/3437, but it didn't seems like it found MoltenVK

cgalvan (approved):

PR #3461, opened 2025-05-13
Make sure renderThread is alive before calling requestExitAndWait()

louwers:
I think this will fix a crash @sjg-wdw reported.

OsmAnd applied this exact fix. https://github.com/osmandapp/OsmAnd/issues/16161


adrian-cojocaru (approved):

TimSylvester:
Is there still a race, if the thread terminates between `isAlive` and `requestExitAndWait`?

TimSylvester (approved):

PR #3465, opened 2025-05-14
Improve core static build release

louwers:
Makes sure ICU and nunicode are included in static library.

Adds some additional headers to the distributed archive with headers.


nyurik (approved):
thx!

cgalvan (approved):

acalcutt:
Would this work for the node linux build? icu is one of the libraries I have trouble with because the build relies on the specific version that was on the server it was built with.

louwers:
@acalcutt Yes I think so.

PR #3469, opened 2025-05-15
Clean legacy code

alexcristici:
Removed legacy shaders/programs.


include/mbgl/shaders/gl/raster.hpp
```diff
@@ -1,89 +0,0 @@
-// Generated code, do not modify this file!
```
louwers: Did you change the script the generates this file?

include/mbgl/shaders/shader_manifest.hpp
```diff
@@ -35,7 +35,6 @@
 #include <mbgl/shaders/gl/drawable_wide_vector.hpp>
 #include <mbgl/shaders/gl/prelude.hpp>
 #include <mbgl/shaders/gl/clipping_mask.hpp>
-#include <mbgl/shaders/gl/raster.hpp>
```
louwers, line 58: This file is also generated.

include/mbgl/shaders/gl/raster.hpp
louwers: Ah yes I see it now.

louwers (approved):

PR #3478, opened 2025-05-16
Clean more legacy code

alexcristici:
Removed legacy texture.


test/renderer/shader_registry.test.cpp
```diff
@@ -206,6 +209,7 @@ TEST(ShaderRegistry, NamedReplace) {
     ASSERT_NE(progA, progB);
 }
 
+#ifdef MLN_RENDER_BACKEND_OPENGL
```
louwers, line 212: It's already disabled.

Does it pass if you enable it now?

louwers (approved):

test/renderer/shader_registry.test.cpp
alexcristici: Still disabled, probably not passing, I didn't fix them. Should compile only for GL only, the code is GL specific.

adrian-cojocaru (approved):

PR #3479, opened 2025-05-16
Validate version android-ci and android-release workflow

louwers:
For some reason the android-release workflow is complaining that it cannot find the `VERSION` file even though the working directory is set to `platform/android`.

### DIFF
--- a/.github/scripts/validate-version.sh
+++ b/.github/scripts/validate-version.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+validate_version() {
+  local version="$1"
+  if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-pre.*)?$ ]]; then
+    echo "::error::Invalid version format in VERSION file. Expected: X.Y.Z or X.Y.Z-pre*"
+    return 1
+  fi
+  return 0
+}
+
+version=$(cat VERSION 2>/dev/null)
+if [ -z "$version" ]; then
+  echo "::error::VERSION file not found or empty"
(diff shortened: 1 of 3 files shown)


.github/scripts/validate-version.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+validate_version() {
+  local version="$1"
+  if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-pre.*)?$ ]]; then
+    echo "::error::Invalid version format in VERSION file. Expected: X.Y.Z or X.Y.Z-pre*"
```
HarelM, line 6: A more useful error message would also add the version that was provided.

.github/scripts/validate-version.sh
```diff
@@ -0,0 +1,23 @@
...
+    return 1
+  fi
+  return 0
+}
+
+version=$(cat VERSION 2>/dev/null)
+if [ -z "$version" ]; then
+  echo "::error::VERSION file not found or empty"
```
HarelM, line 14: Which file?

HarelM (approved):

.github/scripts/validate-version.sh
louwers: Literally the file called `VERSION`.

.github/scripts/validate-version.sh
louwers: I agree, I will add it in another PR since I don't want to wait for CI to run through again.

PR #3482, opened 2025-05-18
add `maplibre-native-base` to `maplibre-native-headers.tar.gz` included in the core release

CommanderStorm:
I think this is one of the header-directories that I am still missing.
Not sure if this the last you are hearing from me about this (hope it is).

We will see ^^


louwers (approved):

CommanderStorm:
Thanks ❤️

PR #3487, opened 2025-05-19
modify the transform implementation to allow for concurrent animations

yousifd:
The previous implementation of the transform class overrides the current animation with the new one. This new implmentation changes that to allow for concurrent transformations to occur for different types of transfomrations at the same time. For example, you can have two concurrent animations where one modifies the zoom while the other modifies the position of the camera without overriding the other.

This change is necessary to allow for concurrent camera animations during navigation.


louwers (changes requested):
Allowing concurrent animations is a nice feature, but without any changes to the platform APIs I don't think this implementation is backward compatible? It should be.

Some general notes:

- Implementations should be added to `.cpp` files (unless it's a class/function template implementation), not in headers.
- Do not use single-letter variables.

src/mbgl/map/transform.hpp
```diff
@@ -16,6 +16,30 @@
 
 namespace mbgl {
 
+template <class T>
```
louwers, line 19: Why are these structs with everything public?

Why are you using a shared pointer? Does that data need to be heap allocated?

src/mbgl/map/transform.hpp
```diff
@@ -122,18 +146,26 @@ class Transform : private util::noncopyable {
...
 
-    void startTransition(const CameraOptions&,
-                         const AnimationOptions&,
-                         const std::function<void(double)>&,
-                         const Duration&);
+    void startTransition(const CameraOptions&, const Duration&);
+    bool animationTransitionFrame(std::shared_ptr<PropertyAnimation>&, double);
+    void animationFinishFrame(std::shared_ptr<PropertyAnimation>&);
+
+    void visit_pas(const std::function<void(std::shared_ptr<PropertyAnimation>&)>& f) {
```
louwers, line 153: Nit: but we tend to use camelCase.

src/mbgl/map/transform.hpp
yousifd: > Why are these structs with everything public?

They don't have to be public, will move them to be private.

> Why are you using a shared pointer? Does that data need to be heap allocated?

It needs to be shared across all `PropertyAnimations` member variables. We need it so that a `latlng` and `zoom` that share the same `PropertyAnimation` (such as when you update the camera to translate and zoom at the same time) you don't have to keep track of two copies of the same `PropertyAnimation` and only have to update one when necessary and the other would be updated as well. For example, if an animation has a finish frame callback and that animation updates both the latlng and the zoom values of the camera we only call the finish frame callback once for either of the latlng or zoom `PropertyAnimation` objects and we wouldn't call it again for the other `PropertyAnimation`. This is necessary to keep backwards compatibility for animations that span multiple properties.

yousifd:
@louwers 
> but without any changes to the platform APIs I don't think this implementation is backward compatible? It should be.

This implementation is backwards compatible. What changes to the platform APIs are you expecting?

src/mbgl/map/transform.hpp
yousifd: @louwers I've exposed a flag to enable or disable this feature during navigation for both android and ios. I've also cleaned up large parts of the code and fixed some of the failing tests. Please let me know if you need help understanding the changes or if you think some of the changes should be done differently.

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,77 @@ class Transform : private util::noncopyable {
...
+    };
+
     TransformObserver& observer;
     TransformState state;
 
-    void startTransition(const CameraOptions&,
-                         const AnimationOptions&,
-                         const std::function<void(double)>&,
-                         const Duration&);
+    void startTransition(const CameraOptions&, const Duration&, std::shared_ptr<Animation>&);
```
louwers: Passing a shared pointer by reference does not make sense I think. 

Just take a normal reference and dereference the pointer when calling it.

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,77 @@ class Transform : private util::noncopyable {
...
+                  bool scaling_,
+                  bool rotating_)
+            : start(start_),
+              duration(duration_),
+              options(options_),
+              panning(panning_),
+              scaling(scaling_),
+              rotating(rotating_) {}
+
+        double interpolant(TimePoint);
```
louwers: Typo?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/location/LocationComponentOptions.java
```diff
@@ -1973,6 +1999,16 @@ public LocationComponentOptions.Builder pulseInterpolator(Interpolator pulseInte
       return this;
     }
 
+    /**
+     * Enable or disable the concurrent camera animations during navigation feature
+     *
+     * @return whether concurrent camera animations are enabled or disabled
+     */
+    public LocationComponentOptions.Builder concurrentCameraAnimationsEnabled(Boolean concurrentCameraAnimationsEnabled) {
```
louwers: Enabling concurrent camera animations is not only relevant for the LocationComponent, is it? So it should not be a setting that exists only here.

src/mbgl/map/transform.hpp
yousifd: I wasn't sure about the name myself but that's the name of a function that returns the value used for interpolation https://en.wiktionary.org/wiki/interpolant

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/location/LocationComponentOptions.java
yousifd: Currently that's the only use case I can think of. If you want to have multiple animations at the same time from the application side then you could just bundle the animations into a single `setCamera` call.

If you still think that concurrent camera animations should be universal then I'd also have to change the logic around `cancelTransitions` to not check for the current camera mode.

src/mbgl/map/transform.hpp
yousifd: done https://github.com/maplibre/maplibre-native/pull/3487/commits/0f387f2ad5d1bdf79acef2ef3d0751d2e3ebc940

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,87 @@ class Transform : private util::noncopyable {
...
+
+        bool isAnimated() const { return duration != Duration::zero(); }
+    };
+
 private:
+    template <class T>
+    struct Property {
+        std::shared_ptr<Animation> animation;
```
louwers, line 174: Why is this a shared pointer?

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,87 @@ class Transform : private util::noncopyable {
...
 
-    void startTransition(const CameraOptions&,
-                         const AnimationOptions&,
-                         const std::function<void(double)>&,
-                         const Duration&);
+    void startTransition(const CameraOptions&, const Duration&, Animation&);
+    bool animationTransitionFrame(Animation&, double);
+    void animationFinishFrame(Animation&);
+
+    void visitProperties(const std::function<void(Animation&)>& f) {
```
louwers, line 196: Please don't use single-letter variables.

src/mbgl/map/transform.cpp
```diff
@@ -692,4 +806,15 @@ void Transform::setFreeCameraOptions(const FreeCameraOptions& options) {
     state.setFreeCameraOptions(options);
 }
 
+double Transform::Animation::interpolant(TimePoint now) {
+    bool isAnimated = duration != Duration::zero();
```
louwers: Can use the method you added.

src/mbgl/map/transform.cpp
```diff
@@ -158,37 +158,79 @@ void Transform::easeTo(const CameraOptions& inputCamera, const AnimationOptions&
...
+        if (properties.latlng.animation) {
+            animationFinishFrame(*properties.latlng.animation);
+        }
+        properties.latlng = {
+            .animation = animation,
+            .current = startPoint,
+            .target = endPoint,
+            .set = true,
+            .frameLatLngFunc =
+                [=, this](TimePoint now) {
```
louwers: Please see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-this-capture

src/mbgl/map/transform.cpp
```diff
@@ -519,130 +617,146 @@ ProjectionMode Transform::getProjectionMode() const {
...
+    return properties.latlng.set || properties.zoom.set || properties.bearing.set || properties.padding.set ||
+           properties.pitch.set;
+}
 
-        if (anchor) state.moveLatLng(anchorLatLng, *anchor);
+void Transform::updateTransitions(const TimePoint& now) {
+    if (!activeAnimation) {
+        activeAnimation = true;
+
+        bool panning = false, scaling = false, rotating = false;
```
louwers: Nit: we tend not to use multiple variable declarations.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-name-one

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,87 @@ class Transform : private util::noncopyable {
...
     void setFreeCameraOptions(const FreeCameraOptions& options);
 
+    struct Animation {
+        TimePoint start;
+        Duration duration;
+        AnimationOptions options;
+        bool ran = false;      // Did this property animation run this frame
+        bool finished = false; // Did we execute the finish frame for this property animation this frame
+        bool done = false;     // Did this property animation reach the end of the frame
+        bool panning = false, scaling = false, rotating = false;
```
louwers: Please define separately https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-name-one

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,87 @@ class Transform : private util::noncopyable {
     FreeCameraOptions getFreeCameraOptions() const;
     void setFreeCameraOptions(const FreeCameraOptions& options);
 
+    struct Animation {
```
louwers, line 136: Does this need to be a `struct`? Can it be a class?

src/mbgl/map/transform.hpp
louwers: I think it can be const right?

src/mbgl/map/transform.cpp
```diff
@@ -103,13 +103,13 @@ void Transform::jumpTo(const CameraOptions& camera) {
  * smooth animation between old and new values. The map will retain the current
  * values for any options not included in `options`.
  */
-void Transform::easeTo(const CameraOptions& inputCamera, const AnimationOptions& animation) {
+void Transform::easeTo(const CameraOptions& inputCamera, const AnimationOptions& options) {
```
louwers: Nit: instead of renaming this to `options`, perhaps use `animationOptions`? Goes for several places.

src/mbgl/map/transform.hpp
```diff
@@ -132,22 +132,87 @@ class Transform : private util::noncopyable {
...
+        TimePoint start;
+        Duration duration;
+        AnimationOptions options;
```
louwers: An `Animation` has a `duration`, but an `Animation` also has an `AnimationsOptions` object with an optional `duration`. I think this needs some more refactoring...

louwers (changes requested):
Left some more comments. As discussed on Slack, needs test coverage before we can consider merging this.

General comment: please use `const` for definitions wherever possible.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/location/LocationComponentOptions.java
louwers: > If you want to have multiple animations at the same time from the application side then you could just bundle the animations into a single `setCamera` call.

Why is this not possible in the location component? 

What if someone wants to implement their own location component?

src/mbgl/map/transform.hpp
yousifd: Yes

src/mbgl/map/transform.hpp
yousifd: The `Animation` object is used to keep track of the properties current animation. It keeps track whether we've ran the animation's frame callback, animation's finish callback, and the panning, scaling, and rotating state.

So, to avoid updating the above for every property we use a shared pointer to ensure that if any one of the properties finished running the animation's frame or finish callbacks then the other one doesn't do it again.

src/mbgl/map/transform.hpp
yousifd: I'm not sure if I can come up with a more descrptive name for the function, maybe I can call it `func`. I followed the same style you already have for visit functions similar to `RenderOrchestrator::visitLayerGroups`.

src/mbgl/map/transform.hpp
yousifd: So in the `flyTo` and `easeTo` functions where we initialize the `Animation` object there is logic around setting the `duration` value that doesn't necessarily equal to the `duration` value in `AnimationOptions`. Maybe I could change those to override the value in `AnimationOptions` but that might have unintended side effects.

src/mbgl/map/transform.hpp
yousifd: It's just a container class of related information. I don't think there is any need to make the properties private by default and add accessors. I can make the whole definition of `Animation` be private to avoid others using it if that makes the usage clearer?

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/location/LocationComponentOptions.java
yousifd: I think I understand. Would it make more sense to add a boolean to the java `Transform` class and expose the concurrent animations flag from the `MapView` class instead?

louwers (changes requested):
I just tried the GLFW app, you can build and run it with:

```
cmake --preset macos
cmake --build build-macos --target mbgl-glfw
build-macos/platform/glfw/mbgl-glfw --style https://tiles.openfreemap.org/styles/liberty
```

You can jump between various locations with the 'A' key.

It seems to be broken with this PR.

Before:

https://github.com/user-attachments/assets/afb878e6-244d-4235-b256-8a98490606bf

After:

https://github.com/user-attachments/assets/c2edebc1-0013-4741-9b9c-67e14e8a18cc

I am skeptical of the approach taken by this PR. Whether concurrent camera animations are enabled or not should be set on the C++ level, not defined at the platform level, definitely not tucked away in an internal component only used by the platform itself. Relying on `cancelTransitions` being called is brittle and not backward-compatible on all platforms. Also the Transform class having knowledge of the location component is not a good design. In its current form I do not feel comfortable approving this PR.

What I suggest now and for next time: create an issue describing the problem first. Perhaps with a video or two. I still do not 100% understand the problem being addressed here. Especially when modifying fundamental code in the library that has the potential to disrupt our users, it is better to discuss the issue first or even write a [design proposal](https://github.com/maplibre/maplibre-native/tree/main/design-proposals). That way we can together explore the solution space and choose a good approach before code is written.

It took a while for my opinion to crystalize. I apologize that you went though several rounds of reviews and addressing comments. I aim to do better next time and I hope this does not discourage you from contributing.

yousifd:
@louwers I fully understand and I should have started with an issue or some sort of longer form discussion about the problem. I just got a bit too excited and shared this PR hoping we can have this discussion here. I'll post the issue with all the necessary details and the approach I took here and we can go from there.

I fully understand your reluctance to merge this and I expected it. I'll make sure to follow the same process going forward for other significant changes.

yousifd:
@louwers

- I resolved the fly to camera animation issues in this commit f6a66c18ff105fb7c6168b6445a0eaef66336c20.
- Currently concurrent camera animations is enabled by default to users if they directly transform the camera; however, if they interact with the camera through gestures or by following the user's location then the default behavior if enforced where previous transitions are canceled and the gesture or the animation due to a location update takes priority over any previous animation. However, I've changed this such that if the user has concurrent camera animations enabled then we wouldn't cancel the transitions and an activate animation with a non-conflicting gesture don't overlap each other. You can see this change here [adf734c](https://github.com/maplibre/maplibre-native/pull/3487/commits/adf734cd70d113467cd5aca45dbeb7f973245bf3)
- There is still an issue with two concurrent animations one that animates zoom and the other animates latlng with different anchor values. Will investigate that when I have the time but we can still discuss the other issues until then.

Also, do you have any comments about the implementation?

yousifd:
> * There is still an issue with two concurrent animations one that animates zoom and the other animates latlng with different anchor values. Will investigate that when I have the time but we can still discuss the other issues until then.

I investigated this issue and it also happens in the non-concurrent implementation if you use the `mapViewUserLocationAnchorPoint` delegate method to set the anchor point for the user location.

sjg-wdw:
@mwilsnd and @stefankarschti would you two be willing to review this?  We could use some eyes on this that might have more experience with the cameras.

louwers:
@ntadej Any chance you could try this out with the Qt platform?

ntadej:
I could do it towards the end of the week.

louwers:
@ntadej That would be great, thank you.

sjg-wdw:
@louwers did you ever get feedback on this PR?

PR #3488, opened 2025-05-20
Increase the high-end GCC version in qt-ci

Ferenc-:


ntadej (approved):

ntadej:
The change is ok, but the test will have to be fixed.

Ferenc-:
@ntadej Thanks for the quick feedback!

Are we sure that the test is at fault? Because this is not an assertion failure, but simply a call to `image.resize({0, 0});` raises an error from STL, because of trying to do a memset into an 0 long buffer.


Either the if statement [here](https://github.com/maplibre/maplibre-native/blob/main/include/mbgl/util/image.hpp#L79) should ensure that none of the dimensions is 0:
```cpp
    void fill(uint8_t value) {
        if (valid()) {
```

Or perhaps the [bool valid()](https://github.com/maplibre/maplibre-native/blob/main/include/mbgl/util/image.hpp#L64) should return false if any of the dimensions is `0`.

Ferenc-:
So I have surmised from the `copy` function, that when size is empty we can return, so I did the same for `fill`.
@cgalvan could you check?

include/mbgl/util/image.hpp
```diff
@@ -77,7 +77,7 @@ class Image : private util::noncopyable {
     size_t bytes() const { return stride() * size.height; }
 
     void fill(uint8_t value) {
-        if (valid()) {
+        if (valid() && !size.isEmpty()) {
```
louwers, line 80: Already implied by `valid()`

Ferenc-:
@louwers So this appears to be a compiler bug. But now the amended change avoids that and makes the code perhaps a bit more rational. WDYT?

Ferenc-:
The 2 windows-ci failures are both appear to be caused by timeouts. Perhaps they can be re-triggered?

louwers (approved):

louwers:
@ntadej Could you merge after CI passed?

Issue #3091, opened 2024-12-13
Scaling issue on some devices (Vulkan)


mtpdog:
### MapLibre Android Version

11.7.0-vulkan

### Android SDK Version

Android 10

### Device

Samsung Galaxy M52

### What happened?

There is an issue with map scaling on some devices with Vulkan renderer.

Vulkan (1st device)
https://github.com/user-attachments/assets/4d53fea3-9b26-47f7-874b-ac879ab55757

Vulkan (2nd device)
https://github.com/user-attachments/assets/2dae086e-79eb-4f3b-b20e-5b08a2883603

OpenGL (1st device)
https://github.com/user-attachments/assets/9d4c1641-02ab-4b04-9615-b35d52198a3a

### Steps to reproduce

Here is an example code: 

    @Composable
    fun MaplibreScalableMap() {
        val map = rememberMapView()

        LaunchedEffect(Unit) {
            val maplibreMap = map.awaitMap()
            maplibreMap.setStyle("mapbox://styles/mapbox/satellite-v9")
        }

        // State to track current map size
        var currentMapSize by remember { mutableStateOf(MapSizeNew.QUARTER) }

        // Define constraint sets for different map sizes
        val quarterConstraints = ConstraintSet {
            val mapRef = createRefFor("map")
            val buttonsRef = createRefFor("buttons")

            constrain(mapRef) {
                bottom.linkTo(parent.bottom)
                start.linkTo(parent.start)
                width = Dimension.percent(0.5f)
                height = Dimension.percent(0.25f)
            }

            constrain(buttonsRef) {
                top.linkTo(parent.top, margin = 16.dp)
                start.linkTo(parent.start)
                end.linkTo(parent.end)
            }
        }

        val halfConstraints = ConstraintSet {
            val mapRef = createRefFor("map")
            val buttonsRef = createRefFor("buttons")

            constrain(mapRef) {
                bottom.linkTo(parent.bottom)
                start.linkTo(parent.start)
                width = Dimension.percent(0.5f)
                height = Dimension.percent(0.5f)
            }

            constrain(buttonsRef) {
                top.linkTo(parent.top, margin = 16.dp)
                start.linkTo(parent.start)
                end.linkTo(parent.end)
            }
        }

        val fullConstraints = ConstraintSet {
            val mapRef = createRefFor("map")
            val buttonsRef = createRefFor("buttons")

            constrain(mapRef) {
                bottom.linkTo(parent.bottom)
[...]

PR #3489, opened 2025-05-21
Fix vulkan scaling issue

adrian-cojocaru:
Fixes https://github.com/maplibre/maplibre-native/issues/3091

### DIFF
--- a/src/mbgl/vulkan/pipeline.cpp
+++ b/src/mbgl/vulkan/pipeline.cpp
@@ -6,6 +6,10 @@
 namespace mbgl {
 namespace vulkan {
 
+#ifndef USE_DYNAMIC_VIEWPORT
+#define USE_DYNAMIC_VIEWPORT 0
+#endif
+
 vk::Format PipelineInfo::vulkanFormat(const gfx::AttributeDataType& value) {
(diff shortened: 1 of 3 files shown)


src/mbgl/vulkan/pipeline.cpp
```diff
@@ -6,6 +6,10 @@
 namespace mbgl {
 namespace vulkan {
 
+#ifndef USE_DYNAMIC_VIEWPORT
```
louwers, line 9: Should we even make this configurable?

louwers (approved):

src/mbgl/vulkan/pipeline.cpp
adrian-cojocaru: It doesn't need to be configurable (force of habit).

louwers: let's remove the macro then

adrian-cojocaru: Removed the `#ifndef` part. The macro is still useful for debugging similar issues.

alexcristici (approved):

PR #3491, opened 2025-05-25
Release workflow fixes, release MapLibre Android 11.9.0

louwers:
Simplify changelog extraction.

Makes sure git repo is checked out so that android-ci can run android-release.

Release MapLibre Android 11.9.0


HarelM (approved):
I think I have a similar script in maplibre-gl-js, probably needs this fix as well... I'm guessing a better npm package is what's needed here, but this will probably do for the near-mid future...

PR #3495, opened 2025-05-26
Pattern layout performance improvement

TimSylvester:
In profiling tile parsing with a complex style (Americana), I observed on numerous occasions that the heaviest stack trace was in inserting items into the pattern dependency map, or in the generated move assignment operator.  Even with `unordered_map` the assignment operator moves each individual element.  Holding the map in a `unique_ptr` allows it to be moved at O(1) rather than O(n), which moves feature setup to the second heaviest item, from greater than the expression evaluation stage to less than half of it, in this particular example.

image

&nbsp;

image


louwers (changes requested):
Please correct me if I am wrong, but won't declaring the parameter as `PatternLayerMap &&` solve the problem? Then it should actually do a (cheap) move instead of an expensive copy.

TimSylvester:
It's already being moved, the bare type syntax allows copy elision when possible, though an explicit move would probably be better here.  The problem is that the move still isn't all that cheap, as it's visiting each item.

I think the `PatternLayout` is being moved multiple times, because I never saw the `PatternLayerMap` move in the constructor show up in profiling, only the one in the copy operator.  I had considered using `unique_ptr` even earlier, and passing that in to the constructor instead, but since it didn't seem important I went with a more contained change.

louwers:
Oh yes, copy elision should work there. Maybe not ideal to rely on it though. But I still wonder why the move would be O(n) since according to cppreference the move constructor of std::map should be constant: https://en.cppreference.com/w/cpp/container/map/map.html

There are some other places that take a 'bare' PatternLayerMap.

louwers (approved):
Still don't understand it, but I believe you.

alexcristici:
> One render test failing on Pixel 7 Pro
> 
> image

I've fix it, I will create a PR for it.

alexcristici (approved):

alexcristici:
> > One render test failing on Pixel 7 Pro
> > image
> 
> I've fixed it, I will create a PR for it.

PR that fix it - https://github.com/maplibre/maplibre-native/pull/3504

louwers:

louwers:
Release in MapLibre iOS 6.15.0 https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.15.0

PR #3497, opened 2025-05-27
improve automated MapLibre Android release

louwers:
The `android-release` workflow is triggered from `android-ci`.

But the `gh workflow run android-release.yml --ref {ref}` command requires a tag, not a commit. So create the tag before triggering the workflow. I extracted the `ensure-tag.sh` script and made some improvements to the `validate-version.sh` script.

### DIFF
--- a/.github/scripts/ensure-tag.sh
+++ b/.github/scripts/ensure-tag.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+if [ $# -lt 1 ] || [ $# -gt 2 ]; then
+  echo "Usage: $0 <tag> [<commit-sha>]"
+  echo "  <tag>        Git tag name to create or verify"
+  echo "  <commit-sha> Optional commit SHA to use (defaults to HEAD)"
+  echo ""
+  echo "This script will:"
+  echo "  - Check if the tag exists and matches the specified commit"
+  echo "  - Create and push the tag if it doesn't exist"
+  echo "  - Exit with error if tag exists but points to different commit"
+  exit 1
+fi
+
+tag=$1
+commit_sha=${2:-$(git rev-parse HEAD)}
+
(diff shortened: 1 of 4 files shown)


.github/scripts/ensure-tag.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
HarelM, line 1: Isn't there a GitHub action that is similar to what this script is doing?

.github/scripts/validate-version.sh
```diff
@@ -3,21 +3,40 @@
 validate_version() {
   local version="$1"
   if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-pre.*)?$ ]]; then
-    echo "::error::Invalid version '$version' in $(realpath VERSION). Expected: X.Y.Z or X.Y.Z-pre*"
+    echo "::error::Invalid version '$version' in $(realpath "$version_file"). Expected: X.Y.Z or X.Y.Z-pre*"
```
HarelM, line 6: Isn't this a manual process? When was a non-valid version released that this validation is needed? I must admit I find this odd... but it might just be me...

HarelM (approved):
I think I have a similar process in maplibre-gl-js that only uses github actions, but this might be needed here, IDK...

.github/scripts/validate-version.sh
louwers: Yes it is manual, but people can make mistakes. Especially when making a pre-release, which does not require a PR review for a version release.

.github/scripts/ensure-tag.sh
louwers: Probably. But I prefer not adding another dependency (that can break, or be compromised) for something so simple.

I will look into it if it doesn't work.

PR #3502, opened 2025-05-28
Expose MLNSource.attributionHtmlString

### LINKED ISSUE #3501: iOS: Expose attribution string directly

On Android, we have `Source.attribution` which provides the actual source attribution string.

On iOS, we have `MLNSource.attributionInfos`, which parses out the attribution into instances of `MLNAttributionInfo`. This parsing involves [a lot of complex logic](https://github.com/maplibre/maplibre-native/blob/72598cb7467734f647735019427debf3eb0fe635/platform/darwin/src/MLNAttributionInfo.mm#L68-L83), and it's causing some (very cryptic) crashes for us in maplibre-compose.

We only really want to extract the link titles and URLs, and we parse that out from the String ourselves on Android. It'd be helpful to have access to the underlying attribution string so we can do that on iOS too.

Probably just expose this property: https://github.com/maplibre/maplibre-native/blob/72598cb7467734f647735019427debf3eb0fe635/platform/darwin/src/MLNTileSource.mm#L43-L47

None

sargunv:
Untested, and unsure if this is all that's needed to expose the property or if there's more.

Would resolve #3501

### DIFF
--- a/platform/darwin/src/MLNTileSource.h
+++ b/platform/darwin/src/MLNTileSource.h
@@ -204,6 +204,15 @@ MLN_EXPORT
  */
 @property (nonatomic, copy, readonly) NSArray<MLNAttributionInfo *> *attributionInfos;
 
+/**
+ The attribution HTML string associated with this source. 
+
+ By default, this is nil. If the source is initialized with a
+ configuration URL, this is nil until the configuration JSON file
+ is loaded.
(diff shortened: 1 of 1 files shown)


louwers (approved):
Tested, works.

louwers:

PR #3507, opened 2025-05-29
Fix VulkanSDK version

adrian-cojocaru:
Latest version doesn't provide runtime components for download. Hardcoded to the previous version.

It seems it's relatively common to skip the runtime zip if there are no changes in the newer version. Ideally we should add some retry logic with version downgrading.

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -199,7 +199,7 @@ jobs:
         # 1.4.313.1 doesn't have runtime components?
         # $version = (Invoke-WebRequest -Uri "https://vulkan.lunarg.com/sdk/latest.json" | ConvertFrom-Json).windows
         run: |
-          $version = "1.4.313.0" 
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-ci.yml
```diff
@@ -196,8 +196,10 @@ jobs:
 
       - name: Download and configure Vulkan
         if: matrix.renderer == 'vulkan'
+        # 1.4.313.1 doesn't have runtime components?
```
cgalvan, line 199: Can we open a ticket to track/follow-up on this?

cgalvan (approved):

.github/workflows/windows-ci.yml
adrian-cojocaru: https://github.com/maplibre/maplibre-native/issues/3508
Added more context in the PR's description

louwers:
pre-commit.ci run

PR #3509, opened 2025-05-29
Test case for tile lifetime issue

TimSylvester:
Adds a test case which forces a tile data request to finish after the tile is deleted because the source/layer that created it is removed.

Unfortunately, running this 1000x hasn't  reproduced the target crash, even when using the unsafe weak reference check.  I think that issue has been so elusive because it requires both of these rare cases to occur together.

### DIFF
--- a/src/mbgl/renderer/render_source.hpp
+++ b/src/mbgl/renderer/render_source.hpp
@@ -50,6 +50,11 @@ using RenderTiles = std::shared_ptr<const std::vector<std::reference_wrapper<con
 class RenderSource : protected TileObserver {
 public:
     static std::unique_ptr<RenderSource> create(const Immutable<style::Source::Impl>&, const TaggedScheduler&);
+
+    static bool registerRenderSourceType(const style::SourceType& type,
+                                         std::function<std::unique_ptr<RenderSource>(Immutable<style::Source::Impl>)>);
+    static bool unregisterRenderSourceType(const style::SourceType& type);
+
     ~RenderSource() override;
 
     bool isEnabled() const;
--- a/src/mbgl/util/std.hpp
+++ b/src/mbgl/util/std.hpp
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <memory>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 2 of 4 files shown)


sjg-wdw:
What would we do for a prospective fix for the original problem?

src/mbgl/util/std.hpp
```diff
@@ -25,3 +24,12 @@ void erase_if(Container& container, Predicate pred) {
 
 } // namespace util
 } // namespace mbgl
+
+namespace std {
+#if !__has_cpp_attribute(__cpp_lib_to_underlying)
+template <typename E>
+constexpr auto to_underlying(E e) noexcept {
```
louwers, line 31: Is this not supported on any of the platforms that we support?

TimSylvester: It's C++23

https://en.cppreference.com/w/cpp/utility/to_underlying.html

louwers: Makes sense!

test/map/map.test.cpp
```diff
@@ -1892,3 +1900,195 @@ TEST(BackgroundLayer, StyleUpdateZoomDependency) {
...
+// Forces the case where an asynchronous tile load occurs after the source that created it has been removed
+TEST(Map, RemovedSource) {
+    TaggedScheduler scheduler{Scheduler::GetBackground(), {}};
+
+    // Register the custom source, created tiles will be stored in `tiles`
+    TestInfo info;
+    RenderSource::registerRenderSourceType(customSourceType, [&](Immutable<Source::Impl> impl) {
+        return std::make_unique<TestRenderSource>(staticImmutableCast<TestSource::Impl>(impl), scheduler, info);
+    });
+    // Automatic cleanup.  Use `std::scope_exit` when available.
```
louwers: We have `Scoped` in `mbgl/util/scope.hpp`.

src/mbgl/renderer/render_source.hpp
```diff
@@ -50,6 +50,11 @@ using RenderTiles = std::shared_ptr<const std::vector<std::reference_wrapper<con
 class RenderSource : protected TileObserver {
 public:
     static std::unique_ptr<RenderSource> create(const Immutable<style::Source::Impl>&, const TaggedScheduler&);
+
+    static bool registerRenderSourceType(const style::SourceType& type,
```
louwers, line 54: I would have prefer if this was passed as an argument instead of some global state.

But if you don't want to re-address this PR, maybe add a triple-slash comment that these methods are not thread-safe and to used for testing only.

TimSylvester: It's being used in a static method, so it would have to be provided to `RenderOrchestrator` somehow.

It might end up having some use along with custom sources, relating to Malcolm's work.  But if so, it should really work differently, so I'll have another look at it.

PR #3522, opened 2025-06-04
Make sure AndroidRendererFrontend exists when accessing it

louwers:
This should improve things somewhat since at least `AndroidRendererFrontend` still exists when the async task is run.

I think the `mapRenderer` reference could still dangle.

### DIFF
--- a/platform/android/MapLibreAndroid/src/cpp/android_renderer_frontend.cpp
+++ b/platform/android/MapLibreAndroid/src/cpp/android_renderer_frontend.cpp
@@ -91,13 +91,24 @@ private:
     ActorRef<RendererObserver> delegate;
 };
 
-AndroidRendererFrontend::AndroidRendererFrontend(MapRenderer& mapRenderer_)
+AndroidRendererFrontend::AndroidRendererFrontend(Private, MapRenderer& mapRenderer_)
     : mapRenderer(mapRenderer_),
-      mapRunLoop(util::RunLoop::Get()),
-      updateAsyncTask(std::make_unique<util::AsyncTask>([this]() {
-          mapRenderer.update(std::move(updateParams));
-          mapRenderer.requestRender();
-      })) {}
+      mapRunLoop(util::RunLoop::Get()) {}
+
+std::shared_ptr<AndroidRendererFrontend> AndroidRendererFrontend::create(MapRenderer& mapRenderer) {
+    auto ptr = std::make_shared<AndroidRendererFrontend>(Private(), mapRenderer);
+    ptr->init();
+    return ptr;
+}
+
+void AndroidRendererFrontend::init() {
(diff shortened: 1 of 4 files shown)


platform/android/MapLibreAndroid/src/cpp/android_renderer_frontend.cpp
```diff
@@ -91,13 +91,24 @@ class ForwardingRendererObserver : public RendererObserver {
...
+
+std::shared_ptr<AndroidRendererFrontend> AndroidRendererFrontend::create(MapRenderer& mapRenderer) {
+    auto ptr = std::make_shared<AndroidRendererFrontend>(Private(), mapRenderer);
+    ptr->init();
+    return ptr;
+}
+
+void AndroidRendererFrontend::init() {
+    updateAsyncTask = std::make_unique<util::AsyncTask>([weakSelf = weak_from_this()]() {
+        if (auto self = weakSelf.lock()) {
```
louwers, line 106: How bad is this lock in this hot code path?

mwilsnd:
Is `mapRenderer` under the control of the JVM? If the Java instance was deleted for just one of these objects, what would happen? I would think the [finalizer](https://github.com/maplibre/maplibre-native/blob/main/platform/android/MapLibreAndroid/src/cpp/map_renderer.cpp#L321) might still ruin our day. To be extra safe, I think you could establish a weak reference between the two (`NativeMapView` and `MapRenderer`) [here](https://github.com/maplibre/maplibre-native/blob/main/platform/android/MapLibreAndroid/src/cpp/native_map_view.cpp#L69).

louwers:
Do you mean store it as a `jni::WeakReference<MapRenderer>` instead of a `MapRenderer &`?

And also pass a `jni::WeakReference<MapRenderer>` to `AndroidRendererFrontend`?

mwilsnd:
I'm thinking so but I'm not 100% certain if it will really work. Currently the lifetime of `MapRenderer&` is determined by the Java peer, once the GC runs the [finalizer](https://github.com/maplibre/maplibre-native/blob/b11134a301f720097183adc26c99fd6f7a85cf4f/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/renderer/MapRenderer.java#L171) it'll invoke the C++ peer's destructor. Same goes for `NativeMapView` [here](https://github.com/maplibre/maplibre-native/blob/b11134a301f720097183adc26c99fd6f7a85cf4f/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/NativeMapView.java#L1385), though `MapRenderer` is actually a member of this class. I can't say for certain that the JVM GC is deterministic enough to reason about the ordering of destructors, or that they're called in a timely fashion.

cgalvan (approved):

louwers:
Will address the other (potential) issue in a follow-up.

Issue #3524, opened 2025-06-04
The symbol in lower level will blink when zooming or moving the camera.


qqz003:
**Describe the bug**

Sometimes the symbol in lower level will blink when zooming or moving.

**To Reproduce**

1. Load a style that defines a layer containing a symbol without data in a specific zoom range. (e.g., Layer A's zoom range is 1-6, and it contains a symbol B that only has data in the range of 1-3. )
2. Navigate to the location of the symbol B, zoom level 6, now the symbol B is invisible.
3. Zoom out or moving.
4. The symbol B in lower level will blink.

**Expected behavior**

The symbol should only be visible in the level it

**Screenshots**

Please check the attached video.

**Platform information (please complete the following information):**

 - Operating System: iOS, Android.

https://github.com/user-attachments/assets/e4cbde40-85ab-40cf-af39-d1e365c7c58d

qqz003:
Root cause:

In ML/N, when rendering the tile, if the current tile is still not renderable, the code trying find its ancestor to load and render, so duration the transition, a tile in lower zoom level will be marked to render. (please check the file: update_renderable.cpp)

Potential solutions:

Solution A:
Update the logic for rendering the tile in ML/N to align with the ML/JS by updating the code in the line 112, update_renderable.cpp, to
```
if (!tile && parentIsLoaded)
```
In JS (source_cache.js), the condition is written as `if (!tile && parentWasRequested)`. Here,` parentWasRequested` is similar to `parentIsLoaded` in Native. As a result, the ancestor tile will not be loaded unless the current tile fails to load.

Shortage:
1. Will bring the render effect regression when zooming out. (Please check the attachment).
    ([Before update](https://github.com/user-attachments/assets/4bf93905-e3d8-4670-8d7c-84c647b5a13b), [After update](https://github.com/user-attachments/assets/f88a39d5-fd1a-4883-bf45-6f27c4a5ab19))
2. Unit tests for the method `updateRenderables` will fail.

Solution B:
Add check when placing the label in the file (line 279, placement.cpp):
```
if (ctx.getOverscaledID().overscaledZ != int(placementZoom)) {
    return kUnplaced;
}
```
Only symbols in the tile that isn't in the current zoom level will be filtered out, ensuring that all other elements remain rendered to avoid empty blocks when zooming out. Additionally, this change does not break any test cases.

Shortage:
Looks a little tricky.

[...]

PR #3534, opened 2025-06-06
Fix the symbol blink issue by only placing the symbol in current level

qqz003:
Fix the issue #3524 by choosing Solution B to avoid causing significant disruption to the code.

### DIFF
--- a/src/mbgl/text/placement.cpp
+++ b/src/mbgl/text/placement.cpp
@@ -277,6 +277,7 @@ JointPlacement Placement::placeSymbol(const SymbolInstance& symbolInstance, cons
         return kUnplaced;
     }
 
+    // Place the symbol only at the current level to prevent it from blinking during the transition.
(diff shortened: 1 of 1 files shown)


louwers:
Trying it out!

qqz003:
@louwers, we have tested the code locally and found no regressions. Do you have any other questions? If not, could you please approve and merge this PR?

louwers (approved):

louwers:
I did some testing and also did not see any regressions. 

We might want to port this change to MapLibre GL JS as well?

qqz003:
@louwers Thank you very much! 
We don't need to apply this change to the JS code because it does not have this issue. For more details, please refer to issue #3524. Differences in the implementation mean that the JS code will not prefetch ancient tiles unless the current tile fails to load, so this issue does not occur in JS.

louwers:

PR #3536, opened 2025-06-06
Fix segfault this-capture GeoJsonSource which may be deleted

### LINKED ISSUE #3493: Programatically setting `withClusterProperty` on a GeoJSON Source causes a segfault
### MapLibre Android Version

11.8.8

### Android SDK Version

Android 15

### Device

Galaxy Z Fold 6

### What happened?

When trying to configure cluster properties of a geojson source programmatically, I get a segfault with no other relevant log output.

### Steps to reproduce

See screenshot from debugger below

I attempted to call `withClusterProperty` with the equivalent of `["+", ["get", "current_range_meters"]]`. Specifically, I pass `operatorExpr = LiteralExpression("+")` and `mapExpr = MLNExpression.Converter.convert(...)`

Calling it with the equivalent of `["+", 0]` works fine.

Calling it with the equivalent of `["+", ["+", 1, 1]]` segfaults again.

You can reproduce easily with the maplibre-compose demo app by uncommenting this example: https://github.com/maplibre/maplibre-compose/blob/1932e6ed68a1abf65bde61565136ee4da9d306f3/demo-app/src/commonMain/kotlin/dev/sargunv/maplibrecompose/demoapp/demos/ClusteredPointsDemo.kt#L91-L100

### Renderer

OpenGL (choose this if you are unsure)

### Relevant log output

```shell
2025-05-25 22:11:08.713 10266-10266 libc                    dev.sargunv.maplibrecompose.demoapp  A  Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x8 in tid 10266 (compose.demoapp), pid 10266 (compose.demoapp)
```

### Additional context

Image

louwers:
Stores data the lambda in `GeoJSONDataCallback` needs in `shared_ptr`s so that we can capture a weak reference instead of a `this` pointer. Regression test that segfaults before this fix added.

We cannot use `std::enable_shared_from_this` here because jni.hpp expects all the native peer classes to be stored in a unique_ptr.

Closes #3493, but there are other places where we need to apply a similar fix.


TimSylvester:
Makes sense, but if I understand correctly this relies on the `shared_ptr` never actually being shared, and being destroyed at the same time as the `unique_ptr` so, to make it harder to screw up, I would make the `shared_ptr` private, and only expose weak pointers to subclasses via a `getWeakSource()` or similar.  And maybe make both private, and do `getSource<style::GeoJSONSource>()` instead of `source->get().as<style::GeoJSONSource>()`.

[sub-PR](https://github.com/louwers/maplibre-native/pull/17)

louwers:
Released in MapLibre Android 11.10.2 https://github.com/maplibre/maplibre-native/releases/tag/android-v11.10.2

PR #3545, opened 2025-06-11
Add rendering info reports to `ActionJournal`

adrian-cojocaru:
This PR adds a rendering stats event to the action journal logging. 
The event contains min/max/average encoding and rendering frame times over a configurable time interval.

### DIFF
--- a/include/mbgl/util/action_journal_options.hpp
+++ b/include/mbgl/util/action_journal_options.hpp
@@ -87,6 +87,23 @@ public:
      */
     uint32_t logFileCount() const { return logFileCount_; }
 
+    /**
+     * @brief Set the number of seconds to wait between rendering stats reports.
+     *
+     * @param interval time interval in seconds.
+     * @return ActionJournalOptions for chaining options together.
+     */
+    ActionJournalOptions& withRenderingReportInterval(const uint32_t interval) {
+        renderingStatsReportInterval_ = interval;
+        return *this;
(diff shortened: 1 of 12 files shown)


include/mbgl/util/action_journal_options.hpp
```diff
@@ -87,6 +87,23 @@ class ActionJournalOptions {
      */
     uint32_t logFileCount() const { return logFileCount_; }
 
+    /**
+     * @brief Set the number of seconds to wait between rendering stats reports.
+     *
+     * @param interval time interval in seconds.
+     * @return ActionJournalOptions for chaining options together.
+     */
+    ActionJournalOptions& withRenderingStatsReportInterval(const uint32_t interval) {
```
louwers, line 96: Could you add a note to the iOS and Android documentation pages that a rendering stats report is also output to the action journal by default (but that the interval is configurable)?

adrian-cojocaru: I checked the existing doc and we don't list any of the events logged. They're inferred from `MapObserver` interface. It might be better to provide the full event list and attributes (in another PR)?

louwers: Sure!

louwers:
Released for iOS https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.18.0

PR #3547, opened 2025-06-11
Add polylines in geographic coordinates

adrian-cojocaru:
Enable adding polylines with geographic coordinates

### DIFF
--- a/src/mbgl/style/layers/custom_drawable_layer.cpp
+++ b/src/mbgl/style/layers/custom_drawable_layer.cpp
@@ -519,8 +519,23 @@ util::SimpleIdentity CustomDrawableLayerHost::Interface::addPolyline(const LineS
 
     switch (shaderType) {
         case LineShaderType::Classic: {
-            // TODO: build classic polyline with Geo coordinates
-            return util::SimpleIdentity::Empty;
+            // build classic polyline with Geo coordinates
+            if (!updateBuilder(BuilderType::LineClassic, "custom-lines", lineShaderDefault())) {
+                return util::SimpleIdentity::Empty;
+            }
+
+            // geographic coordinates require tile {0, 0, 0}
+            setTileID({0, 0, 0});
+
+            constexpr int32_t zoom = 0;
+            GeometryCoordinates tileCoordinates;
+            for (const auto& coord : coordinates) {
+                const auto point = Projection::project(LatLng(coord.y, coord.x), zoom);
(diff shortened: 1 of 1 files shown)


src/mbgl/style/layers/custom_drawable_layer.cpp
```diff
@@ -519,8 +519,23 @@ util::SimpleIdentity CustomDrawableLayerHost::Interface::addPolyline(const LineS
 
     switch (shaderType) {
         case LineShaderType::Classic: {
-            // TODO: build classic polyline with Geo coordinates
-            return util::SimpleIdentity::Empty;
+            // build classic polyline with Geo coordinates
```
stefankarschti, line 522: This was left out because the transformation matrixes don't have enough precision at global scale with the standard shader, but it's handled in the Wide Vector.

stefankarschti: check out the usage of diff matrixes to enhance precision in wide vector metal shader
https://github.com/maplibre/maplibre-native/blob/1e78e146f1066320f195496ce9f43825f26488b6/include/mbgl/shaders/mtl/widevector.hpp#L59

adrian-cojocaru: This is mostly a backup for platforms/backends that don't have WideVector implementations (yet).

stefankarschti: ok, but you should check the results, as the lack of precision makes it really unusable past a certain zoom level.

stefankarschti: could we document this limitation, in code comments or other docs?

adrian-cojocaru: I'm thinking of moving this bit in the example code (with comments) and removing the line type selection on `addPolyline` (global) for now.

adrian-cojocaru: I've been looking over the different matrices, but only `mvpMatrix` is used in the shader and this should be the tile matrix (same as the line shader). Is there something else that I'm missing or the extra precision is from the different input types (float3 vs short2)?

stefankarschti: indeed, looks like the diff is not used in the final version.
if the geometry is stable at `zoom >= 20` then go ahead with the changes

stefankarschti (approved):

PR #3550, opened 2025-06-12
Fix crash: do not use ownedSource

louwers:
In #3536 @TimSylvester mentioned:

> Makes sense, but if I understand correctly this relies on the `shared_ptr` never actually being shared, and being destroyed at the same time as the `unique_ptr` so, to make it harder to screw up, I would make the `shared_ptr` private, and only expose weak pointers to subclasses via a `getWeakSource()` or similar.  And maybe make both private, and do `getSource<style::GeoJSONSource>()` instead of `source->get().as<style::GeoJSONSource>()`.

However, ownedSource is invalid after the source is added to the map, causing a crash...


louwers:
We're going to come back to this later, because this fix still has problems. https://github.com/maplibre/maplibre-native/pull/3554

PR #3561, opened 2025-06-16
Improve the logic to let source peers be consistent with C++ peers

grab-liyanjin:


louwers:

PR #3566, opened 2025-06-18
add functionality to metal render pass to cache winding order and cull mode

yousifd:
Instead of making a call to the encoder's `setCullMode` and `setFrontFacingWinding` for every drawable which is expensive we cache the previously set value in the render pass and reset it when necessary.

### DIFF
--- a/include/mbgl/mtl/render_pass.hpp
+++ b/include/mbgl/mtl/render_pass.hpp
@@ -85,8 +85,8 @@ private:
     std::array<MTLTexturePtr, maxBinds> fragmentTextureBindings;
     std::array<MTLSamplerStatePtr, maxBinds> fragmentSamplerStates;
 
-    std::optional<MTL::CullMode> currentCullMode = std::nullopt;
-    std::optional<MTL::Winding> currentWinding = std::nullopt;
+    MTL::CullMode currentCullMode = MTL::CullModeNone;
+    MTL::Winding currentWinding = MTL::WindingClockwise;
 };
 
 } // namespace mtl
--- a/src/mbgl/mtl/render_pass.cpp
+++ b/src/mbgl/mtl/render_pass.cpp
@@ -75,8 +75,8 @@ void RenderPass::resetState() {
         fragmentSamplerStates[i].reset();
     }
 
-    currentCullMode.reset();
-    currentWinding.reset();
+    currentCullMode = MTL::CullModeNone;
+    currentWinding = MTL::WindingClockwise;
 }
 
 namespace {


src/mbgl/mtl/render_pass.cpp
```diff
@@ -197,5 +193,19 @@ void RenderPass::setFragmentSamplerState(const MTLSamplerStatePtr& state, int32_
     }
 }
 
+void RenderPass::setCullMode(const MTL::CullMode mode) {
+    if (mode != currentCullMode) {
```
louwers, line 197: Correct me if I am wrong but this will not cache MTL::CullModeNone right?

adrian-cojocaru: If you're referring to the first call of `setCullMode(MTL::CullModeNone)`, the `std::optional<MTL::CullMode>(std::nullopt) != MTL::CullModeNone` compare will be true and will cache correctly.

include/mbgl/mtl/render_pass.hpp
```diff
@@ -80,6 +84,9 @@ class RenderPass final : public gfx::RenderPass {
 
     std::array<MTLTexturePtr, maxBinds> fragmentTextureBindings;
     std::array<MTLSamplerStatePtr, maxBinds> fragmentSamplerStates;
+
+    std::optional<MTL::CullMode> currentCullMode = std::nullopt;
+    std::optional<MTL::Winding> currentWinding = std::nullopt;
```
adrian-cojocaru: As a very minor nitpick: you can remove the optional on both these values and mirror metal's render pass defaults (`MTL::CullModeNone`, `MTL::WindingClockwise` IIRC)

yousifd: I'm worried that there might be situations where the command encoder state isn't reset correctly and we end up inheriting what was set before rather than setting the actual default value.

src/mbgl/mtl/render_pass.cpp
yousifd: It should cache it just fine, why would `MTL::CullModeNone` be any different?

include/mbgl/mtl/render_pass.hpp
adrian-cojocaru: The metal encoder is created in the `RenderPass` constructor so I don't think it can ever go out of sync.

yousifd: Done https://github.com/maplibre/maplibre-native/pull/3566/commits/b7484b425b8ad7bfb082c682160e22e7b94f1630

src/mbgl/mtl/render_pass.cpp
louwers: Alright!

PR #3568, opened 2025-06-18
add the ability to have ios camera animation durations dynamic during navigation

yousifd:
This change makes it so that camera animations during navigations are not always hard coded to a duration of 1 second. We now replicate what we do for the user location indicator by measuring the difference between the last location's timestamp and the current location update and animate the camera based on that.

This functionality is exposed through a feature flag to maintain the existing functionality for those that need it.


platform/ios/src/MLNMapView.h
```diff
@@ -550,6 +550,15 @@ MLN_EXPORT
...
 @property (nonatomic, assign) BOOL showsUserLocation;
 
+/**
+ A boolean value indicating whether camera animation duration is set based
+ on the time difference between the last location update and the current one
+ or the default animation duration of 1 second.
+
+ The default value of this property is `NO`
+ */
+@property (nonatomic, assign) BOOL dynamicNavigationCameraAnimationDuration;
```
louwers, line 560: Does this need to be initialized?

louwers (commented):
Is there any way to test this?

We fully rely on manual testing with this code I think?

platform/ios/src/MLNMapView.h
yousifd: Yeah, I thought the default initialization value is false. Will add the necessary default value.

yousifd:
> Is there any way to test this?
> 
> We fully rely on manual testing with this code I think?

I'm not sure if there is a mechanism to test any of the ios code but the only real way to ensure this is working is to check how long the camera animations run for when called from the ios side during navigation when the updates happen at a rate faster than 1 second.

platform/ios/src/MLNMapView.h
yousifd: https://github.com/maplibre/maplibre-native/commit/563425abf39f133cd7fff32039273c73f515289e

louwers:
Released https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.18.0

Issue #2783, opened 2024-08-29
RasterDEM custom sources with support for `redFactor`, `blueFactor`, `greenFactor`, `baseShift`


louwers:
Style Spec Definition

- https://maplibre.org/maplibre-style-spec/sources/#encoding

- https://maplibre.org/maplibre-style-spec/sources/#redfactor

MapLibre GL JS PR: https://github.com/maplibre/maplibre-gl-js/pull/3087

Style Spec PR: https://github.com/maplibre/maplibre-style-spec/pull/337

PR #1171, opened 2023-05-24
fix for issue #1170

stefankarschti:
fix for #1170

PR #3570, opened 2025-06-18
Fix for raster dem encoding override in style.json

### LINKED ISSUE #3564: Specifying `encoding` for a RasterDEM source in style.json ignored when providing `url`
### MapLibre iOS Version

6.15.0

### iOS Version

18.5

### Device

iPhone 

### What happened?

I am loading a terrarium-encoded raster dem PMTiles in my style.json. I specify `"encoding": "terrarium"` in the style.json.

## Expected result

Hillshade layer from the raster dem source renders correctly.

## Actual result

Hillshade layer renders as if it's being interpreted as Mapbox-encoded RGB (i.e. exaggerated heights).

## Observations

I spent a bit digging through the code, and I can confirm that the `encoding` field in the style.json sources is ignored when setting a `url` for the source. It seems the pattern for sources is to expect either a TileJSON from the `url` or all of the needed TileJSON fields provided in style.json's source fields. Other than a few special cases (e.g. `tileSize` for raster/raster dem, min zoom/max zoom for vector sources), the style.json fields are not interpreted as overrides to the TileJSON that is retrieved from the `url`. In the case of PMTiles, the TileJSON is constructed dynamically from the metadata in the PMTiles file, but only the common fields (e.g. `bounds`, `center`) are decoded.

MapLibre JS works differently. Here the style.json values do seem to override anything in the TileJSON.

## Possible solutions

1. If the intention is to allow overriding values via style.json (as in the JS version), then a mechanism needs to be provided for the individual source types to override information retrieved from the `url` with values present in the style.json. There are two slightly different patterns in use for this in type of override in `VectorSource` vs. `GeoJSONSource`. A `VectorSource`-like implementation would be pretty simple to implement but less scalable the more options are available. The `GeoJSONSource`-like solution of encapsulating the options in `GeoJSONOption` would be more scalable, but potentially be more complicated of a change.

2. The construction of a dynamic TileJSON from PMTiles metadata should include any extra fields present, such as `encoding`. This would allow for the current method of determining encoding to work and provide an option if the intention is not to allow style.json values to override what is retrieved from `url`.

### Steps to reproduce

* Use terrarium-encoded raster dem PMTiles source. (MBTiles probably produces same result)
* Set encoding to `terrarium` in style.json.
* Create Hillshade layer from raster dem source with default options.
* Observe Hillshade does not render properly.

### Relevant log output

```shell

```

### Additional context

_No response_

scottrhoyt:
Fixes #3564 

**Looking for feedback on this approach. I haven't touched C++ in 2 decades, so help and feedback is appreciated!**

## Motivation

The MapLibre Style Spec allows source options to be provided in the style.json that may supplement or override values found in the TileJSON retrieved from `url`. An example of this is the `encoding` parameter for a raster DEM source. The behavior for what should happen when an option can exist both in the style.json and the TileJSON is not well defined by documentation, but from the JS implementation and other examples in native, it appears the intention if for these style.json values to supplement and override any options in the TileJSON. 

Currently this behavior isn't encapsulated well in the code with varying ad hoc implementations inside `RasterSource` (inherited by `RasterDEMSource`) and `VectorSource`. This PR aims to better abstract the behavior of loading a TileJSON for a source and then allowing source-specific overrides.

## Overview

This change essentially does these things:
1. Exctacts a common super class for all TileJSON-loading sources: `TileSource`. `RasterSource` and `VectorSource` now inherit from this. `RasterDEMSource` continues to inherit from `RasterSource`. Follow the same pattern for the `Impl`s.
2. Provide an override point for subclasses to supplement/override `Tileset` values: `TileSource::setTilesetOverrides`. Use this override to implement the min/maxzoom overrides currently used by `VectorSource`.
3. Define a `RasterDEMOptions` struct to decode and hold style.json raster dem options (only `encoding` at the moment).
4. Use `setTilesetOverride` and `RasterDEMOptions` to implement `encoding` override from style.json

## `RasterDEMOptions`

While this PR only includes `encoding` override support for `RasterDEMSource`, I chose to follow a pattern similar to `GeoJSONOptions` for two reasons:
1. Provide an encapsulation for supporting more options for `RasterDEMSource`, e.g. supporting custom encoding types.
2. Provide a potential pattern to follow to allow all types of TileJSON-based sources to provide overrides in style.json. A base options class could encapsulate the standard TileJSON decoding (perhaps share the current implementation) and the individual source types can extend this for source-specific or non-standard options.

## Other Notes
* There were two design decisions that I made that I'm not fully aware of the implications of and would appreciate review on. Please see comment in source.
* The way that the `Impl`s are used to abstract data/implementation was hard for me to make sense of. I tried to follow the pattern I saw for `tileSize`, `maxzoom`, `minzoom`.

## Tests

I wasn't sure how to run the standard test suites, pointers there would be appreciated. I expect there are at least a couple call sites to fix.

I did run the sample iOS apps and checked out the rendering of both raster and vector tile maps and everything looked good there. The only override I checked was the terrarium `encoding` for raster dem. This does now render correctly:

Simulator Screenshot - iPhone 16 Pro - 2025-06-18 at 15 25 37

## Future Work

If this direction is chosen, I see it unlocking the following work:
1. Continue *Options pattern to encapsulate all TileJSON overrides--including the min/maxzoom currently in `VectorSource`. This would bring native to parity with JS behavior.
2. Modularize and share TileJSON parsing
3. Potentially consolidate shared code for TileSource renderers more.
4. Use `RasterDEMOptions` as a foundation for custom DEM encoding (#2783). The remaining work would be turning `Tileset::encoding` into a variant with a custom encoding struct, and then using this struct in the DEM-decoding phase--both should be pretty easy.


src/mbgl/style/sources/tile_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+            observer->onSourceError(*this, std::make_exception_ptr(std::runtime_error("unexpectedly empty TileJSON")));
+        } else {
+            conversion::Error error;
+            std::optional<Tileset> tileset = conversion::convertJSON<Tileset>(*res.data, error);
+            if (!tileset) {
+                observer->onSourceError(*this, std::make_exception_ptr(util::StyleParseException(error.message)));
+                return;
+            }
+            const auto& tileServerOptions = fileSource.getResourceOptions().tileServerOptions();
+            if (tileServerOptions.uriSchemeAlias() == "mapbox") {
```
scottrhoyt, line 75: Previously, this check for mapbox as the uri scheme was only present in `RasterSource`. It was added in #1171 . This change would effectively add this check for `VectorSource` as well. It appears this only has to do with the built-in tile server schemes like maptiler and mapbox. I verified both raster and vector maps from maptiler uris worked as expected. Let me know if there are other implications to this.

src/mbgl/style/sources/vector_source.cpp
```diff
@@ -18,78 +18,17 @@ VectorSource::VectorSource(std::string id,
                            variant<std::string, Tileset> urlOrTileset_,
                            std::optional<float> maxZoom_,
                            std::optional<float> minZoom_)
-    : Source(makeMutable<Impl>(std::move(id))),
-      urlOrTileset(std::move(urlOrTileset_)),
+    : TileSource(id, urlOrTileset_, util::tileSize_I, SourceType::Vector),
```
scottrhoyt, line 21: I chose to have `tileSize` be a part of the base `TileSource` instead of making it specific to `RasterSource` since previously `VectorSource` used this value for tileSource [here](https://github.com/maplibre/maplibre-native/blob/1899cde3bd19dac0b88536c55d7a42b7507460ff/src/mbgl/style/sources/vector_source.cpp#L80). This also means the `getTileSize` function is available on `VectorSource`, which doesn't seem problematic. An argument can be made for specializing `tileSize` to `RasterSource` and its sub-classes, but that would entail a bigger change and less shared code.

scottrhoyt (commented):
As you are reviewing, I'm specifically looking for feedback on two design decisions I made in the comments here.

louwers (commented):
Check out the Developer Documentation to see instructions on how to run the tests depending on what platform you are on: https://maplibre.org/maplibre-native/docs/book/

scottrhoyt:
@louwers this should be good to go now. I believe the one test failure is unrelated to the changes here.

louwers (commented):
Looks good.

Could you add some test coverage?

scottrhoyt:
Sure thing. I added tests for `RasterDEMOptions` which handles the `encoding` parsing. I couldn't find any tests for general source parsing so I didn't add any for `RasterDEMSource`, but the new parsing logic should be largely covered by the `RasterDEMOptions` tests. If there are tests for source parsing and I missed them, I'd appreciate a pointer. Thanks!

src/mbgl/style/conversion/raster_dem_options.cpp
```diff
@@ -0,0 +1,34 @@
...
+    auto encodingValue = objectMember(value, "encoding");
+    if (encodingValue) {
+        std::optional<std::string> encoding = toString(*encodingValue);
+        if (encoding && *encoding == "terrarium") {
+            options.encoding = {Tileset::DEMEncoding::Terrarium};
+        } else if (encoding && *encoding == "mapbox") {
+            options.encoding = {Tileset::DEMEncoding::Mapbox};
+        } else {
+            error.message =
+                "invalid raster-dem encoding type - valid types are 'mapbox' "
```
louwers, line 23: This is duplicated now, can that be avoided somehow?

test/style/conversion/raster_dem_options.test.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+TEST(RasterDEMOptions, ErrorHandling) {
+    Error error;
+    std::optional<RasterDEMOptions> converted = convertJSON<RasterDEMOptions>(
+        R"JSON({
+        "encoding": "this isn't a valid encoding"
+    })JSON",
+        error);
+    ASSERT_FALSE((bool)converted);
```
louwers: Shouldn't need a cast.

Also you can assert the error message here I think.

src/mbgl/style/sources/tile_source.cpp
```diff
@@ -0,0 +1,108 @@
...
+        }
+    });
+}
+
+void TileSource::setTilesetOverrides(Tileset&) {
+    // Default empty implementation, subclasses can override
+}
+
+bool TileSource::supportsLayerType(const mbgl::style::LayerTypeInfo* info) const {
+    return mbgl::underlying_type(Tile::Kind::Raster) == mbgl::underlying_type(info->tileKind);
```
louwers: I think we need to force derived classes to implement this. It does not make sense to provide a default implementation. 

You can remove it from the `TileSource` header and implementation, because it is already defined in `Source`. 

This will make `TileSet` an abstract class (which is OK, because it is not being instantiated directly anywhere).

src/mbgl/style/conversion/raster_dem_options.cpp
scottrhoyt: Yes. That would be solved by Future Work 2 above--modularizing are reusing TileJSON parsing. I consider that to be outside the scope of this PR though as the change will be at least as significant as this one. I suggest we spin off another issue to track that work.

test/style/conversion/raster_dem_options.test.cpp
scottrhoyt: I was following the pattern as established [here](https://github.com/maplibre/maplibre-native/blob/main/test/style/conversion/geojson_options.test.cpp), but happy to do something different if you think it warrants it.

src/mbgl/style/sources/tile_source.cpp
scottrhoyt: Sounds good!

src/mbgl/style/conversion/raster_dem_options.cpp
louwers: Could you add a FIXME comment or something?

test/style/conversion/raster_dem_options.test.cpp
louwers: Fair enough, see if it works without a cast (I think it should?) otherwise you can maybe use static_cast<bool>(...)

scottrhoyt:
@louwers I made the requested changes. I avoided some of the code duplication by using `RasterDEMOptions` in the main `Tileset` decoding.

scottrhoyt:
@louwers anything else I can do to help get this landed?

louwers:
@scottrhoyt No I think this is good to go. Sorry, forgot to merge it!

scottrhoyt:
It looks like some flaky tests to me?

louwers:
@scottrhoyt Yes, don't worry about it!

PR #3571, opened 2025-06-18
[WIP] Add encoding parameter to RasterDEMSource (simple).

### LINKED ISSUE #3564: Specifying `encoding` for a RasterDEM source in style.json ignored when providing `url`

scottrhoyt:
Fixes #3564 

This is a minimal changeset to be compared to #3570.

Instead of refactoring out a `TileSource` and adding `RasterDEMOptions`, this just adds an `encoding` param to `RasterDEMSource` and copies over the implementation of `loadDescription` from `RasterSource` to insert the encoding param.

### DIFF
--- a/src/mbgl/style/sources/raster_dem_source.cpp
+++ b/src/mbgl/style/sources/raster_dem_source.cpp
@@ -1,3 +1,4 @@
+#include <mbgl/storage/file_source.hpp>
 #include <mbgl/style/conversion/json.hpp>
 #include <mbgl/style/conversion/tileset.hpp>
 #include <mbgl/style/layer.hpp>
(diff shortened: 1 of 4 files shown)


src/mbgl/style/sources/raster_dem_source.cpp
```diff
@@ -1,18 +1,77 @@
...
-RasterDEMSource::RasterDEMSource(std::string id, variant<std::string, Tileset> urlOrTileset_, uint16_t tileSize)
-    : RasterSource(std::move(id), std::move(urlOrTileset_), tileSize, SourceType::RasterDEM) {}
+RasterDEMSource::RasterDEMSource(std::string id,
+                                 variant<std::string, Tileset> urlOrTileset_,
+                                 uint16_t tileSize,
+                                 std::optional<Tileset::DEMEncoding> encoding)
+    : RasterSource(std::move(id), std::move(urlOrTileset_), tileSize, SourceType::RasterDEM),
+      encoding(encoding) {}
+
+void RasterDEMSource::loadDescription(FileSource& fileSource) {
```
louwers, line 24: I don't see any difference with `RasterSource::loadDescription`. Why did you override it?

louwers (changes requested):
Needs a test and duplication removed.

Please check the build error for Android on CI.

src/mbgl/style/sources/raster_dem_source.cpp
```diff
@@ -1,18 +1,77 @@
...
+            if (!tileset) {
+                observer->onSourceError(*this, std::make_exception_ptr(util::StyleParseException(error.message)));
+                return;
+            }
+            const auto& tileServerOptions = fileSource.getResourceOptions().tileServerOptions();
+            if (tileServerOptions.uriSchemeAlias() == "mapbox") {
+                util::mapbox::canonicalizeTileset(tileServerOptions, *tileset, url, getType(), getTileSize());
+            }
+
+            if (encoding) {
```
scottrhoyt, line 58: @louwers This is the difference. We override the encoding if it's different. This is the same approach employed by [VectorSource::loadDescription](https://github.com/maplibre/maplibre-native/blob/ec531339266acf2b8439ab377e6bb77e8ccc849c/src/mbgl/style/sources/vector_source.cpp#L44) which is virtually identical other than overriding the min/max zoom levels.

This duplication of logic across the different tile-based sources is the problem that I tried to address in the other proposal through an extraction of a common super class to hold the shared `loadDescription`.

louwers: Why not keep most of the logic in RasterSource, but extract only this relevant bit in here?

scottrhoyt: I did look at that as well. Turns out the logic is also the pretty much the same as `VectorSource` (or any TileJSON-based source) which is how I got to #3570.

scottrhoyt:
Closing this out to focus on #3570

PR #3572, opened 2025-06-19
[WIP] drawables backend for qt platform

birkskyum:


birkskyum:
The opengl backend is left as it is, except for the rename from `RendererBackend` to `OpenGLRendererBackend`

louwers (commented):
Does it run yet? Is it possible to try it out?

birkskyum:
@louwers  it does run, and it can be tried out. The metal instructions with the quick example found in the top of [this PR](https://github.com/maplibre/maplibre-native-qt/pull/216) is using this MapLibre Native brach associated to this PR.

OpenGL also works for quick/widgets

Vulkan will be improved further soon. That can also happen in a follow-up pr though.

birkskyum:
I moved the platform/qt into the maplibre-native-qt instead (part of https://github.com/maplibre/maplibre-native-qt/pull/216). platform/qt can be safely removed when that update to maplibre-native-qt lands.

PR #3573, opened 2025-06-19
set default move gesture threshold to a small non-zero value

### LINKED ISSUE #2792: `MapLibreMap.OnMoveListener::onMoveBegin` is also called on tap
**Describe the bug**
`onMoveBegin(detector: MoveGestureDetector)` in a `MapLibreMap.OnMoveListener` is also called when the map is not actually moved but when the user just taps the map.

**To Reproduce**
Register an `MapLibreMap.OnMoveListener` and log something on `onMoveBegin`. Start the app and tap the map. Something will be logged.

**Expected behavior**
`onMoveBegin(detector: MoveGestureDetector)` is only called when the map is moved

 - OS: Android, any API level
 - Version 11.2.0

**Additional context**
In StreetComplete, there is a follow-me mode. I.e. the map automatically centers on the current user's location. This is turned off when the user moves the map manually but it should **not** turn off when the user just taps the map.

Cause of this issue: https://github.com/streetcomplete/StreetComplete/issues/5824

westnordost:
fixes #2792


louwers (approved):

louwers:
Thanks!

louwers (commented):
@westnordost Tests are failing, can you have a look?

louwers:

Issue #3506, opened 2025-05-28
Style.java crash that could be a warning


sjg-wdw:
### MapLibre Android Version

11.8.*

### Android SDK Version

*

### Device

*

### What happened?

In Style.java we're calling validateState() for things like addImage.  Often validateState() is triggering for an out of sequence operation for something like, well adding an image.  

Can we turn the exception validateState() is generating into a warning?

### Steps to reproduce

Almost impossible to reproduce.

### Relevant log output

```shell

```

sjg-wdw:
What we noticed with a similar bug was that it was MapLibre doing the calling.  In that case I think it was the user puck.

In any case, I think it's fine to just not do the thing rather than throw an exception.

louwers:
@sjg-wdw Right, that's different. Should I investigate where we do this? Or do you have a stack trace.

sjg-wdw:
@TimSylvester had a good stack trace on this one I think.

TimSylvester:
```
Exception in onDidFinishLoadingStyle
java.lang.IllegalStateException: Calling addImage when a newer style is loading/has loaded.
	at org.maplibre.android.maps.Style.validateState(Style.java:776)
	at org.maplibre.android.maps.Style.addImageAsync(Style.java:469)
	at org.maplibre.android.maps.Style.addImageAsync(Style.java:437)
	... application code ...
	at org.maplibre.android.maps.MapLibreMap.notifyStyleLoaded(MapLibreMap.java:1010)
	at org.maplibre.android.maps.MapLibreMap.onFinishLoadingStyle(MapLibreMap.java:229)
	at org.maplibre.android.maps.MapView$MapCallback.onDidFinishLoadingStyle(MapView.java:1702)
	at org.maplibre.android.maps.MapChangeReceiver.onDidFinishLoadingStyle(MapChangeReceiver.java:219)
	at org.maplibre.android.maps.NativeMapView.onDidFinishLoadingStyle(NativeMapView.java:1148)
	at android.os.MessageQueue.nativePollOnce(Native Method)
	at android.os.MessageQueue.next(MessageQueue.java:335)
	at android.os.Looper.loop(Looper.java:183)
	at android.app.ActivityThread.main(ActivityThread.java:7664)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
[...]

PR #3574, opened 2025-06-20
Prevent `Style.validateState()` exception on location state updates

adrian-cojocaru:
Add check to prevent style exception. https://github.com/maplibre/maplibre-native/issues/3506#issuecomment-2991783147
Still not clear how `StaleMessageHandler.handleMessage` is active without a valid style as their lifetime seems controlled by the same location component methods (`onLocationLayerStart`/`onLocationLayerStop`).


louwers:

PR #3576, opened 2025-06-21
Symbol Shader Performance Improvments

AtlasProgramming:
This adds support for vertex culling in the Symbol shaders.  In our style, some of the symbol layers are showing significant overdraw of entirely transparent glyphs, so this will early-out in the vertex shaders and push the vertices out of scene to allow the GPU to cull those vertices and never send them to the fragment shader.  When profiling, GPU encoding frame time goes down from ~23ms to 12ms and one of the SymbolSDFShader iterations goes from ~60% of the overall frame time to ~13%.  Images attached show the performance changes and also the overdraw of the same buffer.

The frame captures were done from the same exact camera setup and after waiting for all loads/etc are completed.  So, in essence, the exact same scene.

Progressively zooming into a single draw call showing the overdraw
Screenshot 2025-06-21 at 10 53 23 AM

Screenshot 2025-06-21 at 10 53 29 AM

Screenshot 2025-06-21 at 10 53 35 AM

Two frame captures showing the % difference between the two SDF shader implementations as well as overall frame time.
Screenshot 2025-06-21 at 10 52 39 AM

Screenshot 2025-06-21 at 10 51 08 AM

The same draw call after the modification and how much fewer vertices are being rendered
Screenshot 2025-06-21 at 10 55 02 AM

Screenshot 2025-06-21 at 10 55 17 AM


include/mbgl/shaders/mtl/symbol.hpp
```diff
@@ -4,6 +4,8 @@
 #include <mbgl/shaders/shader_source.hpp>
 #include <mbgl/shaders/mtl/shader_program.hpp>
 
+#define MTL_SDF_SHADER_VERTEX_CULLING_ENABLED 1
```
AtlasProgramming: I put this in to allow quick A/B testing

louwers (approved):
LGTM! That is a pretty significant performance gain.

include/mbgl/shaders/mtl/symbol.hpp
louwers: Note: before merging we should remove this `#define`.

@alexcristici Could you try this out and could give this PR a review?

adrian-cojocaru:
Are all quad vertices guaranteed to have the same opacity? I know this issue was raised before for a similar scenario, but after searching a bit it seems that's not the case?

```
// SymbolLayout::addSymbol
auto opacityVertex = SymbolBucket::opacityVertex(true, 1.0);
buffer.opacityVertices().emplace_back(opacityVertex);
buffer.opacityVertices().emplace_back(opacityVertex);
buffer.opacityVertices().emplace_back(opacityVertex);
buffer.opacityVertices().emplace_back(opacityVertex);

// Placement::updateBucketOpacities
bucket.text.opacityVertices().extend(textOpacityVerticesSize, opacityVertex);
iconBuffer.opacityVertices().extend(iconOpacityVerticesSize, opacityVertex);
```

TimSylvester:
If I understand correctly, this is detecting that the opacity is zero and short-circuiting the rendering pipeline for the associated geometry?

I've been experimenting with eliminating the drawables entirely if the opacity is zero, which also eliminates the overhead of setting them up and sending all that geometry to the GPU, so that should probably be what we try first.

[This](https://github.com/maplibre/maplibre-native/compare/main...WetDogWeather:maplibre-gl-native:symbol-opacity-filter?expand=1) is the approach I've been pursuing.  It directly interrogates the per-vertex opacity attribute produced by property evaluation, and skips the segment/drawable if they are all zero.

Technically that's a slightly different result, but only if some of the vertices have zero opacity and some do not, and I'm not sure if that can actually happen.  I think that would produce incorrect results with the shader approach, right?  If we assume that the vertices will have the same opacity, the drawable filtering approach could be more efficient by not needing to consider every vertex in the segment.

TimSylvester:
@adrian-cojocaru To be clear, that means they are always the same?

Too bad we have to repeat them, then, but it doesn't seem practical to avoid that currently.

adrian-cojocaru:
In `Placement::updateBucketOpacities` and `SymbolLayout::addSymbol` all vertices are the same. We can always try and see what we can gain by moving some of the attributes as per instance data.

sjg-wdw:
From discussion I think we're all good with this if it passes the render tests on iOS.

include/mbgl/shaders/mtl/symbol.hpp
```diff
@@ -493,6 +523,22 @@ FragmentStage vertex vertexMain(thread const VertexStage vertx [[stage_in]],
...
+    const float2 fade_opacity = unpack_opacity(vertx.fade_opacity);
+    const float fade_change = (fade_opacity[1] > 0.5) ? paintParams.symbol_fade_change : -paintParams.symbol_fade_change;
+    const half fo = half(max(0.0, min(1.0, fade_opacity[0] + fade_change)));
+
+)"
+#if MTL_SDF_SHADER_VERTEX_CULLING_ENABLED
+                                   R"(
+    if (fo == 0.0) {
+        return {
+            .position     = float4(-2.0, -2.0, -2.0, 1.0),
```
alexcristici: Let's name this position `(-2.0, -2.0, -2.0, 1.0)` in a constant for an easier understanding in the future.

alexcristici (approved):

Issue #3496, opened 2025-05-26
Include Windows build in core release


louwers:
Let's include a Windows build in the core release workflow. See Linux here: https://github.com/maplibre/maplibre-native/blob/5ec8430fa38a890633b5145951ef350b7eb18206/.github/workflows/core-release.yml#L57-L90

I think it should be quite doable, but I am not familiar at all with Windows. I guess we would distribute a ddl? 

cc @sargunv @tdcosta100

sargunv:
Yup, I think a .dll, though I'm also not terribly familiar with windows. I do have an x64 and an ARM Windows machines to test with though

PR #3581, opened 2025-06-24
Windows build fixes and Windows inclusion in core releases

tdcosta100:
This PR brings some improvements and fixes in Windows builds (specially the missing symbols errors when `MLN_USE_BUILTIN_ICU=ON`). As requested in #3496, Windows is included in core releases (only the static library, like other platforms).


louwers (approved):
Thanks! LGTM.

I am actually looking into using Bazel for core builds, because so far I have not been able to make a core build with CMake included all the needed symbols. It seems a bit easier with Bazel.

.github/workflows/node-ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
             arch: arm64
           - runs-on: macos-14-large
             arch: x86_64
-          - runs-on: windows-2022
+          - runs-on: windows-latest
```
louwers: OK with me but might start to fail if a new Windows version is out.

tdcosta100: Hmmm maybe. It's because it was `-latest` in `windows-ci.yml`, and in other places I see `ubuntu-latest` too, so I was thinking it was better to make every Windows as latest. But I can do the opposite, make every Windows 2022, what do you think?

louwers: I sometimes use `ubuntu-latest` in places where it is unlikely to make a difference.

I will leave it up to you, but I think it is unlikely that everything will build cleanly with a major Windows update. Also it is probably better to pin especially releases to specific OS versions so there are no unexpected changes to the build environment between releases.

tdcosta100: Agree, you are right. I will have to make some corrections, so `windows-2022` will also be the standard and will be in the next commit.

tdcosta100: Done! `windows-latest` left only when is irrelevant (taking into consideration other OSes also in their latest version).

PR #3584, opened 2025-06-25
Add option to wait for submitted GL commands before surface destruction

adrian-cojocaru:


louwers:
Can you provide some more context? When is this needed?

sjg-wdw:
We're seeing a case on specialized hardware where one render thread is going and another one gets stuck shutting down in the surface or context destruction.  Specifically in the egl calls.

The thinking is flushing out the queue first might help with that.

platform/android/MapLibreAndroid/src/opengl/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreGLSurfaceView.java
```diff
@@ -274,22 +281,56 @@ void destroySurface() {
...
         }
+
         mEglSurface = null;
       }
     }
 
     public void finish() {
+      if (DEFER_CONTEXT_DESTRUCTION) {
+        MapLibreGLSurfaceView view = mGLSurfaceViewWeakRef.get();
+        if (view != null && (mEglContext != null || mEglDisplay != null)) {
```
cahaverl, line 323: There is a race condition here. To fix it, assign the "final" identifiers first, then use those in the "if" condition.

platform/android/MapLibreAndroid/src/opengl/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreGLSurfaceView.java
```diff
@@ -274,22 +281,56 @@ void destroySurface() {
...
+          final GLSurfaceView.EGLContextFactory factory = view.eglContextFactory;
+          final EGLDisplay display = mEglDisplay;
+          final EGLContext context = mEglContext;
+
+            cleanupExecutor.submit(() -> {
+            if (context != null) {
+              factory.destroyContext(mEgl, display, context);
+            }
+
+            if (display != null) {
```
cahaverl, line 333: In the original code this runs whenever `display != null`. Here we might skip this based on the earlier "if" condition (line 313). I don't see why.

More generally, there are two ways of finishing, by duplication. What about combining these, e.g. always defining and using the "final" identifiers, so that we either do "the thing" in this thread or we do "the thing" in a cleanup thread, but where we always do the same thing?

platform/android/MapLibreAndroid/src/opengl/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreGLSurfaceView.java
```diff
@@ -274,22 +281,56 @@ void destroySurface() {
...
 
         mEgl.eglMakeCurrent(mEglDisplay, EGL10.EGL_NO_SURFACE,
-          EGL10.EGL_NO_SURFACE,
-          EGL10.EGL_NO_CONTEXT);
+                EGL10.EGL_NO_SURFACE,
+                EGL10.EGL_NO_CONTEXT);
         MapLibreGLSurfaceView view = mGLSurfaceViewWeakRef.get();
         if (view != null) {
-          view.eglWindowSurfaceFactory.destroySurface(mEgl, mEglDisplay, mEglSurface);
+          if (DEFER_CONTEXT_DESTRUCTION) {
```
cahaverl, line 305: Nit: To keep these cases from looking more different than they are, than they are supposed to be, how about swapping the "final" identifier assignments with the "if" statement. We could always call `factory.destroySurface(mEgl, display, surface);` either in this thread or in the cleanup thread.

platform/android/MapLibreAndroid/src/opengl/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreGLSurfaceView.java
adrian-cojocaru: The surface/context are touched only by the render thread (and in synchronized blocks). Unless I'm missing something this should guarantee that the values are valid in that scope.

cahaverl: Perhaps, but that requires non-local knowledge. For example, I am not sure if your analysis is correct.

Meanwhile, if you adopt the suggestion, then we won't need to discuss it. It will locally-provable that there is no race condition. That is preferable.

platform/android/MapLibreAndroid/src/opengl/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreGLSurfaceView.java
adrian-cojocaru: Hmm. You're right. The display cleanup will get skipped when there is no valid `mGLSurfaceViewWeakRef`, but at the same time `MapLibreGLSurfaceView` is the owner of the `EglHelper` (and passes itself as a weak reference) and I'm not sure if there is a scenario where the parent can go out of scope and the helper is still alive (this leads to other problematic leaks).

My main concern when adding the workarounds was to keep the existing surface lifecycle code the same since it has been problematic in the past. If it is confirmed that they solve the existing issues this needs to be extended to cover the texture view variant.

platform/android/MapLibreAndroid/src/opengl/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreGLSurfaceView.java
sjg-wdw: We'd like to see if this case is necessary.  It's a weird way to work around a driver bug and, if we're lucky, we can get rid of it completely.  

If it does have to be merged in, then we should look closer.

PR #3601, opened 2025-07-04
feat: add clusterMinPoints option

MarkusWendorf:


MarkusWendorf:
@louwers could you take a look, am I missing something?

I have some issues with the render tests on android locally but this also seems to happen on main, seems unrelated. I checked in the Android test app and looks fine to me (that is with your fix https://github.com/maplibre/maplibre-native/pull/3604)

Screenshot 2025-07-06 at 11 00 20

louwers:
Looks unrelated indeed. I would not worry about those render tests.

louwers:
@MarkusWendorf Did you try clustering on iOS?

MarkusWendorf:
@louwers yeah modified one of the swift ui examples and tested it there, there's no clustering example in the any of the example apps for ios right? 

Maybe I could add one

louwers (approved):

MarkusWendorf:
@louwers the "Android Instrumentation Tests" seems to be stuck / or waiting in queue forever, anything that can be done about that?

MarkusWendorf:
@louwers I wanted to avoid running all of them again but this works too.

But it seems like you have to approve the workflows again. Thank you

louwers:
@MarkusWendorf Thanks for your contribution! This will be released soon.

If you still want to contribute that iOS clustering example, that would be very welcome.

PR #3605, opened 2025-07-07
Fix calling completion callback for setContentInset in specific situation

yousifd:
If for any reason you end up calling `setContentInset` while `self.userTrackingState == MLNUserTrackingStateBegan` the passed in `completion` callback is never called. This change makes it so that even though we don't track the user's location after calling `setContentInset` we still call the passed in `completion` callback.

You can reproduce this by calling `setUserTrackingMode` right before calling `setContentInset` and you can check if whatever in the `completion` callback for `setContentInset` is ever called back and it's usually not.

### DIFF
--- a/platform/ios/src/MLNMapView.mm
+++ b/platform/ios/src/MLNMapView.mm
@@ -6339,7 +6339,8 @@ static void *windowScreenContext = &windowScreenContext;
     CLLocation *location = self.userLocation.location;
     if ( ! _showsUserLocation || ! location
         || ! CLLocationCoordinate2DIsValid(location.coordinate)
-        || self.userTrackingMode == MLNUserTrackingModeNone)
+        || self.userTrackingMode == MLNUserTrackingModeNone
+        || self.userTrackingState == MLNUserTrackingStateBegan)
     {
         if (completion)
         {


louwers (approved):

louwers:
Just wondering now if this is a breaking change? Do you know what the impact might be?

yousifd:
> Just wondering now if this is a breaking change? Do you know what the impact might be?

I can't be 100% sure but I don't think this is a breaking change because of the following:
- If you look at the logic below this changed line you'll see that we do not call the `completion` callback when `self.userTrackingState == MLNUserTrackingStateBegan`, so adding this ensures that we call it.
- This change does not affect the `setContentInset` function since we already update the content inset before calling `didUpdateLocationWithUserTrackingAnimated`.
- `didUpdateLocationWithUserTrackingAnimated` is only called in two places, the first one is in `setContentInset` the other one is in `- (void)locationManager:(__unused id<MLNLocationManager>)manager didUpdateLocations:(NSArray *)locations animated:(BOOL)animated completionHandler:(nullable void (^)(void))completion` which already checks for  `self.userTrackingState != MLNUserTrackingStateBegan` before setting the location
- The call to `unrotateIfNeededAnimated` also does nothing when `self.userTrackingState == MLNUserTrackingStateBegan` since it checks for it at the start.

louwers (approved):
Sounds good enough to me!

PR #3608, opened 2025-07-08
add ios clustering example app

MarkusWendorf:
Add simple geojson clustering example for the iOS sample app. Ugly but I think it gets the point across.

Screenshot 2025-07-08 at 16 32 32


louwers:
Ugly is good!

louwers (approved):

PR #3611, opened 2025-07-10
Font fallback + Harffbuzz integration

stefankarschti:
Implement [fallback mechanism for fonts](https://github.com/maplibre/maplibre-style-spec/pull/1204#issuecomment-3052084229) on top of #1439 codebase.


vendor/BUILD.bazel
```diff
@@ -227,6 +227,12 @@ cc_library(
     ],
 )
 
+filegroup(
```
louwers: I think this can be removed.

vendor/BUILD.bazel
```diff
@@ -235,3 +241,620 @@ cc_library(
...
+        "freetype/src/sdf/sdf.c",
+        "freetype/src/sfnt/sfnt.c",
+        "freetype/src/smooth/smooth.c",
+        "freetype/src/svg/svg.c",
+        "freetype/src/truetype/truetype.c",
+        "freetype/src/type1/type1.c",
+        "freetype/src/type42/type42.c",
+        "freetype/src/winfonts/winfnt.c",
+    ],
+    hdrs = glob([
```
louwers, line 287: All array elements in the `glob` should use a `*`.

We can move all files out e.g. `[file1, file2, file3] + glob(["/some/path/*.h"])`

We can probably use `**` to avoid having to explicitly specify so many files.

cmake/harfbuzz.cmake
```diff
@@ -0,0 +1,58 @@
+set(SHAPE_TARGETS, "")
```
louwers: If possible I think we should use the same pattern as we use for the other vendored dependencies.

I.e. create `vendor/freetype.cmake` and `vendor/harfbuzz.cmake`.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ option(MLN_WITH_WERROR "Make all compilation warnings errors" ON)
 option(MLN_USE_UNORDERED_DENSE "Use ankerl dense containers for performance" ON)
 option(MLN_USE_TRACY "Enable Tracy instrumentation" OFF)
 option(MLN_USE_RUST "Use components in Rust" OFF)
+option(MLN_TEXT_SHAPING_HARFBUZZ "Use haffbuzz to shape complex text" ON)
```
louwers, line 19: I think we should decide: do we make Harfbuzz optional yes or no.

If not, we should remove this option.

If we do decide it should be configurable, we should check on CI if it builds. I am getting linker errors on macOS if I try to build with `-DMLN_TEXT_SHAPING_HARFBUZZ=OFF`.

src/mbgl/style/parser.cpp
```diff
@@ -116,6 +116,74 @@ StyleParseResult Parser::parse(const std::string& json) {
...
+                            fontFace.name = faceName;
+
+                            // Parse url
+                            if (fontFile.HasMember("url")) {
+                                const JSValue& url = fontFile["url"];
+                                if (url.IsString()) fontFace.url = url.GetString();
+                            }
+
+                            // Parse unicode-range
+                            if (fontFile.HasMember("unicode-range")) {
```
louwers, line 143: The `unicode-range` is optional. When not specified, it should be set to `U+0-10FFFF`.

src/mbgl/style/parser.cpp
```diff
@@ -116,6 +116,74 @@ StyleParseResult Parser::parse(const std::string& json) {
...
+                                        if (range.IsString()) {
+                                            std::string rangeString = range.GetString();
+                                            if (rangeString.length() > 2) {
+                                                rangeString = rangeString.substr(2);
+                                                std::string::size_type pos = rangeString.find('-');
+                                                if (pos != std::string::npos) {
+                                                    std::string start = rangeString.substr(0, pos);
+                                                    std::string end = rangeString.substr(pos + 1);
+                                                    if (!start.empty() && !end.empty()) {
+                                                        int startInt = std::stoi(start, nullptr, 16);
```
louwers: This will throw an exception in case of an invalid number. I am not sure if we are handling that well on each platform. At the very least we should log a warning similar to other parts of the style parser.

cmake/harfbuzz.cmake
stefankarschti: of course. please let me know if there are other build system related changes you'd like to make.

CMakeLists.txt
stefankarschti: > I think we should decide: do we make Harfbuzz optional yes or no.

true. Weren't the extra size concerns the reason to make it optional? And if that is still a concern, we should make it optional, therefore I'd look into making `-DMLN_TEXT_SHAPING_HARFBUZZ=OFF` work.

stefankarschti: The [font-faces](https://github.com/maplibre/maplibre-style-spec/pull/1204) implementation is very intertwined with Harbuzz and Freetype. 

If `MLN_TEXT_SHAPING_HARFBUZZ=OFF` do we mean to also not to support `font-faces` in style? Maybe this deserves its own thread for discussions.

CC @HarelM @alanchenboy

alanchenboy (approved):

alanchenboy:
> I think this is pretty much good to go!
> 
> I have some build-related nits, if you want I can help clean the build config up a little bit @stefankarschti.
> 
> We should probably hold off merging until [maplibre/maplibre-style-spec#1204](https://github.com/maplibre/maplibre-style-spec/pull/1204) is merged, but since we addressed the feedback from the earlier proposal, I don't expect that to take too long.

This approach use `MLN_TEXT_SHAPING_HARFBUZZ` to turn on the shaping if this flag is turned off the shaping can be disable.

for bazel use this flag to turn if off
```
string_flag(
    name = "shaping",
    build_setting_default = "harfbuzz",
    values = [
        "legacy",
        "harfbuzz",
    ],
)
```

CMakeLists.txt
louwers: As discussed, I am open to dropping the configuration flag.

If someone cares about binary size they can put in the work to make it configurable.

louwers:
@stefankarschti added some defnes and somehow the tests are now failing with

> `corrupted size vs. prev_size`

We are holding off merging to investigate the issue.

include/mbgl/util/convert.hpp
```diff
@@ -36,5 +39,18 @@ inline constexpr auto to_array(std::tuple<Ts...>&& tuple) {
                       std::forward<std::tuple<Ts...>>(tuple));
 }
 
+template <typename T = int>
+std::enable_if_t<std::is_integral_v<T>, std::optional<T>>
+str_to_int(const std::string& str, int base = 10) noexcept {
+    T result;
+    auto first = str.data();
+    auto last = str.data() + str.size();
+    auto [ptr, ec] = std::from_chars(first, last, result, base);
```
louwers: `std::from_chars` requires at minimum iOS 13 😢 
https://developer.apple.com/xcode/cpp/

We're still targeting iOS 12 in this release series.

louwers:
Released for iOS https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.18.0

PR #3621, opened 2025-07-16
Support for MSYS2 builds

tdcosta100:
Adding MSYS2 support to Windows builds, so we can have one alternative to Microsoft Visual Studio builds.


louwers (approved):
I am unable to test this because I don't have a Windows PC.

But looks good!

Tracking the ios-ci fail here: https://github.com/maplibre/maplibre-native/issues/3629

tdcosta100:
> I am unable to test this because I don't have a Windows PC.
> 
> But looks good!
> 
> Tracking the ios-ci fail here: #3629

And I updated the documentation too, so we are again on par with other platforms!

louwers:
If you can resolve the glslang merge conflict I think CI will pass.

tdcosta100:
> If you can resolve the glslang merge conflict I think CI will pass.

Great! I updated to the same version in my branch, let's see if CI is going well now.

PR #3632, opened 2025-07-17
texture2d - getVulkanImage and support for Texture2DUsage::Attachment

birkskyum:
When working on the vulkan implementation of the updated Qt SDK, my output initially looked like this:

[Screencast_20250717_175543.webm](https://github.com/user-attachments/assets/c616b479-2f0e-4118-8360-f85b7a45b282)

I found I needed the attachment tiling, which is more gpu optimized, and then i could render the map

[Screencast_20250717_175934.webm](https://github.com/user-attachments/assets/17128442-65c4-416b-baae-1fd5be9f1a27)

I added a getVulkanImage header-only getter as well for convenience

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -100,26 +100,27 @@ jobs:
       #- name: Running iOS UI tests (Thread Sanitizer)
       #  run: bazel test //platform/ios/iosapp-UITests:uitest --test_output=errors --//:renderer=metal --features=tsan
 
-      - name: Running iOS UI tests (Address+UB Sanitizer)
-        run: |
-          retry_count=0
-          max_retries=5
-          while [ $retry_count -le $max_retries ]; do
-            bazel test //platform/ios/iosapp-UITests:uitest \
-              --test_output=errors \
-              --//:renderer=metal \
-              --features=include_clang_rt \
-              --copt=-fsanitize=undefined \
-              --linkopt=-fsanitize=undefined \
-              --copt=-fsanitize-recover=null \
-              --linkopt=-fsanitize-recover=null && break
-            retry_count=$((retry_count+1))
-            if [ $retry_count -eq $max_retries ]; then
-              echo "::error::Failed to run iOS UI tests after $max_retries attempts"
-              exit 1
-            fi
-            echo "Attempt $retry_count failed. Retrying..."
-          done
+      # https://github.com/maplibre/maplibre-native/issues/3642
+      # - name: Running iOS UI tests (Address+UB Sanitizer)
+      #   run: |
+      #     retry_count=0
+      #     max_retries=5
+      #     while [ $retry_count -le $max_retries ]; do
+      #       bazel test //platform/ios/iosapp-UITests:uitest \
+      #         --test_output=errors \
+      #         --//:renderer=metal \
+      #         --features=include_clang_rt \
+      #         --copt=-fsanitize=undefined \
+      #         --linkopt=-fsanitize=undefined \
+      #         --copt=-fsanitize-recover=null \
+      #         --linkopt=-fsanitize-recover=null && break
+      #       retry_count=$((retry_count+1))
+      #       if [ $retry_count -eq $max_retries ]; then
+      #         echo "::error::Failed to run iOS UI tests after $max_retries attempts"
+      #         exit 1
+      #       fi
+      #       echo "Attempt $retry_count failed. Retrying..."
+      #     done
 
       # render test
 


adrian-cojocaru:
The `Texture2DUsage::Attachment` flag wasn't meant to have CPU read capabilities since adding `vk::ImageUsageFlagBits::eTransferSrc` can have an impact on performance for the general scenario (GPU only texture). I'd suggest adding a separate value to `Texture2DUsage` that handles that specific case.

I'm curious on how the texture is used. Is the application rendering a bitmap of the `Texture2D::readImage` or using the vulkan texture as an attachment?

src/mbgl/vulkan/texture2d.cpp
```diff
@@ -502,31 +508,97 @@ std::shared_ptr<PremultipliedImage> Texture2D::readImage() {
...
-    const auto& layout = device->getImageSubresourceLayout(imageAllocation->image,
-                                                           vk::ImageSubresource(vk::ImageAspectFlagBits::eColor, 0, 0));
-
     imageData->resize(size);
     const auto& imageSize = getDataSize();
 
-    void* mappedData_ = nullptr;
-    vmaMapMemory(context.getBackend().getAllocator(), imageAllocation->allocation, &mappedData_);
+    // Check if this is an attachment texture that needs staging buffer
+    if (textureUsage == Texture2DUsage::Attachment || textureUsage == Texture2DUsage::AttachmentReadable) {
```
adrian-cojocaru: Since `Texture2DUsage::Attachment` lost the transfer bit this will fail. I can update this after it's merged with some asserts for unsupported usages.

src/mbgl/vulkan/texture2d.cpp
```diff
@@ -502,31 +508,97 @@ std::shared_ptr<PremultipliedImage> Texture2D::readImage() {
...
+        }
 
-    uint8_t* mappedData = reinterpret_cast<uint8_t*>(mappedData_) + layout.offset;
+        // Copy image to staging buffer
+        context.submitOneTimeCommand([&](const vk::UniqueCommandBuffer& commandBuffer) {
+            // Transition image layout for reading
+            const auto barrier = vk::ImageMemoryBarrier()
+                                     .setImage(imageAllocation->image)
+                                     .setOldLayout(imageLayout)
+                                     .setNewLayout(vk::ImageLayout::eTransferSrcOptimal)
```
adrian-cojocaru, line 561: `imageLayout` should be updated to `vk::ImageLayout::eTransferSrcOptimal` after the pipeline barrier.

birkskyum:
Thanks for the review! Itś using the offscreen texture, which has the line texture.setUsage(Texture2DUsage::Attachment);` [here](https://github.com/maplibre/maplibre-native/blob/12e0922fc4cadcd88808830e697cfb1d5206c8c9/src/mbgl/vulkan/offscreen_texture.cpp#L28), so if we introduce another e.g. Texture2DUsage::AttachmentReadable in the texture2d, it would also require some conditional logic to the offscreen texture.

I´m making bitmaps and loading those - itś a bit inefficient to say the least, but it´s a compatibility solution that works for now, which is better than no throughput. The reason for it is that going 100% gpu with vulkan and qt presently doesn´t seem to work on arm linux laptops, and streaming bitmaps, effectively screenshots, for better or worse gives visual output with a usable framerate.

src/mbgl/vulkan/texture2d.cpp
birkskyum: Good catch, i´ve added the imageLayout = barrier.newLayout;

src/mbgl/vulkan/texture2d.cpp
birkskyum: This AttachmentReadable gave me some issues with the offscreen texture that I use, because it use the Attachment, and if I change that to use AttachmentReadable for my use case it also changes it for all other users of the offscreen texture

birkskyum: The offscreen texture that use Texture2DUsage::Attachment, do you have idea how to allow it to use the AttachmentReadable, without breaking existing behavior

adrian-cojocaru: A `bool readable` (defaulting to false) passed to the `OffscreenTexture` constructor that handles the toggle between the two usages would be ideal I think. Would this work for your use case?

`texture.setUsage(readable ? Texture2DUsage::AttachmentReadable : Texture2DUsage::Attachment)` in [vulkan/offscreen_texture.cpp](https://github.com/maplibre/maplibre-native/blob/12e0922fc4cadcd88808830e697cfb1d5206c8c9/src/mbgl/vulkan/offscreen_texture.cpp#L28)

birkskyum: I´d appreciate any help on landing this - it´s not clear to me if itś best to always have the offscreen texture set to AttachmentReadable, have that be set by a parameter in the constructor, or if it´s better to have the attachment be readable as in the initial version of this PR.

I reckon this work also would allow to implement the vulkan offscreen textures´ readStillImage, which is just a stub now, by calling readImage.

adrian-cojocaru: Looks like `OffscreenTexture::readStillImage` is part of the interface, but it's not implemented (the method is a stub), so the `AttachmentReadable` is the right choice in this case. If it's working right now you can leave it as is and I'll update the missing `readStillImage` and `Texture2DUsage` asserts.

PR #3645, opened 2025-07-22
Bug: Don't abort on potential HTTP/2 connections

### LINKED ISSUE #3644: Qt6 causes HTTP/2 errors
**Describe the bug**
I have opened the bug downstream. https://github.com/maplibre/maplibre-native-qt/issues/221

And with a bit of help I found some workarounds. It is obviously that the true solution must be found here (or even more upstream in Qt6).

**To Reproduce**

1. Run the widget test with bit longer timeout 
2. Scroll into map
3. Notice that no new tiles are requested
4. See error in the console

**Expected behavior**

It is obvious that the user wants an interface that works.

**Screenshots**

./test_mln_widgets
********* Start testing of TestWidgets *********
Config: Using QtTest library 6.9.1, Qt 6.9.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 10.3.1 20210422 (Red Hat 10.3.1-1)), arch unknown
PASS   : TestWidgets::initTestCase()
QWARN  : TestWidgets::testGLWidgetStyle() [ INFO ]  "{unknown}[General]: GPU Identifier: llvmpipe (LLVM 20.1.8, 256 bits)"
QCRITICAL: TestWidgets::testGLWidgetStyle() qt.network.http2.connection: [0x7798ac05f8d0] Connection error: HEADERS on invalid stream (1)
QCRITICAL: TestWidgets::testGLWidgetStyle() qt.network.http2.connection: [0x7798ac0786c0] Connection error: HEADERS on invalid stream (1)
QCRITICAL: TestWidgets::testGLWidgetStyle() qt.network.http2.connection: [0x7798ac0736d0] Connection error: DATA on invalid stream (11)
QWARN  : TestWidgets::testGLWidgetStyle() qt.network.http2: stream 13 error: "DATA on invalid stream"
QWARN  : TestWidgets::testGLWidgetStyle() qt.network.http2: stream 13 finished with error: "Server dislikes our behavior, excessive load detected."
QWARN  : TestWidgets::testGLWidgetStyle() qt.network.http2: stream 15 error: "DATA on invalid stream"
QWARN  : TestWidgets::testGLWidgetStyle() qt.network.http2: stream 15 finished with error: "Server dislikes our behavior, excessive load detected."
QWARN  : TestWidgets::testGLWidgetStyle() [ ERROR ]  "{unknown}[Style]: Failed to load tile 2/2/1=>2 for source openmaptiles: HTTP status code 0"
QWARN  : TestWidgets::testGLWidgetStyle() [ ERROR ]  "{unknown}[Style]: Failed to load tile 2/1/1=>2 for source openmaptiles: HTTP status code 0"

** Workarounds **
 1. Disable HTTP/2, with HTTP/1 there are no errors.
```
diff --git a/platform/qt/src/mbgl/http_file_source.cpp b/platform/qt/src/mbgl/http_file_source.cpp
index 2cdb50f734b..fdada5298fe 100644
--- a/platform/qt/src/mbgl/http_file_source.cpp
+++ b/platform/qt/src/mbgl/http_file_source.cpp
@@ -37,7 +37,7 @@ void HTTPFileSource::Impl::request(HTTPRequest* req) {
     networkRequest.setAttribute(QNetworkRequest::FollowRedirectsAttribute, true);
 #endif
 #endif
-
+    networkRequest.setAttribute(QNetworkRequest::Http2AllowedAttribute, false);
     data.first = m_manager->get(networkRequest);
     connect(data.first, &QNetworkReply::finished, this, &HTTPFileSource::Impl::onReplyFinished);
 #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
```
 2. Explicitly handle the errors and clear the connection Cache. The latter is the most important part.
```diff
@@ -74,6 +74,7 @@ void HTTPFileSource::Impl::cancel(HTTPRequest* req) {
 
 void HTTPFileSource::Impl::onReplyFinished() {
     QNetworkReply* reply = qobject_cast<QNetworkReply*>(sender());
+    if (!reply) return;
 
 #if defined(Q_OS_WASM)
     const QUrl& url = reply->url();
@@ -91,14 +92,27 @@ void HTTPFileSource::Impl::onReplyFinished() {
         return;
     }
 
-    QByteArray data = reply->readAll();
-    QVector<HTTPRequest*>& requestsVector = it.value().second;
+    // Error handling
+    if (reply->error() != QNetworkReply::NoError) {
+        qWarning() << "Network error for URL" << url << ":" << reply->errorString();
+
+        QVariant statusCode = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute);
+        if (!statusCode.isValid()) {
+            qWarning() << "No HTTP status code received; possible connection/protocol failure.";
+        }
 
-    // Cannot use the iterator to walk the requestsVector
-    // because calling handleNetworkReply() might get
-    // requests added to the requestsVector.
-    while (!requestsVector.isEmpty()) {
-        requestsVector.takeFirst()->handleNetworkReply(reply, data);
+       m_manager->clearConnectionCache();
+    } else 
+    {
+           QByteArray data = reply->readAll();
+           QVector<HTTPRequest*>& requestsVector = it.value().second;
+
+           // Cannot use the iterator to walk the requestsVector
+           // because calling handleNetworkReply() might get
+           // requests added to the requestsVector.
+           while (!requestsVector.isEmpty()) {
+               requestsVector.takeFirst()->handleNetworkReply(reply, data);
+           }
     }
 
     m_pending.erase(it);
```

 - Operating System: Linux
 - Platform (e.g. Node.js, Qt): Qt6
 - Version: 6.9.1

skinkie:
In recent Qt6 releases HTTP2 is enabled by default, this was not the case for Qt5. With this change cancelling a connection via reply->abort() breaks the protocol. We should not do that.

In addition, if the protocol is broken in an error situation, we could not recover from this.

Fix #3644


skinkie:
@louwers can you rerun the GCC14? It seems that this is unrelated to the change.

https://github.com/maplibre/maplibre-native/actions/runs/16457498317/job/46555453962

ntadej:
Looks OK to me (but I do not have time to test it). Will backport it then later also to the OpenGL2 branch for now.

This is maybe more a question for @louwers as you have a better overview: I guess in worst case now we load too much data? Probably I'm paranoid, but this does not sound ideal for mobile devices?

louwers:
What do you mean?

ntadej:
I think we abort if one zooms or pans while loading to avoid unnecessary data transfer. This MR now unconditionally disables that. I guess on desktops it's fine, but for mobile devices on data every byte saved usually matters.

skinkie:
I wonder if the requests that are upcomming not already removed from the list. I would agree that the current live transfer will be finished in this scenario.

https://github.com/maplibre/maplibre-native/pull/3645/files#diff-0f6edf95426d1644f0d13239a71353940518944ac0b60078d3da88f71d0c0666R59

ntadej:
I was looking into this a bit more. Do you have a server where this can be reproduced, as I can not do it? Also supposedly Qt should handle HTTP/2 aborts just fine.

skinkie:
> I was looking into this a bit more. Do you have a server where this can be reproduced, as I can not do it? Also supposedly Qt should handle HTTP/2 aborts just fine.

Absolutely, this is _mbtileserver_ plus _haproxy_. https://zonekaart.nl/style.json

ntadej:
OK, I can now reproduce this issue both on Linux and macOS using the MapLibre test style. The issue is this seems to break a unit test. I suspect there's some issue with threading and HTTP/2.

ntadej:
There are several fixes that will land in Qt 6.9.2 (in a few days): https://bugreports.qt.io/browse/QTBUG-132124?jql=project%20%3D%20QTBUG%20AND%20text%20~%20%22qt.network.http2.connection%22%20ORDER%20BY%20created%20DESC

Let's test there if the situation improves.

skinkie:
Would you agree even if 6.9.2 would resolve the issue, it should still be mitigated for any version before?

ntadej:
> Would you agree even if 6.9.2 would resolve the issue, it should still be mitigated for any version before?

In principle yes, but the fix should not break timeouts handling. Also everyone reports the issue only for Qt 6.9 when HTTP/2 is used by default (if I understand correctly it was opt-in before).

PR #3646, opened 2025-07-23
Fix Windows core-relase action

tdcosta100:
Fixes core-relase action failing because of Windows jobs:

https://github.com/maplibre/maplibre-native/actions/runs/16421648591


louwers:
Thank you!

PR #3649, opened 2025-07-23
Add Vulkan device sharing example

adrian-cojocaru:
This PR adds the possibility to pass an external instance/device pair to the core backend, enabling resource sharing between the owner and MapLibre.


birkskyum:
Excellent, thanks! This will definitely help move forward the zero-copy vulkan backend for qt.

louwers (approved):

PR #3655, opened 2025-07-24
Add MLN_CREATE_AUTORELEASEPOOL flag to create autoreleasepool in render loop

louwers:
When using the Metal backend, if there is is no AutoreleasePool, it will result
in memory leaks when autoreleased objects are created.

Platforms such as iOS and GLFW use ARC with @autoreleasepool blocks, for these platforms
creating an AutoreleasePool is not neccesary (and in fact will not work).
When using MapLibre Native with Metal on other platforms such as Node.js, you can pass
`MLN_CREATE_AUTORELEASEPOOL` to ensure an AutoReleasePool is created in the render loop.

Doesn't really solve https://github.com/maplibre/maplibre-native/issues/2928, but will be fixed once Node.js builds for macOS use `MLN_CREATE_AUTORELEASEPOOL`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1512,7 +1512,7 @@ if(MLN_WITH_METAL AND MLN_CREATE_AUTORELEASEPOOL)
     # Platforms such as iOS and GLFW use ARC with @autoreleasepool blocks, for these platforms
     # creating an AutoreleasePool is not neccesary (and in fact will not work).
     # When using MapLibre Native with Metal on other platforms such as Node.js, you can pass
-    # this configuration command to ensure an AutoReleasePool is created in the render loop.
+    # this variable to ensure an AutoReleasePool is created in the render loop.
     # See also:
     # https://developer.apple.com/documentation/foundation/nsautoreleasepool
     # https://github.com/maplibre/maplibre-native/issues/2928
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ option(MLN_USE_UNORDERED_DENSE "Use ankerl dense containers for performance" ON)
 option(MLN_USE_TRACY "Enable Tracy instrumentation" OFF)
 option(MLN_USE_RUST "Use components in Rust" OFF)
 option(MLN_CORE_INCLUDE_DEPS "Include depdendencies in static build of core" OFF)
+option(MLN_CREATE_AUTORELEASEPOOL "Create autoreleasepool in render loop" OFF)
 
 if (MLN_LEGACY_RENDERER)
     message(FATAL "The legacy renderer is no longer supported")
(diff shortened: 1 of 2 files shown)


acalcutt:
I can give this a try on my WifiDB fork where I usually make test maplibre-native releases a bit later.

I would imagine we would just be adding this flag to https://github.com/maplibre/maplibre-native/blob/main/CMakePresets.json#L63-L69 since the node workflow users this preset file for macos.

alexcristici (approved):

cgalvan (approved):

acalcutt:
I tried to test this with the test mentioned in https://github.com/maplibre/maplibre-native/issues/2928#issuecomment-3114239944 but I wasn't able to get it to work.

At first it looked like it was missing libuv, so I installed it with brew and it looked like it fixed that error. however when it ran it just tried to generate 1 of 50 images and ended with no error. there was a warning I am not sure if is related to it not rendering.
```
maplibre-node-test % node .
(node:92598) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Rendering 1 of 50
maplibre-node-test %
```

I also tried to add it into tileserver-gl as a test but it didn't seem to render there either.

louwers:
@lseelenbinder Could you try this fix?

For me this worked to get `bin/render.cpp` to render more than 32 times on macOS.

acalcutt:
Just an FYI i made a second node test release with this new option removed, but I had the same trouble with that release. so maybe something broke before this PR like maybe the switch to use libuv or something like that.

louwers:
@acalcutt Yes that seems likely! The tests also fail.

acalcutt:
@louwers I tested stating with the last node 6.0.1 release commit ( https://github.com/maplibre/maplibre-native/commit/6491d98b632d8783068c14925b816a371fa69686 ) then cherry picking your changes from this PR in https://github.com/WifiDB/maplibre-native/tree/node_test_7-28 and it seems to work in that test release on macos

The test release I made with these changes included is [@acalcutt/maplibre-gl-native-test v6.1.1-pre.3](https://www.npmjs.com/package/@acalcutt/maplibre-gl-native-test/v/6.1.1-pre.3) and tested it on the test app in https://github.com/maplibre/maplibre-native/issues/2928 and tileserver-gl and both seemed to render without freezing.

So it does seem like this fixes the issue, at least from my initial testing. unfortunately other changes seems to have broken the node release sometime after 6.0.1.

louwers:
Awesome! Bummer though that something else broke. But it is good to have a working Node.js release out with Metal.

But I guess this can be merged?

acalcutt:
Ya, I think this could be merged since it seemed to work as expected. but it seems some more testing will be needed to find which point the node release broke ( I am still guessing when macos was changed to libuv recently, but that would need testing)

acalcutt:
@louwers  do you think it would it be ok for me to fix the current maplibre 6.1.0 node release by replacing the darwin binaries in https://github.com/maplibre/maplibre-native/releases/tag/node-v6.1.0 with the ones I just generated and tested in https://github.com/WifiDB/maplibre-native/releases/tag/node-v6.1.1-pre.3 . It would fix our current release on macos since when installing the binaries are pulled from there.

It would give us some time to test this better and also time to figure out what is going on in the current version. 

If not i totally understand wanting to wait for a proper release.

lseelenbinder:
@louwers Just confirming this works on our side. Thanks!

PR #3661, opened 2025-07-25
Adds feature collections to plug-in layers

AtlasProgramming:
Updated PR: https://github.com/maplibre/maplibre-native/pull/3698

This adds the ability to latch into the tile loading and receive the geometry and properties for features in the tiles.

### DIFF
--- a/src/mbgl/plugin/feature_collection.hpp
+++ b/src/mbgl/plugin/feature_collection.hpp
@@ -22,7 +22,7 @@ public:
     double _tileY = 0; // Tile coord
 };
 
-// This is a list of coordinates.  Broken out into it's own class because
+// This is a list of coordinates.  Broken out into its own class because
 // a raw bucket feature can have an array of these
 class FeatureCoordinateCollection {
 public:
--- a/src/mbgl/plugin/plugin_layer_render.cpp
+++ b/src/mbgl/plugin/plugin_layer_render.cpp
@@ -127,7 +127,7 @@ void RenderPluginLayer::update([[maybe_unused]] gfx::ShaderRegistry& shaderRegis
                     continue;
                 }
 
-                // See if we already have this tile'a feature collection
+                // See if we already have this tile's feature collection
                 if (_featureCollectionByTile.contains(tileID)) {
                     continue;
                 }


src/mbgl/plugin/feature_collection_bucket.cpp
```diff
@@ -0,0 +1,146 @@
...
+    double mercY = 1.0 - worldPixelY / worldSize * 2.0;
+
+    lon = mercX * 180.0;
+    lat = 180.0 / M_PI * (2.0 * atan(exp(mercY * M_PI)) - M_PI / 2.0);
+}
+
+std::string toString(FeatureIdentifier& v) {
+    std::string tempResult = "";
+    auto ti = v.which();
+    if (ti == 0) {
```
TimSylvester: seems like these could return directly without constructing a temporary

src/mbgl/plugin/feature_collection_bucket.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace mbgl {
+
+class BucketParameters;
+class RenderFillLayer;
+
+class FeatureCollectionBucket final : public Bucket {
+public:
+    ~FeatureCollectionBucket() override;
+    // using PossiblyEvaluatedLayoutProperties = style::FillLayoutProperties::PossiblyEvaluated;
```
TimSylvester: remove commented-out code

src/mbgl/plugin/plugin_layer_factory.hpp
```diff
@@ -23,6 +23,9 @@ class PluginLayerFactory : public LayerFactory {
                        mbgl::style::LayerTypeInfo::CrossTileIndex crossTileIndex,
                        mbgl::style::LayerTypeInfo::TileKind tileKind);
 
+    // Set to false, but if the caller wants to support on features loaded, then set this to true
+    bool _supportsFeatureCollectionBuckets = false;
```
TimSylvester: Why are these prefixed with `_`?  If it's to prevent shadowing, the ML convention is to use a suffix on the parameters.

src/mbgl/plugin/plugin_layer_render.cpp
```diff
@@ -85,6 +85,70 @@ void RenderPluginLayer::update([[maybe_unused]] gfx::ShaderRegistry& shaderRegis
...
+        if (!renderTiles->empty()) {
+            auto drawPass = RenderPass::Pass3D;
+
+            // If we're reading feature collections, go through
+            // and notify the plugin of any new feature collections
+            for (const RenderTile& tile : *renderTiles) {
+                const auto& tileID = tile.getOverscaledTileID();
+
+                if (!hasRenderTile(tileID)) {
+                }
```
TimSylvester: empty conditional?

src/mbgl/plugin/plugin_layer_render.cpp
```diff
@@ -85,6 +85,70 @@ void RenderPluginLayer::update([[maybe_unused]] gfx::ShaderRegistry& shaderRegis
...
+                    continue;
+                }
+                const auto& renderData = *optRenderData;
+                auto& bucket = static_cast<FeatureCollectionBucket&>(*renderData.bucket);
+                auto featureCollection = bucket._featureCollection;
+                if (featureCollection == nullptr) {
+                    continue;
+                }
+
+                // See if we already have this tile'a feature collection
```
TimSylvester:
```suggestion
                // See if we already have this tile's feature collection
```

src/mbgl/plugin/feature_collection.hpp
```diff
@@ -0,0 +1,57 @@
...
+    FeatureCoordinate(double lat, double lon)
+        : _lat(lat),
+          _lon(lon) {}
+    double _lat = 0;
+    double _lon = 0;
+    double _tileX = 0; // Tile coord
+    double _tileY = 0; // Tile coord
+};
+
+// This is a list of coordinates.  Broken out into it's own class because
```
TimSylvester:
```suggestion
// This is a list of coordinates.  Broken out into its own class because
```

src/mbgl/plugin/plugin_layer_factory.hpp
AtlasProgramming: Ah.  Just the standard C++ convention for naming class iVars that I'm used to.  Can change, but all the other iVars in the plugin stuff are prefixed with _.

src/mbgl/plugin/feature_collection_bucket.cpp
AtlasProgramming: Good point.  Probably legacy from debugging

src/mbgl/plugin/feature_collection_bucket.hpp
AtlasProgramming: done!

src/mbgl/plugin/plugin_layer_factory.hpp
AtlasProgramming: updated

src/mbgl/plugin/plugin_layer_render.cpp
AtlasProgramming: good catch.  removed!

AtlasProgramming:

PR #3662, opened 2025-07-25
Add the ability to filter the style data while loading

### LINKED ISSUE #3663: Add the ability to filter the style data before parsing
This was borne out of the need to encrypt our style.  We'd like to intercept the style request and decrypt it before it's parsed.  However, once this was built, it's clear that it can be used for so many other things like custom filters for styles could be created to adjust/add/etc the style before it's passed to the style parser.

I'd like to be able to specify one or more "filters" that are applied to the style data before it's parsed.

There are not other alternatives that allow for this ability.  Custom protocols would lock you into a specific method for fetching the style data/etc.

**Additional context**

AtlasProgramming:
Updated PR: https://github.com/maplibre/maplibre-native/pull/3698

This will add the ability to latch into the style loading and modify the style before it gets to the actual style parsing.  There are two parts to this functionality.  The first is a C++ based latch into the actual style that will allow a lambda to be assigned to each filter that will receive (in order of add) the data after it's loaded from the source (e.g. http/file/etc).  Each filter is passed the output of the previous filter.

The second part is a platform implementation of this for iOS that creates a base filter class that the user can descend from, override a single method and then get the input data and return the data to pass along the filter chain.  Since the filters are added at runtime, this will allow run-time filter plugins to be registered and utilized.


louwers (changes requested):
Networking concerns are currently neatly separated from the `Style` class in its implementation. This PR breaks that separation, the 'style filter' works with a network request.

Instead of adding a hook to modify style requests, I would much rather see a a more generic solution, one that allows adding hooks that change arbitrary network requests. Take a look add [`addProtocol`](https://maplibre.org/maplibre-gl-js/docs/API/functions/addProtocol/) for MapLibre GL JS for example. You could e.g. intercept the `hudhud://` protocol and run a custom hook on that.

I would recommend to create an issue with the use case(s) you envision so we can explore the solution space with other users.

AtlasProgramming:
> Networking concerns are currently neatly separated from the `Style` class in its implementation. This PR breaks that separation, the 'style filter' works with a network request.
> 
> Instead of adding a hook to modify style requests, I would much rather see a a more generic solution, one that allows adding hooks that change arbitrary network requests. Take a look add [`addProtocol`](https://maplibre.org/maplibre-gl-js/docs/API/functions/addProtocol/) for MapLibre GL JS for example. You could e.g. intercept the `hudhud://` protocol and run a custom hook on that.
> 
> I would recommend to create an issue with the use case(s) you envision so we can explore the solution space with other users.

Just to be clear.  This does not require or link the style with the network at all.  This simply gets the "style data" before the parser and has the ability to filter/translate it.

AtlasProgramming:
> Networking concerns are currently neatly separated from the `Style` class in its implementation. This PR breaks that separation, the 'style filter' works with a network request.
> 
> Instead of adding a hook to modify style requests, I would much rather see a a more generic solution, one that allows adding hooks that change arbitrary network requests. Take a look add [`addProtocol`](https://maplibre.org/maplibre-gl-js/docs/API/functions/addProtocol/) for MapLibre GL JS for example. You could e.g. intercept the `hudhud://` protocol and run a custom hook on that.
> 
> I would recommend to create an issue with the use case(s) you envision so we can explore the solution space with other users.

Issue here: https://github.com/maplibre/maplibre-native/issues/3663

AtlasProgramming:
Change the flow a bit to support when loadJSON is called.  Removed references to the Response object which was being passed around needlessly.  So this implementation is more tightly focussed on just filtering the style however it's loaded (either via some kind of URL) or directly via loadJSON.

TimSylvester:
So each filter would need to separately parse the JSON, perhaps apply changes, and convert back to a string?  As long as the filter is on the raw string content, I agree that it's more appropriate as a generic request filter.  I would expect a style filter to pass around the parsed tree, but that makes the ObjC layer tricky.

AtlasProgramming:
> So each filter would need to separately parse the JSON, perhaps apply changes, and convert back to a string? As long as the filter is on the raw string content, I agree that it's more appropriate as a generic request filter. I would expect a style filter to pass around the parsed tree, but that makes the ObjC layer tricky.

Currently yes.  It's less than ideal, but like you mentioned, passing around the parsed JSON would be tricky and that by passing the binary representation around we can have the style, from the source, be in a binary form (e.g. compressed, encrypted, bson, protobuf/etc) down the road.

TimSylvester:
I was thinking more about this, and I guess I'm not clear on the use case.

If I'm not mistaken, styles aren't ever fetched automatically, like spritesheets are as a result of loading a style.  They're only loaded when the app provides them.  So if the app needs to do something with a style before loading it, shouldn't the app do whatever it needs to and then pass the resulting JSON in as a string?

It seems like the only real benefit you're getting here is to have MLN do the downloading.  But ObjC, Swift, Kotlin, etc., have simple facilities for that.

AtlasProgramming:

louwers:
@TellToldTold Could you have a look at this? Sounds like it may fit your use case as well.

PR #3664, opened 2025-07-25
vulkan - Allow opting out of surface based rendering

birkskyum:
Follow up to
- #3649 

I realized that init() calls initSurface(), which looks like this:

```cpp
void RendererBackend::initSurface() {
    getDefaultRenderable().getResource<SurfaceRenderableResource>().createPlatformSurface();
}
```

For offscreen rendering, we can't have references to the SurfaceRenderableResource, so the init() also has to be virtual, so that it can be overwritten for offscreen use.

Also the context.cpp, this PR puts all the surface rendering logic inside conditions.

The result of this is that finally, the Vulkan backend for Qt can [render correctly](https://github.com/maplibre/maplibre-native-qt/pull/216#issuecomment-3120775668), by injecting the Vulkan device from Qt into the MapLibre Native core

### DIFF
--- a/include/mbgl/vulkan/renderer_backend.hpp
+++ b/include/mbgl/vulkan/renderer_backend.hpp
@@ -38,7 +38,7 @@ public:
 
     /// One-time shader initialization
     void initShaders(gfx::ShaderRegistry&, const ProgramParameters& programParameters) override;
-    void init();
+    virtual void init();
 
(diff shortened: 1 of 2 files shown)


birkskyum:
Close in favor of using the SurfaceRenderableResource
- https://github.com/maplibre/maplibre-native-qt/pull/216/commits/afa3707e4849d5ee2118d8bab074bcb1b794f2d4

PR #3665, opened 2025-07-26
Adding support for custom protocols

AtlasProgramming:
This adds support similar to plugin-layers and style filters in exposing the loading paradigm via a plug-in class and lambda implementation.  This does not create or implement a way to chain FileSources together.

```
```

```

        return YES;
    }
    return NO;
}

    response.data = data;
    return response;
    
}

@end
```

```

    }


        // Return some data here
        return null;
    }

}
```

```
mapView.getMapAsync {
```


src/mbgl/storage/file_source_manager.cpp
```diff
@@ -86,4 +87,14 @@ FileSourceManager::FileSourceFactory FileSourceManager::unRegisterFileSourceFact
     return factory;
 }
 
+// Registers a custom file source factory
+void FileSourceManager::registerCustomFileSource(std::shared_ptr<FileSource> fileSource) noexcept {
+    impl->customFileSources.push_back(fileSource);
+}
+
+// Returns an array of custom file sources
+std::vector<std::shared_ptr<FileSource>> FileSourceManager::getCustomFileSources() noexcept {
```
TimSylvester, line 96: This makes a copy, and so isn't `noexcept`, nor is the register above.

The `request`/`canRequest` loops happen on the main thread, right?  If so, there's no need to operate on a snapshot of the collection.  Otherwise, some locking would be necessary when constructing that, anyway.

platform/ios/src/MLNMapView.mm
```diff
@@ -7801,6 +7805,140 @@ -(void)addPluginLayerType:(Class)pluginLayerClass {
...
+    mbgl::ResourceOptions resourceOptions;
+
+    // TODO: Unclear if any of the properties on clientOptions need to be set
+    mbgl::ClientOptions clientOptions;
+
+    // Use weak here so there isn't a retain cycle
+    __weak MLNPluginProtocolHandler *weakHandler = handler;
+    __weak MLNMapView *weakSelf = self;
+
+    std::shared_ptr<mbgl::PluginFileSource> pluginSource = std::make_shared<mbgl::PluginFileSource>(resourceOptions, clientOptions);
```
TimSylvester, line 8016: prefer `auto` in place of complex type names

platform/default/src/mbgl/storage/main_resource_loader.cpp
```diff
@@ -65,54 +65,67 @@ class MainResourceLoaderThread {
...
-                            // the network, we will forward response to the
-                            // requester.
-                            res.priorData = response.data;
-                        }
+        // Go through custom handlers
+        bool requestHandledByCustomHandler = false;
+        auto fm = FileSourceManager::get();
+        for (auto customFileSource : fm->getCustomFileSources()) {
+            if (customFileSource->canRequest(resource)) {
+                tasks[req] = customFileSource->request(resource, callback);
```
TimSylvester, line 73: It looks like `request` is expected to never return null, though the base class implementation does so (the ObjC one, at least).  That base implementation should probably have an assert since it shouldn't be called, and this should assert and check for null and treat it the same as `!canRequest`.

AtlasProgramming:

PR #3676, opened 2025-07-29
Frustum Offset Proposal

yousifd:
This PR adds the following:
- Exposes a `frustumOffset` property in iOS and Android that allows a user to specify how much of the edges of the screen is out of view and shouldn't be rendered in logical pixel units.
- Modifies the projection matrix computation to compute the fov with a top offset from the provided `frustumOffset` property.
- Modifies the Metal, OpenGL, and Vulkan renderers to apply a scissor test given the `frustumOffset` to reject fragments at the edges of the screen which might be behind UI elements.

### DIFF
--- a/include/mbgl/map/map_options.hpp
+++ b/include/mbgl/map/map_options.hpp
@@ -133,6 +133,22 @@ public:
      */
     float pixelRatio() const;
 
+    /**
+     * @brief Sets the frustum offset. Used to avoid rendering edges of the screen
+     * which might be behind UI elements.
+     *
+     * @param offset_ A set of edge insets in logical pixels.
+     * @return reference to MapOptions for chaining options together.
+     */
+    MapOptions& withFrustumOffset(const EdgeInsets& offset_);
+
+    /**
+     * @brief gets the previousl set (or default) frustum offset.
+     *
+     * @return frustum offset value
+     */
+    EdgeInsets frustumOffset() const;
+
 private:
     class Impl;
     std::unique_ptr<Impl> impl_;
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
@@ -117,6 +117,13 @@ public final class MapLibreMap {
     nativeMapView.enableRenderingStatsView(value);
   }
 
+  /**
+   * Set frustum offset to disable rendering for edges of the screen
+   */
+  public void setFrustumOffset(@NonNull RectF offset) {
+      nativeMapView.setFrustumOffset(offset);
+  }
+
   public void setSwapBehaviorFlush(boolean flush) {
     nativeMapView.setSwapBehaviorFlush(flush);
   }
--- a/platform/ios/src/MLNMapView.h
+++ b/platform/ios/src/MLNMapView.h
@@ -508,6 +508,16 @@ MLN_EXPORT
  */
 @property (nonatomic, assign) double tileLodZoomShift;
 
+/**
+ Frustum offset used to disable rendering of elements at the edge of the screen
+
+ Offset applied to camera frustum and scissor rectangle. The camrea frustum is modofied
+ to avoid loading geometry that's behind UI elements at the top of the screen. The scissor
+ rectangle is used to avoid shading fragments that are behind UI elements at the edges of
+ the screen. All values are in logical pixels.
+ */
+@property (nonatomic, assign) UIEdgeInsets frustumOffset;
+
 // MARK: Displaying the User’s Location
 
 /**
--- a/src/mbgl/gfx/scissor_rect.hpp
+++ b/src/mbgl/gfx/scissor_rect.hpp
@@ -1,0 +1,21 @@
+#pragma once
+
+#include <cstdint>
+
+namespace mbgl {
+namespace gfx {
+
+class ScissorRect {
+public:
+    int32_t x, y;
+    uint32_t width, height;
+
+    bool operator==(const ScissorRect& other) const {
+        return x == other.x && y == other.y && width == other.width && height == other.height;
+    }
+
+    bool operator!=(const ScissorRect& other) const { return !(*this == other); }
+};
+
+} // namespace gfx
+} // namespace mbgl
--- a/src/mbgl/vulkan/pipeline.cpp
+++ b/src/mbgl/vulkan/pipeline.cpp
@@ -6,8 +6,6 @@
 namespace mbgl {
 namespace vulkan {
 
-#define USE_DYNAMIC_VIEWPORT 0
-
 vk::Format PipelineInfo::vulkanFormat(const gfx::AttributeDataType& value) {
     switch (value) {
         case gfx::AttributeDataType::Byte:
@@ -108,6 +106,10 @@ vk::CullModeFlagBits PipelineInfo::vulkanCullMode(const gfx::CullFaceSideType& v
         default:
             return vk::CullModeFlagBits::eNone;
     }
+}
+
+vk::Rect2D PipelineInfo::vulkanScissorRect(const gfx::ScissorRect& value) {
+    return {{value.x, value.y}, {value.width, value.height}};
 }
 
 vk::FrontFace PipelineInfo::vulkanFrontFace(const gfx::CullFaceWindingType& value) {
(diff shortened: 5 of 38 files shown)


src/mbgl/vulkan/pipeline.cpp
```diff
@@ -400,10 +406,8 @@ std::size_t PipelineInfo::hash() const {
                       stencilDepthFail,
                       wideLines,
                       VkRenderPass(renderPass),
-#if !USE_DYNAMIC_VIEWPORT
                       viewExtent.width,
                       viewExtent.height,
```
adrian-cojocaru:
```suggestion

```
Since the viewport is now dynamic, these values don't need to be hashed anymore (they don't represent a change to the pipeline object)

include/mbgl/map/map_options.hpp
```diff
@@ -133,6 +133,22 @@ class MapOptions final {
...
     float pixelRatio() const;
 
+    /**
+     * @brief Sets the frustum offset. Used to avoid rendering edges of the screen
+     * which might be behind UI elements.
+     *
+     * @param offset_ A set of edge insets in logical pixels.
+     * @return reference to MapOptions for chaining options together.
+     */
+    MapOptions& withFrustumOffset(const EdgeInsets& offset_);
```
adrian-cojocaru: Using `MapOptions::withFrustumOffset` for static setup sounds like a worse option than resizing the draw surface to match the content. I think `Map::setFrustumOffset` handles all use cases and encourages dynamic updates (when needed).

adrian-cojocaru (approved):

include/mbgl/map/map_options.hpp
louwers: But what if you need to draw to a larger area layer?

adrian-cojocaru: In that case I'd think that most people would solve it via composition instead of searching for this specific option. If `withFrustumOffset` is used the color and depth attachments are still allocated to the full size. My point is it's not the most intuitive option for static layouts and if you still need it you can use `Map::setFrustumOffset`.

src/mbgl/vulkan/pipeline.cpp
yousifd: https://github.com/maplibre/maplibre-native/pull/3676/commits/87176e9237f8eba518df884fc6d83ac8888a88b3

include/mbgl/map/map_options.hpp
yousifd: https://github.com/maplibre/maplibre-native/pull/3676/commits/87176e9237f8eba518df884fc6d83ac8888a88b3

platform/ios/src/MLNMapView.h
```diff
@@ -508,6 +508,16 @@ MLN_EXPORT
  */
 @property (nonatomic, assign) double tileLodZoomShift;
 
+/**
+ Frustum offset used to disable rendering of elements at the edge of the screen
+
+ Offset applied to camera frustum and scissor rectangle. The camrea frustum is modofied
```
louwers:
```suggestion
 Offset applied to camera frustum and scissor rectangle. The camrea frustum is modified
```

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -117,6 +117,13 @@ public void enableRenderingStatsView(boolean value) {
     nativeMapView.enableRenderingStatsView(value);
   }
 
+  /**
+   * Set frustum offset to disable rendering for edges of the screen
```
louwers: Maybe you could make the documentation comments similar for iOS and Android.

src/mbgl/gfx/scissor_rect.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <cstdint>
+
+namespace mbgl {
+namespace gfx {
+
+class ScissorRect {
+public:
+    int32_t x, y;
```
louwers: Please put these on their own lines.

src/mbgl/gfx/scissor_rect.hpp
```diff
@@ -0,0 +1,21 @@
...
+    uint32_t width, height;
+
+    bool operator==(const ScissorRect& other) const {
```
louwers: Since we are using C++20 I think we can use `= default` now for both `operator==` and `operator!=`. https://en.cppreference.com/w/cpp/language/default_comparisons.html

src/mbgl/gfx/scissor_rect.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include <cstdint>
+
+namespace mbgl {
+namespace gfx {
+
+class ScissorRect {
+public:
+    int32_t x, y;
+    uint32_t width, height;
```
louwers: Please put on individual lines.

louwers (approved):
Please consider adding some kind of test coverage, if only for the setters/getters of the platforms.

Android Build needs to be fixed.

yousifd:
> Please consider adding some kind of test coverage, if only for the setters/getters of the platforms.

@louwers Can you share an example for coverage tests for both android and iOS?

platform/ios/src/MLNMapView.h
yousifd: 82ee2461fae457823fde3b90644f416fa53c69a4

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
yousifd: 82ee2461fae457823fde3b90644f416fa53c69a4

src/mbgl/gfx/scissor_rect.hpp
yousifd: 16db56ec12c0f9897993ea01b4cebffdd243f6dd

src/mbgl/gfx/scissor_rect.hpp
yousifd: 16db56ec12c0f9897993ea01b4cebffdd243f6dd

src/mbgl/gfx/scissor_rect.hpp
yousifd: 16db56ec12c0f9897993ea01b4cebffdd243f6dd

louwers:
@yousifd See https://github.com/HudHud-Maps/maplibre-native/pull/68

yousifd:
> @yousifd See [HudHud-Maps#68](https://github.com/HudHud-Maps/maplibre-native/pull/68)

I was referring to another example to base my tests on but I appreciate that you wrote the tests. Will merge them into our fork and pull the commits back here

louwers:
@yousifd I think if you merge that PR they should appear here.

yousifd:
@ntadej Hey, @louwers told me to reach out to you. I'm not sure why the linux qt build is failing after this change and I can't get the linux qt build to work locally using docker. Do you have any idea what the issue might be? or a way to test this locally to attempt to debug it?

ntadej:
I think you added a platform-specific function that is now not implemented for the Qt backend. The error is that a pure virtual function was called.

I can look through the changes later when I'm at the computer. In the meantime if you look at what you did for other platforms, you can replicate it also for Qt.

yousifd:
@ntadej I didn't add any virtual method in this PR. I added methods to specific rendering backends and not to a specific platform.

Also running the test locally in a docker container running macos doesn't result in any issues 

```
ubuntu@a9501f014335:/app/build$ xvfb-run ./mbgl-test-runner --gtest_filter="CustomDrawableLayer.SymbolIcon"
Detected locale "C" with character encoding "ANSI_X3.4-1968", which is not UTF-8.
Qt depends on a UTF-8 locale, and has switched to "C.UTF-8" instead.
If this causes problems, reconfigure your locale. See the locale(1) manual
for more information.
Note: Google Test filter = CustomDrawableLayer.SymbolIcon
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from CustomDrawableLayer
[ RUN      ] CustomDrawableLayer.SymbolIcon
[ INFO ]  "{unknown}[General]: GPU Identifier: llvmpipe (LLVM 20.1.2, 128 bits)"
[       OK ] CustomDrawableLayer.SymbolIcon (134 ms)
[----------] 1 test from CustomDrawableLayer (134 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (134 ms total)
[  PASSED  ] 1 test.
```

louwers:
Merging since CI seems to pass now.

PR #3680, opened 2025-07-29
Add rendering stats tracker example

adrian-cojocaru:
This PR introduces an example class designed to track and report rendering statistics based on `RenderingStats` and offers:
- periodic reports of minimum, maximum, and average values
- callbacks when predefined thresholds are exceeded


louwers (commented):
Do you want to add it to the example docs?

PR #3681, opened 2025-07-29
Fix Android backend cleanup

adrian-cojocaru:
Currently the `AndroidGLRendererBackend` is deleted when the `MapRenderer` object is garbage collected. This forces OpenGL resource cleanup on the finalizer thread after the render thread and it's context are deleted. 
Fixed by explicitly deleting the backend before the rendering surface is destroyed.


louwers (approved):

sjg-wdw:
Do we think this could cause trouble?
I'm getting a request to merge it.

adrian-cojocaru:
The cleanup itself is not likely to cause any problems and if the renderer was reset at that point I think it's safe to assume the backend isn't needed.

sjg-wdw:
@louwers are we okay to merge this?

Issue #3687, opened 2025-07-30
Windows arm64 presets


birkskyum:
We don't have presets yet for windows arm, despite a growing number of  devices having that arch.

In addition, there is now availability of windows-11-arm (arm64) runners for public repos, so when the capacity is sufficient we can expand the CI suite matix.

https://github.com/orgs/community/discussions/155713

birkskyum:
Visual Studio 2022 has stable Arm64 support - https://learn.microsoft.com/en-us/visualstudio/install/visual-studio-on-arm-devices?view=vs-2022

MSYS2 has preliminary ARM64 support - https://www.msys2.org/wiki/arm64/

birkskyum:
### OpenGL

https://github.com/user-attachments/assets/625f1590-92af-4958-9610-990d443fa029

I've built the opengl glfw example file with Visual Studio 2022, following [this guide](https://maplibre.org/maplibre-native/docs/book/platforms/windows/build-msvc.html), on windows Arm

Running the tests, all 997 / 1001 tests passed, with the failing being:
- TileLOD.disabled
- TileLOD.pitchThreshold
- TileLOD.scale
- TileLOD.radius

### Vulkan

GLFW builds

  - 965 tests passed out of 966 total tests
  - Only 1 test failed: Map.SetStyleInvalidJSON
  - 4 tests are disabled
  - The test suite ran successfully despite some "Vulkan layers not found" warnings (these are
   likely validation layer warnings that don't affect the actual functionality)

sargunv:
Did you build with the `opengl` or `egl` preset on ARM? I'm finding that WGL headers aren't found on Windows ARM (works on x64). 

EDIT: In https://github.com/sargunv/maplibre-native-jvm-demo I was able to build for Windows ARM with EGL and Vulkan.
* Vulkan works well (just the same rendering bugs as on other platforms).
* EGL renders the map well for a second or so, but then crashes afterwards.
  * It's possible I'm doing something wrong, but the same code does work properly on the other platforms.

sargunv:
Ah, I see the "Copy OpenGL Headers" step in your PR. Will try that in a bit!

Issue #3670, opened 2025-07-28
windows-build-and-test-msys2 workflow failing (duplicate section)


louwers:
https://github.com/maplibre/maplibre-native/actions/runs/16532952609/job/46762266297

FAILED: [code=1] mbgl-render-test-runner.exe 
[...]

PR #3692, opened 2025-07-31
Windows Arm64 presets 

birkskyum:
Towards
- #3687 

This adds presets for windows-arm64-opengl and windows-arm64-vulkan, as well as some docs on how to build for arm

I had to jump through some hoops which i hope won't be necessary, like manually installing some packages with vcpkg.

This PR is to verify that the windows x86 targets still work after the changes, and to see if it's possible to add a windows arm64 ci test


acalcutt:
I am guessing if we wanted to support windows arm in node we would need something in https://github.com/maplibre/maplibre-native/blob/main/platform/node/cmake/module.cmake#L123-L128 , but @tdcosta100 might be more familiar with that since he added the windows support there.

birkskyum:
The CI here breaks because of:
- https://github.com/maplibre/maplibre-native/issues/3670

birkskyum:
Superseded by
- #3760

PR #3697, opened 2025-08-01
Benchmark cpu/memory usage

adrian-cojocaru:
Added a few extra metrics to the benchmarking runs


platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/benchmark/BenchmarkActivity.kt
```diff
@@ -88,6 +89,7 @@ suspend fun MapView.setStyleSuspend(styleUrl: String): Unit =
  */
 class BenchmarkActivity : AppCompatActivity() {
     private val TAG = "BenchmarkActivity"
+    private val useAdvancedMetrics = false
```
louwers, line 53: Is there any downside to enabling it by default?

adrian-cojocaru: I was getting some error logs when calling `Debug.getMemoryInfo` that caused the fps to take a hit. I'm not sure if it's throttling the access, but it wasn't consistent on all test devices/emulators.

sjg-wdw:
Recommendation from meeting:  Throttle back the memory check and merge.

louwers (approved):

sjg-wdw:
Can we just merge this?

Issue #3685, opened 2025-07-30
Be less restrictive in libpng version check or statically include image libraries into the amalgamation


CommanderStorm:
**Is your feature request related to a problem? Please describe.**

When I try to get the opengl amalgamation running using https://github.com/maplibre/maplibre-native-rs/ on manjaro, I get this error

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  libpng version mismatch: headers report 1.6.43, but library reports 1.6.50
```

This comes from the following code:

https://github.com/maplibre/maplibre-native/blob/a6d7314538e9f5ae372785be7f63eb147938881e/platform/default/src/mbgl/util/png_reader.cpp#L20-L33

**Describe the solution you'd like**

There are two approaches that are viable:

- making the libpng release version check less strict `(version / 100) != (version PNG_LIBPNG_VER / 100)` 
- statically include image libraries into the amalgamation

I really know far too little if the first version would be save for other users.
I think the second option is a better bet, but also far more effort.

If the first option is acceptable, I can do a PR. For the second option, I am a bit out of my depth.

**Describe alternatives you've considered**

Telling our (potential) users to just use the one version we have in CI, but this is rather hard to do given that his might shift and is not really pheasable

louwers:
They need to match. I will include it.

PR #3699, opened 2025-08-01
Add static libpng to Linux amalgamation

louwers:
Closes https://github.com/maplibre/maplibre-native/issues/3685

cc @CommanderStorm

### DIFF
--- a/platform/linux/linux.cmake
+++ b/platform/linux/linux.cmake
@@ -184,6 +184,9 @@ target_link_libraries(
 # Bundle system provided libraries
 if(MLN_CORE_INCLUDE_DEPS AND NOT MLN_USE_BUILTIN_ICU AND NOT "${ARMERGE}" STREQUAL "ARMERGE-NOTFOUND")
     message(STATUS "Found armerge: ${ARMERGE}")
+    include(${CMAKE_CURRENT_LIST_DIR}/cmake/find_static_library.cmake)
+    find_static_library(PNG_STATIC_LIB NAMES png)
+
     add_custom_command(
         TARGET mbgl-core
         POST_BUILD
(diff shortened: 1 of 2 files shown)


platform/linux/linux.cmake
```diff
@@ -184,6 +184,9 @@ target_link_libraries(
 # Bundle system provided libraries
 if(MLN_CORE_INCLUDE_DEPS AND NOT MLN_USE_BUILTIN_ICU AND NOT "${ARMERGE}" STREQUAL "ARMERGE-NOTFOUND")
     message(STATUS "Found armerge: ${ARMERGE}")
+    include(${CMAKE_CURRENT_LIST_DIR}/cmake/find_static_library.cmake)
+    find_static_library(PNG_STATIC_LIB NAMES png)
```
CommanderStorm, line 188: could we add the other image libraries as well?

- libjpeg
- libwebp

If possible via this route (this has not caused an issue for now => I would be happy to go at this one issue at a time if you would be okay with revisiting this 😅):

- libz
- libcurl4
- libuv1

cgalvan (approved):

platform/linux/linux.cmake
louwers: Doing them in one go sounds better.

I hope putting OpenSSL and SQLite in the static build won't cause any issues? Might cause symbol conflicts.

CommanderStorm: If you think that they are risky, we can also move them to when we might need them.

CommanderStorm: actually, sqlite and openssl are not even public symbols 🎉

louwers: added the whole bunch, let's see

PR #3700, opened 2025-08-02
chore(ci): Remove sqlite from the amalgamation

CommanderStorm:
As can be seen in https://github.com/maplibre/maplibre-native/actions/runs/16686869302/job/47237968305 this causes issues.
We are it seems already merging it in via armerge (?)

Anyhow, since maplibre-native-core-amalgam builds work without this, we should not need this step.
Not sure if opensssl is needed here or handled via armerge (?) 🤔


louwers:
Ah this is because we are already vendoring SQLite.

louwers:
Let me make sure the build is run on CI.

PR #3675, opened 2025-07-29
Fix dashed line issue when style change

alexcristici:

PR #3705, opened 2025-08-06
Do not run notify-release-on-prs.ts for forks

louwers:


louwers:
The script is causing comments from forks 😄 https://github.com/maplibre/maplibre-native/pull/3675#issuecomment-3161431914

HarelM (approved):

PR #3706, opened 2025-08-07
Improve Core Builds for Linux

louwers:
- Include freetype, harfbuzz, color parser and crypto in amalgamation.
- Change `find_static_library` function to extend a list to avoid repetition.
- Remove core CMake presets, always build amalgamation if armerge is installed.

### DIFF
--- a/platform/linux/linux.cmake
+++ b/platform/linux/linux.cmake
@@ -175,41 +175,40 @@ target_link_libraries(
         ${WEBP_LIBRARIES}
         $<$<NOT:$<BOOL:${MLN_USE_BUILTIN_ICU}>>:${ICUUC_LIBRARIES}>
         $<$<NOT:$<BOOL:${MLN_USE_BUILTIN_ICU}>>:${ICUI18N_LIBRARIES}>
-        $<$<BOOL:${MLN_USE_BUILTIN_ICU}>:$<IF:$<BOOL:${MLN_CORE_INCLUDE_DEPS}>,$<TARGET_OBJECTS:mbgl-vendor-icu>,mbgl-vendor-icu>>
+        $<$<BOOL:${MLN_USE_BUILTIN_ICU}>:mbgl-vendor-icu>
         PNG::PNG
         mbgl-vendor-nunicode
         mbgl-vendor-sqlite
 )
 
 # Bundle system provided libraries
-if(MLN_CORE_INCLUDE_DEPS AND NOT MLN_USE_BUILTIN_ICU AND NOT "${ARMERGE}" STREQUAL "ARMERGE-NOTFOUND")
+if(NOT MLN_USE_BUILTIN_ICU AND NOT "${ARMERGE}" STREQUAL "ARMERGE-NOTFOUND")
     message(STATUS "Found armerge: ${ARMERGE}")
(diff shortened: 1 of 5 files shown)


acalcutt:
Do you think this will package these files into the linux node binary now? I noticed when testing the new node package on macos I had to add harffbuzz with brew to have the needed lib so i was wondering if it would be the same in linux.

louwers:
> Do you think this will package these files into the linux node binary now?

We are creating a special static library which contains all dependencies as statically (an amalgamation). This is used by maplibre-native-rs so people using that library do not have special system dependencies. This is common in the Rust ecosystem.

platform/linux/linux.cmake
```diff
@@ -175,41 +175,40 @@ target_link_libraries(
...
+
+    find_static_library(STATIC_LIBS NAMES png)
+    find_static_library(STATIC_LIBS NAMES z)
+    find_static_library(STATIC_LIBS NAMES jpeg)
+    find_static_library(STATIC_LIBS NAMES webp)
+    find_static_library(STATIC_LIBS NAMES curl)
+    find_static_library(STATIC_LIBS NAMES uv uv_a)
+    find_static_library(STATIC_LIBS NAMES ssl)
+    find_static_library(STATIC_LIBS NAMES crypto)
+    find_static_library(STATIC_LIBS NAMES sqlite3)
```
CommanderStorm: 🤔 is keeping sqlite here correct, or would this cause the same issue as last time?

CommanderStorm (commented):
Looks good to me, only one question

platform/linux/linux.cmake
louwers: Yes, because I am not using the vendoring SQLite anymore!

platform/linux/linux.cmake
```diff
@@ -175,41 +175,40 @@ target_link_libraries(
         ${WEBP_LIBRARIES}
         $<$<NOT:$<BOOL:${MLN_USE_BUILTIN_ICU}>>:${ICUUC_LIBRARIES}>
         $<$<NOT:$<BOOL:${MLN_USE_BUILTIN_ICU}>>:${ICUI18N_LIBRARIES}>
-        $<$<BOOL:${MLN_USE_BUILTIN_ICU}>:$<IF:$<BOOL:${MLN_CORE_INCLUDE_DEPS}>,$<TARGET_OBJECTS:mbgl-vendor-icu>,mbgl-vendor-icu>>
+        $<$<BOOL:${MLN_USE_BUILTIN_ICU}>:mbgl-vendor-icu>
         PNG::PNG
         mbgl-vendor-nunicode
         mbgl-vendor-sqlite
```
mwilsnd, line 181: Should this get removed since you're merging the system library now?

platform/linux/linux.cmake
louwers: fixed

platform/linux/linux.cmake
louwers: `MLN_USE_BUILTIN_ICU` is `OFF` by default.

If you set `MLN_USE_BUILTIN_ICU` to `ON` it is not making the amalgamation, so I think it's correct this way!

PR #3709, opened 2025-08-08
Cancel pending style request when loading style JSON

### LINKED ISSUE #3708: [iOS] Race condition when initializing map and immediately setting styleJSON
### MapLibre iOS Version

6.18.0

### iOS Version

18.5

### Device

iPhone 13 Pro

### What happened?

I'm trying to write a PR for https://github.com/maplibre/flutter-maplibre-gl to add support for passing raw style JSON on iOS. flutter-maplibre-gl has the somewhat odd behavior of initializing `MLNMapView` with no style and then setting the map style immediately afterwards.

When setting `mapView.styleURL` in this way everything works normally. **However setting `mapView.styleJSON` seemingly has no effect, and the demo style is displayed instead.** After the initial load setting `styleJSON` works just fine.

I suspect that setting `styleJSON` isn't properly cancelling the loading of the demo style URL, causing it to be immediately overwritten when the demo style loads.

It's likely as simple as adding `self.mbglMap.getStyle().loadURL("");` if `styleURL` is set in setStyleJSON. At the moment I don't have the time to run a test build to verify this.

https://github.com/maplibre/maplibre-native/blob/16be52ed55e79ee453b0f2fc734eef2038a405da/platform/ios/src/MLNMapView.mm#L643-L658

### Steps to reproduce

1. Initialize a MLNMapView with `mapView = MLNMapView(frame: frame)`
2. Immediately set `mapView.styleJSON` to your favorite map style string

## Expected behavior

The user-supplied map style is loaded and displayed

## Actual behavior

The default demo style is shown

louwers:
https://github.com/maplibre/maplibre-native/pull/3240 added a new method to load a style JSON from string. However when a map is loading a style asynchronously, this can lead to a race condition where the style set from JSON is overwritten by the incoming style response.

Includes a regression test.

Closes #3708

### DIFF
--- a/test/src/mbgl/test/delayed_file_source.hpp
+++ b/test/src/mbgl/test/delayed_file_source.hpp
@@ -10,7 +10,7 @@
 namespace mbgl {
 
 /*
-  This is a file source that allows control when a request is responded to.
+  This is a file source that allows control of when a request is responded to.
 */
 class DelayedFileSource : public StubFileSource {
 public:


include/mbgl/style/style.hpp
```diff
@@ -71,6 +71,9 @@ class Style {
     void addLayer(std::unique_ptr<Layer>, const std::optional<std::string>& beforeLayerID = std::nullopt);
     std::unique_ptr<Layer> removeLayer(const std::string& layerID);
 
+    // Style loading control
```
louwers:
```suggestion
```

test/src/mbgl/test/delayed_file_source.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+/*
+  This is a file source that allows control when a request is responded to.
+*/
+class DelayedFileSource : public StubFileSource {
+public:
+    DelayedFileSource();
+    ~DelayedFileSource() override = default;
+
+    // Override from FileSource
```
louwers:
```suggestion
```

test/src/mbgl/test/delayed_file_source.hpp
```diff
@@ -0,0 +1,43 @@
...
+  This is a file source that allows control when a request is responded to.
+*/
+class DelayedFileSource : public StubFileSource {
+public:
+    DelayedFileSource();
+    ~DelayedFileSource() override = default;
+
+    std::unique_ptr<AsyncRequest> request(const Resource&, Callback) override;
+
+    // Custom method to respond to pending requests
```
louwers:
```suggestion
    // Custom method to respond to pending request
```

test/src/mbgl/test/delayed_file_source.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <mbgl/util/async_request.hpp>
+#include <mbgl/test/stub_file_source.hpp>
+#include <memory>
+#include <string>
+#include <optional>
+
+namespace mbgl {
+
+/*
+  This is a file source that allows control when a request is responded to.
```
TimSylvester:
```suggestion
  This is a file source that allows control of when a request is responded to.
```

TimSylvester (approved):
That looks good.

Using the destructor to potentially do network stuff makes me nervous, but that's another matter.

cgalvan (approved):

louwers:
@TimSylvester Nice. That destructor deletes the `FileSourceRequest`, which calls the cancellation callback. I don't think that actually cancels in-flight requests so it's not really doing network stuff per se. It also deletes the mailbox. So if there's a network request coming in, it will just send a message to a dead mailbox.

PR #3715, opened 2025-08-12
Use free runners Node.js CI

louwers:


acalcutt:
Seems like jwlawson/actions-setup-cmake  isn't working on arm linux yet https://github.com/jwlawson/actions-setup-cmake/issues/79

I have been using macos-13 on my test branch for macos-arm, so I think that shoud be fine. we moved away from it because it seemed at the time macos-14 fixed the freezing issue, but we later found that it didn't actually do that and that should actually be fixed now.

acalcutt:
Seems fine to me, the only thing I might try is removing 'jwlawson/actions-setup-cmake' all together and see if the built in cmake works now on the other workflows. we had only added that because there was a cmake release that was giving us issues, but if the  arm ubuntu workflow works without it maybe it is no longer needed at all.

louwers:
@acalcutt The GitHub runners come with pretty recent versions of CMake pre-installed.

acalcutt:
> @acalcutt The GitHub runners come with pretty recent versions of CMake pre-installed.

the issue before was the newest version was not working at one point, and we added that action in to downgrade cmake as a workaround. but it seems like that might not be needed anymore if the latest cmake is working in the arm linux runner. we can change that later though.

PR #3717, opened 2025-08-14
[cmake] Declare renderer backend as public

ntadej:
Declare renderer backend as public so clients don't have to do it again. Probably some includes/link libraries can also be added as public, but maybe this will come later.


ntadej:
Hmm, everything else passes besides Windows. Is there a known issue?

louwers:
@ntadej Yes it's a known issue https://github.com/maplibre/maplibre-native/issues/3670

Should not be a blocker.

ntadej:
My initial testing shows that (in simple cases) this reduces the configuration needed so let's merge it.

PR #3726, opened 2025-08-15
Add weak pointer management to RasterSource and derived classes

louwers:
WeakPointerFactory data members need to be last so they destroyed first. It is problematic when a base class uses it, because the derived class' data members will be destroyed first. So here I am moving it up the hierarchy, removing it from `TileSource`. Since `RasterDEM` cannot be made into an abstract class, invalidate weak pointers in its derived classes' destructors.

Issue identified by @TimSylvester introduced by #3570


TimSylvester:
That seems like it should do the trick.

Searching for `WeakPtrFactory<`, it looks like it also needs to be handled in the `style::Layer` and `ThreadedScheduler` trees.  @sjg-wdw These might explain a few of those rare crashes we've seen.

For those, as well as ones where it's not a problem like `AnnotationManager` and `SpriteLoader`, I think the most-derived class should be marked final, with a comment reminding us that any derived class needs to take care of this.  Or perhaps a preprocessor synonym for `final` specifically for that purpose.

TimSylvester (approved):

PR #3727, opened 2025-08-16
[metal] OffscreenTextureResource: fix compilation warnings

ntadej:
Fix unused variable compilation warning in `mbgl::mtl::OffscreenTextureResource`.


ntadej:
> I think [maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html) would work here.

Thanks for the tip, done.

louwers (approved):

PR #3728, opened 2025-08-17
Backport to android-10.x.x: Compile for Android with support for 16 KB page sizes

johan12345:
Equivalent to #2852, but for NDK 25 (see https://developer.android.com/guide/practices/page-sizes#cmake).

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/cpp/CMakeLists.txt
+++ b/platform/android/MapboxGLAndroidSDK/src/cpp/CMakeLists.txt
@@ -216,4 +216,7 @@ target_link_libraries(mapbox-gl
                               mbgl-core
                               mbgl-vendor-unique_resource)
 
+target_link_options(mapbox-gl
+                    PRIVATE "-Wl,-z,max-page-size=16384")
+
 install(TARGETS mapbox-gl LIBRARY DESTINATION lib)


louwers:
I guess this is the best way without updating NDK?

johan12345:
yeah... Google's docs say they still recommend to upgrade NDK to 27 or higher because the included `libc++_shared.so` may not be 16KB aligned. But this should not affect our case, as we are using `ANDROID_STL=c++_static`.

louwers:
OK let's try it. Would be great if you could report back once the release is out.

louwers:
I don't think this did the trick because I still get the warning that it is not aligned.

Edit: never mind, that is just the custom layer in the example app.

image

johan12345:
ah right - for it to apply to that as well, it might have to go into `platform/android/android.cmake` instead of `platform/android/MapboxGLAndroidSDK/src/cpp/CMakeLists.txt`? But yeah, it matters just for the example, not for using it in other apps.

michael-winkler:
Hello,

why is the new release android-v10.3.5 ? The release before was [android-v11.13.0](https://github.com/maplibre/maplibre-native/releases/tag/android-v11.13.0)

So why the version downgrade? @johan12345 @louwers

johan12345:
@michael-winkler As the title says, this backports a change to the 10.x.x branch for those who can't upgrade to 11.x.x yet.

louwers:
@johan12345 Did you try out the release? Curious if it works well for you.

johan12345:
@louwers yes, works perfectly! Android Studio doesn't complain anymore and it also runs fine on the 16 KB page size emulator.

louwers:
Great

PR #3730, opened 2025-08-18
feat: add `onSingleTap` function support for Android Auto

thanhdang198:
This pull request introduces a new method for handling single tap events on the map and adds comprehensive unit tests to ensure correct behavior. The main focus is on improving marker selection/deselection logic and interaction with gesture management when a tap occurs.

**Map tap event handling:**

* Added a new `onSingleTap(float x, float y)` method to `MapLibreMap`, which processes single tap events by delegating to the annotation manager.

**Unit tests for tap handling:**

* Added two tests in `MapLibreMapTest.kt`:
  * `testOnSingleTapHandledByAnnotationManager` verifies that taps handled by the annotation manager do not trigger marker deselection. 
  * `testOnSingleTapNotHandled_DeselectsAndNotifies` verifies that if a tap is not handled, markers are deselected and the gestures manager is notified.

**Reference**
- Based on [`MapGestureDetector.java` in MapLibre](https://github.com/maplibre/maplibre-native/blob/main/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapGestureDetector.java#L359)
- Issue: [maplibre/MapLibre-Android-Auto-Sample#8](https://github.com/maplibre/MapLibre-Android-Auto-Sample/issues/8)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -58,2640 +58,2669 @@
 @UiThread
 public final class MapLibreMap {
 
-  private static final String TAG = "Mbgl-MapLibreMap";
-
-  private final NativeMap nativeMapView;
```
louwers, line 63: Please remove whitespace-only changes.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -58,2640 +58,2669 @@
...
-  public void resetNorth() {
-    notifyDeveloperAnimationListeners();
-    transform.resetNorth();
+   * Handles a single tap event on the map.
+   * <p>
+   * If the tap is not handled by the annotation manager, and the UI settings specify to deselect markers on tap,
+   * this method will deselect any selected markers. It then notifies all threads waiting on the gestures manager.
+   * </p>
+   * <p>
+   * Note: Refer to MapGestureDetector.StandardGestureListener#onSingleTapConfirmed (around line 359)
```
louwers: Adding line numbers to comments is not a good idea because they shift.

thanhdang198: I'll remove it @@

louwers (commented):
I don't really understand this pull request.

> This pull request introduces a new method for handling single tap events 

What I see is a method `onSingleTap(float x, float y)` on `MapLibreMap`. You would not override that class, so it's not a method for **handling** single tap events. It is a method to **simulate** single tap events, correct?

thanhdang198:
> I don't really understand this pull request.
> 
> > This pull request introduces a new method for handling single tap events
> 
> What I see is a method `onSingleTap(float x, float y)` on `MapLibreMap`. You would not override that class, so it's not a method for **handling** single tap events. It is a method to **simulate** single tap events, correct?

On Android Auto, we can’t add a listener for `onMapClick` because the map is only rendered as a bitmap and displayed on the AA surface. With this function, we can notify the `MapView` when the user clicks on the AA surface, and the map controller will take care of the rest of the logic—such as triggering the onMapClick callback (returning latitude and longitude without calling queryFeature), handling annotation gestures, and more.

```kt
        override fun onClick(x: Float, y: Float) {
            mapLibreMap?.onSingleTap(x, y)
        }
```

thanhdang198:
This func meaning:
Hey MapView, the surface was clicked at position (x, y). Please handle all annotations for me.

louwers:
It does not make sense to call such a method `onSingleTap`, because `MapLibreMap` is not a listener.

The method you added has almost the same contents as `StandardGestureListener.onSingleTapConfirmed`. Can't you call that method directly to simulate a click?

thanhdang198:
> It does not make sense to call such a method `onSingleTap`, because `MapLibreMap` is not a listener.
> 
> The method you added has almost the same contents as `StandardGestureListener.onSingleTapConfirmed`. Can't you call that method directly to simulate a click?

No, I couldn’t find a way to directly call this function. I believe it was created to be handled from the Android surface as well. This logic introduces fewer changes and has minimal impact on the current flow, since this code block is only used by Android Auto and does not affect anything else

thanhdang198:
@louwers Could you let us know if any changes are needed before we merge this PR?

louwers:
- It should be called something like `simulateSingleTap`. Not `onSingleTap`.
- I am not sure if MapLibreMap is the right place for it.
- If possible we should try to deduplicate the logic between this method and `StandardGestureListener.onSingleTapConfirmed`.

@frankkienl What do you think?

frankkienl:
the "onX" naming convention is very normal for Android; 
I don't see any problems with readability and understandability 👌 
Calling something `testOnSingleTap` instead `simulateSingleTap` would make perfect sense to me;
As you are testing the `onSingleTap` function here.
Not sure what naming conventions exist outside of Android, or inside of MapLibre. But I would approve this.

frankkienl:
"I am not sure if MapLibreMap is the right place for it." 
Any suggestions? Otherwise, it's probably fine.

frankkienl:
"If possible we should try to deduplicate the logic between this method and StandardGestureListener.onSingleTapConfirmed."
Sounds like a good point; I'm not familiar with the StandardGestureListener, but reducing code duplication sounds good.

thanhdang198:
> "If possible we should try to deduplicate the logic between this method and StandardGestureListener.onSingleTapConfirmed." Sounds like a good point; I'm not familiar with the StandardGestureListener, but reducing code duplication sounds good.

Thanks for reply. Cause I'm not familiar with java or kotlin native, so I think keep it independent of anything else is better than change current logic of MaplibreMap and other class @@

Issue #3733, opened 2025-08-18
Linux fails to build with Vulkan back end


jwinarske:
**Describe the bug**

Build config
```
cmake -GNinja -S /mnt/raid10/workspace-automation/app/maplibre-native -B /mnt/raid10/workspace-automation/app/maplibre-native/build -DBUILD_SHARED_LIBS=OFF -DMLN_WITH_WERROR=OFF -DMLN_WITH_X11=OFF -DMLN_WITH_WAYLAND=ON -DMLN_WITH_EGL=OFF -DMLN_WITH_OPENGL=OFF -DMLN_WITH_VULKAN=ON -DMLN_WITH_GLFW=OFF -DMLN_DRAWABLE_RENDERER=ON -DMLN_LEGACY_RENDERER=OFF -DMLN_WITH_PMTILES=ON -DCMAKE_STAGING_PREFIX=/mnt/raid10/workspace-automation/app/maplibre-native/staging
```

Build output
```
-- The CXX compiler identification is Clang 18.1.8
-- The C compiler identification is Clang 18.1.8
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib64/llvm18/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib64/llvm18/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Configuring Vulkan renderer backend
CMake Deprecation Warning at vendor/maplibre-native-base/CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found CURL: /usr/lib64/libcurl.so (found version "8.11.1")
-- Found ICU: /usr/include (found version "76.1") found components: i18n
-- Found ICU: /usr/include (found version "76.1") found components: uc
[...]

PR #3734, opened 2025-08-18
Vulkan backend fails to build on Linux (Fedora 42)

jwinarske:
resolves https://github.com/maplibre/maplibre-native/issues/3733

### DIFF
--- a/platform/linux/linux.cmake
+++ b/platform/linux/linux.cmake
@@ -16,7 +16,7 @@ pkg_search_module(ICUUC icu-uc)
 pkg_search_module(ICUI18N icu-i18n)
 find_program(ARMERGE NAMES armerge)
 
-if(MLN_WITH_WAYLAND)
+if(MLN_WITH_WAYLAND AND NOT MLN_WITH_VULKAN)
     # See https://github.com/maplibre/maplibre-native/pull/2022
 
     # MLN_WITH_EGL needs to be set for Wayland, otherwise this CMakeLists will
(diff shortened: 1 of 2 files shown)


platform/linux/linux.cmake
```diff
@@ -16,7 +16,7 @@ pkg_search_module(ICUUC icu-uc)
 pkg_search_module(ICUI18N icu-i18n)
 find_program(ARMERGE NAMES armerge)
 
-if(MLN_WITH_WAYLAND)
```
louwers, line 19: If it is really not supported at this point, please log a `message(FATAL "Vulkan backend not supported with Wayland, set either MLN_WITH_VULKAN=OFF or MLN_WITH_WAYLAND=OFF)`

jwinarske: The logic here is to not implicitly set GLESv3/EGL if Wayland and Vulkan.

louwers: Ah, OK got it.

louwers (approved):

jwinarske:
@louwers Fix pushed for Windows.  CI needs approval to run.

jwinarske:
@louwers just pushed fix for MinGW; ready to run.  I don't see any details about Android tests.  Not sure what's up with those.  Do have more details on these?

louwers:
The instrumented Android tests fail when there are no available devices. We're trying to get a private device. Not related to this PR!

jwinarske:
@louwers `AND` changed to `OR` (sigh), please re-run

louwers:
I'll send you an invite for the repo.

Issue #3704, opened 2025-08-06
Z-fighting or flickering on Vulkan


sargunv:
**Describe the bug**

When zoomed in with the OpenFreeMap Bright style, I see what looks like z-fighting on the building polygons.

**To Reproduce**

1. Load a map with style: https://tiles.openfreemap.org/styles/bright
2. Zoom into a city
3. Rotate the camera to exacerbate the effect

**Expected behavior**

The map should render as nicely as it does with OpenGL or Metal.

**Screenshots**

https://github.com/user-attachments/assets/4ae1193e-5428-4fa0-b30b-f13f58ec0bb2

**Platform information (please complete the following information):**

 - Operating System: Windows 11 (ARM and x64), Fedora 42 (ARM and x64), macOS 26 (ARM)
 - Platform (e.g. Node.js, Qt): JVM via https://github.com/sargunv/maplibre-native-jvm-demo
 - Version: https://github.com/maplibre/maplibre-native/commit/ca5f79c

**Additional info**

I saw https://github.com/maplibre/maplibre-native/issues/3490 but this looks like a different kind of flickering.

louwers:
https://github.com/user-attachments/assets/2eb42818-8ea3-4dfc-ada1-cc64a3eb6fff

Confirmed on Android as well, @adrian-cojocaru could you have a look?

adrian-cojocaru:
I'm seeing the same issue on macOS GLFW + Metal, which utilizes 32 bit depth (compared to 24 bit used by OpenGL/Vulkan). Still checking the exact difference, but playing around with [PaintParameters::depthEpsilon](https://github.com/maplibre/maplibre-native/blob/16be52ed55e79ee453b0f2fc734eef2038a405da/src/mbgl/renderer/paint_parameters.hpp#L133) fixes it.

sargunv:
Oh interesting, I haven't seen the same on Metal:

https://github.com/user-attachments/assets/c923c7e6-260f-4af6-90a4-827f1c505eb0

louwers:
I think this is fixed.

PR #3738, opened 2025-08-19
Tweak layer depth distribution

adrian-cojocaru:
This fixes the z-fighting issues described in https://github.com/maplibre/maplibre-native/issues/3704.


TimSylvester:
Does that depend on the total number of layers?  I suspect we don't have any tests with more than a few layers.

adrian-cojocaru:
@louwers I tried generating a test with close enough values to trigger the same issue, but it ended up with inconsistent results.
@TimSylvester The projection [tweak](https://github.com/maplibre/maplibre-native/blob/98d80825e926f7d7f49ae69671c5983bbaac584e/src/mbgl/renderer/layer_tweaker.cpp#L67) mirrors what's used for [OpenGL](https://github.com/maplibre/maplibre-native/blob/0bcc48f57900b23ea71b6bf185ee645e8057840d/src/mbgl/renderer/paint_parameters.cpp#L102) and both depend on the total number of layers.

louwers:
Trippy!

louwers:
INSTRUMENTATION_STATUS: test=launchActivity[102: Activity class org.maplibre.android.testapp.activity.options.MapOptionsXmlActivity]
INSTRUMENTATION_STATUS_CODE: 1
INSTRUMENTATION_RESULT: shortMsg=Process crashed.

Looks unrelated

PR #3741, opened 2025-08-20
Use `ctcache` to cache `clang-tidy` results

louwers:
This PR integrates a cache for clang-tidy results. clang-tidy is a major slowdown for the linux-ci workflow, a compiler cache does not speed things up at all. From my testing:

> 42 minutes 18 seconds with sccache (compiler cache, 100% hit rate).
> 2 minutes 50 seconds with sccache + ctcache :sparkles:

For future reference here is how the server is set up:

```
python -m venv venv
source venv/bin/activate
git clone https://github.com/matus-chochlik/ctcache.git
cd ctcache
pip install -r requirements.txt
CTCACHE_WEBROOT=$PWD python src/ctcache/clang_tidy_cache_server.py --auth-key-writes=<secret key>
```

The secret key is set as `CTCACHE_AUTH_KEY` in the repo and is used when building from this repo where that secret is available. ctcache did not have any authentication so I contributed this as a [PR](https://github.com/matus-chochlik/ctcache/pull/94), I hope it is accepted in some form. I will update the workflow when this change has been upstreamed.

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -128,6 +128,9 @@ jobs:
           if [ -n "${{ secrets.CTCACHE_AUTH_KEY }}" ]; then
             echo "setting CTCACHE_AUTH_KEY" 
             export CTCACHE_AUTH_KEY=${{ secrets.CTCACHE_AUTH_KEY }}
+          else
+            # do not write to cache if CTCACHE_AUTH_KEY not set
+            export CTCACHE_HOST_READ_ONLY=1
           fi
 
           cmake --preset linux-${{ matrix.variant.renderer }} \
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -91,9 +91,9 @@ jobs:
 
       - name: Install ctcache
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/linux-ci.yml
```diff
@@ -112,8 +120,22 @@ jobs:
             export SCCACHE_S3_NO_CREDENTIALS=1
           fi
 
+          # ctcache configuration
+          export CTCACHE_HOST=34.229.50.221
```
frankkienl, line 124: Maybe put the host ip-address also in a `secret` ? 
Not because we are not allowed to know, but to make configuration changes easier.

louwers: It is not a secret. Developers are welcome to use this on their own machines to speed up local builds as well.

louwers: I thought about it.

If I set this as a secret or a variable then PRs from forks will not have access to the server. That defeats the purpose of this since most PRs come from forks! 

If we use ctcache in another workflow in the furute I will factor this code in a reusable action.

mwilsnd (approved):

PR #3742, opened 2025-08-20
remove fetch-depth: 0 from workflows

louwers:
Confusingly, setting `fetch-depth: 0` fetches all history for all branches and tags...

https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4


.github/workflows/ios-release-cocoapods.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         working-directory: platform/ios
     runs-on: macos-14
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
```
HarelM, line 17: I honestly prefer the version numbers as the comment for `# v4` will stay even if it were to be undated to 5.
I honestly am not sure what's wrong with using a full version number like everywhere else when it comes to package management.

HarelM (approved):

PR #3743, opened 2025-08-20
Use sccache with S3 backend for Android CI

louwers:
Some optimizations to speed up Android CI.

Use sccache with S3 for all builds. Unfortunately the config needs to be copy pasted in all jobs/workflows, since reusable actions cannot modify environment variables.

Brings the total time of android-ci down from **1h 51m 42s** to **33m 41s** (with 100% cache hit rate).

Cache is updated when workflows run in the context of this repository (needs to authenticate with AWS, for that secrets are required).


louwers:
pre-commit.ci run

louwers:
pre-commit.ci run

PR #3750, opened 2025-08-24
Added information about maplibre-compose to readme

michalgwo:
I think it's worth mentioning that there is a maplibre-compose project in the README file, but feel free to reject it if it doesn't belong there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Add a `MapView` to your layout XML file:
 ```
 
 > [!TIP]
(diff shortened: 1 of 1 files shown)


louwers:
@michalgwo How about we add a new section? It's kind of a platform on its own. https://github.com/maplibre/maplibre-native/pull/3751

michalgwo:
Right, makes more sense to add a new section

PR #3752, opened 2025-08-24
Compatibility fixes for WebAssembly and WebGL 2

ntadej:
Compatibility fixes for WebAssembly and WebGL 2.

Note that I used GPT-5 mini/copilot to actually deal with the fixes so I'm happy for better suggestions.

FYI @birkskyum

### DIFF
--- a/src/mbgl/gl/buffer_allocator.cpp
+++ b/src/mbgl/gl/buffer_allocator.cpp
@@ -262,13 +262,24 @@ public:
         }
         MBGL_CHECK_ERROR(glBindBuffer(type, buffer->id));
 
-        // Map the next available slice of memory
+        // Map the next available slice of memory. On WASM/emscripten we don't have
+        // glMapBufferRange available, so use glBufferSubData as a fallback (upload
+        // directly from client memory).
+        const auto writtenIndex = buffer->pointer;
+#ifdef __EMSCRIPTEN__
+        // glBufferSubData accepts the real size to upload; no need to provide
+        // alignedSize as the upload size — upload only `size` bytes.
+        MBGL_CHECK_ERROR(glBufferSubData(type, buffer->pointer, static_cast<GLsizeiptr>(size), data));
+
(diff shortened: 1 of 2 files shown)


ntadej:
Demo: https://experiments.vremenar.app/wasm/QMapLibreExampleQuickStandalone.html

louwers:
How did you build the demo?

mwilsnd:
If fences and mapBufferRange aren't available, you might also consider bypassing the allocator altogether. `gl::UniformBuffer` takes in the allocator as a constructor argument and `gl::Context` creates and manages the allocator, an `#ifdef` can be inserted in those locations to just shim around it and do regular UBO construction inside `gl::UniformBuffer` instead. `gl::UniformBuffer` already manages regular UBOs in cases where the requested size is too large for the sub-allocation scheme.

birkskyum:
The demo is built using this branch
- https://github.com/maplibre/maplibre-native-qt/pull/216

The specific steps needed aren't published yet, but it's along the lines of [this](https://birkskyum.github.io/maplibre-native-wasm/qt-for-webassembly/webgl1-from-opengl2-legacy-renderer.html), but using the `quick` example instead of `widgets`, and maybe few other adjustments like adding the -DMLN_WITH_OPENGL=ON flag.

@ntadej , do you have the specific build steps anywhere so that we can repro?

ntadej:
> If fences and mapBufferRange aren't available, you might also consider bypassing the allocator altogether. `gl::UniformBuffer` takes in the allocator as a constructor argument and `gl::Context` creates and manages the allocator, an `#ifdef` can be inserted in those locations to just shim around it and do regular UBO construction inside `gl::UniformBuffer` instead. `gl::UniformBuffer` already manages regular UBOs in cases where the requested size is too large for the sub-allocation scheme.

Sorry to get back to this just now. Do you mean then that just the regular `gl::BufferAllocator` should be used for WASM?

ntadej:
I think I understand what the idea was so now I just force disable the managed allocation and it seems to work just fine. This PR is ready to go in.

PR #3756, opened 2025-08-25
format Bazel files with buildifier

louwers:
I hope that it is no longer causing problems on pre-commit.ci


mwilsnd:
Looks like you might need to add rules_shell to `MODULE.bazel`?

louwers:
@mwilsnd Looks like it...

mwilsnd (approved):

PR #3760, opened 2025-08-26
Update node to v24. Add support for Windows arm64 node binary build and publish. Update Linux node binary to be built on ubuntu 24.04

acalcutt:
This PR updates node current version to v24 and adds a windows arm64 binary to the build. It also updates the Linux binary to ubuntu 24.04. It includes changes from https://github.com/maplibre/maplibre-native/pull/3759 and changes from https://github.com/maplibre/maplibre-native/pull/3692 with some of the (Get-VendorPackages.ps1) changes reverted.

Closes https://github.com/maplibre/maplibre-native/issues/3687

This PR does the following

1. Updates the built node versions to node 20, 22, and 24 . Note that node 18 was removed because it is no longer supported and also did not seem to have node build candidates for windows arm64

2. Updates NAN to 2.23.0 to support node v24

3. Reverts the change to use @mapbox/node-pre-gyp and points back to my @acalcutt/node-pre-gyp. This is to allow node-pre-gyp to work in node v24. The [fix](https://github.com/mapbox/node-pre-gyp/pull/904) has been committed in the main branch of @mapbox/node-pre-gyp , but as noted in https://github.com/mapbox/node-pre-gyp/issues/907 we can't seem to get anyone who has permission to get a new version published there.  It could make sense in the future to make a @maplibe version of node-pre-gyp instead of using my version.

4. Adds presets for Linux and windows node builds in CMakePresets.json like existed for macOS already. Builds upon https://github.com/maplibre/maplibre-native/pull/3692 form windows arm64 node version.

5. For windows arm64 workflows, "ilammy/msvc-dev-cmd" was changed to add "arch: amd64_arm64" option, which sets VSCMD_ARG_TGT_ARCH environment variable and allows 'Get-VendorPackages.ps1' to built for the proper arm64 target/triplet. Without doing this Get-VendorPackages.ps1 was picking an x64 triplets.

6. I added some debug information to the windows triplet files to help understand what was going on with the builds. At first it was hard to tell the VSCMD_ARG_TGT_ARCH  was causing the wrong triplet to be loaded, which cause it to build for the wrong platform. This shows in the configure step on windows, like
image

7.  Updated platform/node/cmake/module.cmake to add support for windows arm64

8. Removes some old workarounds from the workflow which no longer seem to be needed, like setting a specific cmake version and strawberry perl fixes.

9.  Updated version of boost to fix a lot of "[The contents of <stdfloat> are available only with C++23 or later.] warnings I was seeing in the windows arm64 build. I found in  https://github.com/lballabio/QuantLib/issues/1827 that error existed in boost 1.83, which is what version our maplibre-native boost package was tacked to, and was fixed in 1.84.  https://github.com/maplibre/maplibre-gl-native-boost was already updated to 1.86, so I just updated to that latest commit.

10.  When running tests on windows arm, the version of pal1000/mesa-dist-win was failing because that mesa3d release is not compatible with arm. I changed this to use mmozeiko/build-mesa for windows arm tests, which was suggested for arm64 on the [pal1000 release](https://github.com/pal1000/mesa-dist-win/releases/tag/25.2.1)


acalcutt:
@louwers @birkskyum @tdcosta100 . Do any of you have any reviews on this PR. It seems like the windows arm binary is building and everything seems to be working on node v24. I don't have a system that can test this new windows arm build but the [ci tests](https://github.com/maplibre/maplibre-native/actions/runs/17364812314/job/49289981077?pr=3760) are promising.

acalcutt:
@birkskyum  one thing I wonder about is the [documentation ](https://github.com/maplibre/maplibre-native/blob/2139247d4778799dc2245a05867bf060705088ef/docs/mdbook/src/platforms/windows/build-msvc-arm64.md) I pulled in from https://github.com/maplibre/maplibre-native/pull/3692 . Because I put back Get-VendorPackages.ps1 on windows arm64 and it works as long as "ilammy/msvc-dev-cmd" was changed to add "arch: amd64_arm64" option, which sets VSCMD_ARG_TGT_ARCH environment variable before running the maplibre build. I wonder if there is a way to set VSCMD_ARG_TGT_ARCH in a regular (manual) build and not have to build all those dependencies beforehand (like i am doing it the workflow)

I did try setting VSCMD_ARG_TGT_ARCH  from our CMakeLists.txt (like I think your PR did also) but in testing that never seemed to actually work and it kept picking the wrong x64 triplet.

.github/actionlint.yaml
```diff
@@ -1,8 +1,7 @@
 self-hosted-runner:
   # Labels of self-hosted runner in array of strings.
   labels:
-    - macos-14  # can be removed once actionlint is updated
```
birkskyum, line 4: Should these stay?

- macos-14
- ubuntu-24.04

CMakePresets.json
```diff
@@ -60,9 +60,11 @@
...
     },
     {
-      "name": "macos-node",
+      "name": "macos-metal-node",
       "inherits": "macos",
       "binaryDir": "${sourceDir}/build",
+      "displayName": "MacOS Node",
+      "description": "MacOS Node build (Metal renderer)",
       "cacheVariables": {
         "MLN_WITH_NODE": "ON",
```
birkskyum, line 69: I expected a MLN_WITH_METAL = ON in this preset

.github/actionlint.yaml
acalcutt: I removed them because it was noted they could be removed once action lint was updated and removing them didn't seem to make a difference so I assume it was updated. I added the same note to the windows-11-arm I added, because at some point I assume action lint will be updated to support that as well.

.github/workflows/node-release.yml
```diff
@@ -140,22 +141,24 @@ jobs:
         if: runner.os == 'MacOS' || runner.os == 'Linux'
         uses: hendrikmuhs/ccache-action@63069e3931dedbf3b63792097479563182fe70d1 # v1
```
louwers: Consider using sccache instead. See `linux-ci.yml`.

The GitHub Actions cache fills up quite quickly, so it is not always used. sccache stores the cache on S3 so there is always enough space.

louwers: Probably better for a follow-up PR though.

acalcutt: Does it work for macos as well? we are already using sscache on windows in the workflow

louwers: Windows is using sscache but it is not using S3 as a storage backend yet.

Yes it works with macOS as well

acalcutt: I think i am going to wait for another PR to try this

louwers: Sounds good

acalcutt:
Can anyone approve this? Also, @louwers there are two failing tests that i don't think are related to this PR as I see them failing elsewhere


I did have one final question about the presets. Do you think the presets should include "CMAKE_CXX_COMPILER_LAUNCHER": "sccache" or should that be added in the workflow?

louwers (approved):
I can't really review the Windows stuff...

So... LGTM. I'll get it merged, will look into the linux workflow.

louwers:
https://github.com/maplibre/maplibre-native/pull/3825 wil probably fix linux-ci

PR #3765, opened 2025-08-27
Add Sentry build tags

adrian-cojocaru:
Add version name, build variant and commit hash to Sentry setup.

### DIFF
--- a/platform/android/MapLibreAndroidTestApp/build.gradle.kts
+++ b/platform/android/MapLibreAndroidTestApp/build.gradle.kts
@@ -32,6 +32,7 @@ android {
         versionName = file("../VERSION").readText().trim()
 
         manifestPlaceholders["SENTRY_DSN"] = ""
+        manifestPlaceholders["SENTRY_ENV"] = ""
     }
 
     nativeBuild(listOf("example-custom-layer"))
--- a/platform/android/MapLibreAndroidTestApp/build.gradle.kts
+++ b/platform/android/MapLibreAndroidTestApp/build.gradle.kts
@@ -27,9 +27,10 @@ android {
         minSdk = 21
         targetSdk = 33
         versionCode = 14
-        versionName = "6.0.1"
         testInstrumentationRunner = "org.maplibre.android.InstrumentationRunner"
         multiDexEnabled = true
+        versionName = file("../VERSION").readText().trim()
+
         manifestPlaceholders["SENTRY_DSN"] = ""
     }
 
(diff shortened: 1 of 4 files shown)


platform/android/MapLibreAndroidTestApp/build.gradle.kts
```diff
@@ -27,9 +27,10 @@ android {
         minSdk = 21
         targetSdk = 33
         versionCode = 14
-        versionName = "6.0.1"
         testInstrumentationRunner = "org.maplibre.android.InstrumentationRunner"
         multiDexEnabled = true
+        versionName = file("../VERSION").readText().trim()
```
louwers, line 32: We should use the GitHub SHA, as the release cycle of the app is not tied to the MapLibre Android version.

adrian-cojocaru: I wanted the version name to be something user friendly that changes from time to time. `SENTRY_ENV` has detailed tags. Should I concatenate the SHA to the version name?

louwers: So the SHA is already there?

adrian-cojocaru: SENTRY_ENV = `buildVariant` + `git rev-parse HEAD`

louwers: Ah neat.

louwers (approved):

platform/android/buildSrc/src/main/kotlin/SentryConfigPlugin.kt
```diff
@@ -58,5 +63,20 @@ class SentryConditionalPlugin : Plugin<Project> {
...
+            val commitHash = ByteArrayOutputStream()
+
+            project.exec {
+                commandLine("git", "rev-parse", "HEAD")
+                standardOutput = commitHash
+                workingDir = project.rootDir
+                isIgnoreExitValue = true
+            }
+
+            androidExtension.applicationVariants.forEach { variant ->
```
louwers, line 77: Looks like we need to move this to `platform/android/MapLibreAndroidTestApp/build.gradle.kts`

@adrian-cojocaru

adrian-cojocaru: `manifestPlaceholders["SENTRY_ENV"] = ""` in the `defaultConfig` should be enough. Checking

louwers: Yes will probably get overwritten.

PR #3768, opened 2025-08-28
Create macOS amalgamation for core releases

louwers:
This PR adds `libmaplibre-native-core-amalgam-macos-arm64-metal.a` to the core releases. I.e. a static core build for macOS + Metal so that users of that static library do not need to link with many other dependencies.

When I try to use the resulting amalgamation with maplibre-native-rust, I got linker errors from maplibre-native-rust itself. So needs some more work.

```
git clone https://github.com/maplibre/maplibre-native-rs
cd maplibre-native-rs
export MLN_CORE_LIBRARY_PATH=...
export MLN_CORE_LIBRARY_HEADERS_PATH=...../maplibre-native-headers.tar.gz
just build metal
```

Also modifies the workflow such that when you modify the workflow, it runs in a PR context. This is convenient for catching issues when working on the workflow. Also it adds the builds as artifacts.

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -78,9 +78,14 @@ jobs:
       - name: Install dependencies
         run: .github/scripts/install-linux-deps
 
-      - uses: taiki-e/install-action@020705266844b275b711684191e11281be48860c # v2
-        with:
-          tool: armerge@2
+      - run: cargo install armerge@2.2.0
+      # TODO: use below once
+      # https://github.com/tux3/armerge/issues/14
+      # is fixed
+
+      # - uses: taiki-e/install-action@020705266844b275b711684191e11281be48860c # v2
+      #   with:
+      #     tool: armerge@2
 
       - if: matrix.variant.rust
         run: cargo install cxxbridge-cmd --version 1.0.157 --locked


.github/workflows/core-release.yml
```diff
@@ -44,16 +49,34 @@ jobs:
         with:
           submodules: recursive
 
+      - run: brew install webp libuv
+
+      - run: cargo install armerge@2.2.0
```
CommanderStorm: We don't have to improve this here, but `cargo install ..` is usually not what you want.

We want `cargo binstall` as, this way we don't have to compile `armerge`.

Below is the way that I currently would do this to skip the compilation overhead.
With a few minues of work, `armerge` could be added under `taiki-e/install-action` instead of going the `binstall` route, but small steps.

```suggestion
      - uses: taiki-e/install-action@bb5b6ec0fe226d4254f80ec39794c3a8c1b6525f # v2
        with: { tool: 'cargo-binstall' }
      - run: cargo binstall armerge@2.2.0
```

.github/workflows/core-release.yml
```diff
@@ -44,16 +49,34 @@ jobs:
...
 
       - name: Rename artifact
-        run: cp build-macos/libmbgl-core.a libmaplibre-native-core-macos-arm64-metal.a
+        run: |
+          cp build-macos/libmbgl-core.a libmaplibre-native-core-macos-arm64-metal.a
+          cp build-macos/libmbgl-core-amalgam.a libmaplibre-native-core-amalgam-macos-arm64-metal.a
+
+      - name: Upload macos-metal-build artifact
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v4
```
CommanderStorm: let's have this step of CI consistent across all options => upload the artefact for windows, linux and macos ^^

CommanderStorm (commented):
> When I try to use the resulting amalgamation with maplibre-native-rust, I got linker errors from maplibre-native-rust itself. So needs some more work.

I don't own an mac unfortunately. Which linker issues are you experiencing?

CC @flowchartsman

.github/workflows/core-release.yml
louwers: Yes good idea.

.github/workflows/core-release.yml
louwers: Unfortunately armerge does not put out binary releases. Will this do anything then? I have created an issue https://github.com/tux3/armerge/issues/13

PR #3769, opened 2025-08-29
chore(ci): pin actions to shas

CommanderStorm:
Our CI uses a few actions.
For these actions, we currently just use the mutable github tag.

Since we use dependabot to update the versions, we should use shas.
This makes sure that we are not affected by a certain class of supply chain vulterabiity where attackers re-publish bad tags.

Using sha's matches githubs recomendations and is a part of the OpenSSFs Scorecard.


CommanderStorm:
Any pointers welcome. I have no clue why this can fail CI. I neither changed code nor versions.

Is said CI broken? @louwers

louwers (approved):

PR #3771, opened 2025-08-30
fix(metal) add missing DepthMode range

birkskyum:
This PR fixes a compilation error I hit in the Metal backend:

```sh
  error: missing field 'range' initializer [-Werror,-Wmissing-field-initializers]
  error: no matching constructor for initialization of 'Range<float>'
```

It's caused by the DepthMode struct initialization, which was missing the required range field, causing build failures on macOS.


birkskyum:
Oh, this is because I ran with MLN_WITH_METAL=ON, but MLN_WITH_OPENGL is also ON by default, so I have to run it with MLN_WITH_OPENGL=OFF. Probably best to have all OFF by default.

PR #3772, opened 2025-08-30
set MLN_WITH_OPENGL=OFF by default, like Metal/Vulkan

birkskyum:
To avoid issues like these
- #3771 

Simply make OpenGL Opt-In like Metal and Vulkan are.

Until now, to build with e.g. Metal, you had to opt into Metal (MLN_WITH_METAL=ON), but a number of issues would occur if you forget to also opt-out of OpenGL (MLN_WITH_OPENGL=OFF). This removes the need for opting out.

Added MLN_WITH_OPENGL=ON to Qt, Node CI and Android App which didn't have explicit setting of the flag.

Also change a few "else // OpenGL" to "elif OPENGL", for strict conditional logic


louwers:
Might need to update CMakePresets.json. But good to merge if CI passes.

birkskyum:
All pass, except for Windows MSYS2 CI, which appear to fail because of MSYS2's toolchain is linking both static (libz.a, libstdc++.a) and dynamic (libz.dll.a) versions, causing duplicate symbols. I don't think it's caused by this PR - has it been reported before?

Setting the --allow-multiple-definitions doesn´t seem to resolve it.

louwers:
Yes it is a known issue. https://github.com/maplibre/maplibre-native/issues/3670

PR #3774, opened 2025-09-01
Fix compile error when building with MLN_USE_TRACY

sargunv:
Was: `wrap_` is not found. I assume this was meant to be `tile_.items`, not `wrap_.releases`.

(I'm not actually sure what tracy is; I just saw some related macros in the code and was experimenting with them)


louwers:
@sargunv It is a frame profiler: https://maplibre.org/maplibre-native/docs/book/profiling/tracy-profiling.html

PR #3783, opened 2025-09-02
Add expected image Pixel 6

louwers:
@adrian-cojocaru I think it's probably a bug that it doesn't show at all anymore. What do you think?


adrian-cojocaru:
https://github.com/maplibre/maplibre-native/pull/3198 had similar results for `render-tests/icon-text-fit/textFit-grid-long` and we ended up updating the expected images.

adrian-cojocaru (approved):

PR #3791, opened 2025-09-06
fix some missing public headers referenced by other public headers (macos)

### LINKED ISSUE #3790: public header mbgl/mtl/context.hpp includes other headers missing from public headers
**Describe the bug**

I'm working on building maplibre-compose against the prebuilt native core, starting first by trying to build without the nonpublic headers (see [wip branch](https://github.com/maplibre/maplibre-compose/pull/583)) and on macos, am blocked by this header.

The affected include: 
https://github.com/maplibre/maplibre-native/blob/9b6325a14e2cf1cc29ab28c1855ad376f1ba4903/include/mbgl/mtl/context.hpp#L3C1-L4C1

Wondering, is there a way to validate in CI that the public headers only reference other public headers?

**To Reproduce**

```cpp
#include <mbgl/mtl/context.hpp>
```

**Expected behavior**

The above include should be valid when building with the public headers..

 - Operating System: macOS

sargunv:
Resolves #3790

The qt-ci failures looks like missing libc++, unrelated to this pr


sargunv:
The qt-ci errors looks like missing libc++, unrelated to this PR?

PR #3793, opened 2025-09-08
Apply just the changes from the MLT branch which aren't really part of that work

TimSylvester:
This is an attempt to approach a CI build issue on #3246 by simplifying the change set.

- Fix a variety of warnings and style issues.
  - Templated virtual method portability in style expression evaluation which has been ignored for a long time
  - Remove unnecessary includes
  - Use `contains` instead of `find(...)!=end()` or `count(...)!=0`
  - Use ranges
  - Use anonymous namespaces instead of statics
  - Use named initializers
- Enable recursive checkout in Node and Qt builds
- Make Xcode version options more explicit
- Move `is_compressed` to `compression.cpp`
- Eliminate double-lookup in format section override
- Use `SYSTEM` property on harfbuzz and freetype in an attempt to suppress warnings from external code


platform/default/src/mbgl/util/compression.cpp
```diff
@@ -31,6 +31,25 @@ namespace util {
 // cause a link error.
 #undef compress
 
+bool is_compressed(const std::string &v) {
```
louwers, line 34: Consider using `static_cast` for the static casts.

PR #3800, opened 2025-09-11
Fix clang-tidy 18.1.8 issues

louwers:
Fixes all issues that `clang-tidy` reports.


louwers:
I disabled `modernize-use-constraints` because it is (still) crashing in this Clang version...

PR #3810, opened 2025-09-16
Set up auto-labeler for PRs

louwers:


.github/labeler.yml
```diff
@@ -0,0 +1,58 @@
...
+  - any-glob-to-any-file:
+    - 'platform/qt/**/*'
+windows:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'platform/windows/**/*'
+windows:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'platform/windows/**/*'
```
adrian-cojocaru: Duplicate?

.github/labeler.yml
```diff
@@ -0,0 +1,58 @@
...
+- changed-files:
+  - any-glob-to-any-file:
+    - '**/CMakeLists.txt'
+    - '**/*.cmake'
+    - '**/BUILD.bazel'
+    - '**/*.bzl'
+OpenGL:
+  - any-glob-to-any-file:
+    - '**/*opengl*'
+    - '**/*opengl*/**'
```
adrian-cojocaru, line 46: `include/mbgl/gl`, `src/mbgl/gl`, `include/mbgl/shaders/gl` and `src/mbgl/shaders/gl` use just `gl`

.github/labeler.yml
```diff
@@ -0,0 +1,58 @@
+android:
+- changed-files:
+  - any-glob-to-any-file: 'platform/android/**/*'
+iOS:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'platform/ios/**/*'
+    - 'platform/darwin/**/*'
+cpp-core:
```
adrian-cojocaru, line 9: Does it make sense to include `platform/default` here?

.github/labeler.yml
louwers: Good one!

.github/labeler.yml
louwers: That would create some false positives.

PR #3813, opened 2025-09-17
Android Release Workflow Fixes

louwers:
Close SonarType repository as a final step. 

Pass GITHUB_TOKEN to `ensure-tag.sh` script.


.github/scripts/ensure-tag.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
HarelM, line 3: Is this intentional?

louwers: Yes, it was failing and then continuing the script before.

cgalvan (approved):

PR #3820, opened 2025-09-18
chore(security): remove unnessesarily persisted permissions

CommanderStorm:
I think we are currently not using these permissions.
Lets remove them.

Given how complex the CI is, this is really not as simple as it sounds.

THis also seems to be a [best practice](https://docs.zizmor.sh/audits/#artipacked), but that is secondary to this PR.

When removing them, I was conservative, there are other occurrences where we might be able to remove them. This is the larger part that we can just remove without deeply looking at.

### DIFF
--- a/vendor/maplibre-native-base/.github/workflows/ci.yml
+++ b/vendor/maplibre-native-base/.github/workflows/ci.yml
@@ -27,7 +27,6 @@ jobs:
       - uses: actions/checkout@v3
         with:
           persist-credentials: false
-          fetch-depth: 0
 
       - name: Build and run tests
         run: |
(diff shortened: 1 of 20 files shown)


louwers:
Ha, I just removed `fetch-depth: 0`. It doesn't mean what you think it does:

```
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

vendor/maplibre-native-base/.github/workflows/ci.yml
```diff
@@ -25,6 +25,8 @@ jobs:
     runs-on: ${{ matrix.os }}
```
louwers, line 25: Oh this used to be a submodule. Need to move these workflows. Let me take a not of that.

louwers (approved):

PR #3821, opened 2025-09-18
chore(security): enforce pinning in CI via pre-commit style ergonomics

CommanderStorm:
This PR automates the fixing of SHAs.


.github/workflows/pinact.yml
```diff
@@ -0,0 +1,16 @@
...
+  pinact:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Pin actions
+        uses: suzuki-shunsuke/pinact-action@d735505f3decf76fca3fdbb4c952e5b3eba0ffdd # v0.1.2
```
CommanderStorm:
```suggestion
        uses: suzuki-shunsuke/pinact-action@d735505f3decf76fca3fdbb4c952e5b3eba0ffdd # v0.1.2

```

louwers (commented):
I don't think this will work.

It needs a token. We can add it, but it will not be available on forks (where most PRs are coming from).

CommanderStorm:
Yea I also noticed this just now.
Would failing this action be an ok compromise?

Alternatively, we could run on pushes to `main` and create a PR.

.github/workflows/pinact.yml
CommanderStorm, line 19:
```suggestion
        uses: suzuki-shunsuke/pinact-action@d735505f3decf76fca3fdbb4c952e5b3eba0ffdd # v0.1.2
        with:
          skip_push: "true"
```

.github/workflows/pinact.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  pinact:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Pin actions
```
CommanderStorm:
```suggestion
      - name: Check if all actions versions are immutable SHAs
```

louwers:
I like this:

image

Maybe tokens from PRs can create review comments?

.github/workflows/pinact.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Check if all actions versions are immutable SHAs
+        uses: suzuki-shunsuke/pinact-action@d735505f3decf76fca3fdbb4c952e5b3eba0ffdd # v0.1.2
+        with:
+          skip_push: "true"
```
CommanderStorm:
```suggestion
          github_token: ${{secrets.GITHUB_TOKEN}}
```

louwers:
I think this CLI is needed for that https://github.com/suzuki-shunsuke/pinact?tab=readme-ov-file#create-reviews

.github/workflows/pinact.yml
```diff
@@ -0,0 +1,20 @@
+name: Pin GitHub Actions to SHAs
+on:
+  pull_request:
+    paths:
+      - .github/**
+permissions:
+  contents: write
```
CommanderStorm, line 7:
```suggestion
  contents: write
  actions: write
```

CommanderStorm:
Yea seems like it. Sad.
I am going to bed now, good night

CommanderStorm:
This needs a bit more experimentation and a different approach for review comments => lets close for now

PR #3841, opened 2025-09-28
Remove bazel-out from xcframework

zarechnyy:
This PR removes bazel-out dir from xcframework for iOS release. 

**BEFORE:**
image

**AFTER:**
image

Diff size:
image


louwers (approved):
Thanks!

PR #3862, opened 2025-10-06
Fix static XCFramework creation

louwers:
Create static XCFramework with:

```
bazel build //platform/ios:MapLibre.static --//:renderer=metal --compilation_mode="opt" --copt -g --copt="-Oz" --strip never --output_groups=+dsyms --apple_generate_dsym
```

Add to Xcode, in General, make sure "Embed & Sign" is set to **Do Not Embed**.

image

In Build Settings set Other Linker Flags to `-ObjC`.

image

In Build Phases, add Mapbox.bundle to Copy Bundle Resources (still needs to be renamed).

image

Draft PR until https://github.com/bazelbuild/rules_apple/pull/2791 is upstreamed and released.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -83,6 +83,12 @@ git_override(
     tag = "1.0.157",
 )
 
+git_override(
+    module_name = "rules_apple",
+    commit = "5005493",
+    remote = "https://github.com/louwers/rules_apple.git",
+)
+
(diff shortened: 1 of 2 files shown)


louwers:
OK, the upstream PR was merged and released.

Let's merge this then I will work on the release.

PR #3868, opened 2025-10-09
[node] Add options parameter to addImage method in Node.js type definitions

NotFounds:
The addImage method signature was missing the optional third parameter for image options.
This update aligns the TypeScript definitions with the actual implementation.

ref: https://github.com/maplibre/maplibre-native/blob/59aca67f394ebf52eeab997566bc0f4ab50fb362/platform/node/src/node_map.cpp#L804-L806

### DIFF
--- a/platform/node/index.d.ts
+++ b/platform/node/index.d.ts
@@ -156,7 +156,7 @@ declare module '@maplibre/maplibre-gl-native' {
     /**
      * Add image to map's style
      */
-    addImage: (imageId: string, image: any) => void;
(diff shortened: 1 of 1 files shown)


NotFounds:
@louwers 
Thank you for your comment.

This variable is not optional but required.
I named it "option" to match the C++ implementation, but would "sizeInfo" or something similar be better?

ref: https://github.com/maplibre/maplibre-native/blob/main/platform%2Fnode%2Fsrc%2Fnode_map.cpp#L820-L832

louwers:
Apologies, I misunderstood.

Issue #3766, opened 2025-08-27
Deadlock in queryPointAnnotations


michalgwo:
### MapLibre Android Version

11.12.1

### Android SDK Version

Android 13

### Device

Xiaomi Redmi Note 10 Pro

### What happened?

I got an ANR on Crashlytics. It looks like it was triggered by `queryPointAnnotations` when the user tapped on the map.
[Full Crash Log.txt](https://github.com/user-attachments/files/22009334/anr.txt)

### Steps to reproduce

I didn't manage to reproduce it on my end

### Renderer

OpenGL (choose this if you are unsure)

### Relevant log output

```shell
main (native):tid=1 systid=14050 
#00 pc 0x89af0 libc.so (syscall + 32) (BuildId: 449f781894033dce6346794a1ee593e0)
#01 pc 0x8e4d8 libc.so (__futex_wait_ex + 148) (BuildId: 449f781894033dce6346794a1ee593e0)
#02 pc 0xf489c libc.so (pthread_cond_wait + 80) (BuildId: 449f781894033dce6346794a1ee593e0)
#03 pc 0xa55e7c split_config.arm64_v8a.apk (std::__ndk1::condition_variable::wait + 24) (BuildId: 8e61794ea890f451d21f5cd55f4a0525660e5c1c)
#04 pc 0xa5665c split_config.arm64_v8a.apk (std::__ndk1::__assoc_sub_state::__sub_wait + 44) (BuildId: 8e61794ea890f451d21f5cd55f4a0525660e5c1c)
#05 pc 0x4e02b8 split_config.arm64_v8a.apk + 32768 (BuildId: 8e61794ea890f451d21f5cd55f4a0525660e5c1c)
#06 pc 0x4d6744 split_config.arm64_v8a.apk (mbgl::android::AndroidRendererFrontend::queryPointAnnotations const + 96) (BuildId: 8e61794ea890f451d21f5cd55f4a0525660e5c1c)
#07 pc 0x5149b0 split_config.arm64_v8a.apk (mbgl::android::NativeMapView::queryPointAnnotations + 148) (BuildId: 8e61794ea890f451d21f5cd55f4a0525660e5c1c)
[...]

PR #3874, opened 2025-10-12
Add query annotation timeout

adrian-cojocaru:
Possible workaround for https://github.com/maplibre/maplibre-native/issues/3766

I couldn't find the root cause of the ANR since the stack points to a valid `AndroidRendererFrontend`/`MapRenderer` instance, but no active `RenderThread` to process the request. Adding a timeout to the native calls generated by `AnnotationManager.onTap` unblocks the main thread if this is a lifecycle issue between the render thread and `MapGestureDetector` (or something similar).


sjg-wdw:
If this one is okay with you @louwers we'd like to get them a release in the next day or two.

PR #3879, opened 2025-10-14
Improve CMake build setup Harfbuzz and Freetype deps

louwers:
This fixes Qt CI on macOS, which chokes on warnings coming from these projects otherwise.

### DIFF
--- a/.github/workflows/qt-ci.yml
+++ b/.github/workflows/qt-ci.yml
@@ -73,14 +73,16 @@ jobs:
             deployment_target: 12.0
             deployment_arch: "x86_64;arm64"
             compiler: ""
-          - name: macOS_LLVM
-            os: macos-14
-            build_type: RelWithDebInfo
-            qt_version: 6.9.1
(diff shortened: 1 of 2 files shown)


ntadej:
Looks OK to me if it helps. Alternatively we could just disable this for now (I did it for now in the `maplibre-native-qt` repo).

TimSylvester:
I tried making those `SYSTEM` at some point and it seemed to have no effect, but I didn't look closer.  Perhaps it only applies to clang and not clang-tidy?

louwers:
Will try some other things... Otherwise will disable it.

I can't repro it locally.

ntadej:
The previous issue was fixed, but now we're affected by https://github.com/llvm/llvm-project/issues/77653

Can you just disable the LLVM test for now (but keep the other changes)?

ntadej:
If the CI passes, I think this is good to go in, thanks Bart!

louwers:
OK, looks like it passes again.

PR #3882, opened 2025-10-15
Add padding support to MapSnapshotter Android

grab-liyanjin:
Here the different between no padding and padding(left:20, top:60, right:20, bottom:60)
maplibre_snapshot2
maplibre_snapshot2_padding

### DIFF
--- a/platform/android/MapLibreAndroid/src/cpp/snapshotter/map_snapshotter.cpp
+++ b/platform/android/MapLibreAndroid/src/cpp/snapshotter/map_snapshotter.cpp
@@ -26,10 +26,10 @@ MapSnapshotter::MapSnapshotter(jni::JNIEnv& _env,
                                jni::jboolean _showLogo,
                                jni::jboolean _showAttribution,
                                const jni::String& _localIdeographFontFamily,
-                               jni::jfloat regionPaddingLeft,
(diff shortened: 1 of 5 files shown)


louwers (changes requested):
In my opinion the padding should be a second (optional) argument to `withRegion` and `setRegion`. Since they are coupled and cannot be used independently (as the comment indicates).

louwers:
Another consideration: the padding and the pixel ratio are interrelated. I think the padding values need to be divided by the pixel ratio similar to what `nativeGetCameraForLatLngBounds` does.

grab-liyanjin:
> Another consideration: the padding and the pixel ratio are interrelated. I think the padding values need to be divided by the pixel ratio similar to what `nativeGetCameraForLatLngBounds` does.

Thanks @louwers for the careful review, I didn't consider the pixelRatio with the padding, and you are right, let me change the PR

grab-liyanjin:
Changed, @louwers , pls review when you have time

louwers:
@grab-liyanjin OK thank you for fixing that.  Created an issue for the corresponding functionality in iOS. https://github.com/maplibre/maplibre-native/issues/3893

This will be part of the next release for Android.

PR #3838, opened 2025-09-26
WebGPU Backend for MapLibre Native

birkskyum:
### Motivation

Using MapLibre Native in browsers is quite a challenge, but there is progress made on it, and I try to keep an overview of findings in the [maplibre-native-wasm](https://birkskyum.github.io/maplibre-native-wasm/) repo.

- July 2024 - WebGL1 - Using the [opengl-2](https://github.com/maplibre/maplibre-native/tree/opengl-2) compat branch of Native, through Qt for WebAssembly, we managed to land a WebGL1 build - [demo](https://birkskyum-maplibre-native-wasm.pages.dev/qt-for-webassembly/webgl1-from-opengl2-legacy-renderer-qt6-9-1/)
- May-August 2025 - WebGL2 - After updating the Qt SDK to the latest MapLibre Native , the OpenGL3+ backend allowed for a WebGL2 build - [demo](https://experiments.vremenar.app/wasm/QMapLibreExampleQuickStandalone.html).
  - https://github.com/maplibre/maplibre-native-qt/pull/216/
- Now, Sep 2025 - WebGPU - To get any further pushing this boundary, a WebGPU backend in Native is needed. 

### PR Content

This PR contains a full implementation of a WebGPU backend for MapLibre Native - tested with [Dawn](https://dawn.googlesource.com/dawn) (abstracted away, allowing for later addition of wgpu).

https://github.com/user-attachments/assets/fa4fe582-e712-47b3-9739-c7b846d3f47c

https://github.com/user-attachments/assets/31f46a2a-13d1-4123-8083-148071df17da

The supported layer types are:

- ✅ background
- ✅ fill
  - ✅ fill-outline
  - ✅ fill-pattern
- ✅ fill-extrusion
- ✅ symbol
- ✅ circle
- ✅ line
  - ✅ line-pattern
- ✅ hillshade
- ✅ hillshade-prepare
- ✅ raster
- ✅ heatmap

Not tested
- location indicator
- custom drawable

## Review styles

I left some styles in the ./review-styles that are handy for testing things, like the heatmap - these will be deleted before merging, and account for ~ 10K LOC

## Build instructions

I have tested this on macOS with Metal, and on linux (Asahi) with Vulkan

**Dawn**
- Make sure dawn is in ./vendor/dawn , the submodule should work, otherwise just clone it in from https://dawn.googlesource.com/dawn
- [Build dawn](https://dawn.googlesource.com/dawn/+/HEAD/docs/building.md#without)

**Core**
[...]

PR #3892, opened 2025-10-16
Exclude node 21 23 25

birkskyum:
The node-ci is breaking, because of node 25 with ubuntu 24. Example of PR where that is the case is e.g.:

- #3838 

There's a yearly LTS release of Node, with Node 26 coming out April 2026, and little reason to use the intermediate "uneven" releases.

Release Schedule:
https://nodejs.org/en/about/previous-releases

This PR excludes the 21, 23, 25 releases

### DIFF
--- a/platform/node/CMakeLists.txt
+++ b/platform/node/CMakeLists.txt
@@ -7,6 +7,8 @@ endif()
 
 include(${CMAKE_CURRENT_LIST_DIR}/cmake/module.cmake)
 
+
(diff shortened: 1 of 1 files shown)


louwers (approved):

acalcutt:
I left the non lts versions so someone could still build that version locally if they wanted. the publish.sh script only includes the LTS versions we specify when it publishes the binaries.

louwers:
@acalcutt Sorry if we were too quick to merge!

PR #3896, opened 2025-10-17
Fix race between setURL and loadDescription

xabbu42:
A follow up for #3891 fixing another race between setURL and loadDescription. Without this patch it is possible for loaded to be set to true from an older loadDescription call after setURL set it to false if the older request gets handled before loadDescription is called again.


louwers:
Is it possible to add a regression test?

xabbu42:
I'm trying to add a regression test based on the test Source.GeoJSonSourceUrlUpdate. Currently I'm stuck because it seems that test no longer works as originally intended. As far as I can tell, the EXPECT_EQ in sourceRequest is never reached, but I don't know why.

TimSylvester:
I think that test is just broken, and maybe never worked.

If I understand correctly, by the time the stub source calls the `sourceResponse` callback, the `test.end()` call is already shutting things down so it never happens.  If I comment out `test.end()`, I see the callback run and the test fail.

The `EXPECT_EQ("url", resource.url);` seems to have been copied from another test and not updated to the URL used in that one.

TimSylvester:
Oh, and add a call `loadDescription` again.  This seems to work:

```cpp
TEST(Source, GeoJSonSourceUrlUpdate) {
    SourceTest test;

    test.fileSource->sourceResponse = [&](const Resource& resource) {
        EXPECT_EQ("http://source-url.ext", resource.url);
        Response response;
        response.data = std::make_unique<std::string>(
            R"({"geometry": {"type": "Point", "coordinates": [1.1, 1.1]}, "type": "Feature", "properties": {}})");
        test.end();
        return response;
    };

    test.styleObserver.sourceDescriptionChanged = [&](Source& src) {
        // Should be called (test will hang if it doesn't)
        //test.end();
    };

    GeoJSONSource source("source");
    source.setObserver(&test.styleObserver);

    // Load initial, so the source state will be loaded=true
    source.loadDescription(*test.fileSource);

    // Schedule an update
    test.loop.invoke([&]() {
        // Update the url
        source.setURL(std::string("http://source-url.ext"));
        source.loadDescription(*test.fileSource);
    });

    test.run();
}
```

But probably it should be using a flag to check that `sourceDescriptionChanged` was called.

xabbu42:
Thanks that unblocked me for a time. Still I could not get a test for this particular race to work probably because most cases are handled by the req.reset() call already. But I can assure you that the race exists in our application. I'll make a separate pull request for some improvements to the tests independent of this fix.

xabbu42:
Hmm I wanted to come back to this topic but can no longer reproduce the problem. I'm not sure what exactly changed.

PR #3899, opened 2025-10-17
WebGPU - Add support for wgpu

### LINKED ISSUE #3869: WebGPU Backend - wgpu support
Dawn is the the preferred implementation when building for web, because it's what's running in Chrome, but for certain native targets, there are cases where we need wgpu support as well.

A case where we need the wgpu support is e.g. for the [MapLibre Native Slint SDK](https://github.com/maplibre/maplibre-native-slint). It's currently doing resource intensive pixel copy when it's rendering, because it unfortunately can't tap into our dedicated metal/vulkan/opengl backends in a similar fashion to [Qt](https://github.com/maplibre/maplibre-native-qt/pull/216)/Flutter/Compose, due to limitations of the Slint API. What Slint does offer instead, is support for embedding wgpu based applications (release post [here](https://slint.dev/blog/slint-1.12-released))

This PR adds support for WebGPU Dawn implementation:
- #3838 

Dawn and wgpu should be abstracted away, so that a user at compile time can decide which to use

Unlocks https://github.com/maplibre/maplibre-native-slint/issues/42

birkskyum:
Follow-up to:
- #3838 

Closes #3869 
- #3869 

Needed for e.g. the maplibre native slint sdk

This PR adds support for wgpu, so that a user can select between running webgpu with dawn or wgpu.

There are compile bool flags for this, one of which should be set ON when using MLN_WITH_WEBGPU:
- MLN_WEBGPU_IMPL_DAWN
- MLN_WEBGPU_IMPL_WGPU

For convenience there are also the cmake presets
- macos-webgpu-dawn
- macos-webgpu-wgpu
- linux-webgpu-dawn
- linux-webgpu-wgpu

I locally measure the performance of wgpu to be comparable to dawn for the GLFW example with the default OpenFreeMap Liberty

I've added the wgpu to the macos-ci matrix, next to dawn, so both will now be running the full test suites for render / expression / unit (mbgl-test-runner). Despite running tests, the WebGPU flows have much shorter runtime, compared to the metal flow.

Most of the deleted lines is because I removed a debug triangle example from the webgpu GLFW backend, that was used for development of dawn, and it's not needed anymore.

### DIFF
--- a/src/mbgl/webgpu/offscreen_texture.cpp
+++ b/src/mbgl/webgpu/offscreen_texture.cpp
@@ -213,12 +213,10 @@ public:
         WGPUBufferMapCallbackInfo callbackInfo = {};
         callbackInfo.mode = WGPUCallbackMode_WaitAnyOnly;
         callbackInfo.callback =
-            [](WGPUMapAsyncStatus status, WGPUStringView message, void* userdata1, void* userdata2) {
+            [](WGPUMapAsyncStatus status, WGPUStringView /*message*/, void* userdata1, void* /*userdata2*/) {
                 auto* ctx = static_cast<MapContext*>(userdata1);
                 ctx->status = status;
                 ctx->completed = true;
-                (void)message;
-                (void)userdata2;
             };
         callbackInfo.userdata1 = &mapContext;
         callbackInfo.userdata2 = nullptr;
@@ -246,12 +244,10 @@ public:
         WGPUBufferMapCallbackInfo callbackInfo = {};
         callbackInfo.mode = WGPUCallbackMode_AllowProcessEvents;
         callbackInfo.callback =
-            [](WGPUMapAsyncStatus status, WGPUStringView message, void* userdata1, void* userdata2) {
+            [](WGPUMapAsyncStatus status, WGPUStringView /*message*/, void* userdata1, void* /*userdata2*/) {
                 auto* ctx = static_cast<MapContext*>(userdata1);
                 ctx->status = status;
                 ctx->completed = true;
-                (void)message;
-                (void)userdata2;
             };
         callbackInfo.userdata1 = &mapContext;
         callbackInfo.userdata2 = nullptr;


birkskyum:
Is the "Android Instrumentation Tests (Vulkan)" failing because of this PR, or something else?

louwers:
@birkskyum It started to fail since we moved to three devices. I will change it so that it passes if 2/3 devices pass...

birkskyum:
Okay - I re-ran it, and now it's passing

src/mbgl/webgpu/offscreen_texture.cpp
```diff
@@ -229,6 +239,40 @@ class OffscreenTextureResource final : public webgpu::RenderableResource {
                 mbgl::Log::Error(mbgl::Event::Render, "WebGPU: Buffer mapping wait failed");
             }
         }
+#elif MLN_WEBGPU_IMPL_WGPU
+
+        // wgpu-native: Use polling approach (WaitAny not implemented yet)
```
louwers, line 242: Is there a tracking issue?

src/mbgl/webgpu/offscreen_texture.cpp
```diff
@@ -229,6 +239,40 @@ class OffscreenTextureResource final : public webgpu::RenderableResource {
...
+        // Use AllowProcessEvents callback mode for synchronous polling
+        WGPUBufferMapCallbackInfo callbackInfo = {};
+        callbackInfo.mode = WGPUCallbackMode_AllowProcessEvents;
+        callbackInfo.callback =
+            [](WGPUMapAsyncStatus status, WGPUStringView message, void* userdata1, void* userdata2) {
+                auto* ctx = static_cast<MapContext*>(userdata1);
+                ctx->status = status;
+                ctx->completed = true;
+                (void)message;
```
louwers: These can be removed I think?

src/mbgl/webgpu/offscreen_texture.cpp
birkskyum: Yes, there are mentions of WaitAny being a bit "theoretical still" in this ticket:
https://github.com/webgpu-native/webgpu-headers/issues/491

birkskyum:
> I think there is something to be said for only supporting one WebGPU implementation. That makes hacking on it a bit easier now that it is still experimental. There's quite a few #ifdefs that are required. However, since you are pushing this effort forward I will leave that decision to you.

I agree it's a bit much conditional logic, but I'd prefer to keep wgpu around. These 62 cases of `#if MLN_WEBGPU_IMPL_DAWN` are a result of trying to add wgpu without making any changes to the dawn path, but it's possible to cut down on that in following work. I'll be actively engaging with the wgpu team on any discrepancies to get things aligned, since these two implementations should be converging in their public api. I also found that wgpu e.g. has stricter shader passing, and that benefits the dawn code too.

birkskyum:
>I think we should move the WebGPU CMake files out of vendor. And we should replace the submodules with FetchContent_Declare in my opinion. Reason: both Dawn and WGPU are not actually vendored dependencies (since they download their own dependencies).

>If you want, I can address this in a follow-up.

Regarding the cmake and vendor modules, I think that sounds like a good idea. Now that there's a baseline established with automated testing, it would be very helpful if you'd be able to address it in a follow-up,

PR #3901, opened 2025-10-19
Add macos-metal cmake preset

birkskyum:
Until now there was a `macos` preset, which acted both as a base preset, but also as a Metal preset. This assumption that all macos presets should have `MLN_WITH_METAL=ON` doesn't hold, which means that several other presets (macos-vulkan, macos-webgpu-x)  wrongly inherit it and has to set it off again `MLN_WITH_METAL=OFF`

This preset adds a `macos-metal` preset (similar to `macos-vulkan`), which will be used for `macos-metal-xcode` and `macos-metal-node`.

I also renamed `macos-xcode` -> `macos-metal-xcode`, so that it's similar to `macos-vulkan-xcode`

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -48,7 +48,7 @@
       "name": "macos-metal-xcode",
       "displayName": "macOS (Xcode)",
       "generator": "Xcode",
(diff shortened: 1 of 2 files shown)


louwers:
Thanks!

PR #3903, opened 2025-10-20
Updates drawable size of metal view after it is attached to display hierarchy.

StephanPartzsch:
Updating contentScaleFactor does not update the drawableSize of the metal view automatically. This results in wrong fullscreen size of mapView on iOS 26 in combination with smart zoom.


KasparElmans (approved):

boldtrn (approved):
This worked in my tests, thanks 👍

PR #3911, opened 2025-10-26
Fix `expected-lite` path in `core-release` workflow

Turbo87:
The `expected-lite` dependency was moved from `vendor/maplibre-native-base/extras/expected-lite` to `vendor/expected-lite` in commit bb42ed111bb, but the workflow was not updated.


louwers:
Thanks

Turbo87:
> This workflow requires approval from a maintainer.

@louwers I guess you might have to press one more button to trigger workflow 😅

PR #3914, opened 2025-10-27
Add configurable threshold for rotate gesture snapping to north

JesseCrocker:
Add configurable threshold for rotate gesture snapping to north on iOS. This is equivalent to the [MapOptions.bearingSnap ](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/) in maplibre-gl-js

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -40,6 +40,7 @@ build:
     - '**/*.cmake'
     - '**/BUILD.bazel'
     - '**/*.bzl'
+    - CMakePresets.json
 OpenGL:
 - changed-files:
   - any-glob-to-any-file:
--- a/.github/workflows/core-release.yml
+++ b/.github/workflows/core-release.yml
@@ -57,13 +57,13 @@ jobs:
 
       - name: Build mbgl-core for macOS
         run: |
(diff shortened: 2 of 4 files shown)


platform/ios/src/MLNMapView.h
```diff
@@ -855,6 +855,15 @@ vertically on the map.
  */
 @property (nonatomic, getter=isRotateEnabled) BOOL rotateEnabled;
 
+/**
+The threshold, measured in degrees, that determines when the map's bearing will snap to north.
+For example, with a bearingSnap of 7, if the user rotates the map within 7 degrees of north,
```
louwers: It should either be called `bearingSnap` or the documentation comment needs to be updated to the actual name.

I think `toleranceForSnappingToNorth` is a bit verbose but the most descriptive, so it would have my preference. But up to you.

louwers (changes requested):
Please see comment.

Looks good otherwise!

platform/ios/src/MLNMapView.h
JesseCrocker: renamed to toleranceForSnappingToNorth

louwers (approved):

PR #3915, opened 2025-10-29
Pass proper argument to delegate method on macos platfrom

michaelkirk:
This is a draft for two reasons:

1. I haven't tested this. I'm a downstream user of maplibre native, but I don't have my build environment wired up to build and consume home cooked maplibre-native. That said, it looks like a trivial typo, introduced when this code was first written in ce4dcb7a9b76

2. There's a bit of a mystery here: I noticed the errant behavior on iOS, but the iOS platform code looks good: https://github.com/maplibre/maplibre-native/blob/a13472a410a246dca5529f4d3a80ea8fb14ef1f5/platform/ios/src/MLNMapView%2BImpl.mm#L167-L170

 Only the macos platform code looks problematic. Is iOS using the macos platform code?

### DIFF
--- a/platform/macos/src/MLNMapView+Impl.mm
+++ b/platform/macos/src/MLNMapView+Impl.mm
@@ -162,7 +162,7 @@ void MLNMapViewImpl::onGlyphsRequested(const mbgl::FontStack& fontStack, const m
 void MLNMapViewImpl::onTileAction(mbgl::TileOperation operation, const mbgl::OverscaledTileID& tile, const std::string& sourceID) {
     [mapView tileDidTriggerAction:MLNTileOperation(static_cast<int>(operation))
                                 x:tile.canonical.x
-                                y:tile.canonical.x
+                                y:tile.canonical.y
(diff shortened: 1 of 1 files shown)


michaelkirk:
> There's a bit of a mystery here: I noticed the errant behavior on iOS, but the iOS platform code looks good:

Ah! I think I've explained this. The issue was recently fixed in the iOS platform (but not macos) in https://github.com/maplibre/maplibre-native/pull/3793/files#diff-c72fe1027f9d5d9f28ffae53b5c81a35d55de7b498b2ae6b49a758e6c6868547R169, which is more recent than the v6.18.0 build I was using.

louwers:
Looks good, happy to merge.

The macOS code is not shared with iOS, only `platform/darwin` is.

michaelkirk:
I still haven't been able to build and test the macos platform code, but feel free to merge it if you are confident.

louwers:
@michaelkirk Feel free to hit me up on Slack if you need help with that.

PR #3916, opened 2025-10-29
Throw exception on `System.loadLibrary` fail

adrian-cojocaru:
Catching the `UnsatisfiedLinkError` thrown by `System.loadLibrary` hides the original exception (and message) while still crashing on the next native function call. This PR adds a toggle for the exception handling logic (disabled by default).

### DIFF
--- a/platform/android/MapLibreAndroid/src/test/java/org/maplibre/android/BaseTest.kt
+++ b/platform/android/MapLibreAndroid/src/test/java/org/maplibre/android/BaseTest.kt
@@ -1,0 +1,13 @@
+package org.maplibre.android
+
+import org.junit.BeforeClass
+
+open class BaseTest {
+    companion object {
+        @JvmStatic
+        @BeforeClass
+        fun setup() {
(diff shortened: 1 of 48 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/LibraryLoader.java
```diff
@@ -30,10 +31,17 @@ public static void setLibraryLoader(LibraryLoader libraryLoader) {
     loader = libraryLoader;
   }
 
+  /**
+   * Catch UnsatisfiedLinkErrors on load
+   */
+  public static void enableErrorHandling(boolean value) {
+    handleLoadError = value;
```
louwers, line 38: What is the reason for making it configurable?

adrian-cojocaru: Initially it was only to keep the exiting logic in place in case it's needed. After checking the CI it seems it might be needed for tests to pass (the library isn't loaded).

PR #3919, opened 2025-10-30
Add Android core library builds

Turbo87:
As suggested in https://github.com/maplibre/maplibre-native-rs/issues/98, I've attempted to build the core library for Android devices.

It looks like I've managed to run `maplibre_native::set_log_thread_enabled(true)` via maplibre-native-rs with the build artifacts from this, so this does not seem to be completely wrong, but my experience in this area is admittedly limited.

### DIFF
--- a/.github/workflows/core-release.yml
+++ b/.github/workflows/core-release.yml
@@ -146,15 +146,6 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Set up NDK
-        uses: nttld/setup-ndk@afb4c9964b521afb97c864b7d40b11e6911bd410 # v1.5.0
-        with:
-          ndk-version: r28c
-          add-to-path: true
-
-      - name: Set NDK environment
-        run: echo "ANDROID_NDK_ROOT=\"$ANDROID_NDK_HOME\"" >> "$GITHUB_ENV"
-
       - uses: taiki-e/install-action@d6912b47771be2c443ec90dbb3d28e023987e782 # v2.62.1
         with:
           tool: armerge@2


Turbo87:
I fail to understand what `actionlint` wants from me. I've tried four different variations now, but it's still not happy. Any clue how to resolve this? 🤔

louwers:
You need to quote all `"$variables"`.

.github/workflows/core-release.yml
```diff
@@ -135,6 +135,57 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        renderer: [opengl, vulkan]
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
+        with:
+          submodules: recursive
+
+      - name: Set up NDK
```
louwers: No need, NDK is pre installed I think.

.github/workflows/core-release.yml
```diff
@@ -135,6 +135,57 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
 
+  build-android:
```
louwers, line 138: Could this be combined with the linux build job above by adding more matrix variants?

.github/workflows/core-release.yml
Turbo87: indeed, looks like it works without it :)

Turbo87: and that also resolves the actionlint issue 🎉

.github/workflows/core-release.yml
Turbo87: I guess technically it could, but the same could be said for the macos job above. I think it would probably become to complex to read/unterstand if we combined all of them into the same job.

louwers: Fair point.

I think it looks similar enough though!

If you want I can try it, let me know.

Turbo87: feel free to give it a try

Turbo87:
> I am not actually sure this will actually work. Did try it out?

as I said in https://github.com/maplibre/maplibre-native-rs/issues/98#issuecomment-3467394241:

> I think I managed to integrate it in maplibre-native-rs too, but so far I've only run maplibre_native::set_log_thread_enabled(true) to verify that it doesn't crash immediately.

in other words: not sure 😅

louwers:
Note that maplibre-native-rs currently doesn't support real time rendering. It only supports rendering static maps and tiles (mainly so we can add support for that to [Martin](https://github.com/maplibre/martin)). Adding support for Android and real time rendering will require significantly more leg work. And on top of that you will need to build out the entire Rust API for interacting with the C++ Core.

If you want to build an Android app you are likely much better served by using the MapLibre Android platform, MapLibre Compose, MapLibre React Native, MapLibre Flutter or MapLibre Qt.

Turbo87:
> If you want to build an Android app you are likely much better served by using the MapLibre Android platform, MapLibre Compose, MapLibre React Native, MapLibre Flutter or MapLibre Qt.

yeah, I realized that by now. I've pivoted for the app to using https://v2.tauri.app/ and maplibre-gl-js instead of the native renderer.

louwers:
@Turbo87 Sounds like a sensible choice!

I will close this for now, if we distribute Android builds we should have some sort of build verification in place.

Turbo87:
yeah, makes sense. no worries :)

Issue #3935, opened 2025-11-10
Crash `org.maplibre.android.maps.Projection.getContentPadding` reported by Grab


louwers:
### MapLibre Android Version

?

### Android SDK Version

?

### Device

?

### What happened?

<img width="2656" height="1512" alt="Image" src="https://github.com/user-attachments/assets/deee17f4-3248-428f-a9ab-c4abec904529" />

### Steps to reproduce

?

### Relevant log output

```shell

```

PR #3937, opened 2025-11-10
Add null check to Projection.getContentPadding()

louwers:
Grab is reporting a crash related to `padding` being `null` here.

Would be good to get a fix out. I don't know what combination of factors is causing `null` to be returned here, but let's add a defensive check here to make sure it doesn't crash. #3935


grab-liyanjin:
LGTM, thanks, Bart

PR #3942, opened 2025-11-12
Disable `UnsatisfiedLinkError` during local tests

adrian-cojocaru:
Follow-up to https://github.com/maplibre/maplibre-native/pull/3916. 
Detect and disable `UnsatisfiedLinkError` when running in an environment that doesn't require the native library.

### DIFF
--- a/platform/android/MapLibreAndroid/src/test/java/org/maplibre/android/BaseTest.kt
+++ b/platform/android/MapLibreAndroid/src/test/java/org/maplibre/android/BaseTest.kt
@@ -7,7 +7,7 @@ open class BaseTest {
         @JvmStatic
         @BeforeClass
         fun setup() {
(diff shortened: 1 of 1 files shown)


platform/android/MapLibreAndroid/src/test/java/org/maplibre/android/BaseTest.kt
```diff
@@ -7,7 +7,7 @@ open class BaseTest {
         @JvmStatic
         @BeforeClass
         fun setup() {
-            LibraryLoader.enableErrorHandling(true)
+            //LibraryLoader.enableErrorHandling(true)
```
louwers: Does it make sense to keep this?

louwers (approved):

PR #3958, opened 2025-11-19
Adds plugin protocol handlers

AtlasProgramming:
### Protocol Handler
This adds support similar to plugin-layers and style filters by exposing a data loading paradigm via a plug-in class and lambda implementation.  This does not create or implement a way to chain FileSources together.

```
```

```

        return YES;
    }
    return NO;
}

    response.data = data;
    return response;
    
}

@end
```

```

    }


        // Return some data here
        return null;
    }

}
```

```
mapView.getMapAsync {
```


sjg-wdw:
Looks reasonable.  Do the MapLibreJS folks have any thoughts?  For some reason I thought they had a way of doing this too.

PR #3560, opened 2025-06-16
Expose sync methods in GeoJsonSource 

grab-liyanjin:
Synchronized methods can avoid [the concurrent issue mentioned](https://github.com/maplibre/maplibre-native/issues/3485) , all changes are new, so it won't impact the old flows

PR #3968, opened 2025-11-24
Implement synchronous updates for GeoJSON sources

alexcristici:
Synchronous update for GeoJSON source will fix issues related to updating features per frame smoothly.

iOS usage: `MLNShapeSource(..., options: [.synchronousUpdate: true])`
Android usage: `GeoJsonSource(..., new GeoJsonOptions().withSynchronousUpdate(true))`

It is recommended to be used only for small updates, otherwise it will affect the map rendering frame rate.

One example is SymbolLocationIndicator:

Before:

https://github.com/user-attachments/assets/c13bd156-2f36-4a61-a6f0-e5a000479739

After:

https://github.com/user-attachments/assets/4e70508c-3534-411e-ac0a-28033ba9c532

### DIFF
--- a/test/tile/geojson_tile.test.cpp
+++ b/test/tile/geojson_tile.test.cpp
@@ -87,7 +87,7 @@ TEST(GeoJSONTile, SynchronousUpdate) {
     tileParameters.isUpdateSynchronous = true;
     GeoJSONTile tile(OverscaledTileID(0, 0, 0), "source", tileParameters, data);
     Immutable<LayerProperties> layerProperties = makeMutable<CircleLayerProperties>(
-            staticImmutableCast<CircleLayer::Impl>(layer.baseImpl));
+        staticImmutableCast<CircleLayer::Impl>(layer.baseImpl));
     std::vector<Immutable<LayerProperties>> layers{layerProperties};
     tile.setLayers(layers);
     ASSERT_TRUE(tile.isComplete());


louwers (commented):
Could you provide some more context in the PR description, and maybe the video you shared?

sjg-wdw:
We did discuss that change.  I think the conclusion was it was only partly synchronous as it doesn't reach all the way down into the renderer.  But it won't hurt if @alexcristici takes a closer look.

louwers:
@sjg-wdw With the current API it looks like when you are not using `setUpdateSynchronous(true)`, then `setGeoJsonSync()` is still using that old path. I think that is confusing. Ideally we align the two approaches and undo some of the changes in #3560 (particularly those to [`geojson_source.cpp`](https://github.com/maplibre/maplibre-native/pull/3560/files#diff-366d7b22496345a8a77afeaeedea24fc1ac94c5a8ae3c2fab9b2fb99d7499940)) if Alex' approach is better.

Maybe we don't even need to change the API?

alexcristici:
The difference between https://github.com/maplibre/maplibre-native/pull/3560

```
    /**
     * Updates the GeoJson with a single geometry. The update is performed synchronously,
     * so the data will be immediately visible and available to query when this method returns.
     *
     * @param geometry the GeoJSON [Geometry] to set
     */
    fun setGeoJsonSync(geometry: Geometry?) {
```

and this PR

```
    /**
     * Set a flag defining whether or not the fetched tiles for the given source
     * should be synchronously updated on the render thread
     *
     * @param value current setting for synchronous.
     */
    fun setUpdateSynchronous(value: Boolean) {
```

First refers only to `setGeoJson` to be sync, to make the data available synchronously, but the second one it's more deeper, includes first one and force also the tile updates for this specific source to be updated on the render thread.

Now we have 3 ways to set the geojson data:
`setGeoJson(feature)`
`setGeoJsonSync(feature)`
`setGeoJsonAsync(feature)`

where

```
    fun setGeoJson(feature: Feature?) {
        if (detached) {
            return
        }
        if (isUpdateSynchronous()) {
            setGeoJsonSync(feature)
        } else {
            setGeoJsonAsync(feature)
        }
    }
```

My initial idea was to remove `setGeoJsonSync(feature)` and `setGeoJsonAsync(feature)`, but I was not sure if there is an use case when somebody wants to have the tile updates async but only the response of `setGeoJson` to be sync, so that's why I kept all 3. 

I wanted to bring that up, so maybe we should remove those 2 and keep only `setGeoJson(feature)` which depends on `isUpdateSynchronous()`?

louwers:
@alexcristici What if we keep the API the same, but set sync/async mode internally depending on which method is called? Would that work?

alexcristici:
I don't think it will be correct, because setGeoJsonSync refers only to the GeoJson data update, compare to a state of the source to update tiles on render thread. 

So setting setGeoJsonSync and then setGeoJsonAsync it will create the impression that one update will be sync and one async, which will not be true because the second call will replace the source sync state to async.

In order to work like this I have to create a state in GeoJsonData, instead of a state per source.

alexcristici:
Maybe one idea is to add the source update sync mode in source options when it's created, instead exposing new methods like setUpdateSynchronous(bool).

adrian-cojocaru:
Can the `setUpdateSynchronous` value be flipped after the source is added to the map? Adding the sync mode in source options/constructor does sound cleaner if not.

alexcristici:
Yes, the sync state could be changed but this will propagate async, so leaving the option to change it needs to be explicit and to make the dev be aware of this.

Otherwise we could limit to a state that cannot be changed, by putting it in options/constructor.

louwers:
It sounds like mixing synchronous and asynchronous updates is something you would not want to do anyway, so adding an option to the constructor sounds like a good solution to me.

The question remains if we want to keep `setGeoJsonSync`? Otherwise we can mark it as deprecated, I wouldn't remove it for now.

alexcristici:
We should keep it only if somebody wants the tile update async but only this call `setGeoJsonSync` to be synced. Probably deprecated sound the best solution.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/sources/GeoJsonSource.kt
```diff
@@ -266,22 +276,44 @@ class GeoJsonSource : Source {
...
+        }
+    }
 
     /**
      * Updates the GeoJson with a single feature. The update is performed synchronously,
      * so the data will be immediately visible and available to query when this method returns.
      *
      * @param feature the GeoJSON [Feature] to set
      */
+    @Deprecated("use {@link #setGeoJson(Feature)} instead.")
```
louwers: I think a more accurate deprecation message would be: "use `withSynchronousUpdate` to enable synchronous updates on construction". Or something similar.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/sources/GeoJsonSource.kt
```diff
@@ -228,36 +228,46 @@ class GeoJsonSource : Source {
...
+        } else {
+            nativeSetFeature(feature)
+        }
     }
 
     /**
-     * Updates the GeoJson with a single geometry. The update is performed asynchronously,
-     * so the data won't be immediately visible or available to query when this method returns.
+     * Updates the GeoJson with a single geometry.
+     * The update is performed synchronously or asynchronously, based on the source synchronous update flag.
```
louwers, line 252: Could you keep the original text.  Something like

> In the case of asynchronous updates, the data won't be immediately visible or available to query when this method returns.

louwers (commented):
Could you add some test coverage?

Much of the changed / updated C++ code has existing tests. For Android the GeoJsonSource also had extensive tests.

Maybe we can also update the test app to use synchronous updates (or randomly enable them)?

sjg-wdw:
A specific test might be good, but you probably don't want to randomly enable the option on any given GeoJSON source.

alexcristici:
The test app uses synchronous updates when the non specialised location indicator is visible and this is already happening in the Long running Maps test.

alexcristici:
I updated the LocationLayerControllerTest.kt to use the synchronous source.

sjg-wdw:
How are we feeling about this one?  It would be nice to report it was merged, but if it breaks something that's obviously not good.

alexcristici:
From my side should be good to go.
Waiting for more reviews. 👍

louwers:
In terms of test coverage, I am not really sure what `LocationLayerControllerTest.kt` tests. Does it actually perform synchronous GeoJSON updates?

Does it make sense to to add a test in `test/tile/geojson_tile.test.cpp` that performs a synchronous `GeoJSONTile` update? And just checks if the data is immediately available? For Android, I think [`GeoJsonSourceTests.java`](https://github.com/maplibre/maplibre-native/blob/f25a4581fefea2e38521ec9be17b3c558137160c/platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/testapp/style/GeoJsonSourceTests.java) could be updated to (also) add synchronous GeoJSON sources.

Since this implementation replaces [#3560](https://github.com/maplibre/maplibre-native/pull/3560) maybe we should create a tracking issue to clean up the C++ code when we remove those methods? 

Other than that I think this is good to go. Thanks!

alexcristici:
This implementation is not replacing https://github.com/maplibre/maplibre-native/pull/3560, it's using that code. We will remove only the exposed methods.

alexcristici:
Regarding the tests, this synchronous update is not only about having the data available when the method returns, it's mostly about updating tiles on render thread, instead worker thread.

louwers:
> This implementation is not replacing https://github.com/maplibre/maplibre-native/pull/3560, it's using that code. We will remove only the exposed methods.

OK makes sense, thanks!

> it's mostly about updating tiles on render thread, instead worker thread.

@alexcristici Right, that was one suggestion to also test the implementation. It would be good to have the new code path somewhat covered by tests. A call to `tile.updateData(data, false, true);`, a variant of the Android GeoJSON tests that adds a synchronous source. If you want I can add it, and have you look over it, and then we can merge it.

alexcristici:
I updated `GeoJsonSourceTests.java` to use `new GeoJsonOptions().withSynchronousUpdate(true)` instead of deprecated `setGeoJSONSync`.

include/mbgl/actor/optional_actor_ref.hpp
```diff
@@ -0,0 +1,40 @@
...
+        if (objectRef) {
+            (objectRef->*fn)(std::forward<Args>(args)...);
+        } else if (actorRef) {
+            actorRef->invoke(fn, std::forward<Args>(args)...);
+        }
+    }
+
+private:
+    Object* objectRef = nullptr;
+    std::optional<ActorRef<Object>> actorRef;
```
adrian-cojocaru, line 37: `objectRef` can be updated to a `std::optional<std::reference_wrapper<Object>>` to have the same structure as the `actorRef` and avoid using a raw pointer.

alexcristici:
@louwers I implemented the `TEST(GeoJSONTile, SynchronousUpdate)` in `geojson_tile.test.cpp`.

include/mbgl/actor/optional_actor_ref.hpp
alexcristici: `ActorRef` is using a raw pointer, so I kept the same pattern.

alexcristici: Changing like this it will force me to update the constructor like this:
`objectRef = std::make_optional<std::reference_wrapper<Object>>(&object);`
I don't see the benefit in doing this.

louwers (approved):
Thanks!

include/mbgl/actor/optional_actor_ref.hpp
adrian-cojocaru: It was mostly just to avoid the raw pointer usage. I think `objectRef  = object` should work in the constructor. Just a small nit, so it can be safely ignored.

radityagumay:
@alexcristici thank you so much for the improvement. i do have question on this line

> It is recommended to be used only for small updates, otherwise it will affect the map rendering frame rate.

are you referring to the small updates on the location updates?

alexcristici:
> @alexcristici thank you so much for the improvement. i do have question on this line
> 
> > It is recommended to be used only for small updates, otherwise it will affect the map rendering frame rate.
> 
> are you referring to the small updates on the location updates?

No, I'm referring to the GeoJsonSource (that you create) to contain small data that needs to be updated, because if is synchronous the data for this GeoJsonSource will not be processed on a background thread, but instead of the rendering thread. So if the GeoJsonSource contains big data that needs to be updated could end up in affecting the render thread.

The SymbolLocationIndicator, not specialized one, is already using synchronous update and should be fine because its GeoJsonSource is small so the updates it triggers should not affect the rendering fps.

radityagumay:
understood, @alexcristici thank you

PR #3974, opened 2025-11-27
Include static XCFramework in GitHub iOS release

louwers:
First need to update it here before I can update https://github.com/maplibre/maplibre-gl-native-distribution/blob/main/Package.swift


.github/workflows/ios-ci.yml
```diff
@@ -360,7 +366,7 @@ jobs:
             -d '{"ref":"main","inputs":{
               "changelog_url": "https://maplibre-native.s3.eu-central-1.amazonaws.com/changelogs/ios-${{ env.version }}.md",
               "version":"${{ env.version }}",
-              "download_url": "https://github.com/maplibre/maplibre-native/releases/download/ios-v${{ env.version }}/MapLibre.dynamic.xcframework.zip"'
+              "download_url": "https://github.com/maplibre/maplibre-native/releases/download/ios-v${{ env.version }}/MapLibre.dynamic.xcframework.zip"}}'
```
mwilsnd, line 369: Extra `}}`, intentional?

louwers: Yes... That is an unrelated but needed fix.

mwilsnd (approved):

PR #3996, opened 2025-12-09
Fix pure virtual function call v2

adrian-cojocaru:
Follow-up to https://github.com/maplibre/maplibre-native/pull/3979.

`VectorTile` class has a `TileLoader` member that goes out of scope in the base class, after `~VectorMVTTile` / `~VectorMLTTile` is called and can still handle requests before its parent gets destroyed. Incoming requests in this time frame resolve `loadData` calls to it's pure virtual definition.

- This PR enforces the `TileLoader` destruction in the derived classes
- Since the object is being destroyed (tile was marked as obsolete), an alternative fix is to change `VectorTile::loadData` to an empty implementation and ignore the response data.

### DIFF
--- a/src/mbgl/tile/vector_mlt_tile.cpp
+++ b/src/mbgl/tile/vector_mlt_tile.cpp
@@ -16,6 +16,7 @@ VectorMLTTile::VectorMLTTile(const OverscaledTileID& id_,
     : VectorTile(id_, std::move(sourceID_), parameters_, tileset_, observer_) {}
 
 VectorMLTTile::~VectorMLTTile() {
+    // this needs to be explicitly deleted in the most-derived destructor
     loader.reset();
 
     // Don't rely on `~TileLoader` to close, it's not safe to call there.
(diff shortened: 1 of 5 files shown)


src/mbgl/tile/vector_mlt_tile.cpp
```diff
@@ -16,6 +16,8 @@ VectorMLTTile::VectorMLTTile(const OverscaledTileID& id_,
     : VectorTile(id_, std::move(sourceID_), parameters_, tileset_, observer_) {}
 
 VectorMLTTile::~VectorMLTTile() {
+    loader.reset();
```
louwers, line 20: Why does this need to be called manually?

adrian-cojocaru: To enforce the order and stop any pending request in the first destructor of the inheritance chain. If it remains tied to the parent scope it will get deleted in `~VectorTile` instead of `~VectorMLTTile`

PR #3998, opened 2025-12-09
update linux-ci to Ubuntu 24.04, apply clang-tidy fixes

louwers:
Most of them are quite pedantic, but these fixes are needed so that I can update clang-tidy.

### DIFF
--- a/platform/default/src/mbgl/storage/online_file_source.cpp
+++ b/platform/default/src/mbgl/storage/online_file_source.cpp
@@ -125,7 +125,7 @@ public:
 
     void activateOrQueueRequest(OnlineFileRequest* req) {
         assert(allRequests.contains(req));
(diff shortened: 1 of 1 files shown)


platform/default/src/mbgl/storage/online_file_source.cpp
```diff
@@ -124,8 +124,8 @@ class OnlineFileSourceThread {
     }
 
     void activateOrQueueRequest(OnlineFileRequest* req) {
-        assert(allRequests.find(req) != allRequests.end());
-        assert(activeRequests.find(req) == activeRequests.end());
+        assert(allRequests.contains(req));
+        assert(activeRequests.contains(req));
```
louwers:
```suggestion
        assert(!activeRequests.contains(req));
```

PR #4007, opened 2025-12-11
Remove MLN_CORE_INCLUDE_DEPS add MLN_CREATE_AMALGAMATION

louwers:
This PR removes the `MLN_CORE_INCLUDE_DEPS` CMake flag, since CMake's object libraries don't really work reliably. Instead we decided to use armerge. Still need to figure out what to do with Windows.

Not happy how we are doing 'core' releases. We need to do some build verification to make sure the libraries we put out even work...

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -322,6 +322,11 @@
       "targets": ["mbgl-core"]
     },
     {
+      "name": "windows-opengl-core",
+      "configurePreset": "windows-opengl",
+      "targets": ["mbgl-core"]
+    },
+    {
       "name": "windows-egl-core",
       "configurePreset": "windows-egl",
       "targets": ["mbgl-core"]


cmake/clang-tidy.cmake
```diff
@@ -1,6 +1,6 @@
 if (MLN_WITH_CLANG_TIDY OR MLN_CLANG_TIDY_COMMAND)
     if(MLN_CLANG_TIDY_COMMAND)
-        set(CLANG_TIDY_COMMAND MLN_CLANG_TIDY_COMMAND)
+        set(CLANG_TIDY_COMMAND "${MLN_CLANG_TIDY_COMMAND}")
```
louwers, line 3: unrelated change, but this would try to call a binary `MLN_CLANG_TIDY_COMMAND` otherwise when set.

Issue #4004, opened 2025-12-10
Update node npm release to Trusted Publishing


birkskyum:
This token workflow won't work anymore for the node package, because npm has sunset the classic tokens

https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/

https://github.com/maplibre/maplibre-native/blob/0dd94eb7057ec36aadce2425d6861ffdafeb0f6f/.github/workflows/node-release.yml#L303-L319

See migrations like:

- https://github.com/maplibre/maplibre-gl-js/pull/6840
- https://github.com/maplibre/maplibre-gl-inspect/pull/341
- https://github.com/maplibre/ngx-maplibre-gl/pull/262

HarelM:
Recent changes to npm security requires the following changes in each npm related repo:
1. Update node to version 24.11 or later
2. Change the url in package.json to match the repo address (without git+ and .git, etc)
3. Workflow permissions need to be 
```
permissions:
  id-token: write  # Required for trusted publishing
  contents: write
```

Since these need to change, I want to see if I can leverage the opportunity to consolidate the release workflows for `@maplibre` npm packages.
For this I have created two workflows to allow easy publishing of npm packages.
The workflows can be found here:
https://github.com/maplibre/reusable-workflows/tree/main/.github/workflows

1. A create version bump PR which changes the npm version in the package json, updates the changelog and creates a PR
2. a release workflow that should be ran on every merge to main in order to check if the version has changed and release a new version if needed.

This flow is the standard flow I'm using for most of the repos I maintain under maplibre.
Let me know if this can work for you as well.

HarelM:
There's another part in my message above to consider consolidating some workflows...

acalcutt:
The node release already does 2 at least, so that should be all set. Right now there isn't a version bump script though. Usually I follow our steps in https://github.com/maplibre/maplibre-native/blob/main/platform/node/RELEASE.md and use it as a chance to review the changelog as part of the PR.

HarelM:
That's what I'm doing as well, the bump version PR simply opens a PR for review of the changelog and updates the package json and lock file. 
The advantage is that it helps avoid to forget to do either of the above. 
I also want to write something to update the changelog from github PRs and allow attribution to contributors as part of this action. 
In any case I might be able to open a PR once I test everything out if you are interested.

acalcutt:
[...]

PR #4010, opened 2025-12-14
[Node ] Add create version bump PR workflow

acalcutt:
Add a version bump workflow for the node release as discussed in https://github.com/maplibre/maplibre-native/issues/4004

It is based on the workflows at https://github.com/maplibre/reusable-workflows/tree/main/.github/workflows , but is modified to allow it to create a PR on whatever branch it was run from (which is helpful when I make test releases in my fork, which I usually do in a branch)

Tested in https://github.com/WifiDB/maplibre-native/actions/runs/20216335275/job/58029875149

### DIFF
--- a/.github/workflows/node-create-bump-version-pr.yml
+++ b/.github/workflows/node-create-bump-version-pr.yml
@@ -1,4 +1,4 @@
-name: Create Node.js bump version PR
+name: node-create-bump-version-pr
 on:
   workflow_dispatch:
     inputs:


HarelM:
I'm wondering if there's something we can change in the reusable workflow to avoid the need to copy the code from there.
If this is something you're interested in do let me know and we can solve this together.

HarelM:
This looks good nonetheless and aligns the methodology of other repos, so definitely a step in the right direction.

acalcutt:
I think the reusable version bump workflow could work since you allow specifying folder and changelog, but because this package isn't in root having to enter all the paths is a bit of a pain (not a huge one but it seems like we are trying to remove manual entry with this)

The things I had to change to work with our package were small

1.) When running 'npm publish' I need to set preid  to 'pre' for node-pre-gyp-github to set the prerelease flag when publishing to gitbub. There was no way to set this in the other workflow

2.)  I did not like that the reusuable workflow uses a string for version type and changed it to use a combo box

3.) When testing on my fork I needed to be able to push from a non-main branch. I updated the workflow to checkout the branch it was run from and get the heading for the changelog based on that branch as well. It makes a PR against the branch it was run from.

3 also brings up another concern I have. If this is made to use a re-useable workflow, can it still be run on a non-maplibre fork?

----

For the release workflow I don't think it would be possible to use the reusable workflow with this package, just because of the way it needs to build binaries for each platform. it isn't really a standard package since it uses the binary building and node-pre-gyp

acalcutt:
If possible I wouldn't mind moving to this first so we can try it with a test pre-release, then finally push out 6.3.0 since I figured out the issue I was having testing it was not related to maplibre-native but was a canvas issue.

HarelM:
I agree the release flow is probably not something that can be shared. 
Regarding the bump version maybe we can specify the base folder to avoid a lot of other configurations. 
Regarding the version combo (2), you can take a look at the other repos I migrated, there's a need for input there, which is a combo, but the shared workflow receives a string, so this is obviously a non-issue. 
In any case, I'm open to change requests for the reusable workflows. Let me know how you want to play it.

HarelM:
Oh, and regarding the branch, we can probably omit the hard coded "main" to facilitate running this on other branches.

PR #4015, opened 2025-12-17
Fix issue caused by frustum offset when map is resized

alexcristici:
Fix scissor rect when frustum offset is 0.

image

### DIFF
--- a/src/mbgl/renderer/render_target.cpp
+++ b/src/mbgl/renderer/render_target.cpp
@@ -71,9 +71,8 @@ void RenderTarget::render(RenderOrchestrator& orchestrator, const RenderTree& re
                                                                   .clearDepth = {},
                                                                   .clearStencil = {}});
 
-    const gfx::ScissorRect prevSicssorRect = parameters.scissorRect;
-    const auto& size = getTexture()->getSize();
-    parameters.scissorRect = {.x = 0, .y = 0, .width = size.width, .height = size.height};
+    const gfx::ScissorRect prevScissorRect = parameters.scissorRect;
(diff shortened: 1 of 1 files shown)


adrian-cojocaru (approved):

cgalvan (approved):

src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -188,16 +188,20 @@ void Renderer::Impl::render(const RenderTree& renderTree, const std::shared_ptr<
     const TransformState& state = renderTreeParameters.transformParams.state;
     const Size& size = state.getSize();
     const EdgeInsets& frustumOffset = state.getFrustumOffset();
-    const gfx::ScissorRect scissorRect = {
-        .x = static_cast<int32_t>(frustumOffset.left() * pixelRatio),
+    gfx::ScissorRect scissorRect = {.x = 0, .y = 0, .width = 0, .height = 0};
```
yousifd, line 191: I think setting the scissor rect to all 0 for vulkan breaks things. I think we can fix this by checking if the scissor rect is set to all 0 when calling `buffer->setScissor(0, scissorRect, dispatcher);` in `PipelineInfo::setDynamicValues`

We also have to add the same check to the webgpu drawable around `wgpuRenderPassEncoderSetScissorRect`.

I was initially trying to avoid that by setting the the scissor to the screen size at the start but I guess that breaks in some situations.

alexcristici: Yes, only OpenGL platform disable scissor test for rect 0, so I have to add this check for other platforms as well: vulkan, metal, webgpu.

PR #4018, opened 2025-12-22
Vulkan backend: setSize also for the texture

Murmele:
otherwise the wrong size of the image gets copied.
This does not solve the problem completely because the map is not completely shown, only the text

grafik

### DIFF
--- a/.github/workflows/gh-pages-mdbook.yml
+++ b/.github/workflows/gh-pages-mdbook.yml
@@ -23,8 +23,8 @@ jobs:
       - name: Install Dependencies
         shell: bash
         run: sudo apt-get install -y libwayland-dev libxkbcommon-dev # Required for winit
-      - name: Install mdbook-alerts, mdbook-mermaid
-        run: cargo install mdbook-alerts mdbook-mermaid
+      - name: Install mdbook-mermaid
+        run: cargo install mdbook-mermaid
       - name: Build
         working-directory: docs/mdbook
         shell: bash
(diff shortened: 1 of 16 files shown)


adrian-cojocaru:
Moving the texture resize outside it's initialization [block](https://github.com/maplibre/maplibre-native/blob/1f99707581fc555517e877cb0eceb9c59a3e5ff0/platform/default/src/mbgl/vulkan/headless_backend.cpp#L84) is enough to resize it on demand when a read is requested (`setSize` can be called multiple times per frame without a read). https://github.com/maplibre/maplibre-native/pull/4030

From my testing this puts it on par with other backends, but without the issue in the description. If you can provide a repro I'll be happy to take a look.

Murmele:
@adrian-cojocaru Thanks for viewing into it. So this PR is anymore required with your PR? I will try to reproduce. It was during bringing maplibre-native for rust

adrian-cojocaru:
They're mostly the same so it's one or the other, but the `Texture2D::setSize` call triggered some asserts that needed fixing.

Murmele:
@adrian-cojocaru Thanks for viewing into it. So this PR is anymore required with your PR? I will try to reproduce. It was during bringing maplibre-native for rust

louwers:
Closing since @adrian-cojocaru's PR got merged.

Let us know if it indeed solves your problem @Murmele.

Murmele:
> Closing since @adrian-cojocaru's PR got merged.
> 
> Let us know if it indeed solves your problem @Murmele.

Thanks I will try out next days

PR #4020, opened 2025-12-24
Metal backend resource for MacOS fixes

callmeahab:
I tried creating a custom layer on MacOS and had issues with missing backend resource for Metal renderer, these fixes worked for me so thought I'd submit this fix to the main codebase.


louwers:
@callmeahab Thanks for your contribution!

PR #4028, opened 2026-01-05
Add multithreaded upload to dynamic texture on Vulkan [Core feature]

alexcristici:
`DynamicTexture` has been split into platform-specific implementations, keeping the existing functionality on Metal and OpenGL.
Vulkan implementation changed, so now it uploads the image on a worker thread to a temporary texture and then on the render thread it will `blit` the temporary texture to the dynamic texture region.

### DIFF
--- a/src/mbgl/vulkan/context.cpp
+++ b/src/mbgl/vulkan/context.cpp
@@ -179,9 +179,7 @@ void Context::submitOneTimeCommand(const vk::UniqueCommandPool& commandPool,
 #endif
 
     const vk::CommandBufferAllocateInfo allocateInfo(
-        commandPool ? commandPool.get() : backend.getCommandPool().get(),
-        vk::CommandBufferLevel::ePrimary,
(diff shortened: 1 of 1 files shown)


include/mbgl/vulkan/dynamic_texture.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <mbgl/gfx/dynamic_texture.hpp>
+#include <mbgl/vulkan/texture2d.hpp>
+
+#define DYNAMIC_TEXTURE_VULKAN_MULTITHREADED_UPLOAD 1
```
louwers, line 6: Do you want to keep this in or did you just use it for developing?

alexcristici: @adrian-cojocaru Mentioned that debugging is much easier when single threaded, so we will disable it locally when need it.

PR #4029, opened 2026-01-05
Disable scissor test when clearing on OpenGL

adrian-cojocaru:
Follow-up to https://github.com/maplibre/maplibre-native/pull/4015

OpenGL is the only supported backend in which the scissor test rect affects the render pass clear (and where the issue is present). 
Fix: disable the scissor test prior to clearing the framebuffer.

### DIFF
--- a/src/mbgl/renderer/renderer_impl.cpp
+++ b/src/mbgl/renderer/renderer_impl.cpp
@@ -188,26 +188,16 @@ void Renderer::Impl::render(const RenderTree& renderTree, const std::shared_ptr<
     const TransformState& state = renderTreeParameters.transformParams.state;
     const Size& size = state.getSize();
     const EdgeInsets& frustumOffset = state.getFrustumOffset();
-    gfx::ScissorRect scissorRect;
-
+    const gfx::ScissorRect scissorRect = {
+        .x = static_cast<int32_t>(frustumOffset.left() * pixelRatio),
 #if MLN_RENDER_BACKEND_OPENGL
-    if (frustumOffset.isFlush()) {
-        scissorRect = {.x = 0, .y = 0, .width = 0, .height = 0};
-    } else
-#endif
-    {
-        scissorRect = {
-            .x = static_cast<int32_t>(frustumOffset.left() * pixelRatio),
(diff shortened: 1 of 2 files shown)


src/mbgl/renderer/renderer_impl.cpp
```diff
@@ -188,8 +188,14 @@ void Renderer::Impl::render(const RenderTree& renderTree, const std::shared_ptr<
...
     const EdgeInsets& frustumOffset = state.getFrustumOffset();
-    gfx::ScissorRect scissorRect = {.x = 0, .y = 0, .width = 0, .height = 0};
-    if (!frustumOffset.isFlush()) {
+    gfx::ScissorRect scissorRect;
+
+#if MLN_RENDER_BACKEND_OPENGL
+    if (frustumOffset.isFlush()) {
+        scissorRect = {.x = 0, .y = 0, .width = 0, .height = 0};
+    } else
+#endif
```
alexcristici: The best way will be to avoid using `#if`, maybe we could unify how scissor works for all platforms.

Otherwise, I would say to name a macro that suggests what the code does and not use directly the backend macros. 
Like we have here:
`#define MLN_UBO_CONSOLIDATION (MLN_RENDER_BACKEND_METAL || MLN_RENDER_BACKEND_VULKAN || MLN_RENDER_BACKEND_WEBGPU)`

adrian-cojocaru: OpenGL is the only one where clears are limited by the scissor test. It also has the Y axis flip a few lines below so there are 2 differences. 
Adding a backend specific implementation for `ScissorRect` is one solution, but I was hoping we can find the general fix for GL frustum offset since the current one targets only the {0,0,0,0} case (disabled) and unify it with the rest this way.

Currently this is meant to fix some Vulkan errors triggered by some mismatches between pipeline state and `vkCmdSetScissor`.

adrian-cojocaru: Update: All backends will clear the full framebuffer (ignoring the scissor rect).

yousifd (approved):

PR #4030, opened 2026-01-06
Fix Vulkan headless backend resize

adrian-cojocaru:
Similar to https://github.com/maplibre/maplibre-native/pull/4018 this forces the read texture to resize and match the renderable resource. 
Minor rendering stats memory fixes exposed by `Texture2D::setSize`.


platform/default/src/mbgl/vulkan/headless_backend.cpp
```diff
@@ -75,16 +75,21 @@ gfx::Renderable& HeadlessBackend::getDefaultRenderable() {
...
     auto& resourceImpl = static_cast<HeadlessRenderableResource&>(*resource);
 
     if (!texture) {
         texture = std::make_unique<Texture2D>(contextImpl);
         texture->setFormat(gfx::TexturePixelType::RGBA, gfx::TextureChannelDataType::UnsignedByte);
-        texture->setSize(size);
         texture->setUsage(Texture2DUsage::Read);
     }
 
+    texture->setSize(size);
```
Murmele, line 91: Is it required to set it all the time, even when not changed?

adrian-cojocaru: `setSize` has a check and sets the dirty flag only if the new size is different than the current one.

louwers (approved):

CommanderStorm:
@louwers might it be possible for you to do a core release? I think this would help https://github.com/maplibre/maplibre-native-rs/pull/115

Murmele:
> @louwers might it be possible for you to do a core release? I think this would help [maplibre/maplibre-native-rs#115](https://github.com/maplibre/maplibre-native-rs/pull/115)

Would be nice, so I don't have to compile my self

louwers:
Yes will do, let me get one more fix in.

PR #4046, opened 2026-01-13
fix wrong result cameraForLatLngs when pitch != 0

louwers:
Fixes the tests added in #4040 by @NathanMOlson.

It ignores the padding when it is excessive to make sure `cameraForLatLngs` always includes the requested bounds.

This is entirely written by Claude and needs some more testing.

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -207,6 +207,30 @@ CameraOptions cameraForLatLngs(const std::vector<LatLng>& latLngs,
         return {};
     }
     Size size = transform.getState().getSize();
+
+    // When padding is excessive (exceeds viewport dimensions), ignore it entirely
+    // to ensure bounds can still be displayed
+    double effectivePaddingLeft = padding.left();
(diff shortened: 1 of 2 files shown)


src/mbgl/map/map.cpp
```diff
@@ -225,20 +249,15 @@ CameraOptions cameraForLatLngs(const std::vector<LatLng>& latLngs,
...
+        scaleY -= (effectivePaddingTop + effectivePaddingBottom) / height;
         minScale = util::min(scaleX, scaleY);
     }
 
     double zoom = transform.getZoom();
     if (minScale > 0) {
         zoom = util::clamp(
             zoom + util::log2(minScale), transform.getState().getMinZoom(), transform.getState().getMaxZoom());
-    } else {
-        Log::Error(Event::General,
```
louwers, line 238: Should we move this warning to the effective padding calculation?

louwers:
I tested it with StreetComplete, this needs a different approach.

Issue #3658, opened 2025-07-24
PMTiles runtime crash


cobra811:
### MapLibre Android Version

11.12.1

### Android SDK Version

24.1.1

### Device

Samsung s20

### What happened?

Runtime crash using hosted PMTiles source.
"terminating due to uncaught exception of type std::runtime_error: incorrect header check"

I did not expect a hard crash. I expected a graceful warning or error indicating the incorrect header check.

### Steps to reproduce

1. Use R2 storage to host a tippecanoe generated PMTiles source.
2. Add a PMTiles source through a hosted style.json.
3. Run the MapLibre app, navigate the map to load/view PMTiles, works fine.
4. Keep the app open and the screen on.
5. Generate a newly updated PMTiles source with tippecanoe and replace source on R2 with new source.
6. Navigate the app to load new PMTiles.
7. Crash (incorrect header check)

### Relevant log output

```shell
2025-07-21 01:54:21.841 24911-24970 libc++abi               ...               E  terminating due to uncaught exception of type std::runtime_error: incorrect header check
2025-07-21 01:54:21.841 24911-25994 TrafficStats            ...               D  tagSocket(191) with statsTag=0xffffffff, statsUid=-1
2025-07-21 01:54:21.841 24911-24970 libc                    ...               A  Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 24970 (PMTilesFileSour), pid 24911 (...)
```

louwers:
Do you have a PMTiles file that crashes you can share?

cobra811:
> Do you have a PMTiles file that crashes you can share?

Yes, i have two files, one original, one for the update. No single pmtiles file will cause a crash. It seems like a caching issue. When the pmtiles file is updated while a user has the map open, the next time the user pans/zooms in and out maplibre will crash. The pmt_test_2.pmtiles file name will need to be changed to pmt_test.pmtiles in order to replace the original when updating.

[pmt_test.zip](https://github.com/user-attachments/files/21440000/pmt_test.zip)

Yamayu0504:
Hi, what is the current status of this issue? 
We are experiencing this on iOS in our app as well. Thanks.

TimSylvester:
[...]

PR #4048, opened 2026-01-13
fix crash when source PMTile file changed

louwers:
Fix for #3658.

AI generated, still need to review this myself.


TimSylvester:
This only applies when the source is a local file.  That is one small part of the fix, but I expect pmtiles is not often used locally.

Detecting when the remote file changes is more difficult, especially when there are multiple requests being made in parallel, and various HTTP servers may produce a variety of headers for the same situation.

louwers:
@TimSylvester True, maybe it's better to just avoid the crash.

PR #3965, opened 2025-11-22
core: Update Hillshade Algorithms and add Color-Relief Layer support

acalcutt:
# Advanced Terrain Rendering: Multiple Hillshade Algorithms & Color-Relief Layer

## Overview

This PR introduces advanced terrain rendering capabilities to MapLibre Native, implementing multiple hillshade algorithms and a new color-relief layer type for hypsometric tinting. These features enable advanced cartographic workflows using terrain-RGB tiles.

## Closes

- Closes #3396 - Multiple hillshade algorithm options
- Closes #3408 - Hypsometric tint from terrain-RGB tiles

### 🏔️ Multiple Hillshade Algorithms

Added support for five hillshade rendering algorithms, matching GDAL's capabilities:

1. **Standard** (default) - MapLibre's traditional hillshade algorithm
2. **Basic** - Simple physics-based model (similar to GDAL default)
3. **Combined** - Combines slope and oblique shading for enhanced terrain visualization
4. **Igor** - Minimizes effects on underlying map features (based on [Maperitive's algorithm](http://maperitive.net/docs/Commands/GenerateReliefImageIgor.html))
5. **Multidirectional** - Combines illumination from multiple light sources (up to 4 independent light sources, based on [USGS methodology](http://pubs.usgs.gov/of/1992/of92-422/of92-422.pdf))

**Style API:**
```json
{
  "type": "hillshade",
  "paint": {
    "hillshade-method": "multidirectional",
    "hillshade-illumination-direction": [225, 270, 315, 0],
    "hillshade-illumination-altitude": [30, 30, 30, 30],
    "hillshade-highlight-color": ["#FF0000", "#00FF00", "#0000FF", "#FFFF00"],
    "hillshade-shadow-color": ["#000088", "#008800", "#880000", "#888800"],
    "hillshade-exaggeration": 0.5
  }
}
```

**Paint Properties:**
| Property | Type | Default | Description |
|----------|------|---------|-------------|
| `hillshade-method` | enum | `standard` | Algorithm selection (standard, basic, combined, igor, multidirectional) |
| `hillshade-exaggeration` | number | `0.5` | Height exaggeration factor |
| `hillshade-accent-color` | color | `#000000` | Accent color for flat areas |
| `hillshade-illumination-direction` | number[] | `[335]` | Light source direction(s) in degrees (1-4 values) |
| `hillshade-illumination-altitude` | number[] | `[45]` | Light source altitude(s) in degrees (1-4 values) |
[...]

PR #4050, opened 2026-01-14
Add script for validating PR title

louwers:
This script adds some requirement for PR titles. Now, the labels of a PR must match the PR title.

It is not very strict yet. For example if a PR has `android` label then Android must exist somewhere in the PR title.

This makes it easier to gather the correct commits for the changelog. Right now I use the `generate-changelog.mjs` script for that but it needs to read all PRs from the GitHub API which is slow and brittle compared to just using git directly. This PR also modifies that script to get rid of the Octokit dependency.

We already have an auto labeler in place that ensures PRs have the right labels.

One point of discussion is what labels should match the title. For now I just went with Qt, Node, iOS, Android and 'Core' for changes to the C++ Core (those are in principle relevant for all platforms).

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -69,7 +69,7 @@ jobs:
           node-version-file: ".nvmrc"
 
       - name: npm install
-        run: npm install --ignore-scripts
+        run: npm ci --ignore-scripts
         working-directory: .
         shell: bash
 


.github/scripts/validate-pr-title.ts
```diff
@@ -0,0 +1,38 @@
...
+const labelsString = process.env.PR_LABELS.toLowerCase();
+const labels = labelsString.split(',').map(l => l.trim()).filter(l => l);
+// these labels are the ones we care about
+const keywords = ['node', 'core', 'ios', 'android', 'qt'];
+
+const mismatches = [];
+for (const keyword of keywords) {
+  if (labels.includes(keyword) && !prTitle.includes(keyword)) {
+    mismatches.push(keyword);
+  }
```
adrian-cojocaru: If PRs touch core + multiple platforms (and get labeled as such by the auto labeler) doesn't this add a lot of requirements for the title? https://github.com/maplibre/maplibre-native/pull/3965 currently needs 4/5 keywords.

louwers: OK I made now only 'core' required when the PR has the 'core' label.

I also added a test.

Please have another look!

PR #4057, opened 2026-01-16
Distance-based tile LOD

NathanMOlson:
This PR adds an alternative tile LOD algorithm, which is off by default. The new algorithm scales tiles according to their distance from the camera (as opposed to the distance from the screen center). Changes were also made to allow requesting higher-than-nominal zoom tiles so that tiles in the foreground can be rendered at higher detail.

See also:

- https://github.com/maplibre/maplibre-gl-js/pull/4779
- https://github.com/maplibre/maplibre-gl-js/pull/5719
- #2958

Existing "reduced" profile:
image

Existing "aggressive" profile:
image

New distance-based algorithm with default parameters (note the higher detail at the bottom of the screen):
image


louwers (commented):
How does this compare to the MapLibre GL JS tile LOD algorithm?

Would it be possible to get the [debug tiles](https://demotiles.maplibre.org/debug-tiles/#9/0/0/0/70) working for MapLibre Native so we can add some render tests for LOD? Or maybe we can just use the existing debug layer.

Before this is merged we should ideally add relevant APIs for iOS and Android. If this is not relevant for your use case then I can assist there.

NathanMOlson:
> How does this compare to the MapLibre GL JS tile LOD algorithm?

The algorithm core is the same as GL JS, splitting tiles to achieve (as best as possible) `tileScale = k / distanceToTile * cos(tileOffNadirAngle)^N`.

GL JS provides control over the exponent `N` via the user-facing parameter `maxZoomLevelsOnScreen` and auto-scales `tileLodScale` as a function of pitch using the user-facing parameter `tileCountMaxMinRatio`. In the current Native implementation, I left `N` hardcoded to 1/2 and kept `tileLodScale` manual instead of auto.

As discussed in https://github.com/maplibre/maplibre-gl-js/pull/5719, I don't like the `maxZoomLevelsOnScreen` parameter. In my opinion it is not the best way  to adjust the behavior, I prefer to modify `N` directly, which I feel provides more understandable behavior. Also, in the current implementation `maxZoomLevelsOnScreen` is only accurate when tiles are very small compared to the viewport, meaning is practice it leads to questions like "why do I only see 5 zoom levels on the screen when I set `maxZoomLevelsOnScreen` to 9?".

> Would it be possible to get the [debug tiles](https://demotiles.maplibre.org/debug-tiles/#9/0/0/0/70) working for MapLibre Native so we can add some render tests for LOD? Or maybe we can just use the existing debug layer.

Yes, it's possible. GL JS has the number tiles [included in the repo](https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/assets/tiles/number), is that a good approach to take here as well?

louwers:
Since @alasram and @yousifd worked on LOD in the past, I wonder what you think about this implementation.

> Yes, it's possible. GL JS has the number tiles [included in the repo](https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/assets/tiles/number), is that a good approach to take here as well?

Yes, that would be a great approach I think!

louwers:
@NathanMOlson Are you able to join the TSC call today? Might be good to have a chat about this PR.

NathanMOlson:
I added render tests that use the various tile LOD parameters. There are small differences at the horizon on Android, for example: 
image

Is this something that needs to be fixed? Or can I resolve it by either reducing the required exactness, or adding Android-specific "expected" images?

louwers:
@NathanMOlson Awesome. I can take it from here and add the alternate Android expected images.

include/mbgl/map/map.hpp
```diff
@@ -194,6 +201,8 @@ class Map : private util::noncopyable {
     double getTileLodPitchThreshold() const;
     void setTileLodZoomShift(double shift);
     double getTileLodZoomShift() const;
+    void setUseDistanceBasedTileLod(bool val);
```
louwers: Sorry, I have one more API nit.

I think it would be a bit cleaner if we would change this API to just `setTileLodMode(TileLodMode)` and `getTileModMode`. What do you think?

NathanMOlson: Yup, I think that's an improvement. Done.

render-test/parser.cpp
```diff
@@ -612,6 +612,15 @@ TestMetadata parseTestMetadata(const TestPaths& paths) {
     return metadata;
 }
 
+mbgl::TileLodMode parseTileLodMode(const std::string& mode_str) {
```
NathanMOlson: Is there a standard way to do string-to-enum in Maplibre that I should follow?

louwers: Yes. You can search for `MBGL_DEFINE_ENUM` to see it.

louwers:
@NathanMOlson Thanks for another great contribution.

Good to merge from your side?

include/mbgl/map/map.hpp
louwers: Thanks!

NathanMOlson:
Yup, good to merge.

PR #4059, opened 2026-01-19
Make rendering stats thread safe [Core feature]

alexcristici:
Rendering stats are called from different threads, because dynamic texture uploads new images multithreaded on Metal and Vulkan.

### DIFF
--- a/.github/scripts/validate-pr-title.ts
+++ b/.github/scripts/validate-pr-title.ts
@@ -36,12 +36,12 @@ export function validateTitle({
 }
 
 if (import.meta.main) {
-  if (!process.env.PR_TITLE) {
+  if (process.env.PR_TITLE === undefined) {
     console.log('::error::PR_TITLE environment variable is not set');
     process.exit(1);
   }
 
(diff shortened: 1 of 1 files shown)


include/mbgl/gfx/context.hpp
```diff
@@ -93,6 +94,16 @@ class Context {
     gfx::RenderingStats& renderingStats() { return stats; }
     const gfx::RenderingStats& renderingStats() const { return stats; }
 
+    void threadSafeAccessRenderingStats(const std::function<void(RenderingStats&)>& function) {
+        std::scoped_lock lock(renderingStatsMutex);
+        function(stats);
+    }
+
+    RenderingStats threadSafeCopyRenderingStats() {
+        std::scoped_lock lock(renderingStatsMutex);
```
louwers: Could probably use

```suggestion
        std::shared_lock lock(renderingStatsMutex);
```

But not really performance critical code.

louwers:
Looks like you need a `std::shared_mutex` for that.

Issue #4054, opened 2026-01-15
Update iOS Objective-C API to allow setting roll, fov and centerAltitude


louwers:
#3011 added support for camera roll, and added some additional APIs for Android.

We should port those to iOS as well.

PR #4061, opened 2026-01-20
ios: add basic support for setting camera roll

louwers:
Adds basic support for setting a roll value on `MLNMapView`. Also adds an example.

Simulator Screenshot - BAZEL_TEST_iPhone Xs_18 5 - 2026-01-20 at 16 44 14

Partially completes https://github.com/maplibre/maplibre-native/issues/4054


cgalvan:
Are there unit tests we can add for this as well?

louwers:
@cgalvan Good call, let me check.

louwers:
@cgalvan Added some tests.

PR #4068, opened 2026-01-21
Split static and dynamic XCFramework build in separate jobs (`ios-release.yml`)

louwers:
It takes about 2 hours to build each, so by putting the builds in separate jobs the time to run a release is halved.

### DIFF
--- a/.github/workflows/ios-release.yml
+++ b/.github/workflows/ios-release.yml
@@ -65,6 +65,10 @@ jobs:
         working-directory: platform/ios
         shell: bash
     steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v4
+        with:
+          submodules: recursive
+
       - name: Build XCFramework (Static)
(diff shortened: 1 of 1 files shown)


.github/workflows/ios-release.yml
```diff
@@ -94,25 +153,15 @@ jobs:
...
-
       - name: Release (GitHub) - Create Draft
         id: github_release
         uses: softprops/action-gh-release@a06a81a03ee405af7f2048a818ed3f03bbf83c7b # v2.5.0
         with:
           name: ios-v${{ env.version }}
           files: |
-            MapLibre.dynamic.xcframework.zip
-            MapLibre.static.xcframework.zip
-            ${{ env.debug_symbols_ios }}
```
cgalvan, line 115: Are the debug symbols still being included?

louwers: Yes, the name is just hardcoded two lines down.

PR #4082, opened 2026-02-01
Fix incorrect shader selection for text labels with inline images

ciscorn:
Fixes the exact same issue and root cause as https://github.com/maplibre/maplibre-gl-js/pull/6956

- Prevent the feature loop from resetting the `iconsInText` flag once an inline image is seen
- Add the `text-field/formatted-images-mixed` render test

← Before fix | After fix →

diff image


TimSylvester (approved):

louwers:
@ciscorn Thanks for your contribution.

No need to merge main back in, it is not a requirement for merging to be up to date.

ciscorn:
Sorry. I sometimes get random mobile notifications about CI failures in this repo (maybe a GitHub bug?). I tried updating this branch in case it would fix it.

PR #4091, opened 2026-02-09
fix Android Vulkan instrumentation tests

adrian-cojocaru:
Fixed a few issues that were merged while the flaky [sequential tests](https://github.com/maplibre/maplibre-native/pull/4031) were running (making real issues hard to identify).


include/mbgl/vulkan/renderer_backend.hpp
```diff
@@ -19,7 +19,7 @@
 #define VMA_DYNAMIC_VULKAN_FUNCTIONS 0
 #include "vk_mem_alloc.h"
 
-#ifndef NDEBUG
+#if !defined(NDEBUG) && !defined(__ANDROID__)
```
louwers, line 22: Why is this only relevant for Android?

adrian-cojocaru: [HWASan](https://source.android.com/docs/security/test/hwasan) was generating some errors for it (nothing on other platforms).

PR #4092, opened 2026-02-09
Fix laggy map when using synchronous GeoJSON source [Core fix]

alexcristici:
Fixed synchronous GeoJSON source by using OptionalActor for converter, instead of waiting for Actor result and  blocking the thread on which it was called.

This PR replaces https://github.com/maplibre/maplibre-native/pull/3560.


louwers:
We will make a pre-release with this change.

Maybe @Williamrai can try it out in the WIkipedia app and @grab-liyanjin for Grab?

Will post here when it is released!

louwers:
Pre-release is out: https://github.com/maplibre/maplibre-native/releases/tag/android-v13.0.0-pre0

Williamrai:
> Pre-release is out: https://github.com/maplibre/maplibre-native/releases/tag/android-v13.0.0-pre0

@alexcristici @louwers Thank you for fixing this issue. I tested it on my end and didn’t notice any lag and it’s working as expected now.

louwers:
@Williamrai Thanks a lot for reporting back!

systemed:
Could you clarify how to use this on Android when not explicitly constructing a `GeoJsonSource`?

I have a source in my stylesheet defined like this:

```json
"route_highlights":{"type":"geojson","data":{"type":"FeatureCollection","features":[]}},
```

I then populate it like this:

```kotlin
val src = map.style?.getSource("route_highlights") as GeoJsonSource? ?: return
src.setGeoJsonSync(someFeatureCollection)
```

Because I'm not using the `GeoJsonSource` constructor, but obtaining the source via `map.style.getSource`, I can't see any way of passing the `.withSynchronousUpdate` option. (But I may be missing something.)

louwers:
I don't think is possible at the moment (anymore). @alexcristici can confirm.

Maybe we need a style spec property for this...

louwers:
Issue: https://github.com/maplibre/maplibre-native/issues/4187

PR #4093, opened 2026-02-10
Bake build ID into test app for iOS

louwers:
This is needed so that each release has a unique build identifier.

### DIFF
--- a/platform/ios/app-swift/BUILD.bazel
+++ b/platform/ios/app-swift/BUILD.bazel
@@ -63,7 +63,7 @@ ios_application(
     ],
     infoplists = [":Info-swift.plist"],
     minimum_os_version = "16.0",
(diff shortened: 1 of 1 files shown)


platform/ios/app-swift/BUILD.bazel
```diff
@@ -49,10 +63,11 @@ ios_application(
     ],
     infoplists = [":Info-swift.plist"],
     minimum_os_version = "16.0",
-    provisioning_profile = "xcode_profile",
+    provisioning_profile = "MapLibre.mobileprovision",
```
adrian-cojocaru: Does this break the Xcode generated profile when running locally?

louwers: Whoops, good catch!

cgalvan (approved):

PR #4097, opened 2026-02-12
Support for WebGPU backend in Android SDK

birkskyum:
Similar to 
- #4103 

This PR allow a user of Android SDK to use Vulkan through WebGPU as an alternative to the Vulkan backend. This PR:

- Adds WebGPU as a rendering backend on Android, with support for both Dawn and wgpu-native implementations
- New Gradle product flavors webgpuDawn and webgpuWgpu for library and test app
- Adds CMake presets android-webgpu-dawn and android-webgpu-wgpu
- Extends vendor/wgpu.cmake with Android cross-compilation support
- Adds docs for rendering backend selection and WebGPU build instructions

readFrameBuffer / Snapshot Activity
- Implements readFramebuffer() with BGRA swizzle, deferred present, and synchronous buffer mapping
- Fixes scissor validation errors from Vulkan swapchain extent mismatch

Screenshot 2026-02-12 at 11 25 32

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -399,9 +399,10 @@ jobs:
           LATEST_NDK=$(find "$ANDROID_SDK_ROOT/ndk" -mindepth 1 -maxdepth 1 -type d | sort -V | tail -n 1)
           NDK_PREBUILT="$LATEST_NDK/toolchains/llvm/prebuilt/linux-x86_64"
           RESOURCE_DIR=$("$NDK_PREBUILT/bin/clang" --print-resource-dir)
-          export BINDGEN_EXTRA_CLANG_ARGS=$(echo "${{ env.RUST_TARGET }}" | tr '-' '_')
+          BINDGEN_EXTRA_CLANG_ARGS=$(echo "${{ env.RUST_TARGET }}" | tr '-' '_')
+          export BINDGEN_EXTRA_CLANG_ARGS
           ENV_VAR="BINDGEN_EXTRA_CLANG_ARGS_$BINDGEN_EXTRA_CLANG_ARGS"
-          export $ENV_VAR="--sysroot=$NDK_PREBUILT/sysroot --target=${{ env.BINDGEN_TARGET }} -isystem $RESOURCE_DIR/include"
+          export "${ENV_VAR}=--sysroot=$NDK_PREBUILT/sysroot --target=${{ env.BINDGEN_TARGET }} -isystem $RESOURCE_DIR/include"
           cargo build --release --target ${{ env.RUST_TARGET }} --no-default-features --features vulkan,wgsl,spirv,glsl
 
       - name: Configure AWS Credentials


platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,565 @@
...
+    if (impl->currentTexture) {
+        wgpuTextureRelease(impl->currentTexture);
+        impl->currentTexture = nullptr;
+    }
+
+    configureSurface(impl->framebufferSize.width, impl->framebufferSize.height);
+}
+
+PremultipliedImage AndroidWebGPURendererBackend::readFramebuffer() {
+    // TODO not implemented
```
louwers: Will you still implement this?

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+    setDevice(static_cast<WGPUDevice>(impl->device));
+    setQueue(static_cast<WGPUQueue>(impl->queue));
+    setInstance(static_cast<WGPUInstance>(impl->instance));
+#endif
+    setDepthStencilFormat(depthStencilFormat);
+
+    createSurface();
+
+    int w = ANativeWindow_getWidth(window);
```
louwers: Can both be const.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+#include <cassert>
+
+#if MLN_WEBGPU_IMPL_WGPU
+#include <android/log.h>
+#include <unistd.h>
+#include <thread>
+
+static void wgpuLogCallback(WGPULogLevel level, WGPUStringView message, void* /*userdata*/) {
+    int androidLevel = ANDROID_LOG_INFO;
```
louwers: Can be const.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,565 @@
...
+            androidLevel = ANDROID_LOG_DEBUG;
+            break;
+        case WGPULogLevel_Trace:
+            androidLevel = ANDROID_LOG_VERBOSE;
+            break;
+        default:
+            break;
+    }
+    if (message.data && message.length > 0) {
+        __android_log_print(androidLevel, "wgpu-native", "%.*s", (int)message.length, message.data);
```
louwers: Prefer `static_cast`.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,565 @@
...
+    }
+}
+
+// Redirect stderr to logcat so Rust panic messages are visible
+static void startStderrToLogcat() {
```
louwers, line 49: Should be wrapped in a `std::call_once`.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
birkskyum: I didn't plan for it in this PR because the webgpu backend needs to be prepared for it, but I think it's a useful thing to have, so I'll circle back to it.

birkskyum:
> We should definitely get a build going on CI before merging this.

I've added android CI build tests for wgpu (17 min) and dawn (33 min)

platform/android/MapLibreAndroid/src/webgpu/java/org/maplibre/android/maps/renderer/surfaceview/MapLibreWebGPUSurfaceView.java
adrian-cojocaru, line 1: Are there differences between this and  `MapLibreVulkanSurfaceView` (can be extended to all `src/webgpu` files)? If not we can use a common `sourceSet` for both.

birkskyum: That's a good point - it now direct the webgpu source sets to the vulkan java files, and got rid of the webgpu version.

```gradle
    sourceSets {
        getByName("opengl") {
            java.srcDirs("src/opengl/java/")
        }
        listOf("webgpuDawn", "webgpuWgpu").forEach {
            getByName(it) {
                java.srcDirs("src/vulkan/java")
                manifest.srcFile("src/vulkan/AndroidManifest.xml")
            }
        }
    }
```

birkskyum:
There's a one-line change to the heatmap view in the test app, so that it works.

It's currently not visible in the openfreemap liberty, and most other styles, because it look for a layer id called `country_label` which most styles doesn't have. This change just adds the heatmap and circles on top of any chosen style.

Screenshot 2026-02-13 at 07 13 52

 I find, at least in emulation, that it's much easier to run the heatmap with vulkan though webgpu, than directly with the vulkan-backend moltenvk.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
birkskyum: I have added this functionality to the PR, so the Snapshot Activity works now.

birkskyum:
The live readFrameBuffer (Snapshot Activity working)

https://github.com/user-attachments/assets/2dfbda1b-01fc-4cea-b04c-4dccbe17c582

vendor/wgpu.cmake
```diff
@@ -31,38 +31,58 @@ if(APPLE)
         set(_wgpu_lib_arch "x86_64-apple-darwin")
     endif()
     set(_wgpu_lib_name "libwgpu_native.a")
-    set(_wgpu_lib_suffix ".dylib")
+elseif(ANDROID)
+    if(CMAKE_ANDROID_ARCH_ABI STREQUAL "arm64-v8a")
```
louwers: Could you move these changes to `wgpu.android.cmake`?

docs/mdbook/src/platforms/android/android-rendering-backends.md
```diff
@@ -0,0 +1,53 @@
...
+./gradlew :MapLibreAndroidTestApp:installWebgpuWgpuDebug
+```
+
+The same pattern applies to the library module (`MapLibreAndroid`):
+
+```bash
+./gradlew :MapLibreAndroid:assembleVulkanRelease
+```
+
+## Selecting in Android Studio
```
louwers, line 40: It's not showing up for the render test app for me.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+    AndroidWebGPURenderableResource(AndroidWebGPURendererBackend& backend_)
+        : backend(backend_) {}
+
+    void bind() override {}
+
+    void swap() override { backend.markNeedsPresent(); }
+
+    const mbgl::webgpu::RendererBackend& getBackend() const override { return backend; }
+
+    const WGPUCommandEncoder& getCommandEncoder() const override {
```
louwers, line 118: Should we throw or at least put an `assert(false)` in the unimplemented methods?

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+    instanceDesc.requiredLimits = &instanceLimits;
+    impl->instance = std::make_unique<dawn::native::Instance>(&instanceDesc);
+
+    std::vector<dawn::native::Adapter> adapters = impl->instance->EnumerateAdapters();
+    if (adapters.empty()) {
+        Log::Error(Event::Render, "WebGPU: no adapters found");
+        return;
+    }
+
+    size_t selectedIndex = 0;
```
louwers: This can be tidied a bit by factoring it into a function.

This way the adapter can be returned and the `selectedIndex` variable is no longer needed.

Can also use a range-based for loop.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+    }
+
+    WGPUSupportedFeatures features = {};
+    wgpuAdapterGetFeatures(static_cast<WGPUAdapter>(impl->adapter), &features);
+#endif
+
+    std::vector<wgpu::FeatureName> requiredFeatures;
+    wgpu::TextureFormat depthStencilFormat = wgpu::TextureFormat::Depth24PlusStencil8;
+    if (features.featureCount > 0 && features.features) {
+        for (size_t i = 0; i < features.featureCount; ++i) {
```
louwers, line 240: Prefer using a range-based for-loop.

.github/workflows/android-ci.yml
```diff
@@ -311,6 +311,118 @@ jobs:
...
+
+      - name: Add Android Rust target
+        run: rustup target add aarch64-linux-android
+
+      - name: Install Android NDK
+        run: yes | "$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager" "ndk;28.1.13356709" > /dev/null || true
+
+      - name: Configure cargo for Android NDK linker
+        run: |
+          NDK_BIN="$ANDROID_SDK_ROOT/ndk/28.1.13356709/toolchains/llvm/prebuilt/linux-x86_64/bin"
```
louwers: Could you replace this with a non-hardcoded NDK version?

Maybe use the latest one that is available in the SDK root?

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
louwers: Nevermind, I don't think `WGPUSupportedFeatures` can be iterated over.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+    deviceDesc.SetUncapturedErrorCallback([](const wgpu::Device&, wgpu::ErrorType type, wgpu::StringView message) {
+        Log::Error(Event::Render,
+                   std::string("Dawn error [") + std::to_string(static_cast<int>(type)) + "] " +
+                       (message.data ? std::string(message.data, message.length) : std::string()));
+    });
+    deviceDesc.SetDeviceLostCallback(
+        wgpu::CallbackMode::AllowSpontaneous,
+        [](const wgpu::Device&, wgpu::DeviceLostReason reason, wgpu::StringView message) {
+            Log::Error(Event::Render,
+                       std::string("Dawn device lost [") + std::to_string(static_cast<int>(reason)) + "] " +
```
louwers: Are there any helpers in wgpu.h to create a human-readable string from a `DeviceLostReason`? If not, consider creating them.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+        }
+        wgpuSupportedFeaturesFreeMembers(features);
+    }
+
+    wgpu::DeviceDescriptor deviceDesc = {};
+#if MLN_WEBGPU_IMPL_DAWN
+    deviceDesc.label = "MapLibre Android WebGPU Device";
+    deviceDesc.SetUncapturedErrorCallback([](const wgpu::Device&, wgpu::ErrorType type, wgpu::StringView message) {
+        Log::Error(Event::Render,
+                   std::string("Dawn error [") + std::to_string(static_cast<int>(type)) + "] " +
```
louwers: Same here (as below) maybe we can create a small helper to make this error code in a human-readable string.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+        WGPUWaitStatus waitStatus = wgpuInstanceWaitAny(instance, 1, &waitInfo, UINT64_MAX);
+        if (waitStatus != WGPUWaitStatus_Success || !waitInfo.completed) {
+            Log::Error(Event::Render, "WebGPU readFramebuffer: WaitAny failed");
+        }
+    }
+#elif MLN_WEBGPU_IMPL_WGPU
+    callbackInfo.mode = WGPUCallbackMode_AllowProcessEvents;
+    wgpuBufferMapAsync(stagingBuffer, WGPUMapMode_Read, 0, alignedDataSize, callbackInfo);
+
+    for (int i = 0; i < 1000 && !mapContext.completed; ++i) {
```
louwers, line 627: I think this came up before. But you're sure this is the way to do it?

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
```diff
@@ -0,0 +1,738 @@
...
+                    std::memcpy(dst, src, rowStride);
+                }
+            }
+        }
+    } else {
+        Log::Error(Event::Render, "WebGPU readFramebuffer: buffer mapping failed");
+    }
+
+    wgpuBufferUnmap(stagingBuffer);
+    wgpuBufferRelease(stagingBuffer);
```
louwers: Are there any RAII helpers available for WebGPU?

If not, consider using the `Scoped` helper in `mbgl/util/scoped.hpp`.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
birkskyum: Yes, until wgpu-native support WaitAny like Dawn, this is the workaround.

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
birkskyum: went for `Scoped`

platform/android/MapLibreAndroid/src/cpp/android_webgpu_renderer_backend.cpp
birkskyum: Created a helper.

vendor/wgpu.cmake
```diff
@@ -31,38 +31,49 @@ if(APPLE)
...
     "${_mln_wgpu_source_dir}/build/release"
 )
 
-find_library(WGPU_LIBRARY
-    NAMES wgpu_native libwgpu_native
-    PATHS ${_wgpu_lib_search_paths}
-    NO_DEFAULT_PATH
-)
+# On Android, prefer the static library to avoid needing to ship a separate .so
+if(ANDROID)
```
louwers, line 62: this can also move to `wgpu.android.cmake`

vendor/wgpu.cmake
```diff
@@ -235,6 +246,14 @@ elseif(WIN32)
         bcrypt
         ntdll
     )
+elseif(ANDROID)
+    find_package(Threads REQUIRED)
```
louwers: this can also move to `wgpu.android.cmake`

.github/workflows/android-ci.yml
```diff
@@ -311,6 +311,117 @@ jobs:
...
+          linker = "$NDK_BIN/aarch64-linux-android26-clang"
+          EOF
+
+      - name: Pre-build wgpu-native for aarch64-linux-android
+        working-directory: vendor/wgpu-native
+        run: |
+          LATEST_NDK=$(find "$ANDROID_SDK_ROOT/ndk" -mindepth 1 -maxdepth 1 -type d | sort -V | tail -n 1)
+          NDK_PREBUILT="$LATEST_NDK/toolchains/llvm/prebuilt/linux-x86_64"
+          RESOURCE_DIR=$("$NDK_PREBUILT/bin/clang" --print-resource-dir)
+          export BINDGEN_EXTRA_CLANG_ARGS_aarch64_linux_android="--sysroot=$NDK_PREBUILT/sysroot --target=aarch64-linux-android26 -isystem $RESOURCE_DIR/include"
```
louwers: This target should ideally not be hardcoded.

.github/workflows/android-ci.yml
```diff
@@ -311,6 +311,117 @@ jobs:
...
+        run: rustup target add aarch64-linux-android
+
+      - name: Configure cargo for Android NDK linker
+        run: |
+          LATEST_NDK=$(find "$ANDROID_SDK_ROOT/ndk" -mindepth 1 -maxdepth 1 -type d | sort -V | tail -n 1)
+          NDK_BIN="$LATEST_NDK/toolchains/llvm/prebuilt/linux-x86_64/bin"
+          mkdir -p ../../vendor/wgpu-native/.cargo
+          cat > ../../vendor/wgpu-native/.cargo/config.toml << EOF
+          [target.aarch64-linux-android]
+          linker = "$NDK_BIN/aarch64-linux-android26-clang"
```
louwers: The linker Android version should ideally not be hardcoded.

louwers (approved):

PR #4103, opened 2026-02-16
Support for WebGPU backend in iOS SDK, core

birkskyum:
Similar to 
- #4097 

Adds iOS platform support for the WebGPU rendering backend, with both Dawn and wgpu-native implementations.

- New MLNMapView+WebGPU backend using UIView + CAMetalLayer
- CMake presets: ios-webgpu-dawn and ios-webgpu-wgpu
- iOS cross-compilation support in vendor/dawn.cmake and vendor/wgpu.cmake
- Follows the same `#if MLN_WEBGPU_IMPL_DAWN` / `#elif MLN_WEBGPU_IMPL_WGPU` pattern used in the Android and desktop backends.
- Added new Live Snapshot view which continuously takes screenshots (like the android Snapshot Activity view) 

I've tested this on iOS simulator and iPhone 17 Pro

One thing that's not supported yet, but which is a bigger effort that goes beyond this pr, is custom drawables.

Screenshot 2026-02-16 at 01 53 08

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -257,6 +257,7 @@ jobs:
         run: cmake --preset ios-webgpu-dawn
 
       - name: Build mbgl-core and iOS SDK
+        continue-on-error: true
         run: cmake --build build-ios-webgpu-dawn --target mbgl-core ios-sdk-static
 
   ios-build-webgpu-wgpu:
@@ -281,6 +282,7 @@ jobs:
         run: cmake --preset ios-webgpu-wgpu
 
       - name: Build mbgl-core and iOS SDK
+        continue-on-error: true
         run: cmake --build build-ios-webgpu-wgpu --target mbgl-core ios-sdk-static
 
   ios-ci-result:
@@ -294,5 +296,5 @@ jobs:
       - ios-build-webgpu-wgpu
     steps:
       - name: Mark result as failed
-        if: needs.ios-build.result != 'success' || needs.ios-build-cmake.result != 'success' || needs.ios-build-webgpu-dawn.result != 'success' || needs.ios-build-webgpu-wgpu.result != 'success'
+        if: needs.ios-build.result != 'success' || needs.ios-build-cmake.result != 'success'
         run: exit 1


birkskyum:
The live readFrameBuffer (Live Snapshot ) working



https://github.com/user-attachments/assets/989f1359-0f5d-48e0-bf6e-c15d11afef22

sjg-wdw:
Is there someone who is asking for this?
I'm not opposed to personal projects, but I'm curious if this is something MapLibre is going to support moving forward.

birkskyum:
Personal project. There haven't been commercial requests.

It gives us a way to benchmark Metal vs. WebGPU-Metal on iOS, and Vulkan vs. WebGPU-Vulkan on Android, and wgpu vs. Dawn. I'm just being proactive about finding the limits of WebGPU and Wasm for the MapLibre renderers, keeping tabs on what's possible with these emerging technologies.

sjg-wdw:
Cool.  I say merge a PR if it doesn't cause too much trouble.

When my group's developers create and change shaders it'll be at their discretion to update the WebGPU version.  If not, we'll just make a note of it as we go.

.github/workflows/ios-ci.yml
```diff
@@ -243,14 +243,56 @@ jobs:
...
     needs:
       - pre_job
       - ios-build
       - ios-build-cmake
+      - ios-build-webgpu-dawn
+      - ios-build-webgpu-wgpu
     steps:
       - name: Mark result as failed
-        if: needs.ios-build.result != 'success' || needs.ios-build-cmake.result != 'success'
+        if: needs.ios-build.result != 'success' || needs.ios-build-cmake.result != 'success' || needs.ios-build-webgpu-dawn.result != 'success' || needs.ios-build-webgpu-wgpu.result != 'success'
```
louwers: Should we fail the build completely if it fails?

louwers (approved):

.github/workflows/ios-ci.yml
birkskyum: Actually no, I've removed that criteria now, and added a `continue-on-error: true` for dawn/wgpu as well

PR #4111, opened 2026-02-17
Detect changes in PMTiles requests

tdcosta100:
This PM addresses the problem described in #3658, where a change in the PMTiles file, either local or remote, causes the PMTiles file source to crash. The response fields to validate against are: ETag, Expires and Modified. If any of them changes in a request, the cache is invalidated and the request is repeated with fresh new data. I had to introduce the Modified field to local file requests, but fortunately the modified time was already available when checking for the file validity.

### DIFF
--- a/platform/default/src/mbgl/storage/pmtiles_file_source.cpp
+++ b/platform/default/src/mbgl/storage/pmtiles_file_source.cpp
@@ -234,7 +234,8 @@ private:
 
     bool ensureUrlCacheIsValid(const std::string& url, const Response& response) {
         if (!url_cache.contains(url)) {
-            url_cache.emplace(url, URLCacheEntry{.modified = response.modified, .expires = response.expires, .etag = response.etag});
+            url_cache.emplace(
+                url, URLCacheEntry{.modified = response.modified, .expires = response.expires, .etag = response.etag});
             return true;
         }
 
@@ -530,8 +531,7 @@ private:
                 }
 
                 if (!ensureUrlCacheIsValid(url, response)) {
-                    callback(std::make_unique<Response::Error>(Response::Error::Reason::Other,
-                                                               error_cache_miss));
+                    callback(std::make_unique<Response::Error>(Response::Error::Reason::Other, error_cache_miss));
                     return;
                 }
 


platform/default/src/mbgl/storage/pmtiles_file_source.cpp
```diff
@@ -30,6 +30,8 @@
 #endif
 
 namespace {
+constexpr const char* error_cache_miss = "CACHE_MISS";
```
louwers, line 33: Nit: would just make this a std::string.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
```diff
@@ -218,6 +232,44 @@ class PMTilesFileSource::Impl {
         return fileSource;
     }
 
+    bool validateUrlCache(const std::string& url, const Response& response) {
```
louwers: Is "validate" the right word here? Because it's also updating the cache.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
```diff
@@ -218,6 +232,44 @@ class PMTilesFileSource::Impl {
...
+            entry.modified = response.modified;
+            url_cache.emplace(url, entry);
+            return true;
+        }
+
+        auto cache_entry = url_cache.at(url);
+
+        if (cache_entry.etag != response.etag || cache_entry.expires != response.expires ||
+            cache_entry.modified != response.modified) {
+            url_cache.erase(url);
```
louwers: Consider adding clearFromCaches(url) method.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: I would too, but when I did a little research about doing this, I felt discouraged to do so. The most close to `std::string` I can do without making the compiler warn against it is to make a `std::string_view`. I will test with it and see if it works ok.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: The concept behind this method is to check if the cache is still valid. It creates a cache entry if it doesn't exist, and this is a valid situation. It clears the cache if it's not valid anymore. I can move these lines to a new method, like you suggested in other comment.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: That's a good idea, I will include this in the next commit.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
louwers: Ah yes that should work. https://stackoverflow.com/questions/27123306/is-it-possible-to-use-stdstring-in-a-constant-expression

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Done!

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Done!

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
```diff
@@ -218,6 +232,47 @@ class PMTilesFileSource::Impl {
         return fileSource;
     }
 
+    bool validateUrlCache(const std::string& url, const Response& response) {
+        if (!url_cache.contains(url)) {
+            URLCacheEntry entry;
+            entry.etag = response.etag;
```
louwers: Can use C++20 designated initializers

URLCacheEntry entry{ .modified = response.modified, .expires = response.expires, .etag = response.etag };

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
```diff
@@ -218,6 +232,47 @@ class PMTilesFileSource::Impl {
...
+            return false;
+        }
+
+        return true;
+    }
+
+    void clearUrlCaches(const std::string& url) {
+        url_cache.erase(url);
+
+        if (header_cache.contains(url)) {
```
louwers: this check is not needed, it is safe to call erase without

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
louwers: You could even do it inline in the emplace call or use std::move.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
louwers: My point is that I don't expect a function that 'validates' to have side effects. To validate something means to check, I'm thinking more read only.

Maybe `ensureUrlCacheValid` fits better?

tdcosta100: Now I understand better your considerations. `ensureUrlCacheValid` is a good name for this method considering what it does. But if instead we move the `clearUrlCaches` to outside it, would be a better approach for understandability, maybe? Let me know. I will accept either option.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Nice! I will do it inline so.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Great! I didn't know this check was not needed. So much cleaner code now.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Unfortunately I cannot use it as direct replacement of std::string =(. Reverting back to `const char*`.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Done!

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: Done!

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
louwers: If it can be made completely read only then using "validate" may be appropriate, but it's still calling these functions that modify state so I would personally go with 'ensure'.

bdon:
> Maybe @bdon also can take a look.

I can reproduce the original crash on main, I'm testing by uploading an archive to an R2 bucket e.g. `example.com/city.pmtiles` and then replacing it in-place while the app is running.

Currently #4111 does not load the map at all on pmtiles archives, is it currently in a testable state?

I'm using the macOS glfw build:

```
bazel run --//:renderer=metal //platform/glfw:glfw_app -- --style http://localhost:8500/style.json
```

where `style.json` references a `pmtiles://https://...` URL in a source.

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
tdcosta100: I will keep it as ensure, for consistency.

tdcosta100:
> > Maybe @bdon also can take a look.
> 
> I can reproduce the original crash on main, I'm testing by uploading an archive to an R2 bucket e.g. `example.com/city.pmtiles` and then replacing it in-place while the app is running.
> 
> Currently #4111 does not load the map at all on pmtiles archives, is it currently in a testable state?
> 
> I'm using the macOS glfw build:
> 
> ```
> bazel run --//:renderer=metal //platform/glfw:glfw_app -- --style http://localhost:8500/style.json
> ```
> 
> where `style.json` references a `pmtiles://https://...` URL in a source.

Could you please provide the style.json and if possible, the pmtiles files? So I can reproduce better your scenario.

bdon:
Yes, I should have provided that before, here's how I'm running the macOS app:

```
bazel run --//:renderer=metal //platform/glfw:glfw_app -- --style https://r2-public.protomaps.com/tmp/maplibre-native-pmtiles.json
```

```json
"protomaps": {
            "type": "vector",
            "attribution": "<a href=\"https://github.com/protomaps/basemaps\">Protomaps</a> © <a href=\"https://openstreetmap.org\">OpenStreetMap</a>",
            "url": "pmtiles://https://r2-public.protomaps.com/tmp/hawaii.pmtiles"
        }
```

bdon:
using the example linked here:

https://maplibre.org/maplibre-native/android/examples/data/PMTiles/

you can also try:

```
bazel run --//:renderer=metal //platform/glfw:glfw_app -- --style https://raw.githubusercontent.com/wipfli/foursquare-os-places-pmtiles/refs/heads/main/style.json
```

Issue #4109, opened 2026-02-17
Map cropping occurring on CarPlay screen


cachly:
### MapLibre iOS Version

Latest in main branch

### iOS Version

26.2.1

### Device

iPhone 14 Pro

### What happened?

Today I created a new `xcframework` file for my app which includes CarPlay. It appears that with the most recent code in `main` that some clipping/cropping of the map is occurring:

<img width="1920" height="1000" alt="Image" src="https://github.com/user-attachments/assets/faa7a8e6-921e-42a8-9d9f-48c38f295ba1" />
<img width="1920" height="1000" alt="Image" src="https://github.com/user-attachments/assets/1109d8f1-c8c8-4869-90a7-a8e1f695d7f5" />

This did not happen previously when I was using `main` with commits on Dec 18 and built an `xcframework` on that date. This one worked fine.

### Steps to reproduce

Build `xcframework` of current `main` branch. Test on CarPlay to see clipping of map.

### Relevant log output

```shell

```

louwers:
@cachly Could you check if this is fixed in the latest release?

fonok3:
I also had the problem before on customer devices. The cases they could see the issue before are now fixed with [this release](https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.24.0).

Issue #3851, opened 2025-10-01
CarPlay map rendering is broken on iOS26, linked to new Smart zoom setting


MaartenZonneveld:
### MapLibre iOS Version

6.18.0

### iOS Version

26.0.1

### Device

iPhone 16 PRO

### Expected Behavior
The map covers the entire screen from edge to edge.
<img width="800" height="480" alt="Image" src="https://github.com/user-attachments/assets/b42f2d96-78bc-4a31-8c18-e33c10a1fc66" />

### Actual Behavior
The map does not cover the entire screen but stops before the edges.
<img width="800" height="480" alt="Image" src="https://github.com/user-attachments/assets/ec9cf657-acaf-4a4a-936d-edc056b87e16" />

In this screenshot the background of the MLNMapView has been given a red color. It is visible that the view resizes correctly. Only the actual map itself does not render correct.
<img width="800" height="480" alt="Image" src="https://github.com/user-attachments/assets/469b54e9-321c-4007-b4e5-fd9a37b2d059" />

This is on the screen of a customer with a wider screen. Note that only the map rendering is broken. Other view like the max speed view works fine.
<img width="2074" height="964" alt="Image" src="https://github.com/user-attachments/assets/ad2ac77d-8b60-4327-9721-7fb5847b74a1" />

### Steps to reproduce

Open the app on a wide CarPlay screen. Per default the CarPlay setting: Smart zoom is enabled. When Smart zoom is enabled the bug appears. When Smartzoom is disabled the bug is "fixed", but the car does not use the optimal resolution for the screen.
To toggle the setting on CarPlay navigate to Settings > Display > Smart zoom

Use the code below and display this view on CarPlay.
```
let mapView = MLNMapView(frame: .zero)
view.insertSubview(mapView, at: 0)
mapView.translatesAutoresizingMaskIntoConstraints = false
mapView.topAnchor.constraint(equalTo: view.topAnchor, constant: 0).isActive = true
mapView.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: 0).isActive = true
mapView.leftAnchor.constraint(equalTo: view.leftAnchor, constant: 0).isActive = true
mapView.rightAnchor.constraint(equalTo: view.rightAnchor, constant: 0).isActive = true
mapView.backgroundColor = .red
```
Draw a MLNMapView and constrain it to top, bottom, leading trailing constrains to make the map cover the entire screen.
Download CarPlay simulator: https://developer.apple.com/download/all/
[...]

Issue #3214, opened 2025-02-13
contentScaleFactor independent from UIScreen mainScreen - issue with secondary screens like CarPlay


boldtrn:
**Is your feature request related to a problem? Please describe.**

Maplibre-Native calculates the scaleFactor like this:

https://github.com/maplibre/maplibre-native/blob/a8f4f70636aa5230328b8aa0b7bd1e274c120e9e/platform/ios/src/MLNMapView%2BMetal.mm#L39

As far as I understand, this can lead to issues if your phone has a different scale than your CarPlay screen.

As described in the [apple dev forum](https://developer.apple.com/forums/thread/726497?answerId=747819022#747819022):

> UIScreen.main will always return a UIScreen instance that corresponds to the iPhone's built-in display.
> If your CarPlay-capable app is a Navigation app (and therefore receives a CPWindow to draw your map content into), you can get the UIScreen for the CarPlay display through your CPTemplateApplicationScene:

**Describe the solution you'd like**

Don't use the mainScreen, but use the screen where the map is drawn on (and maybe fallback to main if none is given).

The dev forum describes the solution like:

`let carPlayScreen = templateApplicationScene.carWindow.screen`

**Describe alternatives you've considered**

I assume there is also a way to get the screen from the view tree / hierarchy?

Alternatively it could also work that implementers pass a scale from the outside into MlnMapview?

**Additional context**

In our case, the CarPlay map is rendered too large proportionally to the screen size. So there is less area covered.

cachly:
You should be able to just add this:

```
// Set the correct scale of the map
let carPlayScreen = self.templateApplicationScene?.carWindow.screen
if let scale = carPlayScreen?.scale {
    mapView.contentScaleFactor = scale
}
```

StephanPartzsch:
Doesn't this scales the map on UIKit level, means after the map content was rendered? The metal rendering does not respect the `contentScaleFactor` of UIKit and therefore takes the wrong scaleFactor. This could lead to wrong sizes and unneeded performance overhead (if I am not completely wrong)

PR #4112, opened 2026-02-18
Fixed issue #4109 CarPlay screen getting clipped

nnhubbard:
This issue fixes the recently introduced bug in CarPlay where the screen is being clipped. See #4109.

### DIFF
--- a/src/mbgl/mtl/render_pass.cpp
+++ b/src/mbgl/mtl/render_pass.cpp
@@ -230,7 +230,9 @@ void RenderPass::setScissorRect(MTL::ScissorRect rect) {
         if (rect.height + rect.y > height) {
             rect.height = height - rect.y;
         }
-        encoder->setScissorRect(rect);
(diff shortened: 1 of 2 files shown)


adrian-cojocaru:
This fix works only when the frustum offset is disabled (0,0,0,0) and the issue will reappear when a custom offset is provided.
Possibly related to https://github.com/maplibre/maplibre-native/issues/3214 https://github.com/maplibre/maplibre-native/issues/3851 as `pixelRatio` (or `size`) isn't updated when the scissor rect is computed.

sjg-wdw:
Can we help out a little bit here @adrian-cojocaru or @alexcristici ?
Our sponsor is interested in CarPlay working, even if they don't use it that way.

louwers:
Last CarPlay fix was from @StephanPartzsch.

nitrag:
I've had reports of this bug occurring in my app.

Issue #4088, opened 2026-02-05
Add automated native macOS XCFramework releases


fonok3:
**Is your feature request related to a problem? Please describe.**

There are no official macOS xcframework releases. The current iOS releases (`ios-v6.x.x`) only include `ios-arm64` and `ios-arm64_x86_64-simulator`.

Developers building native macOS applications must build MapLibre from source, which requires:
- CMake/Bazel expertise
- ~10GB disk space
- 1-2 hours build time

This creates a significant barrier to adoption for macOS developers.

> [!note]
> I'm happy to implement this if the approach is acceptable.

**Describe the solution you'd like**

Add automated CI releases for native macOS using a separate release tag pattern `macos-v{VERSION}`, as suggested in #5.

Implementation:
1. New GitHub Actions workflow `.github/workflows/macos-release.yml`
2. Trigger on tags matching `macos-v*`
3. Build using existing Bazel target: `bazel build //platform/macos:macos_framework --//:renderer=metal`
4. Package as xcframework and upload as release asset
5. Update [maplibre-gl-native-distribution](https://github.com/maplibre/maplibre-gl-native-distribution) Package.swift to add a `MapLibre-macOS` target

**Describe alternatives you've considered**

| Approach | Pros | Cons |
|----------|------|------|
| **Separate `macos-v*` releases** (proposed) | Low risk, no breaking changes, follows #5 recommendation | Two release tracks to maintain |
| Combined `apple-v*` xcframework | Single artifact, matches Firebase/Realm | Breaking change for iOS users, APIs differ between platforms |
| Source distribution only | No CI changes needed | Poor developer experience |
| Wait for Mac Catalyst | — | Blocked by bazelbuild/rules_apple#658, no timeline |

**Additional context**

- Native macOS builds already work with Bazel (`bazel build //platform/macos:macos_framework --//:renderer=metal`)
- This is independent of Mac Catalyst support (#3), which is blocked by external toolchain issues
- Mac Catalyst binaries would belong in the iOS xcframework (since Catalyst uses iOS APIs), not the macOS xcframework proposed here
- #5 was closed with the suggestion that macOS should have separate distribution due to API differences

louwers:
Hi @fonok3, could you clarify how you are using MapLibre Native for macOS? Are you using AppKit?

There is no `//platform/macos:macos_framework` target.

fonok3:
Thanks for the fast response!

[...]

PR #4113, opened 2026-02-18
macos: add xcframework Bazel targets

fonok3:
This PR is a proposal for the first part of #4088.

It adds Bazel targets for building macOS xcframeworks (`MapLibre.static`, `MapLibre.dynamic`).

For this to work I added a `--//:macos_loop=cfrunloop` build flag. The default event loop (libuv) does not integrate with the native run loop of AppKit/SwiftUI apps, so async tasks and rendering callbacks do not fire without a manual workaround. With cfrunloop the xcframework uses CFRunLoop directly and works out of the box. Default stays libuv so existing targets keep working without breaking changes.

Also fixes header visibility to match iOS and removes unused OpenGL/Metal file lists.

Bazel build command: `bazel build //platform/macos:MapLibre.static --//:renderer=metal --//:macos_loop=cfrunloop`


BUILD.bazel
```diff
@@ -246,6 +246,36 @@ config_setting(
     },
 )
 
+# Selects the macOS event loop implementation
```
louwers: I think this can move to `platform/darwin/BUILD.bazel`. What do you think?

fonok3: For me both places work. As all flags are in the root BAZEL file I put it there but as far as I see this is the first flag which is not for all platforms. Then I'd recommend putting it in `platform/macos/BUILD.bazel` as iOS always uses `CFRunLoop`.

fonok3: @louwers would `platform/macOS/BUILD.bazel` work for you?

louwers: Yes!

louwers: I think technically libuv also supports iOS, but I have never tried it.

fonok3: Doesn't really work with UIKit.

louwers:
@fonok3 Thanks, did you try building all the new targets?

fonok3:
@louwers yes all new targets are building locally. (including proper dsym export for the dynamic variant)

PR #4119, opened 2026-02-20
Show progress on device when Android render test is running

louwers:
Makes it a bit easier to see the progress of render tests, as it will show on the device with this PR.

https://github.com/user-attachments/assets/6720e2b9-8c66-42d3-a717-702d1a91b775


louwers:
I can try to show test results in a follow up.

PR #4124, opened 2026-02-23
core: Use interleaved buffer for paint property binders

adrian-cojocaru:
This PR batches all paint properties vertex buffers in an interleaved format to reduce the number of binds. 
~~`USE_INTERLEAVED_BINDER_BUFFER` can toggle between functionalities, since some use cases can benefit from separate buffers, especially if the paint property usage changes and can detect if an update is required (currently the buffers are reuploaded without any changes to their data).~~


include/mbgl/gfx/vertex_attribute.hpp
```diff
@@ -452,6 +452,20 @@ class VertexAttributeArray {
             return;
         }
 
+#ifdef USE_INTERLEAVED_BINDER_BUFFER
```
louwers: Can you remove the `#ifdefs`.

include/mbgl/gfx/vertex_attribute.hpp
```diff
@@ -452,6 +452,20 @@ class VertexAttributeArray {
...
+        if (binder->interleavedVertexBuffer) {
+            const auto& interleavedSharedVector = binder->interleavedVertexBuffer->sharedVertexVector;
+
+            const auto rawSize = static_cast<uint32_t>(binder->interleavedVertexBuffer->stride);
+            const bool isInterpolated = binder->isInterpolated();
+            const auto dataType = isInterpolated ? InterpType::DataType : Type::DataType;
+            // assert(rawSize == static_cast<uint32_t>(isInterpolated ? sizeof(typename InterpType::Value)
```
louwers: Commented out code.

include/mbgl/gfx/vertex_attribute.hpp
adrian-cojocaru: The description covers this a bit. 
The separate buffers (previous implementation) are still better if we lower the update frequency (detecting when changes are needed instead of per frame uploads).

louwers: I would suggest we go with the implementation that makes most sense for us right now. And revert if we ever change the update frequency.

louwers (approved):

PR #4129, opened 2026-02-24
android: make RemovedUnusedImagesTest less flaky?

louwers:
This test was failing in the instrumentation tests on AWS Device Farm.

Increase the timeout a bit, and clean up listeners.


adrian-cojocaru:
From the failed device farm runs this is somewhat recent. The only time I had it trigger in a local build it was reported as an ANR.

Issue #4132, opened 2026-02-26
BUG: Raster and Vector layers desynchronize during map interaction


UlysselaGlisse:
# Bug description

When interacting with the map, there is a desynchronisation between raster and vectors layers.

# Where to find this bug
I've found this bug in many applications using MapLibre Native for map rendering across both IOS and Android.
The issue is more prononced on Android: the misalignment triggers immediately upon touching the screen. 
On IOS, we have to move the map to see it.

For example : 
* CFF (Swiss train company):
    - Android: 
    
https://github.com/user-attachments/assets/84ff82a6-ea89-4adc-a6bb-322e1728ad04

    - IOS:
    
https://github.com/user-attachments/assets/1d2c67b5-3051-4fa8-857c-77baa5820d33

* Mappy:
  - Android:
 
https://github.com/user-attachments/assets/2b55726e-5920-4fbc-9fb6-34f8009e578e

  - IOS:

https://github.com/user-attachments/assets/d0da9b1d-ffa1-457d-933f-92982ca6ba54

# Proposed fix:
I've found a workaround to avoid this dealignment by deleting this condition:
```C++
// src/mbgl/renderer/layers/raster_layer_tweaker.cpp 
// l.97

matrix = getTileMatrix(
        tileID,
        parameters,
        {0.f, 0.f},
        TranslateAnchorType::Viewport,
        false,
        false,
        drawable,
        !parameters.state.isChanging()); // <- That one
        }
```

louwers:
Thanks a lot for the detailed report!

Do you feel comfortable making a PR with your fix?

Pinging @mwilsnd authored this particular line of code. 🙂 

https://github.com/maplibre/maplibre-native/blob/166593f0fb2a43b34eb3353bb8a86a20decacb09/src/mbgl/renderer/layers/raster_layer_tweaker.cpp#L97-L104

UlysselaGlisse:
Thank you! Yes, i will be happy to submit a PR. I have already tested this fix locally on both Android and iOS using a custom build, and it successfully resolve the desynchronization without any immediate performance regressions. I will prepare the PR shortly.

PR #4137, opened 2026-02-28
Fix raster layer tweaker to avoid desynchronization issue 

UlysselaGlisse:
Solution found for fixing #4132


louwers (approved):

louwers:
The render tests need to be updated, although the image looks blurry?

PR #4146, opened 2026-03-05
Enable FastPFOR encodings in MLT

TimSylvester:


TimSylvester:
I previously had this working, but the latest version of fast-pack/FastPFOR [removed SIMDE](https://github.com/fast-pack/FastPFOR/pull/145), which allowed it to compile on ARMv7.  To get it building again I had to fork just before that commit and tweak some `#if` conditions.  I also get some errors locally (MacOS) about insufficient registers for inline assembly when building the Android i686 target, but apparently the same thing doesn't happen on CI.

@louwers I don't think we want to use a fork, so I'm not sure how to proceed here.  I guess we propose undoing that PR and applying my changes upstream?

@nyurik Does the Rust target build on all the supported platforms and architectures, particularly Android-ARMv7?

Apparently NEON is supported on "almost all" ARMv7 devices: [https://developer.android.com/ndk/guides/cpu-arm-neon](https://developer.android.com/ndk/guides/cpu-arm-neon)

But the problem here is not NEON, it's the explicit use of things like [vqtbl1q_u8](https://developer.arm.com/architectures/instruction-sets/intrinsics/vqtbl1q_u8).

I suspect it's theoretically possible to make it work, but the developers do not seem interested in supporting it: [https://github.com/fast-pack/FastPFOR/issues/81](https://github.com/fast-pack/FastPFOR/issues/81)

nyurik:
Rust target uses pure rust implementation without C/C++, so it should. I would need to set up some testing / CI to actually validate it

.gitmodules
```diff
@@ -102,8 +102,8 @@
 	url = https://github.com/gulrak/filesystem.git
 [submodule "vendor/maplibre-tile-spec"]
 	path = vendor/maplibre-tile-spec
-	url = https://github.com/maplibre/maplibre-tile-spec.git
-	branch = cpp
+    url = https://github.com/TimSylvester/maplibre-tile-spec.git
```
louwers: this can be updated now I think

TimSylvester:
Since we made it disabled by default, for this to actually be usable I also need to make the enable option accessible from a higher level than the source, like `MapOptions`, and wire it up to the platform APIs.

louwers (approved):

vendor/maplibre-tile-spec.cmake
```diff
@@ -1,15 +1,20 @@
...
-endif()
+string(REPLACE ";" " " TEMP "${CMAKE_CXX_COMPILER_PREDEFINES_COMMAND}")
+message(STATUS "Clang predefines command: ${TEMP}")
+
+execute_process(COMMAND bash -c "${TEMP}" OUTPUT_VARIABLE CLANG_DEFINES ERROR_VARIABLE CLANG_DEFINES_ERR OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_STRIP_TRAILING_WHITESPACE)
+message(STATUS "Clang defines: ${CLANG_DEFINES}")
+message(STATUS "Clang errors: ${CLANG_DEFINES_ERR}")
+execute_process(COMMAND cmd /c "${TEMP}" OUTPUT_VARIABLE CLANG_DEFINES ERROR_VARIABLE CLANG_DEFINES_ERR OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_STRIP_TRAILING_WHITESPACE)
+message(STATUS "Clang defines: ${CLANG_DEFINES}")
+message(STATUS "Clang errors: ${CLANG_DEFINES_ERR}")
```
ntadej: Debugging leftovers?

ntadej:
The Qt build error is:
```
D:\a\maplibre-native\maplibre-native\source\vendor\maplibre-tile-spec\cpp\include\mlt/polyfill.hpp(20): error C2672: 'std::bit_cast': no matching overloaded function found
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\bit(31): note: could be '_To std::bit_cast(const _From &) noexcept'
D:\a\maplibre-native\maplibre-native\source\vendor\maplibre-tile-spec\cpp\include\mlt/polyfill.hpp(20): note: the associated constraints are not satisfied
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\bit(30): note: use of undefined type 'std::array<std::byte,4>'
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\utility(167): note: see declaration of 'std::array<std::byte,4>'
D:\a\maplibre-native\maplibre-native\source\vendor\maplibre-tile-spec\cpp\include\mlt/polyfill.hpp(20): note: the template instantiation context (the oldest one first) is
D:\a\maplibre-native\maplibre-native\source\vendor\maplibre-tile-spec\cpp\src\mlt\decode\int.cpp(39): note: see reference to function template instantiation 'T std::byteswap<unsigned int>(T) noexcept' being compiled
        with
        [
            T=unsigned int
        ]
D:\a\maplibre-native\maplibre-native\source\vendor\maplibre-tile-spec\cpp\include\mlt/polyfill.hpp(21): error C3536: 'value_representation': cannot be used before it is initialized
```

This usually means some header is missing as MSVC is often a bit more strict.

TimSylvester:
@ntadej Hm, I observed that with the MSVC build previously, `note: use of undefined type 'std::array...`, but including <array> was in the last MLT PR, so I must have messed up the submodule reference somehow.

vendor/maplibre-tile-spec.cmake
TimSylvester: yeah, good catch

louwers (commented):
One nit: we have `compassEnabled`, `scrollGesturesEnabled` etc. etc. For consistency I would say let's call it `fastPFOREnabled` instead of `enableFastPFOR`?

hectocore:
Hello there ! Thanks for fixing this. We've recently switched over to MLT for our Motorcycle GPS app and it broke our 3D buildings display (extrusions are missing a side on their polygon), and it seems the fix that will resolve our issue and allow us to safely enable 3D buildings again is included in this PR (comes from https://github.com/maplibre/maplibre-tile-spec/pull/816).

Do you know if it will it be part of the next stable iOS build and when it will land ? Thanks a lot !

Issue #4071, opened 2026-01-26
Implement `unit` option for `number-format`


ginnyTheCat:
**Is your feature request related to a problem? Please describe.**
Being able to format quantities in a localized way: 5000 meters as `5.000 m` in `de-DE`, `5,000米` in `cmn-CN` and `5 000 м` in `uk-UA`.

**Describe the solution you'd like**
Adding a `unit` option to `number-format` based on CLDR and ECMA-402.

**Describe alternatives you've considered**
Not localizing or changing the style dynamically depending on user language.

**Additional context**
Design proposal in https://github.com/maplibre/maplibre-style-spec/issues/1336

alwaysalearner1234:
Hi @ginnyTheCat,@louwers I’ve opened a PR implementing the `unit` option for `number-format` based on the design proposal discussed in this issue.

The implementation adds parsing, evaluation, equality checks, and serialization support for the `unit` field.
Currently the unit string is appended to the formatted number.

Please let me know if any changes or adjustments are needed. Thanks!

ginnyTheCat:
Looks great. I just feel like without the ICU support this feature is not quite complete, but idk how difficult it is to add that, especially on Apple.

alwaysalearner1234:
Good point. ICU support would definitely make this feel more complete. I'll take a look at what would be required to add it, though I'm not sure yet how straightforward it would be on Apple platforms.

PR #4151, opened 2026-03-07
Add unit option to number-format (#4071)

alwaysalearner1234:
## Description

This PR adds support for the `unit` option in `number-format` expressions.

Currently `number-format` supports formatting numbers with locale, currency, and fraction digit options, but does not support displaying units such as meters, kilometers, etc.

This change allows developers to specify a `unit` option that will be appended to the formatted number.

Example:

```json
["number-format", 5000, {
  "locale": "de-DE",
  "unit": "meter"
}]
```

Output:

5.000 meter

## Changes

- Added support for the `unit` option in `number-format` expressions
- Updated `NumberFormat` constructor to include `unit`
- Added dependency tracking for the `unit` expression
- Implemented parsing for the `unit` option in `NumberFormat::parse`
- Evaluated the `unit` expression in `NumberFormat::evaluate`
- Updated `eachChild()` to include the `unit` expression
- Updated `operator==` to compare the `unit` field
- Added serialization support for the `unit` option

## Future Improvements

- Integrate ICU-based unit formatting so that units are localized according to the selected locale
- Support automatic pluralization and localized unit symbols (e.g., `m`, `km`, `米`)
- Improve formatting logic to handle spacing and locale-specific formatting rules

## Testing

- Verified that the project compiles successfully after adding the unit option
- Confirmed that expressions using unit are parsed correctly
- Verified that the unit string is appended to the formatted number during evaluation
- Tested serialization and equality comparisons for expressions containing the unit option


louwers:
AI slop, closing.

PR #4155, opened 2026-03-08
Fix: Missing data for single tiles when using GeoJsonSource

buldogin-cloud:
## Changes
- Modified GeoJSONTile::updateData to properly reset tile state when data is updated
- Added condition to reset() when this is the first data update (data is null) or when needsRelayout is true
- This ensures that existing tiles are properly invalidated when new GeoJSON data is set

This fixes the issue where single tiles would not display updated data after switching overlays in GeoJsonSource, particularly noticeable when switching between different overlays like street lighting data.

### DIFF
--- a/src/mbgl/tile/geojson_tile.cpp
+++ b/src/mbgl/tile/geojson_tile.cpp
@@ -23,14 +23,14 @@ void GeoJSONTile::updateData(std::shared_ptr<style::GeoJSONData> data_, bool nee
     MLN_TRACE_FUNC();
 
     assert(data_);
-    
+
     // Reset state if relayout is needed or if this is the first data update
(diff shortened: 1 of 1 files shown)


louwers (commented):
Thanks for your contribution. I can test this later this week.

Do you have a style or anything I can test it with?

sjg-wdw:
Can you take a look @alexcristici

alexcristici:
@buldogin-cloud I have some questions:
1. Why do we need to reset state when it's the first data update?
2. How GeoJSONTile::updateData properly reset tile state when data is updated - by moving up the `reset()` method?
Thanks!

alexcristici (commented):
I posted some questions.

PR #4178, opened 2026-03-17
Handle allocation failure [Core feature]

alexcristici:
Catch exceptions when rendering a layer or drawable to prevent the app from crashing, and instead skip displaying the layer or drawable that generated the exception.

New `onRenderError` method in `MapObserver` to receive rendering errors on iOS and Android.
iOS: `- (void)mapViewRendererDidError:(MLNMapView *)mapView;` 
Android: `public void addOnRenderErrorListener(MapView.OnRenderErrorListener callback)`

Some screenshots of the app catching the exception without crashing, while skipping rendering of the affected layer or drawable:

Screenshot_20260409_174932
Screenshot_20260409_223847


src/mbgl/gl/resource_pool.cpp
```diff
@@ -154,6 +154,9 @@ TextureID Texture2DPool::allocateGLMemory(const Texture2DDesc& desc) {
                                   Enum<gfx::TexturePixelType>::to(desc.pixelFormat),
                                   Enum<gfx::TextureChannelDataType>::to(desc.channelType),
                                   nullptr));
+    if (glGetError()) {
+        throw std::bad_alloc();
+    }
```
adrian-cojocaru, line 159: This will throw `std::bad_alloc` on other usage errors (invalid value/operation/enum). `GL_OUT_OF_MEMORY` should be the trigger for it.

alexcristici: Yes, because the buffer will not be allocated if any error appears.

adrian-cojocaru (approved):

louwers (approved):

PR #4190, opened 2026-03-26
Terrain 3D

NathanMOlson:
Working in `mbgl-render`!

image

- [ ] Symbols, Circles, and other layers that are rendered on top are not displayed at the correct altitude. We need to get the elevation at the symbol position in the shader.
- [ ] Line layers need different shader behavior when rendering to terrain vs not. GL JS achieves this using `#ifdef TERRAIN3D`.
- [ ] 3D buildings: not started.  Also need to get elevation in the shader.
- [ ] Support for tiles rendering onto smaller terrain tiles. The current implementation supports multiple tiles being rendered to a single terrain tile, but not a single tile being rendered to multiple terrain tiles. Changing this will likely require significant redesign: `getTileMatrix()` will need more information from somewhere.
- [ ] `coveringTiles()`: use terrain in calculation of covering tiles.
- [ ] Use lower res terrain tiles? And figure out how to set the tile size for the render-to-texture buffers.
- [ ] CustomLayer/PluginLayer: make render-to-terrain optional? Does this require a style spec change?
- [ ] Object re-use. All testing so far has been done using `mbgl-render`, which uses `Static` map mode. Performance will likely be bad in `Continuous` mode, and there may be memory leaks.
- [ ] Support transparency


stefankarschti:
@NathanMOlson  @louwers : what's the strategy to keep this in sync with main? I've fixed the current compile issues, should I open a PR from my fork to target this branch?

NathanMOlson:
> should I open a PR from my fork to target this branch?

That sounds good to me. I'm happy to entertain alternatives but I don't currently have any better ideas.

bonesyjr:
Is this still in progress?

NathanMOlson:
> Is this still in progress?

I'm not currently working on it.

yaing11:
> Is this still in progress?

It seems to me all work moved to PR 4389.

acalcutt:
My https://github.com/maplibre/maplibre-native/pull/4389 should now be merged in + the latest main. I fixed a few small issues I saw in the ci and local testing after the merge, but I think those are fixed now.

Most of the original checkboxes in the OP have been covered now (except making this a plugin). Buildings, circles, symbols, roads all respect the terrain, symbol and text occlusion have been added. it is now using built in dem decoding instead of its own.

I added a few render tests for buildings on terrain, but a few are failing due to slight renderer differences, most pretty minor. there are a few failing tests I brought form maplibre-gl-js that I think might be camera inside terrain differences since that still needs handling

I also added in the android sample app menu items

  - `TerrainActivity` - self-contained style with color-relief + hillshade over
    3D terrain (Matterhorn); a bright hypsometric colour ramp to show color-relief.
  - `TerrainVectorMapActivity` - a full-planet vector basemap (OpenFreeMap
    Liberty) draped over terrain, DEM/hillshade/terrain added at runtime.
  - `TerrainOsmRasterActivity` - OSM raster draped over terrain (Innsbruck);
    exercises the raster-draping cases.
  - `TerrainDebugTilesActivity` - the gl-js terrain debug-tiles style (numbered
    tiles over synthetic terrain), for comparing tile zoom / draping against
    gl-js.
  - `TerrainFlightActivity` - an automated FPV "drone" camera flight starting on
    the ground at Innsbruck and touring the surrounding Alps, sweeping a range of
    zoom / altitude / pitch / rotation over terrain (baked, elevation-planned
    path that stays above ground); useful for eyeballing draping, occlusion, and
    frame pacing in motion, especially with the rendering-stats HUD on.
    
<p float="left">
  
  
  
</p>
<p float="left">
  
  
  
</p>

acalcutt:
Just to note, it seems like the bump to a higher sdk in main made the map/compass/stats go above the toolbar (you can see it in the screenshots). this wasn't happening before the merge with a lower sdk. however I am not sure that should be fixed in this PR since that is an issue in main as well.

julianjc84:
**Zoom-0 intensity bug — terrain half fixed (2026-08-27).** New symptom: at low zoom,
  *rotating* the camera scales terrain height — the elevation scale is bearing-dependent.
  `Camera::getWorldToCamera` derives `pixelsPerMeter` from the **camera's** latitude, and
  `updateCameraState` orbits the camera around the centre by bearing/pitch, so `1/cos(lat)`
  rescales every metre-valued z as you rotate. gl-js uses the **centre** latitude
  (`mercatorZfromAltitude(1, this.center.lat)`), which is rotation-invariant. Visible only
  at low zoom: the offset is a fixed pixel count over `worldSize = 512 * 2^zoom` (pitch 70,
  equator: z8 1.02x, z6 1.31x, z4 11x). Fixed by passing the centre latitude in; verified on
  device (Pixel 9 Pro, Vulkan). Upstream code untouched by this branch (`19b84464a3`, 2023) —
  standalone PR material. **Supersedes** the hillshade shared-cause guess; that half is open.

daviani:
## Tested on device: iOS 26.6, Metal renderer

@acalcutt — since #4389 landed here, this seems the right place for it.

`TERRAIN.md` says *"Only GL and Vulkan are buildable/testable in this
environment (Metal needs macOS)"*, and the status table reads `builds: not here`
with three `untested` rows for Metal. I built this branch (`039c94b`) for a
physical device and ran it on an **iPhone 16 Pro Max, iOS 26.6, Metal**, so
here is what those rows look like from here. Happy to re-run anything with
different settings.

### 1. It builds and it renders ✅

Taking the Metal column of the status table row by row:

- **builds** — **yes**, via bazel with `--//:renderer=metal`, built for device
- **terrain renders** — **yes**
- **terrain off→on** — **yes**, by rebuilding the style JSON at runtime
- **symbol occlusion** — **not established**, see below

1-terrain-off-exaggeration-1 5

2-terrain-on-exaggeration-1 5



Raising `exaggeration` also behaves as expected.

I could not get a conclusive read on **symbol occlusion**, and I would rather
say so than claim a row I did not fill. Three attempts, all defeated by my own
setup rather than by the renderer: markers spread along a line are raised onto
the terrain surface, so they are never behind it; and packing them across a
ridge instead puts them in a narrow band near the horizon at high pitch, where
they overlap and run off the top of the screen. I could not tell an occluded
marker from one that was simply out of frame.

Reading the code rather than the screen is more informative: the Metal path
looks complete. `src/mln/shaders/mtl/terrain_depth.cpp` is there and is the same
shape as the Vulkan and WebGPU ones, all three symbol shaders bind
`idSymbolDepthTexture` (slots 2/2/3 against Vulkan's 3/3/3, which is just Metal
numbering without the gap Vulkan leaves for the icon texture), and nothing in
`render_terrain.cpp` branches away from Metal. So this is not a missing
implementation — I simply failed to build a test that could see it.

Your `TerrainFlightActivity` describes a "baked, elevation-planned path" — that
is the piece I was missing: querying terrain elevation to place the probes,
instead of guessing at coordinates. If you want that row filled and can point me
at the right approach, I have the hardware and I am happy to run it.

### 2. Lowering `exaggeration` leaves the map unusable 🔴

Raising the exaggeration works. **Lowering it does not**, and the map does not
recover:

- starting at 1.5 and raising to 4.0 — works as expected
- bringing it back down towards 1.0 — **the map goes blank**
- turning terrain **off** afterwards — **does not recover**, still blank
- pinch/zoom gestures — **stop responding entirely**
- restarting the app — the only thing that brings it back

3-exaggeration-4 0-works

4-exaggeration-1 0-map-blank

5-terrain-off-still-blank




The style itself carries the decreasing value correctly — I logged the JSON
being applied and it steps down `4.0 → 3.9 → 3.2 → … → 2.0`. So the value
reaches the style; the map does not follow it back down.

This looks related to the `camera-terrain anchoring` item already described in
`TERRAIN.md`, but two details there do not seem to be covered: the loss of
gesture input, and the fact that **disabling terrain does not restore the map**.
If those are expected consequences of the same root cause, feel free to ignore
this half of the report.

### 3. Performance cost scales with camera pitch 📊

Since Metal had no numbers attached, I measured. Method: zoom 12 over
Innsbruck, camera rotating continuously at 24°/s (a still map does not redraw,
so it would collect nothing), 8 s per pass after a 2 s warm-up, sampling
`MLNRenderingStats` from
`mapViewDidFinishRenderingFrame(_:fullyRendered:renderingStats:)`.
All no-terrain passes run before all terrain passes, so terrain is never
lowered mid-campaign.

Two consecutive runs, so each figure below is given as `run 1 / run 2`:

- **Pitch 25°** — fps **54.8 / 58.8** with terrain against **61.2 / 58.9**
  without, i.e. **−11% / −0%**. Encode CPU **×1.39 / ×1.50**. Draw calls
  **6 with terrain, 6 without**.
- **Pitch 45°** — fps **40.9 / 42.1** against **60.5 / 59.6**, i.e.
  **−32% / −29%**. Encode CPU **×1.73 / ×1.82**. Draw calls **8 and 8**.
- **Pitch 65°** — fps **21.5 / 22.0** against **53.8 / 56.2**, i.e.
  **−60% / −61%**. Encode CPU **×2.35 / ×2.30**. Draw calls
  **28 with terrain against 19 without**.
  
6-pitch-benchmark-run-1
7-pitch-benchmark-run-2

  
The 65° figures were also reproduced four times before this campaign
(21.0–21.8 fps, ×2.30–2.35, always 28/19), the first two with an earlier
version of the harness, so I am fairly confident they are stable.

Two things I did not expect:

- **At 25° and 45° the draw call count is identical with and without terrain**
  (6/6, then 8/8), yet encode time is still ×1.4–1.8. That part of the cost is
  not extra tiles.
- At 65° terrain draws 28 against 19, **+47%** — the horizon moves far enough
  out that noticeably more terrain has to be loaded and drawn.

Caveats. The harness rotates the camera, and @julianjc84 reported on 26 August
that bearing rescales terrain height at low zoom (z8 1.02x, z6 1.31x, z4 11x).
These passes are at z12, so if I read that correctly the effect should be well
under the 1.02x measured at z8 — but tell me if it invalidates the method and I
will re-run without rotation. Beyond that: the no-terrain passes sit at the
60 Hz vsync ceiling, so the fps deltas are a **lower bound** on the real cost. And the style used here keeps a
`hillshade` layer in both cases, so this measures terrain *on top of* hillshade,
not against a bare map. Memory counters are process-wide gauges and I could not
attribute them to a pass, so I have left them out rather than publish a number
I cannot defend.

### 4. Three corrections for `TERRAIN.md`

Following the document's example verbatim gives a blank map with no error, which
cost me a while before I looked at the tile URLs:

- The background tiles at `demotiles.maplibre.org/tiles/{z}/{x}/{y}.png`
  return **404** — that source is vector, not raster.
  Check: `curl -I https://demotiles.maplibre.org/tiles/10/537/365.png`
- The DEM encoding is given as `terrarium`, but it is **Terrain-RGB**
  (`"encoding": "mapbox"`).
  Check: `curl -s https://demotiles.maplibre.org/terrain-tiles/tiles.json`
- The `bounds` are worldwide, but the DEM (`jaxa_terrainrgb_N047E011`) only
  covers **one degree around N47 E11** and stops at z12. Same `tiles.json`.

Happy to send these three as a small PR against this branch if you prefer them
that way rather than fixing them yourself.

### 5. Offer: an iOS example

There are now five terrain activities on the Android side and none on iOS.
 While
testing I wrote a small `TerrainExample` for `app-swift` (a map, a terrain
toggle and an exaggeration slider, ~110 lines, driven entirely from style JSON
since there is no Objective-C terrain API). It is what produced the screenshots
above.

I also have the harness from section 3 — it is roughly the iOS counterpart of
your `TerrainFlightActivity`: scripted camera motion plus `MLNRenderingStats`,
which nothing in `platform/ios` currently uses.

Happy to open a PR against this branch for either or both — just say which, and
whether you would rather have them as their own examples or folded into
existing ones.

### 6. Two notes on building it (unpaid Apple account)

The iOS guide says a paid or unpaid Apple Developer account both work. On an
unpaid one, three things blocked me: the default profile name
`"iOS Team Provisioning Profile: *"` is compared **literally** by
`local_provisioning_profile_finder.py`, and that wildcard profile only exists on
a paid account; the documented workaround (build from the generated Xcode
project) does not help because `rules_xcodeproj` delegates signing to bazel;
and opening the project in Xcode creates five App IDs while a free profile
allows only three installed apps. What worked was letting Xcode create the
profile, then naming that exact profile in `config.bzl`. Happy to open a docs
issue if that is useful.

Unrelated, but worth knowing: `platform/ios/BUILD.bazel:216` and
`platform/ios/app-swift/BUILD.bazel:59` both use
`bundle_id = "{}.maplibre.app"`, so installing one demo app silently replaces
the other. Renaming a bundle id is your call, not something I would touch.

---

**AI disclosure.** I used Claude Code (Opus 5) for this work: it set up the iOS
build, wrote the example and the measurement harness, parsed the logs, and
drafted this comment in English, which is not my first language. The testing on
device, the screenshots, and the observations in section 2 — in particular that
raising exaggeration works while lowering it does not, and that gestures stop
responding — are mine. I have reviewed everything above and can answer questions
on any of it.

acalcutt:
I'd say open a PR with your test additions and fixes suggested against this branch and I can look it over and try to help get it merged into this.

Just to note on section 4, at the demotiles site there a few different kinds of demotiles you can see at https://github.com/maplibre/demotiles#maplibre-demo-tiles. Under 'tiles' it is right that there are only world/country vector tiles, but there are also terrain-tiles (TerrainRGB tiles around Innsbruck ) , tiles-omt (open map tiles around Innsbruck).  The examples there, like in https://github.com/maplibre/demotiles/tree/gh-pages/terrain-tiles#how-to-use-the-demo-terrain-tiles use OpenStreetMap raster tiles.

In my android examples, I opted to use https://mapterhorn.com/data-access/ for terrain so we could have the full planet for testing, those are terrarium encoded. In the raster test I used OpenStreetMap like the example above. The was also an existing full planet vector tiles set (osm liberty i think) which i used as the base of my vector tile test since it was used in other tests already

Most of the tests I made were based of the styles in the [demotiles examples](https://github.com/maplibre/demotiles#maplibre-demo-tiles), but with the mapterhorn terrain and osm liberty in place of the narrower demotiles i mentioned

skifahrer:
Awesome! Looking forward when it will be  official part.

pdrhlik:
Hi. I've been building an iOS hiking app (MapRiot, code not public, app in development) on this branch, and it now does something I couldn't do without 3D terrain: a "replay" flight that follows a recorded track over the real landscape, both live on screen and exported as a video. Here's a 45-second export, made on an iPhone SE (2nd gen) with satellite tiles from [ČÚZK - Czech land survey office](https://geoportal.cuzk.gov.cz/default.aspx?lng=EN&mode=TextMeta&text=ortofoto_info&side=ortofoto&head_tab=sekce-02-gp&menu=23):

https://github.com/user-attachments/assets/9b02c9c3-8e4e-4412-aa1a-3c2351dbba14

A few tiles are a bit blurry there which happened because of some connection issues during the video render.

My fork is here: https://github.com/mapriot/maplibre-native, branch `mapriot/terrain`.

The base is tagged `mapriot-integration`, which is feature/terrain-3d at 2ef164d with these pull requests merged in, so the tag marks exactly where other people's work ends and mine begins: #4556, #4557, #4558, #4559 (thanks @AndrewN81), #4615 (@koekiedaf, closed unmerged) and #4650 (@johncarmack1984, since merged to main). Everything after the tag is mine: 14 commits, one change per commit, with the reason in the message - [the full list is here](https://github.com/mapriot/maplibre-native/compare/mapriot-integration...mapriot/terrain).

pre-commit is clean across all 14, and `mbgl-test-runner` (macOS, Metal) passes apart from two tests that are already red on the base: `TerrainCamera.CentreRidesTheTerrainSurface` and `TerrainCamera.CentreElevationSettlesWithoutDrift` expect the centre clamp to be on, and 2ef164d ("core: default the terrain centre clamp off") turned it off without updating them. @acalcutt - do you want the tests updated to the new default, or is the default meant to come back on?

Running a real app on it turned up a handful of things, most of them in areas the checklist above still has open - "coveringTiles(): use terrain in calculation of covering tiles" and "object re-use ... performance will likely be bad in Continuous mode":

- tile cover: tile elevation was compared in the wrong units against the frustum, so raised ground near the camera was not covered and tiles went missing ([e34c4ae](https://github.com/mapriot/maplibre-native/commit/e34c4aecde), test included)
- the camera's centre altitude is kept in metres across zoom changes ([6408ffa](https://github.com/mapriot/maplibre-native/commit/6408ffa288)), and pan/anchor points are unprojected at that altitude ([add843e](https://github.com/mapriot/maplibre-native/commit/add843e878)) - without this, gestures on pitched terrain slide away under your finger on iOS ([c2d4be4](https://github.com/mapriot/maplibre-native/commit/c2d4be4862)). Tests included for the first two.
- drapes are re-baked when a style's layers or sources change ([2606701](https://github.com/mapriot/maplibre-native/commit/2606701b1f)) or a layer's paint changes ([cd2aee3](https://github.com/mapriot/maplibre-native/commit/cd2aee3402)), instead of showing the old picture
- terrain mesh limits that a phone can live with in Continuous mode: mesh past the DEM maxzoom ([eefd367](https://github.com/mapriot/maplibre-native/commit/eefd36768a)), drape resolution falling off with distance ([7e5e772](https://github.com/mapriot/maplibre-native/commit/7e5e772112)), a mesh-tile cap by camera distance ([bbbe273](https://github.com/mapriot/maplibre-native/commit/bbbe273da0)) that grows with the viewport ([1039611](https://github.com/mapriot/maplibre-native/commit/1039611540), test included)
- two build/debug fixes that are not terrain-specific: the shared placeholder texture is released before the context's "no resources left" assert ([1089de0](https://github.com/mapriot/maplibre-native/commit/1089de0c64) - every debug build aborted), and an unused Metal encoder that fails -Werror ([6891a3a](https://github.com/mapriot/maplibre-native/commit/6891a3a949)). Both touch code from #4556 and #4627 - I've commented there rather than opening competing PRs.
- an iOS addition: -[MLNMapView captureNextFrameWithHandler:], which is what makes offline video export possible ([5082c57](https://github.com/mapriot/maplibre-native/commit/5082c57d3b), CHANGELOG entry included)

There's one more commit on the branch, "a frame is fully rendered only once the terrain has caught up" ([4c7f285](https://github.com/mapriot/maplibre-native/commit/4c7f285a42)). It is what made video export reliable for me, but I have no regression test that fails without it yet, so I'm not offering it as a PR - it's there to look at.

I'd be more than happy to help you finish this feature. @acalcutt, you suggested above that fixes and test additions against this branch are welcome - I'm glad to open these as small PRs, one at a time, each with its test. Tell me which you'd like first.

I use 3D terrain in web apps, so I wanted it in a native app too and hadn't realised it wasn't there yet. I'll also be interested in adding the Android layer once the core is ready and I'm happy with my app on iOS.

Disclosure, per MapLibre's AI policy: I used Claude Code with Opus 5 to create this work. I reviewed the changes and tested everything in a simulator, on the mentioned iPhone SE and on a 2025 iPad, both iOS/iPadOS 26.

Patrik

georgbachmann:
I love this! I am hoping for 3d terrain since a loooong time now! And this is awesome news! Thanks for your effort!
I know it's hard to answer, but is there any likleyhood if this will be merged back? Or even any ETA?
This would be such a massive improvement!

PR #4208, opened 2026-04-04
Update Bazel and dependencies, fix Xcode 26 compilation

louwers:


platform/ios/src/MLNMapView.mm
```diff
@@ -6365,7 +6365,7 @@ - (void)didUpdateLocationSignificantlyAnimated:(BOOL)animated
       completionHandler:^{
         MLNMapView *strongSelf = weakSelf;
         if (strongSelf.userTrackingState == MLNUserTrackingStateBegan ||
-            strongSelf.userTrackingState == MLNDistanceThresholdForCameraPause) {
+            strongSelf.userTrackingState == MLNUserTrackingStateBeginSignificantTransition) {
```
louwers, line 6368: This was an error `MLNDistanceThresholdForCameraPause` was some unrelated double. But `MLNUserTrackingStateBeginSignificantTransition` has the same value.

platform/darwin/src/MLNMapSnapshotter.mm
```diff
@@ -892,8 +892,9 @@ - (void)configureWithOptions:(MLNMapSnapshotOptions *)options {
   }
 
   // Create the snapshotter
-  auto localFontFamilyName =
-      config.localFontFamilyName ? std::string(config.localFontFamilyName.UTF8String) : nullptr;
+  auto localFontFamilyName = config.localFontFamilyName
+                                 ? std::string(config.localFontFamilyName.UTF8String)
+                                 : std::string();
```
stefankarschti: https://github.com/maplibre/maplibre-native/blob/main/platform/default/include/mbgl/map/map_snapshotter.hpp#L44 the constructor expects `std::optional<std::string>`. Will the empty string be ok?

louwers: You're right! I added an explicit type and `std::nullopt`.

stefankarschti (approved):

adrian-cojocaru (approved):

PR #4216, opened 2026-04-07
ios: fix build issues Xcode 26

louwers:
The Bazel update I did in #4208 is not strictly needed. This seems to be enough now that Boost is updated.


louwers:
Dawn fail is unrelated to this PR I think.

stefankarschti (approved):

PR #3858, opened 2025-10-04
Add feature state functionality to iOS

louwers:
Implements https://github.com/maplibre/maplibre-native/issues/1698 for iOS. `feature-state` was already implemented in the core.

Also ports the GLFW example to the Swift-based example app.

https://github.com/user-attachments/assets/4dc9fd97-8513-4e69-be3b-145a3fa8395e

TODO:

- [ ] Think about the API.
- [ ] Tests

Issue #1698, opened 2023-09-27
Feature parity for "feature-state" expression


ThibaudM:
**Is your feature request related to a problem? Please describe.**
`feature-state` expression is not yet supported for Android, iOS (and macOS) SDKs. [See here](https://maplibre.org/maplibre-style-spec/expressions/#feature-state)

**Describe the solution you'd like**
Feature parity with gl-js implementation

**Related issues**
- #185

louwers:
@ThibaudM Could you make a proposal what the API should look like?

ThibaudM:
> @ThibaudM What do you envision the API should look like?

I am not sure to understand your question. 

In terms of usage/scenarios, I imagine the same approach as the gl-js one (in order to keep the parity):
- [a piece of the code related to feature state](https://github.com/maplibre/maplibre-gl-js/blob/main/src/source/source_state.ts)
- and samples like [this one](https://maplibre.org/maplibre-gl-js/docs/examples/hover-styles/)

In terms of implementation, I really have no idea. Maybe the gl-js approach can be mimicked if there is no huge divergence in this part of the code

louwers:
Before thinking about the implementation, I think it is good to think about the (platform) APIs.

So specifically: what method(s) should be added to which interface(s) of the iOS and/or Android SDKs? What are the method signature(s)?

MarkusWendorf:
With https://github.com/maplibre/maplibre-native/pull/3406 being merged is the support for feature state now complete? Or is there something missing @louwers?

If it is supported the style spec page should be updated https://maplibre.org/maplibre-style-spec/expressions/#feature-state

louwers:
@MarkusWendorf No it is not complete, there are still no APIs for Android and iOS.

louwers:
@ThibaudM If you're still interested in this, I just made a PR that implements this. Feedback on the API is welcome. https://github.com/maplibre/maplibre-native/pull/3858

ThibaudM:
It's so cool!
I'm still very interested but I currently don't have the time to have a look. I will try my best.

kylerchin:
Is there the API for android yet?

gabbopalma:
Will this feature (with PR #3858) be implemented soon? We wanted to add this implementation to the `flutter-maplibre-gl` API as well

PR #4219, opened 2026-04-09
Add feature state functionality to Android

brandonroeder:
- Implements https://github.com/maplibre/maplibre-native/issues/1698 for Android
- This aims to match the syntax in the open [iOS PR](https://github.com/maplibre/maplibre-native/pull/3858) for the same issue.

Also uses the same GLFW example

https://github.com/user-attachments/assets/e5248e56-d938-47d9-b59c-db6fad1d9e4e

## AI usage

Per the [AI Policy](https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md):
Both Claude Code (Opus) and Codex gpt-5.4 were used during development.
- Opus did the OG work to port the iOS PR to Android
- codex gpt-5.4 was used on subsequent PR feedback changes


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMap.java
```diff
@@ -2285,6 +2286,55 @@ public List<Feature> queryRenderedFeatures(@NonNull RectF coordinates,
...
   }
 
+  /**
+   * Sets the state of a feature identified by source id, optional source layer id and feature id.
+   *
+   * <p>Feature state can be read in style expressions through {@code ["feature-state", key]}.</p>
+   * <p>The target feature must already have an id in the underlying source data. For vector tile
+   * sources, {@code sourceLayerId} is required.</p>
+   */
+  public void setFeatureState(@NonNull String sourceId,
```
louwers: Tested it. Works great.

However, if the source ID is required for all these, wouldn't it make sense to add these methods on source classes instead?

brandonroeder: @louwers thanks! good callout, i reworked things to move them to the source class. Should be good to go

louwers: Thanks, will have another look.

platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/testapp/style/GeoJsonSourceTests.java
```diff
@@ -276,6 +307,42 @@ public void testDuplicateSourceDuringSyncSetGeoJson() {
     });
   }
 
+  private String createFeatureStateStyleJson() {
```
louwers: There's not really a reason for this to be a method.

There are better ways to load resources like styles.

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/feature/FeatureStateActivity.kt
```diff
@@ -0,0 +1,186 @@
...
+import org.maplibre.android.style.layers.PropertyFactory.lineColor
+import org.maplibre.android.style.layers.PropertyFactory.lineWidth
+import org.maplibre.android.style.sources.GeoJsonSource
+import org.maplibre.android.testapp.R
+import org.maplibre.android.testapp.styles.TestStyles
+import java.net.URI
+
+/**
+ * Test activity showcasing interactive styling with feature state,
+ * matching the iOS feature-state example.
```
louwers: The comment that it "matches the iOS feature-state example" is out of place. That one is not even merged yet.

louwers (commented):
Could you explain the reasoning behind having different APIs for GeoJSONSource and VectorTileSource? Could these be merged somehow?

I think the API is an improvement, but I would like so avoid using nullable parameters if possible.

Also I think the APIs for setting and removing feature state should probably return some value that indicates success or failure.

Also if AI was used please disclose some details in the PR description, as per our (new) [AI Policy](https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md).

platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/testapp/style/GeoJsonSourceTests.java
brandonroeder: Thanks, pulled it out into `test_feature_state_style.json` and load with `ResourceUtils.readRawResource`

brandonroeder:
@louwers hi! I addressed your feedback and added the AI disclaimer. Let me know if I can do anything else to get this in.

louwers:
@brandonroeder Could you merge in main to trigger a CI run?

Then once CI passes, looks good to me!

louwers (approved):

TimSylvester:
The feature state tests are failing intermittently in CI.  See #4344, #4146.

```
testFeatureStateKeyRemovalAcrossSource
testFeatureStateUpdatesStyleOwnedSourceImmediately
```

A stack trace excerpt from the AWS runner is:

```
Caused by: java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.maplibre.android.testapp.style.GeoJsonSourceTests.lambda$testFeatureStateUpdatesStyleOwnedSourceImmediately$6(GeoJsonSourceTests.java:143)
	at org.maplibre.android.testapp.style.GeoJsonSourceTests.$r8$lambda$jRgUaK8jNRTV3QGX1PyDroAtPqY(Unknown Source:0)
	at org.maplibre.android.testapp.style.GeoJsonSourceTests$$ExternalSyntheticLambda9.onInvokeAction(D8$$SyntheticClass:0)
	at org.maplibre.android.testapp.action.MapLibreMapAction.perform(MapLibreMapAction.java:39)
	at androidx.test.espresso.ViewInteraction$SingleExecutionViewAction.perform(ViewInteraction.java:429)
	at androidx.test.espresso.ViewInteraction.doPerform(ViewInteraction.java:297)
```

I see them fail unpredictably when I run them locally as well.

Extending the timeout in `pollFeatureCount` doesn't seem to help, so it appears that the changes are not appearing at all in some cases.

louwers:
Thanks @TimSylvester, I'll have a look and revert if needed.

PR #4220, opened 2026-04-09
Bundle wgpu-native into Linux WebGPU amalgamation

yuiseki:
## Summary

Fix Linux WebGPU amalgamation builds so `libmbgl-core-amalgam.a` also includes `libwgpu_native.a` when building with:

- `MLN_CREATE_AMALGAMATION=ON`
- `MLN_WITH_WEBGPU=ON`
- `MLN_WEBGPU_IMPL_WGPU=ON`

## Motivation

This was needed for downstream consumers such as `maplibre-native-rs`.

Before this change, the Linux WebGPU amalgam archive could be produced, but it still contained unresolved `wgpu*` symbols. As a result, `maplibre-native-rs` could not use it through `MLN_CORE_LIBRARY_PATH` alone and still needed an extra `wgpu_native` linker dependency.

## Changes

- resolve a dedicated `WGPU_STATIC_LIBRARY` for Linux `wgpu-native`
- include that static archive in the Linux amalgamation step
- add a Linux-only check script to fail if unresolved `wgpu*` symbols remain
- add a Linux `wgpu` lane to `core-release.yml`

## Validation

Validated locally on Linux by:

- building `maplibre-native` with Linux `wgpu + amalgamation`
- confirming `libmbgl-core-amalgam.a` has no unresolved `wgpu*` symbols
- linking it into `maplibre-native-rs` via `MLN_CORE_LIBRARY_PATH`
- running a successful headless render smoke test downstream

## Size impact

This increases the Linux x64 WebGPU amalgam size, which is expected because `libwgpu_native.a` is now bundled into the archive.

For reference:

- bundled `libwgpu_native.a`: ~51.6 MiB
- resulting Linux x64 WebGPU amalgam: ~385.5 MiB

Compared with existing Linux x64 amalgam artifacts from `core-release`:

- OpenGL amalgam: ~326.7 MiB
- Vulkan amalgam: ~352.6 MiB
- WebGPU (`wgpu`) amalgam after this change: ~385.5 MiB

So the artifact does get larger, but the increase is in line with the size of the bundled `wgpu-native` static archive and does not appear abnormal.

  | Artifact | Before | After |
  |---|---:|---:|
  | Linux x64 OpenGL amalgam | 326.7 MiB | 326.7 MiB |
  | Linux x64 Vulkan amalgam | 352.6 MiB | 352.6 MiB |
  | Linux x64 WebGPU amalgam |  | 385.5 MiB |

- Reference: bundled libwgpu_native.a in the local build is about 51.6 MiB

## Note

This change bundles `wgpu-native` into the amalgam archive, but does not try to bundle all Linux system libraries.

### DIFF
--- a/.github/scripts/check-linux-webgpu-amalgam
+++ b/.github/scripts/check-linux-webgpu-amalgam
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ $# -ne 1 ]]; then
+  echo "usage: $0 <path-to-libmbgl-core-amalgam.a>" >&2
+  exit 2
+fi
+
+archive_path="$1"
+
+if [[ ! -f "$archive_path" ]]; then
+  echo "archive not found: $archive_path" >&2
+  exit 2
+fi
+
+unresolved_symbols="$(
+  nm -g --undefined-only "$archive_path" \
+    | grep -E '(^|[[:space:]])U[[:space:]]+wgpu([A-Z]|_)' \
+    || true
+)"
+
+if [[ -n "$unresolved_symbols" ]]; then
+  echo "found unresolved WebGPU symbols in $archive_path" >&2
+  echo "$unresolved_symbols" >&2
+  exit 1
+fi
+
+echo "no unresolved WebGPU symbols found in $archive_path"
--- a/.github/scripts/check-linux-webgpu-amalgam
+++ b/.github/scripts/check-linux-webgpu-amalgam
@@ -1,16 +1,17 @@
 #!/usr/bin/env bash
+# Linux-only helper: requires GNU nm long options.
 set -euo pipefail
 
 if [[ $# -ne 1 ]]; then
   echo "usage: $0 <path-to-libmbgl-core-amalgam.a>" >&2
-  exit 2
+  exit 1
 fi
 
 archive_path="$1"
 
 if [[ ! -f "$archive_path" ]]; then
   echo "archive not found: $archive_path" >&2
-  exit 2
+  exit 1
 fi
 
 unresolved_symbols="$(
--- a/.github/workflows/core-release.yml
+++ b/.github/workflows/core-release.yml
@@ -85,6 +85,8 @@ jobs:
   build-linux:
     needs: [create-release]
     runs-on: ${{ matrix.runner.os }}
+    env:
+      PRESET_RENDERER: ${{ matrix.preset_renderer || matrix.renderer }}
     strategy:
       matrix:
         include:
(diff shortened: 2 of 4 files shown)


yuiseki:
@birkskyum This feels like a necessary packaging follow-up to your WebGPU work in #3899, and also relevant to the Rust-first downstream direction you described in [maplibre/maplibre-native-slint#46: Use Rust instead of C++](https://github.com/maplibre/maplibre-native-slint/issues/46), where [maplibre-native-slint](https://github.com/maplibre/maplibre-native-slint) is moving toward a more purely Rust-based stack via [maplibre-native-rs](https://github.com/maplibre/maplibre-native-rs).

I think bundling `libwgpu_native.a` into the Linux WebGPU amalgam is the missing step to make downstream Rust consumers work cleanly.

Could you take a look and tell me whether you think this is the right direction?

If so, I’d also be happy to work on the macOS amalgam side in a separate PR.

CommanderStorm:
How much does this increase tha amalgam size?

yuiseki:
@CommanderStorm Thank you for the comment.

This PR does not change any existing amalgam artifacts. It only adds a new Linux x64 `wgpu` amalgam artifact.

So the practical answer is: no existing amalgam gets larger; this PR adds a new one. Its size (~385.5 MiB) looks consistent with bundling `libwgpu_native.a` (~51.6 MiB), so the size impact does not look abnormal to me.

CommanderStorm (approved):
In that case I am fine with it. Sorry, should have read your PR.
LGTM

@louwers obviously needs to +1 since I am not an cmake expert, but should be good.

yuiseki:
Tested the macOS side as well, building on top of this branch.

With `cmake --preset macos-webgpu-wgpu -DMLN_CREATE_AMALGAMATION=ON`, the `WGPU_STATIC_LIBRARY` resolution from `vendor/wgpu.cmake` works as-is on macOS. Adding the same conditional block to `platform/macos/macos.cmake` produces an amalgam with zero unresolved `wgpu*` symbols.

Also verified downstream by building `maplibre-native-rs` against the resulting `libmbgl-core-amalgam.a` via `MLN_CORE_LIBRARY_PATH`. Links and tests pass on macOS arm64.

I'm planning to open a follow-up PR for the macOS side once this lands. The reason for a separate PR is that the macOS amalgam currently does not include `mbgl-vendor-icu` or `mlt-cpp`, which causes unresolved symbols when linking from `maplibre-native-rs`. Adding those to the armerge command fixes the issue, but this is a pre-existing gap in the macOS amalgam unrelated to WebGPU, so I think it deserves its own review rather than being mixed into this PR.

yuiseki:
@louwers Sorry to ask again. :pray:

What I mainly want to understand is whether adding and maintaining a Linux WebGPU amalgam artifact is actually a direction you want for `maplibre-native`.

My impression is that some downstream users, especially `maplibre-native-rs`, are moving more toward local / non-amalgam builds by default. If that is the direction you want to encourage, I would rather stop here than continue with the planned macOS follow-up on top of the same idea.

So the real question for me is simply: do you want this direction, or not?

CommanderStorm (approved):
still lgtm, but @louwers is the cmake expert

yuiseki:
@louwers While resolving conflict of this PR, I noticed the following comment in PR #4222 :

https://github.com/maplibre/maplibre-native/pull/4222/changes#diff-6c4eafaf50e6a8379b9b803b15e59eae119b9046ffec5d93339c74d4521e7149R92-R93
```cmake
    # Use a specific version of WGPU. This is required when to a rust application
    # which uses wgpu directly instead of wgpu-native
```

I believe that #4222 already points us in the direction we should be going.

And `maplibre-native-rs` is stubbed there waiting on exactly this:
https://github.com/maplibre/maplibre-native-rs/blob/6782d1998439c96af5e4c6cbcc69160ab3141e9a/build.rs#L392

As I mentioned earlier, I have already prepared a diff that performs the same amalgamation on macOS.

This PR is a contained CMake change (4 files, no existing artifact affected). Since it is purely build/packaging, please call out anything off in the cmake and I will revise. Mostly I just need a thumbs up on the direction.

louwers (approved):
Feel free to merge.

yuiseki:
@louwers Thanks for the approval!

@CommanderStorm Thanks for the merge!

I’m sorry to ask for one more thing. Could one of you trigger https://github.com/maplibre/maplibre-native/actions/workflows/core-release.yml ?
This GitHub Action appears to be triggered only manually.
Edit: GitHub Action triggered by this PR, but release artifacts are publish only when manually triggered this workflow.

Once `core` is released, I’ll be able to start working on `maplibre-native-rs` right away.

---

About the macOS version of amalgamation: 
another issue has arisen, so I've created an issue for it.

- https://github.com/maplibre/maplibre-native/issues/4340

CommanderStorm:
Done

yuiseki:
@CommanderStorm Thanks for triggering core-release.yml 🙏 

One step seems to be missing. The Create Release step uses gh release create … --draft=true with no publish step, so the run produced a draft.

https://github.com/maplibre/maplibre-native/releases/tag/untagged-91b5f1cd9ab154d7257e

That is why the core-<sha> tag 404s: drafts don't create the tag until published. The Linux WebGPU amalgam is already attached to it.

Could a maintainer publish that draft? maplibre-native-rs only consumes published releases (its updater filters out drafts), so once it's live I can wire up the wgpu feature downstream.

CommanderStorm:
Ah, sorry. done

Issue #3749, opened 2025-08-22
Support OpenHarmony or provide documentation for cross-platform build


Astral-Yang:
**Is your feature request related to a problem? Please describe.**
I want to use `mplibre-native` in a cross-platform project that also targets [OpenHarmony](https://www.openharmony.cn/). Currently, there is no clear guidance or support for running on OpenHarmony, and this makes it hard to integrate.
I’m frustrated because I don’t know if the library can be compiled or adapted smoothly for OpenHarmony.

**Describe the solution you'd like**
It would be great if:

* Official support for OpenHarmony could be added, or
* At least some documentation/guidance could be provided for cross-platform builds (e.g., how to configure or what to modify).

**Describe alternatives you've considered**

* Trying to manually adjust the CMake/build system, but without documentation, this is trial-and-error.
* Searching community forums, but couldn’t find much information.

**Additional context**

* OpenHarmony is becoming more popular in IoT and cross-device projects.
* Adding official or partial support could help expand the library’s adoption.
* Even a minimal doc/example would be very helpful.

sargunv:
I'd love to help you get a working build for OpenHarmony, as I'm interested in Huawei's efforts on HarmonyOS Next from afar.

I live in the US so I don't easily have access to purchase Huawei devices to test on myself. I did pick up a couple of Huawei devices in Europe earlier this year, including a Pura 70 Pro on EMUI 14 and a Watch GT 5 Pro on HarmonyOS 5.1.0, but as a US resident I don't think I can create a Huawei developer account to access their latest documentation and tooling.

I know OpenHarmony runs on non-Huawei hardware too; what kinds of OpenHarmony devices are you trying to build for? What kind of CPU architecture (aarch64? risc-v?) do they run on? Do they have SDKs and emulators freely available? Do these systems support OpenGL or Vulkan? What is the UI toolkit like, and how is it typically integrated with native rendering surfaces (like videos or games)?

Astral-Yang:
> I'd love to help you get a working build for OpenHarmony, as I'm interested in Huawei's efforts on HarmonyOS Next from afar.
> 
> I live in the US so I don't easily have access to purchase Huawei devices to test on myself. I did pick up a couple of Huawei devices in Europe earlier this year, including a Pura 70 Pro on EMUI 14 and a Watch GT 5 Pro on HarmonyOS 5.1.0, but as a US resident I don't think I can create a Huawei developer account to access their latest documentation and tooling.
> 
[...]

PR #4231, opened 2026-04-17
fix(build,Qt,android,iOS,core): add the accept header to http requests

### LINKED ISSUE #4235: Supporting content negotiation via `Accept`
**Describe the bug**

As asked in https://github.com/maplibre/maplibre-style-spec/pull/1251#discussion_r2289493068 it is currently not possible to support MLT in tile servers currently.
This is because the tile server does not know if MVT or MLT should be returned.

This is a simple fix, just including the `Accept` header in the request for tiles.
It is a bit more complicated, because the cache/offline mode is not currently MVT/MLT aware and likely should, but that is a different issue.

**To Reproduce**

1. No `Accept` header
1. Don't know if MLT or MVT should be served
1. Tileserver needs to assume MVT for backwards compatability

**Expected behavior**

`Accept` header

### LINKED ISSUE #4236: Supporting content negotiation via `Accept`

Yes. It is currently not possible to correctly support MLT in tile servers because clients do not provide any signal indicating whether they expect MVT or MLT.
As a result, servers must guess and typically default to MVT for backwards compatibility.
Guessing is bad, we should not guess if we can avoid it.

From a tile server perspective, this is effectively a bug: if a client operates in an MLT-only mode and the server returns MVT, the request fails. There is no reliable way to serve the correct format, making MLT support non-functional in practice.

Tile requests should include an HTTP Accept header to explicitly negotiate the tile format (e.g. `application/vnd.mapbox-vector-tile` vs an MLT media type). This allows servers to deterministically return the correct format using standard HTTP content negotiation.
If this is not done this way and the client does expet MLT, but we trigger the fallback to return MVT this creates unnesary errors. Not good.

- don't support MLT (not an option)
- make mlt vs mvt a global switch in the tileserver (strictly worse, no benefit)
- Inferring the format from file extensions or endpoints (brittle at best, requires more changes to add MLT support, bad error messages, no "just works")
- ???

**Additional context**

This approach was previously discussed but not adopted, which now blocks interoperability between clients and tile servers for MLT. As a tile server maintainer, there is currently no OGC-Compliant correct way to implement MLT support.

A related issue is that cache and offline modes are also not format-aware (MVT vs MLT), but that does not change the need for telling the server what is being expected.

Rewrite of https://github.com/maplibre/maplibre-native/issues/4235

CommanderStorm:
As requested in https://github.com/maplibre/maplibre-style-spec/pull/1251#discussion_r2289493068 we should include the Accept header.

To be able to support MLT <-> MVT trans-ecoding in `maplibre/martin`, we need to know which format/mode MLN is currently in.
Without it, we need to guess, and we will be wrong in some cases... not great.

Transcoding via our new encoder is resonably cheap with the new rust encoder, so we should just do this if clients are in the mode to accept this.

I have reviewed all code, but Opus wrote it.
I think it is backwards compat, the API is not ideal due to this in a few places.
I have no idea which parts are under public api and what the rules are here in either cpp and java, so I would like to have advice on this

Resolves https://github.com/maplibre/maplibre-native/issues/4235
Resolves https://github.com/maplibre/maplibre-native/issues/4236


src/mbgl/tile/tile_loader_impl.hpp
```diff
@@ -27,7 +28,8 @@ TileLoader<T>::TileLoader(T& tile_,
                               id.canonical.y,
                               id.canonical.z,
                               tileset.scheme,
-                              Resource::LoadingMethod::CacheOnly)),
+                              Resource::LoadingMethod::CacheOnly,
+                              acceptHeader)),
```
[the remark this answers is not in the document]
CommanderStorm: not sure if "I" came up with a good solution to this. It looks good and the tests seem to work, but this entire offline/sideloading system seems a bit weirdly designed.

I don't quite get why it was written this way..

CommanderStorm:
No clue why CI is failing on the change. i just tasked claude to fix it and am going to take a nap

louwers:
Converted to draft, because this PR is not ready for review.

> Without it, we need to guess, and we will be wrong in some cases... not great.

I'd say: don't guess. Serve MLT and MVT on different endpoints, like @TimSylvester suggests in the thread you linked.

@CommanderStorm Please create an issue so we can discuss this change first.

CommanderStorm:
> I’d say: don’t guess

Agreed - the server shouldn’t have to guess.

The correct way to handle this is for the client to specify the desired format via the `Accept` header. Without that, it’s impossible to reliably determine whether MVT or MLT should be returned.

Inferring the format from the file extension (e.g. `.mvt` vs `.mlt`) is at best a weak heuristic and not something we should rely on.

If we use the `Accept` header as intended, this works cleanly and predictably out of the box.

test/BUILD.bazel
```diff
@@ -88,6 +88,7 @@ cc_test(
         "//:scripts/style-spec-reference/v8.json",
     ],
     tags = ["no-sandbox"],
+    timeout = "eternal",
```
CommanderStorm:
```suggestion
```

louwers (commented):
If you can get this  PR green we can merge it.

Also the cache handling still needs to be implemented.

include/mln/util/image.hpp
```diff
@@ -179,4 +180,6 @@ using AlphaImage = Image<ImageAlphaMode::Exclusive>;
 PremultipliedImage decodeImage(const std::string&);
 std::string encodePNG(const PremultipliedImage&);
 
+std::string_view rasterAcceptHeader();
```
louwers: This class has nothing to do with HTTP. Seems weird to add it here.

include/mln/util/constants.hpp
```diff
@@ -71,6 +71,11 @@ constexpr uint32_t DEFAULT_MAXIMUM_CONCURRENT_REQUESTS = 20;
 
 constexpr uint8_t TERRAIN_RGB_MAXZOOM = 15;
 
+constexpr const char* MIME_TYPE_MVT = "application/vnd.mapbox-vector-tile";
+constexpr const char* MIME_TYPE_MLT = "application/vnd.maplibre-tile";
+constexpr const char* MIME_TYPE_RASTER = "image/webp, image/jpeg, image/png";
+constexpr const char* MIME_TYPE_RASTER_NO_WEBP = "image/jpeg, image/png";
```
louwers: This smells a little bit. Would be cleaner to create a constexpr function for it if it can be known at compile time.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/HttpRequest.java
```diff
@@ -26,6 +26,24 @@ public interface HttpRequest {
...
+   * @param httpRequest  callback to be invoked when we receive a response
+   * @param nativePtr    the pointer associated to the request
+   * @param resourceUrl  the resource url to download
+   * @param dataRange    http header, used to indicate the part of a resource that the server should return
+   * @param etag         http header, identifier for a specific version of a resource
+   * @param modified     http header, used to determine if a resource hasn't been modified since
+   * @param offlineUsage flag to indicate a resource will be used for offline, appends offline=true as a query parameter
+   * @param acceptHeader the Accept header value to send with the request
+   */
+  default void executeRequest(HttpResponder httpRequest, long nativePtr, String resourceUrl,
```
louwers, line 41: Why add another method that is essentiallly the same?

louwers (changes requested):
I don't see any changes to the cache handling, so I think this will serve MVTs from the cache if you need MLTs...

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/HttpRequest.java
CommanderStorm: to not break existing callers, since that does not sound so semver, or is this not public api?

include/mln/util/constants.hpp
CommanderStorm: Does not work, we won't know if we are on the ios version that does not support webp.
But I can move it do an different module if you like.
I instructed claude to do that.

include/mln/util/image.hpp
CommanderStorm: I mean I can keep the reason why it was added.. to be able to differentiate between webp and not webp platforms..

src/mln/util/accept_header.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+/// The value of an HTTP `Accept` header offering the given MIME types, most preferred first.
+/// Assembled at compile time and null terminated, so it can be handed to C APIs as well.
+template <const std::string_view& first, const std::string_view&... rest>
+inline constexpr std::string_view acceptHeader{detail::acceptHeaderStorage<first, rest...>.data(),
+                                               detail::acceptHeaderStorage<first, rest...>.size() - 1};
+
+/// The `Accept` header for raster tile requests, limited to what this platform can decode.
+inline std::string_view rasterAcceptHeader() {
+    return supportsWebPDecoding() ? acceptHeader<MIME_TYPE_WEBP, MIME_TYPE_JPEG, MIME_TYPE_PNG>
```
CommanderStorm, line 47: I need to get this information from the platform specific image module, so either I have the information about the image module here or the information about headers there..

I don't really think the constexpr is cleaner than just my version (the two static strings) that I need to construct, but that has been rejected, so 🤷🏻

CommanderStorm:
> I don't see any changes to the cache handling, so I think this will serve MVTs from the cache if you need MLTs...

I would prefer to do this in another PR, so fix one bug at a time.
They are unrelated bugs, but if you insist I can do it here.

platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/http/HttpRequest.java
CommanderStorm: `HttpRequest` is public, so therefore I think public api, therefore should not get changed according to what you said about not breaking consumers..

platform/default/src/mln/storage/offline_database.cpp
```diff
@@ -26,7 +26,7 @@ std::string OfflineDatabase::cacheKey(const Resource& resource) {
 
 std::string OfflineDatabase::cacheKey(const Resource::TileData& tile) {
     if (tile.vectorEncoding != Tileset::VectorEncoding::MLT) return tile.urlTemplate;
-    return tile.urlTemplate + tile.urlTemplate.find('?') == std::string::npos ? "#mln" : "&mln";
+    return tile.urlTemplate + (tile.urlTemplate.find('?') == std::string::npos ? "#mln" : "&mln");
```
louwers: Probably time to factor out the query parameter handling...

I can help if you want.

src/mln/util/accept_header.hpp
louwers: Busting out the template parameter packs I see.

platform/default/src/mln/storage/offline_database.cpp
CommanderStorm: okay, I guess I tasked claude to pull the query char handling to the line before this - if you are thinking of something different, could you be more specific what kind of refactoring you would like (or instruct claude to do that refactoring)

src/mln/util/accept_header.hpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/// The encoding a tileset asks for, defaulting to MVT
+inline Tileset::VectorEncoding vectorEncodingOf(const Tileset& tileset) {
+    return tileset.vectorEncoding.value_or(Tileset::VectorEncoding::Mapbox);
+}
+
+constexpr std::string_view vectorAcceptHeader(Tileset::VectorEncoding encoding) {
+    return encoding == Tileset::VectorEncoding::MLT ? acceptHeader<MIME_TYPE_MLT> : acceptHeader<MIME_TYPE_MVT>;
+}
```
ciscorn: Should we keep compatibility with servers that serve MVT using other content types?

```suggestion
constexpr std::string_view vectorAcceptHeader(Tileset::VectorEncoding encoding) {
    constexpr std::string_view mvtAcceptHeader =
        "application/vnd.mapbox-vector-tile, application/x-protobuf;q=0.9, "
        "application/octet-stream;q=0.8, */*;q=0.1";
    return encoding == Tileset::VectorEncoding::MLT ? acceptHeader<MIME_TYPE_MLT> : mvtAcceptHeader;
}
```

src/mln/style/sources/vector_source.cpp
```diff
@@ -32,6 +32,7 @@ void VectorSource::setTilesetOverrides(Tileset& tileset) {
     if (minZoom) {
         tileset.zoomRange.min = static_cast<uint8_t>(*minZoom);
     }
+    tileset.vectorEncoding = encoding;
```
ciscorn: This also overwrites TileJSON’s encoding when the style omits it. Could `encoding` be optional, with tests? The offline path has the same issue.

platform/default/src/mln/storage/http_file_source.cpp
```diff
@@ -285,6 +285,11 @@ HTTPRequest::HTTPRequest(HTTPFileSource::Impl *context_, Resource resource_, Fil
         headers = curl_slist_append(headers, time.c_str());
     }
 
+    if (!resource.acceptHeader.empty()) {
```
ciscorn: Could OHOS forward this header too? Its backend currently doesn’t.

src/mln/util/accept_header.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <array>
+#include <cstddef>
+#include <string_view>
+
+namespace mln {
+namespace http {
+
+/// MIME types of the tile formats we know how to request.
+inline constexpr std::string_view MIME_TYPE_MVT = "application/vnd.mapbox-vector-tile";
+inline constexpr std::string_view MIME_TYPE_MLT = "application/vnd.maplibre-tile";
```
ciscorn: Is this media type defined in the MLT spec? GL JS doesn’t send it for tile requests yet.

include/mln/util/image.hpp
```diff
@@ -179,4 +179,7 @@ using AlphaImage = Image<ImageAlphaMode::Exclusive>;
 PremultipliedImage decodeImage(const std::string&);
 std::string encodePNG(const PremultipliedImage&);
 
+/// Whether `decodeImage` can decode WebP on this platform.
+bool supportsWebPDecoding();
```
ciscorn: OHOS has no definition for this, so its NAPI module should fail to link. Could we add one?

src/mln/tile/vector_tile.cpp
```diff
@@ -13,7 +14,8 @@ VectorTile::VectorTile(const OverscaledTileID& id_,
                        const Tileset& tileset,
                        TileObserver* observer_)
     : GeometryTile(id_, std::move(sourceID_), parameters_, observer_),
-      loader(std::make_unique<TileLoader<VectorTile>>(*this, id_, parameters_, tileset)) {}
+      loader(std::make_unique<TileLoader<VectorTile>>(
```
ciscorn: Could we test this through `VectorTile` → `TileLoader`? The current test creates `Resource` directly.

platform/default/src/mln/storage/http_file_source.cpp
CommanderStorm: I have no idea, this PR is not targeting that platform since it was added after this PR was openend. I also don't have a way to test on that platform I think..

src/mln/style/sources/vector_source.cpp
CommanderStorm: I don't quite follow architecturally.
Making `encoding` optional in the constructor and all sounds like a lot of work for what? We know the encoding, it is not optional in the default because it has an default..

platform/default/src/mln/storage/http_file_source.cpp
CommanderStorm: @sargunv maybe you know more how that device works.
I don't think they work in android studio or the emulator at least.

include/mln/util/image.hpp
CommanderStorm: It is not failing CI... I think it is linking fine, or?

I have no clue how to test this. @sargunv can you help me set up an dev environment? That plaform seems a bit esoteric to setup an dev environment for and I don't really know where to get started for this..

platform/default/src/mln/storage/http_file_source.cpp
CommanderStorm: > The emulator is available only "[on the Chinese mainland](https://developer.huawei.com/consumer/en/doc/harmonyos-guides/ide-tools-overview#:~:text=The%20following%20features%20are%20available%20only%20in%20the%20Chinese%20mainland%3A)"

I am not on the chinese mainland, so don't have emulator access or a way to test this change as far as I read https://github.com/maplibre/maplibre-native/issues/3749#issuecomment-3215566861 that is required for this platform.

src/mln/util/accept_header.hpp
CommanderStorm: I can remove the `MIME_TYPE_MLT` and image header handling entirely if you'd prefer.
I only want to know if as a server I can send MLT or if this would cause an really bad error message.

PR #3541, opened 2025-06-10
Add `MapRenderer` dangling reference check

adrian-cojocaru:
Follow-up to https://github.com/maplibre/maplibre-native/pull/3522

Issue #3517, opened 2025-06-02
Crash on Android: FORTIFY: pthread_mutex_lock called on a destroyed mutex


rickyputrah:
### MapLibre Android Version

11.9.0

### Android SDK Version

Android 10 - 14

### Device

Various Device ( we can't repro it by our own)

### What happened?

We're seeing crashes in production reported via Google Play Console with the following message:

```
FORTIFY: pthread_mutex_lock called on a destroyed mutex
```

This started appearing with MapLibre SDK version 11.8.7 (we just started using MapLibre in our app) and continues after updating to 11.9.0.
Unfortunately, we're unable to reproduce the crash locally so far.

The crash seems to be related to native threading/mutex handling, potentially within the SDK's native layer.

### Steps to reproduce

We haven't been able to reproduce this issue on our end. The crash only appears in production

### Relevant log output

```shell
backtrace:
  #00  pc 0x0000000000059b38  /apex/com.android.runtime/lib64/bionic/libc.so (abort+164)
  #01  pc 0x000000000005b8bc  /apex/com.android.runtime/lib64/bionic/libc.so (__fortify_fatal(char const*, ...)+124)
  #02  pc 0x00000000000bfb38  /apex/com.android.runtime/lib64/bionic/libc.so (HandleUsingDestroyedMutex(pthread_mutex_t*, char const*)+60)
  #03  pc 0x00000000000bf9c8  /apex/com.android.runtime/lib64/bionic/libc.so (pthread_mutex_lock+280)
  #04  pc 0x0000000000a1d078  /data/app/~~FRRbnHghBaRvUqVfstVbEA==/com.xxx.xxx-5Q62mE1q3Ap5yKH8U_zI6g==/lib/arm64/libmaplibre.so (std::__ndk1::mutex::lock()+12) (BuildId: 35cd4f667f754a0f5f902d3359c3804c722da49d)
  #05  pc 0x00000000004e6c34  /data/app/~~FRRbnHghBaRvUqVfstVbEA==/com.xxx.xxx-5Q62mE1q3Ap5yKH8U_zI6g==/lib/arm64/libmaplibre.so (mbgl::android::MapRenderer::update(std::__ndk1::shared_ptr<mbgl::UpdateParameters>)+44) (BuildId: 35cd4f667f754a0f5f902d3359c3804c722da49d)
[...]

PR #4233, opened 2026-04-17
fix(renderer): guard ImageManager callback against use-after-free

### LINKED ISSUE #4232: FORTIFY pthread_mutex_lock on destroyed recursive_mutex inside ImageManager callback on Android (13.0.2)
## What happened

Native crash on Android with maplibre-native 13.0.2 (opengl variant):

```
FORTIFY: pthread_mutex_lock called on a destroyed mutex
    libc.so  HandleUsingDestroyedMutex
    libc.so  pthread_mutex_lock
    libmaplibre.so  std::__ndk1::recursive_mutex::lock()
    libmaplibre.so  <lambda>
    libmaplibre.so  mbgl::android::MapRendererRunnable::run(JNIEnv&)
    <jit>           MapLibreGLSurfaceView$GLThread.guardedRun
    base.apk        MapLibreSurfaceView$RenderThread.run
```

## Root cause

`ImageManager::checkMissingAndNotify` schedules a `removePendingRequests`
callback via `Scheduler::bindOnce`:

```cpp
auto removePendingRequests = [this, missingImage] {
    std::scoped_lock readWriteLock(rwLock);
    auto existingRequest = requestedImages.find(missingImage);
    ...
};
observer->onStyleImageMissing(missingImage,
    Scheduler::GetCurrent()->bindOnce(std::move(removePendingRequests)));
```

The lambda captures `this` (the `ImageManager`) by raw pointer. On
Android, `Scheduler::bindOnce` eventually routes the inner lambda through
`MapRenderer::schedule` to a `MapRendererRunnable` in the Java `queueEvent`
queue. If the Renderer/ImageManager is destroyed while the runnable is
still pending, the next invocation on the render thread dereferences a
destroyed `ImageManager` and locks a destroyed `std::recursive_mutex`.

## How we hit it

Real mobile map with `@maplibre/maplibre-react-native 11.0.0-beta.30`,
React Native New Architecture, release build, Pixel 7a / Android 16:

1. Style has symbol layers that reference icons not yet in the style —
   `onStyleImageMissing` fires and queues at least one
   `removePendingRequests` callback via `bindOnce`.
2. User opens a `@react-navigation/drawer`.
3. The drawer's Reanimated-driven parent layout triggers a `SurfaceView`
   destroy/recreate cycle on this device.
4. `MapRenderer::onSurfaceDestroyed` → `resetRenderer()` → `Renderer` dtor
   → `ImageManager` dtor. `rwLock` is destroyed.
5. The still-pending `MapRendererRunnable` runs the old lambda → FORTIFY.

The crash is release-sensitive — release render cadence gives the queued
callback a wide window between ImageManager teardown and renderer
recreate. Debug builds don't reliably hit it. Using `androidView="texture"`
(TextureView variant) avoids it entirely because the TextureView
lifecycle does not drive `onSurfaceDestroyed` on drawer layout changes.

## Why #3522 / #3541 didn't cover this

Those fixes (merged and shipped in 13.0.2) guard
`AndroidRendererFrontend::update` and `MapRenderer::requestRender`
against the Java peer outliving the C++ side. The `ImageManager`
scheduled-callback path is a distinct code path — it still captures
`this` by raw pointer with no weak-reference guard.

## Symbolication

Tombstone frames point to offset `0x7018c0` in `libmaplibre.so` (build id
`6ff505d11b9e06a8cddc89dbbb99a170b170997a`, from the official 13.0.2
opengl AAR). Disassembling that offset shows the canonical
"lock recursive_mutex at `state+0xe8` → `std::map::find` at `state+0x70`
→ iterate → tail-unlock" pattern, which matches the body of
`removePendingRequests` (locking `rwLock`, finding in `requestedImages`,
iterating `ImageRequestor*`). No other `recursive_mutex` callsite in
`libmaplibre.so` matches.

## Proposed fix

`ImageManager` is already owned by `std::shared_ptr` via
`RenderOrchestrator::imageManager`, so the minimal, safe fix is:

- `ImageManager` inherits `std::enable_shared_from_this<ImageManager>`.
- `removePendingRequests` captures `weak_from_this()` instead of `this`,
  locks the weak_ptr, and early-returns if it's expired.

PR: #4233.

## Environment

- `maplibre-native` Android 13.0.2 (opengl variant, arm64-v8a)
- `@maplibre/maplibre-react-native` 11.0.0-beta.30
- Expo SDK 55, React Native 0.83.4, New Architecture enabled
- Reproduced on Pixel 7a, Android 16 (build `BP4A.260105.004.E1`)

## Workarounds (until fix lands)

1. `androidView="texture"` on `<MapView>` (prop from
   maplibre-react-native) — avoids the SurfaceView path entirely.
2. Unmount the `<MapView>` before opening the drawer — destroys the
   renderer cleanly before the racy teardown window opens.

Both work but neither is a product-acceptable long-term answer.

jaltin:
## Summary

`ImageManager::checkMissingAndNotify` schedules a `removePendingRequests`
callback via `Scheduler::bindOnce`, which then dispatches through the
Android `MapRenderer` to a `MapRendererRunnable` sitting in the Java
`queueEvent` queue. The lambda captures `this` (the `ImageManager`) by raw
pointer. When the Renderer – and the ImageManager it owns – is destroyed
while this callback is still pending, the next invocation on the render
thread dereferences a destroyed `ImageManager` and locks a destroyed
`std::recursive_mutex` (`rwLock`), aborting with:

```
FORTIFY: pthread_mutex_lock called on a destroyed mutex
    libc.so  HandleUsingDestroyedMutex
    libc.so  pthread_mutex_lock
    libmaplibre.so  std::__ndk1::recursive_mutex::lock()
    libmaplibre.so  <removePendingRequests lambda>
    libmaplibre.so  mbgl::android::MapRendererRunnable::run
    <MapLibreGLSurfaceView$GLThread.guardedRun>
```

This PR lets `ImageManager` inherit `std::enable_shared_from_this<ImageManager>`
(it is already owned by `std::shared_ptr` via
`RenderOrchestrator::imageManager`) and captures `weak_from_this()` inside
the lambda. The callback bails out before locking if the ImageManager has
been destroyed in the meantime.

## Reproduction

Android 16 on a Pixel 7a, `maplibre-native 13.0.2` (opengl variant) +
`@maplibre/maplibre-react-native 11.0.0-beta.30`, React Native New
Architecture, **release** build:

1. Load a map style whose symbol layers reference icons that are not yet
   in the style image pool — so `onStyleImageMissing` fires and queues at
   least one `removePendingRequests` callback through `bindOnce`.
2. Open a drawer from `@react-navigation/drawer`. On this device class,
   the drawer's Reanimated-driven parent layout change triggers a
   `SurfaceView` destroy/recreate cycle which runs
   `MapRenderer::onSurfaceDestroyed -> resetRenderer() -> Renderer dtor ->
   ImageManager dtor`.
3. The next `MapRendererRunnable` off the GL-thread queue runs the pending
   lambda, which locks the now-destroyed `rwLock` → FORTIFY abort.

The race is release-sensitive because release render cadence + optimized
layout gives the callback a wide window to land right after the teardown.
Debug builds and the TextureView renderer dodge the crash (TextureView's
surface lifecycle doesn't drive `onSurfaceDestroyed` the same way and does
not tear down the renderer in response to layout changes).

## Stack, symbolicated

From a device tombstone after the crash on 13.0.2 (build id
`6ff505d11b9e06a8cddc89dbbb99a170b170997a`):

```
  #00 libc.so  __fortify_fatal
  #01 libc.so  HandleUsingDestroyedMutex
  #02 libc.so  pthread_mutex_lock
  #03 libmaplibre.so  std::__ndk1::recursive_mutex::lock()
  #04 libmaplibre.so  <checkMissingAndNotify::removePendingRequests lambda>
  #05 libmaplibre.so  mbgl::android::MapRendererRunnable::run(JNIEnv&)
  #07 <jit>  MapLibreGLSurfaceView$GLThread.guardedRun
  #09 base.apk  MapLibreSurfaceView$RenderThread.run
```

The lambda at #04 matches the disassembly pattern of "lock mutex at
`state+0xe8` → `std::map::find` at `state+0x70` → iterate → unlock" —
which is exactly the body of `removePendingRequests` (locking `rwLock`,
finding in `requestedImages`, iterating `ImageRequestor*` entries). No
other recursive-mutex lock in the codebase matches that shape.

## Is this the same thing PR #3522 already fixed?

Related but not the same. #3522 + #3541 guarded
`AndroidRendererFrontend::update` / `MapRenderer::requestRender` against
Java-side GC outliving the C++ side. Both shipped in 13.0.2. The
`ImageManager` scheduled-callback path is a distinct code path that
wasn't covered by those PRs.

## Changes

```
src/mbgl/renderer/image_manager.hpp  |  3 ++-
src/mbgl/renderer/image_manager.cpp  | 18 ++++++++++++++----
```

- `ImageManager` gains `std::enable_shared_from_this<ImageManager>`.
- `removePendingRequests` captures `weak_from_this()` instead of `this`,
  locks the weak pointer, and early-returns if the ImageManager is gone.

No behavior change when the ImageManager is alive.

## Verification

Fork of `android-v13.0.2` with this patch, built and installed on the
repro device:

- Real map screen + drawer cycle: crash gone over many repeats.
- `libmaplibre.so` BuildId changed (`6ff505d1…` → `a42d0b95…`),
  confirming the patched binary is what's loaded.
- Full upstream Android unit tests / integration tests have NOT been run
  as part of this fork — happy to run them through CI on this branch.

## Test plan

- [ ] Existing CI passes.
- [ ] An added test that schedules `onStyleImageMissing`, destroys the
      ImageManager (e.g. by resetting the renderer), then flushes the
      scheduler — expects the callback to no-op rather than abort. Happy
      to add this if reviewers point me at the right harness.

## Related

- Related-but-different crash trace: #3517 (fixed by #3522 / #3541, both
  included in 13.0.2; this PR fixes a remaining sibling path).
- Reported in the wild from a React Native mobile app; tracking issue in
  `maplibre-react-native`: maplibre/maplibre-react-native#1455.

Closes #4232.


louwers (changes requested):
Thanks for the PR. Seems like a good problem to fix, but I think the comment is unnecessary.

Also, would it be possible to make the constructor(s) of ImageManager private? See https://en.cppreference.com/cpp/memory/enable_shared_from_this

jaltin:
Thanks @louwers! Both addressed in 53328f0:

1. The explanatory comment is gone — the rationale lives in the PR body and linked issue #4232.
2. `ImageManager`'s constructor is now private and creation goes through a static `ImageManager::create()` factory returning `std::shared_ptr<ImageManager>`. This matches the [cppreference guidance](https://en.cppreference.com/cpp/memory/enable_shared_from_this) and statically rules out stack-/`unique_ptr`-constructed instances (which would have made `weak_from_this()` return an empty weak_ptr and silently no-op our callback).

Ripple was small: one production call site (`render_orchestrator.cpp` switched from `std::make_unique<ImageManager>()` to `ImageManager::create()`) and a handful of test TUs that were using `std::make_shared<ImageManager>()` (which needs a public ctor). Those all now use the factory too.

No behavior change to the fix itself. CI should tell us if I missed a call site somewhere.

src/mbgl/renderer/image_manager.hpp
```diff
@@ -57,6 +58,10 @@ class ImageManager {
     void clear();
 
 private:
+    // Private so instances can only be created via `create()`, guaranteeing
```
louwers: This is pretty standard for enable_shared_from_this, so comment can be removed I think.

src/mbgl/renderer/image_manager.hpp
```diff
@@ -57,6 +58,10 @@ class ImageManager {
...
+    // shared_ptr ownership and well-defined `weak_from_this()` behavior.
```
louwers: Can be removed

test/renderer/image_manager.test.cpp
```diff
@@ -16,13 +16,15 @@
 using namespace mbgl;
 
 TEST(ImageManager, Missing) {
-    ImageManager imageManager;
+    auto imageManagerPtr = ImageManager::create();
+    auto& imageManager = *imageManagerPtr;
```
louwers: This test and others: why not use ->

louwers (commented):
Please see minor comments but looks good.

jaltin:
Thanks — addressed in 1e2b53a:

- Dropped the 2-line explanatory comment above the private `ImageManager()` ctor.
- Rewrote the `image_manager.test.cpp` fixtures to use the `shared_ptr` directly via `->` instead of the awkward `auto& = *ptr;` alias I'd introduced. The tests now read as `auto imageManager = ImageManager::create();` followed by `imageManager->foo(...)`. No behavior change.

jaltin:
@louwers — follow-up commits `53328f0` and `1e2b53a` address all of your comments (private ctor + factory, dropped the explanatory comments, switched the tests to `->`). Would you mind re-reviewing when you get a chance? The PR is still showing "Changes requested" from the first pass.

On the red Android CI: `android-build (opengl)` and `android-build (vulkan)` are failing in `MapLibreAndroidTestApp:compileVulkanReleaseKotlin` with

> Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0.

…for `kotlin-stdlib 2.2.20` and `kotlinx-coroutines 1.11.0-rc01`. That's a Kotlin toolchain/transitive-dep mismatch in the Android test-app module, which this C++-only PR doesn't touch. The native builds (bloaty + benchmark) succeeded. The branch is already on top of current `origin/main` (`dacb7d4e`), which was green on android-ci on 2026-04-17 — so this looks like a new transitive or cached-dep regression rather than anything from this PR. Happy to rebase/force-push to trigger a fresh run if that would help, or defer to you on the right fix.

jaltin:
@louwers — when you have a moment, would you mind taking another look? The original review on Apr 17 left “Changes requested” for two things, both addressed:

Drop the redundant comment → done in 53328f0 and 1e2b53a
Private constructor per enable_shared_from_this guidance → done in 53328f0 (ImageManager constructors are now private, creation goes through ImageManager::create())
Your Apr 18 follow-up (drop two more comment lines, use -> in tests) was also handled in 1e2b53a. The “Changes requested” status is just the original review that hasn’t been dismissed — there’s nothing left outstanding from my side. Thanks!

louwers:
Thanks @jaltin let me have another look.

louwers (approved):

louwers:
Will merge once CI passes, thanks for your contribution.

Sorry for not reviewing it earlier.

louwers:

PR #4238, opened 2026-04-18
feat(linux): add Nix flake for reproducible build environments

jamesarich:
## Description

Adds a `flake.nix` that provides reproducible dev shells and build packages for Linux, managing all native dependencies through Nix instead of system packages or CMake FetchContent.

This was suggested by @louwers in [#4229 (comment)](https://github.com/maplibre/maplibre-native/pull/4229#issuecomment-4273217641):

> Also I'm thinking, maybe instead of CMake we should be using a Nix flake for managing these dependencies...

### What it provides

- **`nix develop`** — kitchen-sink shell with all Linux deps (cmake, clang, ninja, armerge, image codecs, X11, OpenGL, Vulkan, Wayland)
- **`nix develop .#amalgamation`** — shell with static `.a` libraries for amalgamation builds
- **`nix build .#mbgl-core-opengl`** / **`nix build .#mbgl-core-vulkan`** — build mbgl-core under Nix
- **`nix build .#mbgl-core-opengl-amalgamation`** — full amalgamation build with armerge

### How it works

- Nix provides all system-level deps (curl, libpng, libjpeg-turbo, libwebp, libuv, zlib, bzip2, ICU, OpenSSL, X11, GL/Vulkan)
- Static library variants are built via Nix overrides (`dontDisableStatic`, `-fPIC`) for the amalgamation path
- `armerge` is packaged from source (not yet in nixpkgs)
- Scoped to `x86_64-linux` and `aarch64-linux` only

### Verification

Tested in Docker (`nixos/nix:2.25.5`, `aarch64-linux`):

- [x] `nix develop` resolves all tools and libraries
- [x] `cmake --preset linux-opengl` configures successfully
- [x] Full build compiles 487 objects
- [x] All static `.a` libs verified PIC-safe (no TEXTREL)
- [x] Amalgamation: 338 MB archive, 4799 `mbgl*` exports, zero leaked vendored symbols
- [x] `nix flake check` evaluation passes, no broken darwin outputs

### Open questions

- Does this direction make sense as a complement to (or replacement for) the CMake FetchContent approach in #4229?
- Should CI adopt `nix develop` instead of `.github/scripts/install-linux-deps`?
- Would a Nix-based CI cache speed up builds?

### Files changed

- `flake.nix` — **New**: dev shells, packages, static lib overrides, armerge packaging
- `flake.lock` — **New**: pinned nixpkgs-unstable + flake-utils

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1776329215,
+        "narHash": "sha256-a8BYi3mzoJ/AcJP8UldOx8emoPRLeWqALZWu4ZvjPXw=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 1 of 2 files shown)


louwers:
Closing for now, I think @sargunv cooked something up with better cross platform support.

PR #4242, opened 2026-04-21
macos: add release workflow for GitHub Releases

fonok3:
Adds a manually-dispatched `macos-release.yml` workflow that builds and publishes native macOS XCFrameworks as GitHub Releases under the `macos-v*` tag namespace, mirroring `ios-release.yml`.

Artifacts per release:
- `MapLibre.dynamic.xcframework.zip` + dSYM
- `MapLibre.static.xcframework.zip`

The workflow follows the same release approach as the iOS workflow — build xcframeworks via Bazel, publish as GitHub Release with changelog extraction, tag creation, and PR notifications. Releasing into an SPM distribution repo is out of scope for this PR and can be added as a follow-up.

Context: #4088

The GitHub Workflow was tested in the fork repository.

## Open topic

@louwers We need to decide on the initial macOS version number. The last macOS release was `0.19.1` (from 2021), but the codebase is largely shared with iOS which is at `6.25.0`. I'd propose aligning with the iOS version (`6.25.0`) rather than continuing from `0.x`. At least I would jump to any `6.x` version to show the parity.

## Summary by CodeRabbit

* **New Features**
  * Added support for manually triggered macOS SDK releases, including full and pre-release options.
  * macOS releases now provide dynamic and static XCFramework archives with debug symbols.

* **Documentation**
  * Updated the macOS SDK changelog with version 6.27.0 highlights.
  * Added clearer macOS release labels to release notifications.


louwers:
@fonok3 Sorry for never getting back to you on this.

I agree that aligning with iOS releases is a good idea, but maybe we would reflect this in the workflow by reading the version file from iOS. What do you think?

fonok3:
@louwers

No problem! Thanks for responding.

Would be okay for me. So would you just read the iOS version file for now, or would you also move it so it’s clear that it’s shared between iOS and macOS?

And would you still keep the changelog split or also shared?

I’d lean towards only reading the version file from the macOS workflow and keeping the changelogs split, so this PR doesn’t change the general structure.

louwers:
> I’d lean towards only reading the version file from the macOS workflow and keeping the changelogs split, so this PR doesn’t change the general structure.

Sounds good to me.

fonok3:
@louwers I updated the workflow to read the version from the iOS VERSION file. The CodeRabbit findings are all inherited from ios-release.yml, which I copied, so none are macOS specific. I’d fix them across all workflows in a separate PR instead of diverging here.

louwers (changes requested):
Do you want to try `macos-26`? I don't know how long the `macos-14` runner will still be around.

louwers:
@coderabbitai help

louwers:
@coderabbitai resolve

louwers (approved):
It will be somewhat annoying to keep the workflows in sync. Let's try to see if we can share some code in a follow up...

fonok3:
Thanks a lot!

louwers:
@fonok3 I will run the workflow later today. I'll message you in case it fails.

fonok3:
@louwers do you have a plan for when you're going to try out the workflow?

PR #4243, opened 2026-04-21
iOS: add white style option for compass button

### LINKED ISSUE #3929: Support for styling MLNCompassButton
As far as I can find, the MLNCompassButton doesn't support any kind of style customization

I would at least support a light vs dark mode similar to the `scaleBarShouldShowDarkStyles` which is available for the scale.
Another (but maybe more complex) option would be to allow specifying the tint color for the various piece of UI.

Implement a custom compass control, but it looks an overkill

**Additional context**
As we show other controls on top of the map and they are all white, I'd like to make the compass white as well

nagisa77:
## Summary
- add a compass white style option on iOS (`MLNCompassButton.shouldWhiteStyle`)
- expose map-level convenience API (`MLNMapView.compassViewShouldWhiteStyle`)
- load a dedicated white-style compass asset (`Compass-white`) instead of runtime tinting
- add/update compass tests for style switching behavior

## Motivation
This addresses the request to make the default compass style configurable for light controls on top of dark maps.

Closes #3929.


nagisa77:
Hello @louwers~ Can you read my PR?

louwers (changes requested):
Generally a good idea I think.

But instead of a boolean, I would use an enum so we can add other styles later.

What would be even better is if people could add their own style externally so we don't bloat the library with different images. What do you think?

PR #4247, opened 2026-04-22
 (android): apply size optimization flags to RelWithDebInfo build config

grab-liyanjin:
## Problem

  Android AGP maps the `release` build variant to the `RelWithDebInfo`
  CMake config — **not** `Release`. All existing size-optimization flags
  in `platform/android/android.cmake` were conditioned on
  `$<$<CONFIG:Release>:...>`, which means they were **never applied**
  during a normal Android release build.

  This was confirmed by inspecting the generated `build.ninja` for a
  release AAR: the actual compile flags were `-O2 -g -DNDEBUG` (the
  RelWithDebInfo defaults), with no `-Oz`, `-flto`, or `--icf=all`.

  ## Fix

  Introduce a `_OPT_CONFIGS` generator expression covering both `Release`
  and `RelWithDebInfo`, and apply it to all existing size-opt flags:

  ```cmake
  set(_OPT_CONFIGS "$<OR:$<CONFIG:Release>,$<CONFIG:RelWithDebInfo>>")

  Also switches the linker from gold to lld (-fuse-ld=lld), which
  is the recommended linker in NDK 27+ and produces slightly smaller output.

  Measured Result

  Tested on a full assembleOpenglRelease build (NDK 28, 4 ABIs):

  ┌─────────────┬─────────┬─────────┬───────┐
  │     ABI     │ Before  │  After  │ Delta │
  ├─────────────┼─────────┼─────────┼───────┤
  │ Full AAR    │ 18 MB   │ 14 MB   │ -22%  │
  └─────────────┴─────────┴─────────┴───────┘

  Notes

  - Single-file change, no behaviour change — build config only
  - Verified via build.ninja inspection that flags are present in the
  release build after this fix
  - Tested on Android device (arm64, OpenGL ES renderer)


louwers (approved):
Thank you, I think you're right here...

PR #4256, opened 2026-04-24
Optimize fill extrusion memory by using instancing [Core optimization]

### LINKED ISSUE #4107: Fill-extrusion memory usage on mobile is prohibitively high at street-level zoom
## Problem

Fill-extrusion layers consume 900MB-1.6GB+ of memory on iOS at zoom 17 with 3D buildings enabled. This makes it impractical to use building extrusions alongside other 3D content (such as GLTF models via the custom layer plugin) in navigation or street-level scenarios on mobile.

Memory scales directly with zoom level. Zooming out reduces memory significantly and improves rendering smoothness, confirming the cost is driven by the tile/extrusion system rather than any custom layer content.

## Observed behavior

| Scenario | Memory (iPhone 15 Pro) |
|---|---|
| Map at zoom 17, no extrusions | ~400MB |
| Map at zoom 17, fill-extrusion enabled | ~900MB-1.6GB |
| Map at zoom 13, fill-extrusion enabled | ~500MB |
| Same scene zoomed out | Drops proportionally |

A 4MB GLTF model added via the plugin layer contributed negligible memory compared to the extrusion layer itself.

## Impact

- **HudHud (production app)** has had to disable 3D buildings during navigation entirely because the system kills the app in the background due to memory pressure. They also encountered z-ordering conflicts between GLTF plugin models and fill-extrusion geometry, which was a secondary reason for disabling extrusions. (Confirmed by @AtlasProgramming and Patrick Kladek in the MapLibre Slack)
- **Our app (Endure)** is building a 3D route replay feature with an animated avatar walking through city streets. We want building extrusions for context but cannot use them at street-level zoom due to memory. Currently falling back to a different SDK for this feature.
- Both apps would benefit from being able to use GLTF models and building extrusions simultaneously without memory explosion or depth conflicts.

## Root cause (high level)

The current fill-extrusion pipeline tessellates every building polygon into full 3D vertex and index buffers and stores them on the GPU. At street-level zoom with pitch, a large number of buildings are in view (especially toward the horizon), and each one carries the full cost of its tessellated mesh in memory.

## Desired outcome

- Fill-extrusion memory usage at zoom 17 on mobile should be significantly lower (ideally under 500MB total for map + extrusions)
- Buildings should be renderable alongside GLTF plugin models without z-fighting
- Bonus: distance-based LOD or culling so buildings far from the camera use less memory or are skipped entirely

## Possible approaches (for discussion)

- Reduce stored geometry per building (e.g. generate wall/roof geometry more efficiently, avoid redundant vertices)
- Frustum culling to avoid processing buildings outside the camera view
- Distance-based LOD: simplify or skip buildings beyond a configurable distance
- Tile render distance cap when pitch is applied (reducing the number of tiles loaded toward the horizon)

## Context

This came up in the MapLibre Slack (#maplibre-native) with input from @AtlasProgramming (GLTF plugin author / HudHud), Patrick Kladek (HudHud), and @sjg-wdw. Steve mentioned his team is about to work on rounded building corners for fill-extrusion, and addressing the general memory problem alongside that work could be a good opportunity.

@AtlasProgramming also noted that tile count increases significantly with pitch due to horizon visibility, and that distance-based tile culling is being explored as a mitigation.

Happy to help with testing, benchmarking, and contributing to a solution. The memory numbers above are easily reproducible and could serve as a baseline for A/B comparison.

cc @louwers

## Environment

- MapLibre Native iOS 6.23.0 (Metal renderer)
- iPhone 15 Pro (A17, 8GB RAM)
- OpenFreeMap tiles with fill-extrusion layer
- GLTF plugin layer (maplibre-gltf-models-plugin)

alexcristici:
Before instancing (39MB):
```
fillExtrusionBucketsMem: 39050310 bytes
fillExtrusionBucketsCount: 22
zoom: 14.9547
```

After instancing (7MB):
```
fillExtrusionBucketsMem: 6719670 bytes
fillExtrusionBucketsCount: 22
zoom: 14.9547
```

Memory decrease more than 5 times: from 39 MB to less than 7 MB.

At zoom level 16:

Before instancing (60MB):
```
fillExtrusionBucketsMem: 61895658 bytes
fillExtrusionBucketsCount: 34
zoom: 16
```

After instancing (10MB):
```
fillExtrusionBucketsMem: 10642026 bytes
fillExtrusionBucketsCount: 34
zoom: 16
```

Screenshot 2026-04-23 at 13 22 25


include/mbgl/shaders/mtl/fill_extrusion.hpp
```diff
@@ -146,7 +145,149 @@ FragmentStage vertex vertexMain(thread const VertexStage vertx [[stage_in]],
...
+#if defined(HAS_UNIFORM_u_height)
+    const auto height = props.height;
+#else
+    const auto height = max(unpack_mix_float(vertx.height, drawable.height_t), 0.0);
+#endif
+
+    float2 p1 = float2(outline[instanceID + 1].pos);
+    float2 p2 = float2(outline[instanceID + 0].pos);
+    float2 perp = p1 - p2;
+    float magnitute = sqrt(perp.x * perp.x + perp.y * perp.y);
```
adrian-cojocaru: Typo

include/mbgl/shaders/mtl/fill_extrusion.hpp
```diff
@@ -370,5 +507,225 @@ fragment FragmentOutput fragmentMain(FragmentStage in [[stage_in]],
...
```
adrian-cojocaru: Typo

adrian-cojocaru (approved):

louwers (approved):

louwers:

PR #4262, opened 2026-04-25
fix(vulkan): recreate staging texture in HeadlessBackend::readStillImage on size change

### LINKED ISSUE #4260: Vulkan HeadlessBackend::readStillImage caches stale Texture2D after setSize()
## Summary

In the Vulkan headless backend, `HeadlessBackend::readStillImage` lazy-initialises a staging `Texture2D` on first use and never recreates it. Calls to `HeadlessBackend::setSize` after the first render do reset the renderable `resource` (framebuffer) but leave the staging texture at its original size, so every subsequent `readStillImage` returns pixels at the first-render size regardless of the current backend size.

Affects Static / Tile render modes (anything that calls `readStillImage` more than once with a resize in between). OpenGL and Metal backends are unaffected.

## Minimal repro

Using Rust bindings for brevity, but the underlying C++ calls are `HeadlessBackend::setSize` → `HeadlessFrontend::render` → `HeadlessBackend::readStillImage`:

```rust
// Builder size 64x64
let mut renderer = ImageRendererBuilder::new()
    .with_size(NonZero::new(64).unwrap(), NonZero::new(64).unwrap())
    .build_static_renderer();
renderer.load_style_from_url(...);

// First render at 64x64 - correct
let first = renderer.render_static(0.0, 0.0, 0.0, 0.0, 0.0).unwrap();
assert_eq!(first.as_image().dimensions(), (64, 64));  // passes

// Resize, render again
renderer.set_map_size(Size::new(Width(128), Height(32)));
let second = renderer.render_static(0.0, 0.0, 0.0, 0.0, 0.0).unwrap();
assert_eq!(second.as_image().dimensions(), (128, 32));  // FAILS: actually (64, 64)
```

`set_map_size_before_first_render` works correctly (first-render state is clean). `set_map_size_between_renders` is the failing case.

## Root cause

[`platform/default/src/mbgl/vulkan/headless_backend.cpp:77-89`](https://github.com/maplibre/maplibre-native/blob/main/platform/default/src/mbgl/vulkan/headless_backend.cpp#L77-L89)

```cpp
PremultipliedImage HeadlessBackend::readStillImage() {
    auto& contextImpl = static_cast<Context&>(*context);
    auto& resourceImpl = static_cast<HeadlessRenderableResource&>(*resource);

    if (!texture) {                      // ← lazy-init, never reset on resize
        texture = std::make_unique<Texture2D>(contextImpl);
        texture->setFormat(gfx::TexturePixelType::RGBA, ...);
        texture->setSize(size);          // ← texture size frozen at first-use size
        texture->setUsage(Texture2DUsage::Read);
    }

    contextImpl.waitFrame();
    texture->copyImage(resourceImpl.getAcquiredImage());

    return std::move(*texture->readImage());
}
```

The parent `gfx::HeadlessBackend::setSize` correctly resets `resource`:

```cpp
// platform/default/src/mbgl/gfx/headless_backend.cpp
void HeadlessBackend::setSize(Size size_) {
    size = size_;
    resource.reset();
}
```

…but doesn't reset the Vulkan-specific `texture` member. The Vulkan destructor does `texture.reset()`, which shows the member is Vulkan-specific — so `gfx::HeadlessBackend::setSize` can't touch it from the shared code path.

OpenGL (`platform/default/src/mbgl/gl/headless_backend.cpp`) reads via `readFramebuffer<PremultipliedImage>(size)`, taking the current `size` each call — no caching, no bug.

Metal (`platform/default/src/mbgl/mtl/headless_backend.cpp`) delegates to `offscreenTexture->readStillImage()` where `offscreenTexture` is tied to the `resource` that `setSize` resets — no bug.

## Proposed fix

Smallest viable fix, one line inside the Vulkan `readStillImage`:

```diff
 PremultipliedImage HeadlessBackend::readStillImage() {
     auto& contextImpl = static_cast<Context&>(*context);
     auto& resourceImpl = static_cast<HeadlessRenderableResource&>(*resource);

-    if (!texture) {
+    if (!texture || texture->getSize() != size) {
         texture = std::make_unique<Texture2D>(contextImpl);
         texture->setFormat(gfx::TexturePixelType::RGBA, gfx::TextureChannelDataType::UnsignedByte);
         texture->setSize(size);
         texture->setUsage(Texture2DUsage::Read);
     }

     contextImpl.waitFrame();
     texture->copyImage(resourceImpl.getAcquiredImage());

     return std::move(*texture->readImage());
 }
```

(Alternative, tidier: override `setSize` in `mbgl::vulkan::HeadlessBackend` to also `texture.reset()`. Either works; the `readStillImage`-side fix is more local.)

## Version / commit

Reproduced against maplibre-native at commit `9b6325a14e2cf1cc29ab28c1855ad376f1ba4903` (the amalgam pinned by `maplibre-native-rs` v0.4.5). Confirmed the offending code is still present on `main`.

## Context

Discovered this while integrating a Rust render worker built on `maplibre-native-rs` into a production tileserver. First render of any given pool was correct; subsequent renders with different viewport sizes all came back at the first-render dimensions. The downstream-side workaround is to cache a renderer per `(width, height)` tuple — functional but wasteful of memory. Happy to open a PR with the fix if the approach above looks right.

Happy to provide the full Rust test binary (builds in Docker against the prebuilt amalgam, reproduces in about 1 minute) if that helps.

jfberry:
Closes #4260.

## What

`mbgl::vulkan::HeadlessBackend::readStillImage` lazy-creates a staging
`Texture2D` on first call and never recreates it. Subsequent calls to
`HeadlessBackend::setSize` reset `resource` (the framebuffer) but leave
the staging texture pinned at its original size, so any `readStillImage`
after a resize returns pixels at the **old** dimensions.

Fix: also reallocate the staging texture when the backend's `size` has
diverged from the texture's `size`. One-line change.

```diff
-    if (!texture) {
+    if (!texture || texture->getSize() != size) {
         texture = std::make_unique<Texture2D>(contextImpl);
         texture->setFormat(gfx::TexturePixelType::RGBA, gfx::TextureChannelDataType::UnsignedByte);
         texture->setSize(size);
         texture->setUsage(Texture2DUsage::Read);
     }
```

`Texture2D::getSize()` is `noexcept`; the steady-state cost on the
cache-hit path is two integer compares.

## Why this only affects Vulkan

- **OpenGL**: `readFramebuffer<PremultipliedImage>(size)` reads at the
  current `size` each call, no caching.
- **Metal**: delegates to `offscreenTexture->readStillImage()` where
  `offscreenTexture` is tied to `resource`, which the shared
  `gfx::HeadlessBackend::setSize` already resets.

## Reproduction

Surfaced through [`maplibre-native-rs`](https://github.com/maplibre/maplibre-native-rs)'s
`ImageRenderer::set_map_size` in `Static` mode: works on the first
render, silently no-ops on every subsequent render — output stays
pinned at the first-render size. See #4260 for the full debug write-up
and a Rust-side regression test.

The contract test on the C++ side could be: render at size A, call
`setSize(B)`, render again, assert the result is at B's resolution. I
haven't added that test in this PR because the test surface for
HeadlessBackend in the upstream tree wasn't immediately obvious to me —
happy to add one with a pointer to the right harness.

## Risk

The guard now recreates `texture` whenever the backend's size has
changed. Existing behaviour (no resize, single allocation) is preserved
because the `getSize() != size` branch is taken only when an actual
size change happened. The new allocation/free pattern is identical to
what the destructor already does on backend teardown, so no new
lifetime invariants.

### DIFF
--- a/platform/default/src/mbgl/vulkan/headless_backend.cpp
+++ b/platform/default/src/mbgl/vulkan/headless_backend.cpp
@@ -82,7 +82,7 @@ PremultipliedImage HeadlessBackend::readStillImage() {
     auto& contextImpl = static_cast<Context&>(*context);
     auto& resourceImpl = static_cast<HeadlessRenderableResource&>(*resource);
 
-    if (!texture) {
+    if (!texture || texture->getSize() != size) {
         texture = std::make_unique<Texture2D>(contextImpl);
         texture->setFormat(gfx::TexturePixelType::RGBA, gfx::TextureChannelDataType::UnsignedByte);
         texture->setUsage(Texture2DUsage::Read);


louwers:
As mentioned by @adrian-cojocaru this should be fixed in main already. Can you confirm that?

Also, please take a moment to read our [AI Policy](https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md). AI generated PR descriptions should not be used.

PR #4263, opened 2026-04-27
Hide vendored HarfBuzz symbols on Apple

### LINKED ISSUE #3777: MapLibre iOS interferes with Compose font rendering since addition of Harfbuzz
### MapLibre iOS Version

6.18.1

### iOS Version

18.6

### Device

iPhone 16 Pro (and simulator)

### What happened?

Since v6.18.0, adding MapLibre Native to a Compose iOS project causes text in Compose to be broken. The map itself looks fine. I'm unsure if this is a bug in MapLibre Native or in Compose, but notably:

* MLN iOS v6.18.0 is the release that introduced Harfbuzz font shaping
* Compose uses [Skia](https://github.com/google/skia) for UI rendering, which also uses Harfbuzz

So, I'm guessing this is most likely a bug in either MapLibre Native or Compose/Skia, related to the use of Harfbuzz in both projects.

Before and after MLN 6.18.0:

Before
After

I have not noticed the same bug on other platforms (Android, Linux, Windows, macOS), just on iOS.

### Steps to reproduce

Use Compose Multiplatform (iOS) alongside MapLibre Native, for example with the MapLibre Compose demo app:

* Clone github.com/maplibre/maplibre-compose
* Update maplibre ios version to 6.18.1 in gradle.properties
* Run the demo app on the iOS simulator using IntelliJ IDEA or Android Studio

### Relevant log output

```shell

```

### Additional context

https://github.com/maplibre/maplibre-compose/issues/551

sargunv:
_Created using OpenCode with GPT-5.5_

Fixes https://github.com/maplibre/maplibre-native/issues/3777.

This hides the vendored HarfBuzz symbols in Apple builds to avoid `_hb_*` symbol conflicts with app dependencies such as Compose/Skia.

Validation:
- Built `//platform/ios:MapLibre.dynamic` locally with Bazel.
- Confirmed exported `_hb_*` symbols dropped from 342 to 0 in the simulator framework.
- Confirmed the MapLibre Compose demo using the [original MapLibre-first linker order](https://github.com/maplibre/maplibre-compose/issues/551#issuecomment-4323283345) renders text correctly with the patched framework.

### DIFF
--- a/vendor/BUILD.bazel
+++ b/vendor/BUILD.bazel
@@ -722,6 +722,13 @@ cc_library(
         # "HB_NO_WIN1256",
         "HB_TINY",
     ],
+    # Keep bundled HarfBuzz symbols private so they do not interpose symbols from
+    # app dependencies. See https://github.com/maplibre/maplibre-native/issues/3777.
+    copts = select({
+        "@platforms//os:ios": ["-fvisibility=hidden"],
+        "@platforms//os:macos": ["-fvisibility=hidden"],
+        "//conditions:default": [],
+    }),
     includes = [
         "freetype/include",
         "harfbuzz/src",
--- a/vendor/harfbuzz.cmake
+++ b/vendor/harfbuzz.cmake
@@ -15,6 +15,12 @@ if (MLN_TEXT_SHAPING_HARFBUZZ)
     target_include_directories(mbgl-harfbuzz PRIVATE ${CMAKE_CURRENT_LIST_DIR}/freetype/include)
     target_compile_definitions(mbgl-harfbuzz PRIVATE -DHAVE_FREETYPE)
 
+    if(APPLE)
+        # Keep bundled HarfBuzz symbols private so they do not interpose symbols from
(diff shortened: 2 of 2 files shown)


vendor/BUILD.bazel
```diff
@@ -722,6 +722,13 @@ cc_library(
         # "HB_NO_WIN1256",
         "HB_TINY",
     ],
+    # Keep bundled HarfBuzz symbols private so they do not interpose symbols from
+    # app dependencies. See https://github.com/maplibre/maplibre-native/issues/3777.
+    copts = select({
```
louwers: Probably good idea to just make this the default here.

louwers (approved):

vendor/BUILD.bazel
sargunv: done

louwers: Sorry, turns out this is causing undefined symbol errors on Linux. Not exactly sure why.

Let's revert and merge.

sargunv: done

PR #4267, opened 2026-04-28
Expose Metal texture from headless backend

sargunv:
While working on a C ABI I found myself needing to expose this.

Filing a PR for visibility into what changes I need to make to make the C ABI work -- I'm not confident it's a correct or mergable change.

The code that uses this is https://github.com/sargunv/maplibre-native-ffi/commit/72f8b8a3a638bdc6d0ce51c0037ed29b14a99db3.

_Created with GPT-5.5 in OpenCode_


platform/default/src/mbgl/mtl/headless_backend.cpp
```diff
@@ -86,6 +91,11 @@ PremultipliedImage HeadlessBackend::readStillImage() {
     return getResource<HeadlessRenderableResource>().readStillImage();
 }
 
+MTL::Texture* HeadlessBackend::getMetalTexture() {
+    getDefaultRenderable();
```
louwers: This is a little bit smelly to me. Maybe this part

```cpp
  if (!resource) {
      resource = std::make_unique<HeadlessRenderableResource>(*this, static_cast<mtl::Context&>(getContext()), size);
  }
```

can be factored out into a `ensureResource()` method.

sargunv: done!

louwers (approved):

PR #4272, opened 2026-04-30
add possibility to link wgpu library from extern

Murmele:
When creating a Rust application using maplibre-native it is desired to share the gpu textures from the C++ side and the rust side. Therefore we need bindings to wgpu. Perhabs wgpu-native is not a C binding to wgpu but to wgpu-core and therefore not compatible with wgpu.

So the rust application must provide their own implementation of the webgpu functions exposed as C ABI compatible functions.

We include here the header only to successfully build maplibre-native but the linking is done on the "user" side.

This is used right now in https://github.com/maplibre/maplibre-native-rs/pull/143

The implementation of the takeTexture() function was done by Copilot


louwers:
Could you explain in a bit more words what this PR does and why it is needed?

The PR seems to include a lot of unrelated changes.

Murmele:
Yes for sure, Sorry this is not yet ready

src/mbgl/gfx/offscreen_texture.hpp
```diff
@@ -23,6 +23,8 @@ class OffscreenTexture : public gfx::Renderable {
     virtual PremultipliedImage readStillImage() = 0;
 
     virtual const gfx::Texture2DPtr& getTexture() = 0;
+
+    virtual gfx::Texture2DPtr takeTexture() { return getTexture(); }
```
Murmele:
```suggestion

    virtual gfx::Texture2DPtr takeTexture() = 0;
```

louwers:
Thanks for providing more details in the PR description.

Please also disclose AI usage in the PR description (if applicable) as per our AI Policy. https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md

vendor/webgpu.cmake
```diff
@@ -3,6 +3,7 @@
 
 set(MLN_WEBGPU_IMPL_DAWN OFF CACHE STRING "WebGPU with Dawn implementation")
 set(MLN_WEBGPU_IMPL_WGPU OFF CACHE STRING "WebGPU with wgpu implementation")
+set(MLN_WEBGPU_IMPL_FFI OFF CACHE STRING "WebGPU with ffi interface")
```
louwers, line 6: I like to avoid creating new options.

Is this something we could just enable always?

Murmele: Perhaps this is not possible, because in this case I wanna just include the header file, but not linking the library

platform/default/include/mbgl/gfx/headless_backend.hpp
```diff
@@ -24,6 +27,12 @@ class HeadlessBackend : public gfx::Renderable {
 
     virtual PremultipliedImage readStillImage() = 0;
     virtual RendererBackend* getRendererBackend() = 0;
+    /*!
+     * \brief takeTexture
+     * Take ownership of the current texture
+     * \return
+     */
+    virtual gfx::Texture2DPtr takeTexture() { return nullptr; }
```
louwers: Why not make this a pure virtual method?

louwers:
There is no need to merge main into this branch.

src/mbgl/style/sources/geojson_source.cpp
```diff
@@ -30,6 +30,11 @@ const GeoJSONSource::Impl& GeoJSONSource::impl() const {
     return static_cast<const Impl&>(*baseImpl);
 }
 
+/*!
+ * \brief GeoJSONSource::setURL
+ * Set the url to a geojson file which can be fetched
```
louwers, line 35: This seems to be an unrelated change.

louwers:
> The implementation of the takeTexture() function was done by Copilot

Does this mean that the rest of the PR was written by you, including the random docstrings?

platform/default/include/mbgl/gfx/headless_backend.hpp
Murmele: I did not dig into the other backends how sharing of textures is done there, so I don't wanted to implement it there as well.

I am wondering how is sharing texture done on Android, do you know?

Murmele:
> > The implementation of the takeTexture() function was done by Copilot
> 
> Does this mean that the rest of the PR was written by you, including the random docstrings?

yes :). I implemented also the GeoJSON and came across of this and then I documented it how I understood it. I can remove it

louwers:
OK, no worries, keep it. I just notice AI making a lot of 'drive by' improvements. :see_no_evil:

Murmele:
> OK, no worries, keep it. I just notice AI making a lot of 'drive by' improvements. 🙈

I can imagine haha

How do we proceed regarding the takeTexture()? See comment above

vendor/webgpu.cmake
```diff
@@ -3,6 +3,8 @@
...
+set(MLN_WEBGPU_IMPL_WEBGPU_HEADER_DIR "" CACHE STRING "If not empty use this path to the webgpu.h file")
```
Murmele, line 7: I had to add this one as well to select a different webgpu.h file than the one from wgpu-native

platform/default/include/mbgl/gfx/headless_backend.hpp
louwers: When do you want to land this?

Let me run this by the graphics engineers.

Murmele: There is no rush

Murmele:
@louwers can we merge this so I can finish the maplibre-native-rs pullrequest?

include/mbgl/gfx/renderer_backend.hpp
```diff
@@ -54,6 +58,11 @@ class RendererBackend {
     virtual void initShaders(gfx::ShaderRegistry&, const ProgramParameters&) = 0;
     const mbgl::util::SimpleIdentity uniqueID;
 
+#if MLN_WEBGPU_IMPL_FFI
+    virtual void setDevice(WGPUDevice device) = 0;
```
louwers: Please move this to the WebGPU specific render backend instead of adding an `#ifdef` here.

platform/default/include/mbgl/gfx/headless_backend.hpp
louwers: Please add this only to the WebGPU headless backend if you are only implementing it for WebGPU.

louwers (changes requested):
@Murmele Please see my comments.

If we adjust some backend agnostic interface then all implementations should be updated. Otherwise move the implementation to the WebGPU specific interface.

include/mbgl/gfx/renderer_backend.hpp
Murmele: But I need it to expose to the extern. I need to have a look if this is possible

Murmele:
> @Murmele Please see my comments.
> 
> If we adjust some backend agnostic interface then all implementations should be updated. Otherwise move the implementation to the WebGPU specific interface.

@louwers I reverted the changes in the general gfx files. So now it is only for webgpu implemented

louwers (approved):

Murmele:
Thanks @louwers

PR #4282, opened 2026-05-09
core: Render hillshade prepare pass past tile interior to fix overzoom seam

### LINKED ISSUE #4281: Hillshade has a zero-width seam at tile boundaries when overzoomed past raster-dem maxzoom
**Human note**

After working with Claude on #4279, we noticed another separate bug causing discontinuities at the edges of tiles in the DEM hill shading layer. Our DEM data tops out at zoom 12, But when zooming in further than that, we see seams between the tiles. I'm not familiar with the internals of the MapLibre codebase, but Claude seems to have worked out what's going on, and fixed the problem. This is a little bit more code than our earlier fix, so probably needs a bit more of a detailed look. The rest of this was written by Claude:

**Describe the bug**

When the hillshade layer is rendered at a zoom level higher than the raster-dem source's `maxzoom` (i.e. the source is overzoomed), every tile boundary visible on screen shows a discontinuity in shading: a sharp step between two adjacent screen pixels at the geometric tile edge, with no width. The interior of each tile is smoothly bilinear-interpolated, so the eye reads the abrupt step as a hard line tracing the underlying tile grid — vertical and horizontal — across the visible map.

**To Reproduce**

1. Configure a raster-dem source with `"maxzoom": 12`, e.g. terrarium tiles
2. Add a `hillshade` layer that reads from the source
3. Zoom the map to z14 or z15 (well past the source's maxzoom)
4. Pan around — every tile-grid line is visible as a discontinuity in the hillshade

**Expected behavior**

Adjacent tiles' hillshading should meet seamlessly at the boundary. The slope value at A's east edge and B's west edge represents the slope at the same world position, so the rendered shading should be continuous across the seam.

**Screenshots**

| Zoom | Before | After |
|-------|--------|------|
| 15       | Image |  Image |
| 16.    |  Image | Image |

**Platform information**

- Operating System: iOS 18 / Android 14 (any platform with the Metal / GL hillshade pipeline — same code path)
- Platform: MapLibre Native iOS XCFramework 6.25.1 and current `main` (commit `6aa804f`)
- Version: 6.25.1

**Additional context**

**Root cause:** the hillshade prepare pass renders to a `dim × dim` derivative texture. Each output pixel computes Sobel slope at the centre of an interior DEM cell — half a DEM-pixel inside the tile boundary on each side. Tile A's last column samples slope at `world (1 − 0.5/dim)` of A; tile B's first column samples slope at `world (0 + 0.5/dim)` of B. Those are 1 DEM-pixel apart in absolute world space, computed from different sets of source pixels, and yield different values.

When the hillshade render shader samples the derivative texture with `Clamp` wrap, it returns A's last edge texel for everything from the boundary westward and B's first edge texel for everything from the boundary eastward. At native zoom that mismatch is sub-screen-pixel and invisible; at z = source.maxzoom + 2 it spreads to 4 screen pixels per DEM pixel and the discontinuity at the boundary becomes clearly visible — a sharp step between adjacent screen pixels meeting at the tile edge.

The bug is independent of `RasterDEMTile` neighbour-backfill state and of the `hillshade-shadow-color` / `-highlight-color` paint properties. It reproduces against any non-flat terrain and any hillshade method (`STANDARD`, `BASIC`, `MULTIDIRECTIONAL`, etc.).

A draft PR with the fix is in flight.

dave:
**Human note**

After working with Claude on #4279, we noticed another separate bug causing discontinuities at the edges of tiles in the DEM hill shading layer. Our DEM data tops out at zoom 12, But when zooming in further than that, we see seams between the tiles. I'm not familiar with the internals of the MapLibre codebase, but Claude seems to have worked out what's going on, and fixed the problem. This is a little bit more code than our earlier fix, so probably needs a bit more of a detailed look. The rest of this was written by Claude:

Fixes #4281.

The hillshade prepare pass renders a `dim²` derivative texture sampled by the per-frame hillshade shader. With `dim²` outputs, each output pixel computes Sobel slope at the centre of an interior DEM cell — half a DEM-pixel inside the tile boundary on each side. Adjacent tiles compute slopes at world positions one DEM-pixel apart at the seam, and the per-tile `Clamp`-wrap sampling at the boundary returns two different edge values. At native zoom this is sub-screen-pixel; once overzoomed (display zoom > source maxzoom) it becomes a visible discontinuity — a sharp step between adjacent screen pixels — traced along every tile boundary.

This PR extends the prepare output by one pixel on the south and east — render to `(dim+1)²`. The new `i = dim` output pixel computes Sobel centred on the inner east-border cell, which after neighbour backfill holds tile B's first interior column. Tile B's `i = 0` output uses the same three source columns from the matching west-border backfill on its side. Same Sobel inputs → same output → continuous shading across the boundary.

To make `i = dim` reachable without the kernel running off the buffer, `DEMData::border` is widened from 1 to 2 pixels. `backfillBorder` copies two columns/rows of neighbour data per axis-aligned side and a 2×2 block for diagonals. The interior content and `DEMData::dim` / `get(x, y)` semantics are unchanged.

Both the GLSL and Metal `hillshade_prepare` vertex shaders pick the new boundary outputs with `v_pos = (a_texture_pos / EXTENT) · (dim+1) / stride + 2 / stride` so the corner outputs land on buffer texels 2 and `dim+2`. The fragment shader's `tileSize` derivation switches from `dimension - 2` to `dimension - 4` to match the wider stride.

## Tests

`test/geometry/dem_data.test.cpp` is updated to cover the 2-pixel border layout end-to-end:
- `ConstructorMapbox` / `ConstructorTerrarium`: stride is `dim + 2 · border`, image bytes match.
- `InitialBackfill`: every cell of the border ring is non-empty, edge columns / rows are edge-clamped two-deep, all four corner blocks are seeded from the nearest interior corner.
- `BackfillNeighbor`: each of the 8 neighbour directions populates the expected `border`-deep stripe (or `border × border` corner) with the matching opposite slice from the neighbour DEMData.

Existing render tests covering the hillshade pipeline pass unchanged.

## Cost

- DEM tile memory: stride `dim + 2 → dim + 4` ⇒ ~1.6 % more bytes per tile (~16 KB for a 256² Terrarium tile)
- Prepare-pass GPU: `(dim + 1)² / dim² ≈ 0.8 %` more fragments per prepare invocation
- Per-frame rendering: zero change

Prepare runs once per tile load and once per neighbour backfill, never per frame, so the GPU cost is unobservable in practice.

## Visual evidence

Before / after screenshots to follow on the linked issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 ### 🐞 Bug fixes
 
 - *...Add new stuff here...*
+- [core] Fix overzoom seam in hillshade rendering by extending the prepare-pass output past the tile interior and widening the DEMData neighbour border to 2 pixels ([#4282](https://github.com/maplibre/maplibre-native/pull/4282)).
 - [core] Fix memory access violation exception in vector_tile_data.cpp [#632](https://github.com/maplibre/maplibre-native/pull/632)
 - [iOS] Fix a bug where the compass was determined to be misplaced when hidden [#498](https://github.com/maplibre/maplibre-native/pull/498).
 - [core] `MaptilerFileSource` renamed to `MBTilesFileSource` [#198](https://github.com/maplibre/maplibre-native/pull/198).
--- a/include/mbgl/shaders/mtl/hillshade_prepare.hpp
+++ b/include/mbgl/shaders/mtl/hillshade_prepare.hpp
@@ -61,9 +61,14 @@ FragmentStage vertex vertexMain(thread const VertexStage vertx [[stage_in]],
 
     const float4 position = drawable.matrix * float4(float2(vertx.pos), 0, 1);
 
-    float2 epsilon = 1.0 / tileProps.dimension;
-    float scale = (tileProps.dimension.x - 2.0) / tileProps.dimension.x;
-    float2 pos = (float2(vertx.texture_pos) / 8192.0) * scale + epsilon;
+    // See shaders/hillshade_prepare.vertex.glsl for the derivation. With a
+    // 2-pixel DEM border (stride = dim+4) and a (dim+1)² render target,
+    // output pixel i (i ∈ [0, dim]) lands on buffer texel i+2 — including
+    // the inner east-border cell at i=dim. Adjacent tiles' boundary
+    // outputs reference identical source pixels and produce identical
+    // slope values, so the rendered hillshade has no seam at overzoom.
+    float scale = (tileProps.dimension.x - 3.0) / tileProps.dimension.x;
+    float2 pos = (float2(vertx.texture_pos) / 8192.0) * scale + 2.0 / tileProps.dimension;
 
     return {
         .position    = position,
@@ -87,7 +92,9 @@ half4 fragment fragmentMain(FragmentStage in [[stage_in]],
 #endif
 
     float2 epsilon = 1.0 / tileProps.dimension;
-    float tileSize = tileProps.dimension.x - 2.0;
+    // dimension is the DEM buffer stride (interior + 2 * 2-pixel border),
+    // so the actual interior tile size is stride - 4.
+    float tileSize = tileProps.dimension.x - 4.0;
 
     // queried pixels (using Sobel operator kernel):
     // +-----------+
--- a/shaders/hillshade_prepare.fragment.glsl
+++ b/shaders/hillshade_prepare.fragment.glsl
@@ -20,7 +20,9 @@ float getElevation(vec2 coord, float bias) {
 
 void main() {
     vec2 epsilon = 1.0 / u_dimension;
-    float tileSize = u_dimension.x - 2.0;
+    // u_dimension is the DEM buffer stride (interior + 2 * 2-pixel border),
+    // so the actual interior tile size is stride - 4.
+    float tileSize = u_dimension.x - 4.0;
 
     // queried pixels (using Sobel operator kernel):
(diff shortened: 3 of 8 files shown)


louwers (changes requested):
Could you add a render test? How to work with render tests is described in the developer documentation:

https://maplibre.org/maplibre-native/docs/book/platforms/android/android-tests.html#render-tests

dave:
Closing in favour of #4280, which now contains the full corrected fix for this issue plus the two other hillshade fixes from the original #4280 scope.

The original approach in this PR (DEMData border = 2, prepare-pass output `(dim+1)²`) was insufficient: the boundary texel matched across tiles, but bilinear filtering in the hillshade-color shader still blended that texel with the next-inward texel, which differed across the seam. Adding the render test you asked for surfaced the residual artifact at the 4-tile corner.

The final fix in #4280 widens the shared region further — `border = 3`, prepare output `(dim+3)²`, and an inset hillshade-color/color-relief vertex shader so the displayed tile maps to the inner texels of the wider output, keeping bilinear's partners in shared-data territory. Both directions of the 4-tile-corner seam are now eliminated, and a programmatic regression test (`test/api/hillshade_overzoom_seam.test.cpp`) is included that fails decisively without the fix.

Consolidating into #4280 since the three fixes (this overzoom seam + the existing two in #4280) share a small set of files and reviewer context.

PR #4287, opened 2026-05-11
Add asynchronous deletion option for Android renderer backend

adrian-cojocaru:
Follow up to https://github.com/maplibre/maplibre-native/pull/3681. 

Added a runtime option to let the application decide where the backend deletion will be executed. By default this is done on the render thread while the GL context is still valid and by enabling `asyncRendererCleanup` the cleanup will run when `MapRenderer` is garbage collected (the GL context has been destroyed and it's resources are no longer valid).

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMapOptions.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMapOptions.java
@@ -837,8 +837,11 @@ public class MapLibreMapOptions implements Parcelable {
   }
 
   /**
-   * Move the native renderer backend deletion from the render thread to the garbage collector,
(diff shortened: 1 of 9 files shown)


platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/MapLibreMapOptions.java
```diff
@@ -828,6 +836,19 @@ public MapLibreMapOptions actionJournalRenderingReportInterval(int actionJournal
     return this;
   }
 
+  /**
+   * Move the native renderer backend deletion from the render thread to the garbage collector,
```
louwers: Could you add a few more words when you might want to use this.

louwers (approved):

louwers:

Issue #3079, opened 2024-12-11
Allow choosing rendering backend on Android during initialization


louwers:
Please upvote this issue if you are interested in this, or leave a comment if you have additional context and implementation ideas.

---

**Is your feature request related to a problem? Please describe.**

Allow a choosing between Vulkan and OpenGL ES rendering backend on Android on map initialization.

**Describe the solution you'd like**

On map initialization we should allow passing an enum value `VULKAN`, `OPENGL` or `AUTO`.

With the `AUTO` option we would detect what rendering backend is supported and default to that.

**Describe alternatives you've considered**

Releasing MapLibre Android with different backends as different libraries. This will happen starting 11.7.0:

- `org.maplibre.gl:android-sdk` (OpenGL)
- `org.maplibre.gl:android-sdk-vulkan` (Vulkan)

The downside of this is that there is some build and distribution complexity involved when building different apks and distributing those. It is also easier for other platforms that depend on MapLibre Android, like current implementations of MapLibre for Flutter, React Native and Kotlin Multiplatform, to intergrate.

The downside of using multiple libraries in one library is that the size of the library will increase. We should give people the option to use only Vulkan or only OpenGL ES, or configure their release process to distribute multiple apks. That is why I think that starting with MapLibre Android [12.0.0](https://github.com/maplibre/maplibre-native/issues/2736) we should publish the library as three different packages.

- `org.maplibre.gl:android-sdk` (OpenGL ES + Vulkan, defaulting to automatic selection)
- `org.maplibre.gl:android-sdk-opengl` (OpenGL ES)
- `org.maplibre.gl:android-sdk-vulkan` (Vulkan)

**Additional context**

A run-time switch (i.e. when the map is already initialized) is out of scope for this issue.

vvinogra:
Hi folks, do you have any updates on this?

louwers:
@vvinogra If there's an update it will be posted here. There are other options for a gradual rollout, for example A/B testing... Even if we add a runtime switch, it will not be the default because it would double the size of the library.

We have only had a very few responses from users that tried out Vulkan unfortunately. So that can mean a few things: it is rock solid, and people who tried it did not bother to report back. Or not many people tried it because they don't know it exists or they have other priorities. Or lastly: people tried it, found something that didn't work, and still did not report back.

So the plan is to make Vulkan the default in 13.0.0. https://github.com/maplibre/maplibre-native/issues/4114

[...]

PR #4288, opened 2026-05-12
Add multiBackend Android flavor for runtime OpenGL/Vulkan switching

vvinogra:
## Summary

Fixes issue: https://github.com/maplibre/maplibre-native/issues/3079

Adds a new `multiBackend` Android product flavor that ships **both** rendering
backends in a single AAR:
- `libmaplibre.so` (Vulkan)
- `libmaplibre-opengl.so` (OpenGL)

`multiBackend` flavor supports a runtime switch between 2 rendering engines (App Restart is required)

Use the new `org.maplibre.gl:android-sdk-vulkan-opengl` dependency for runtime switch maplibre version.

### Change list
-  **Runtime engine selection:** Added a new `RenderingEngine` API. Rendering engine is chosen during the first call to `MapLibre.getInstance` and locked for the process lifetime.
PS: This is required since MapLibre loads the correct `.so` library when `MapLibre.getInstance` is called.
- **Added RendererStrategy to prevent code duplication:** To make as minimal changes as possible in the existing codebase, I've moved shared code for Vulkan and OpenGL rendering to the `sharedRenderer` folder. Respective `sharedRenderer` folder is added in the gradle configuration depending on what build supports.
- **Added a new CMake for dual builds:** Created a new `MultiBackend.cmake` which creates 2 `.so` builds. One for vulkan, second one for OpenGL. `MultiBackend.cmake` build file was mostly generated with Claude Code, but was verified and tested.
- **Test app integration:** Implemented a new menu action to view and switch between rendering engines. If flavor supports only a single map rendering engine, the startup falls back to the default engine.
- **Total `multiBackend` flavor release library size:** ~32Mb

## Video

https://github.com/user-attachments/assets/05789925-6c00-4525-97bf-56526da3bda5

### Future recommendations:
1. Add a multi module structure to Gradle. Right now, both rendering engines are added to build in the `sourceSets`. This approach is not scalable when more logic is added. Some ideas for the next steps:
    1. Move everything from `src` to the new `core` module (Or try to only move `maps/renderer` code into a separate module)
    2. Refactor `sharedRenderer` folder into the new modules
    3. Add an `api` module to define interfaces for the renderer
    4. Update `main` package to use different modules based on the chosen flavor
2. Update `‎platform/android/buildSrc/src/main/kotlin/maplibre.download-vulkan-validation.gradle.kts‎` script to only download Vulkan-ValidationLayers library when `debug` build type is chosen
3. Rework `LibraryLoaderProvider` api to make sure that it's not customizable. I found only a single public repository in github which has a custom `LibraryLoaderProvider` implementation. I didn't want to update this API since it could be used by someone. If someone has a custom `LibraryLoaderProvider` implementation, then runtime rendering engine switch may not work unless it's handled on the client side.
4. Move `LibraryLoader.load()` away from the static code block to have a lifecycle for library initialization. Right now, `LibraryLoader.load()` is called in static code block in multiple places but all these places require  `MapLibre#getInstance()` initialization. This is not safe since static code block is executed when JVM loads the class into memory. There's a chance that `LibraryLoader.load()` could be called before `MapLibre#getInstance` initialization in the future.
5. Add integration tests for the `multiBackend` flavor to CI


vvinogra:
@coderabbitai review

vvinogra:
@louwers @adrian-cojocaru Hey folks, could you please take a look?

louwers:
Yes, I will try it out.

Could you please add some details how you developed this?

I assume you used LLMs? Please have a look at our AI Policy: https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md We also appreciate human written PR descriptions. At the very least indicate which parts you wrote and which parts are written by AI.

Also, thanks for working on this. A lot of people requested this! If it works well we'll put out multi backend releases as part of our regular releases.

adrian-cojocaru (approved):

louwers (changes requested):
Please see the comment above.

Could you add an entry to the [documentation](https://maplibre.org/maplibre-native/android/examples/)? Please explain the APIs that the multi backend flavor exposes as well as the default behavior.

How do we want to call this version? Right now we have:

`org.maplibre.gl:android-sdk` Vulkan since 13.0.0
`org.maplibre.gl:android-sdk-opengl` OpenGL ES
`org.maplibre.gl:android-sdk-vulkan` Vulkan

I think 'multibackend' may be ambigious, because theoretically we could have a 'mutli backend' with for example Vulkan and WebGPU.

I was thinking, maybe:

`org.maplibre.gl:android-sdk-vulkan-opengl-compat`

Since it uses Vulkan by default, but switches to OpenGL for compatibility? Or just:

`org.maplibre.gl:android-sdk-vulkan-opengl`

Lastly, if you can update the release workflow I will put out a release as soon as this is merged. Otherwise I can do that as a follow up. Thanks!

vvinogra:
@louwers Thanks a lot! I have updated PR description.
I have also added `Future recommendations` sections based on my findings that I think could be improved but are out of scope for this PR.

> Could you add an entry to the [documentation](https://maplibre.org/maplibre-native/android/examples/)? Please explain the APIs that the multi backend flavor exposes as well as the default behavior.

I have updated documentation with the new flavor details and examples.

> How do we want to call this version?
I think 'multibackend' may be ambigious, because theoretically we could have a 'mutli backend' with for example Vulkan and WebGPU.

Good point! `org.maplibre.gl:android-sdk-vulkan-opengl-compat` could possibly sounds that it's a separate(non-related) compatibility library, like `androidx.appcompat`, but it's another build variant.

I like the `org.maplibre.gl:android-sdk-vulkan-opengl` naming! I have updated release workflow and Github actions

.github/labeler.yml
```diff
@@ -64,3 +64,8 @@ WebGPU:
     - '**/*webgpu*/**'
     - '**/*wgpu*'
     - '**/*dawn*'
+MultiBackend:
```
louwers: This label can be removed.

docs/mdbook/src/platforms/android/android-rendering-engine-api.md
```diff
@@ -0,0 +1,27 @@
...
+
+| | OpenGL ES | Vulkan |
+|---|---|---|
+| Device compatibility | Any device (widest support) | Requires Vulkan 1.0 hardware (`FEATURE_VULKAN_HARDWARE_VERSION`) and API 24+ |
+| Performance | Baseline | Better on modern devices |
+| Stability | Battle-tested, longest track record | Newer, may have edge cases |
+
+The `multiBackend` flavor can detect whether the device supports Vulkan and fall back to OpenGL ES if it does not. Accordingly, `multiBackend` manifest declares the Vulkan feature as optional, therefore making it available in Play Store listings on devices without Vulkan support.
+
+## Selecting a backend
```
louwers: I would like to see this documentation in the end user documentation here. https://maplibre.org/maplibre-native/android/examples/ End users are not compiling the library themselves.

The mdbook is intended for MapLibre Android developers.

Could you mention the name of the Maven Central package you decided on?

louwers: Here you can see documentation on how to build the documentation. https://maplibre.org/maplibre-native/docs/book/platforms/android/android-documentation.html

louwers (changes requested):
Sounds good!

Please have a look at my comments concerning the documentation.

docs/mdbook/src/platforms/android/android-rendering-engine-api.md
vvinogra: Sure! Updated

.github/labeler.yml
vvinogra: Gotcha, updated

platform/android/docs/data/rendering-engine.md
```diff
@@ -0,0 +1,74 @@
+# Rendering Engine (OpenGL ES / Vulkan)
```
louwers, line 1: I would keep this. But remove the two `mdbooks` pages.

louwers (approved):

platform/android/docs/data/rendering-engine.md
vvinogra: Updated

vvinogra: Agree, this mdbooks looks quite similar to the Android dev documentation. Better to have a single version

PR #3714, opened 2025-08-12
Update PMTiles documentation with caching and offline limitations

scottrhoyt:
As mentioned in #3690, this updates the PMTiles documentation to note the limitations with caching and offline pack downloads.

PR #4290, opened 2026-05-12
core: ambient cache for pmtile sources

### LINKED ISSUE #3690: Allow Caching Requests for PMTiles Sources
**Describe the bug**

PMTiles sources do not appear to be cached in a way that is usable when offline.

**To Reproduce**

1. Load a `MLNMapView` with PMTiles sources with internet connectivity
2. Scroll around an area to download the relevant tiles
3. Force close the app
4. Disable internet connectivity (e.g. via airplane mode)
5. Reopen app
6. Scroll around are previously visited with connectivity
7. Observe that the previously downloaded tiles are not loaded

**Expected behavior**

The previously downloaded tiles are able to be used offline (verified non-PMTiles sources work this way).

**Screenshots**

n/a

 - Operating System: iOS 18.5
 - Platform (e.g. Node.js, Qt): MapLibre Native iOS
 - Version: 6.18.0

**Additional context**

n/a

twopastmidnight:
### AI Usage Disclaimer
The changes in this PR were created by Claude Code (Opus 4.7, Extra High effort). The architectural implementation was created by a human (me).

---

### Description
This PR enables ambient caching for PMTile sources with as little disruption to other parts of the library as possible, and additionally the main goal being to avoid changing the SQLite schema.

To do this, the byte range of the HTTP requests is appended on the offline storage URL as a mock parameter, creating a synthetic unique key. Before beginning work on this feature, I verified that the current Maplibre Native implementation does not perform multi-tile/clustered requests.

#### offline_database.cpp
- A `cacheKey()` helper is added that appends a `_mlnRange=` URL parameter if the queried resource has a data range. If it doesn't, it simply returns `resource.url` as before. Sites that previously used `resource.url` are now routed through this helper instead.

#### pmtiles_file_source.cpp
- Changed LoadingMethod to All across the board.
- Added early `return`s in `getHeader()` and `getMetadata()` cache checks. This seems to be an oversight with the original implementation since it didn't have caching anyway.
- Added a few guards.

### Testing
- I have verified that bazel compiles on macOS targeting iOS. I was not able to check Android due to the long build times.
- I have verified that the new feature works on an iOS simulator via the helper logs.
- I have included a couple of new tests to validate to cover the added logic.

Resolves #3690

### DIFF
--- a/platform/default/src/mbgl/storage/offline_database.cpp
+++ b/platform/default/src/mbgl/storage/offline_database.cpp
@@ -13,18 +13,16 @@
 
 namespace mbgl {
 
-namespace {
-// PMTiles and other range-aware sources issue multiple sub-requests against
-// the same URL with different byte ranges. The `resources` table has a UNIQUE
-// constraint on `url`, so we encode the range into the cache key so each range
-// gets its own row instead of overwriting the previous one.
-std::string cacheKey(const Resource& resource) {
+// Single-point transformer for offline storage keys. For sources where the URL is
+// not enough to guarantee uniqueness (eg. PMTiles where all requests resolve to
+// the same URL), this function creates a synthetic key based on the requested
+// data range. Can be expanded in the future to accommodate other formats.
+std::string OfflineDatabase::cacheKey(const Resource& resource) {
     if (!resource.dataRange) return resource.url;
     const auto sep = resource.url.find('?') == std::string::npos ? '?' : '&';
     return resource.url + sep + "_mlnRange=" + std::to_string(resource.dataRange->first) + "-" +
            std::to_string(resource.dataRange->second);
 }
-} // namespace
 
 OfflineDatabase::OfflineDatabase(std::string path_, const TileServerOptions& options)
     : path(std::move(path_)),
@@ -391,10 +389,6 @@ std::optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const 
     query.bind(1, cacheKey(resource));
 
     if (!query.run()) {
-        if (resource.dataRange) {
-            // TODO: Remove after PR review
-            Log::Info(Event::Database, "[ambient cache] miss: " + cacheKey(resource));
-        }
         return std::nullopt;
     }
 
@@ -415,12 +409,6 @@ std::optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const 
     } else {
         response.data = std::make_shared<std::string>(*data);
         size = data->length();
-    }
-
-    if (resource.dataRange) {
-        // TODO: Remove after PR review
-        Log::Info(Event::Database,
-                  "[ambient cache] hit: " + cacheKey(resource) + " (" + std::to_string(size) + " bytes)");
     }
 
     return std::make_pair(response, size);
@@ -457,10 +445,6 @@ bool OfflineDatabase::putResource(const Resource& resource,
         notModifiedQuery.bind(3, response.mustRevalidate);
         notModifiedQuery.bind(4, cacheKey(resource));
         notModifiedQuery.run();
-        if (resource.dataRange) {
-            // TODO: Remove after PR review
-            Log::Info(Event::Database, "[ambient cache] revalidated (304): " + cacheKey(resource));
-        }
         return false;
     }
 
@@ -497,12 +481,6 @@ bool OfflineDatabase::putResource(const Resource& resource,
 
     updateQuery.run();
     if (updateQuery.changes() != 0) {
-        if (resource.dataRange) {
-            // TODO: Remove after PR review
-            Log::Info(
-                Event::Database,
-                "[ambient cache] refresh: " + cacheKey(resource) + " (" + std::to_string(data.size()) + " bytes)");
-        }
         return false;
     }
 
@@ -529,12 +507,6 @@ bool OfflineDatabase::putResource(const Resource& resource,
     }
 
     insertQuery.run();
-
-    if (resource.dataRange) {
-        // TODO: Remove after PR review
-        Log::Info(Event::Database,
-                  "[ambient cache] insert: " + cacheKey(resource) + " (" + std::to_string(data.size()) + " bytes)");
-    }
 
     return true;
 }
--- a/platform/default/src/mbgl/storage/offline_database.cpp
+++ b/platform/default/src/mbgl/storage/offline_database.cpp
@@ -13,6 +13,19 @@
 
 namespace mbgl {
 
+namespace {
+// PMTiles and other range-aware sources issue multiple sub-requests against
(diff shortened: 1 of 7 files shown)


twopastmidnight:
I think the new unit tests are failing because of the temporary cache hit/miss logs firing

platform/default/src/mbgl/storage/offline_database.cpp
```diff
@@ -13,6 +13,19 @@
 
 namespace mbgl {
 
+namespace {
+// PMTiles and other range-aware sources issue multiple sub-requests against
+// the same URL with different byte ranges. The `resources` table has a UNIQUE
+// constraint on `url`, so we encode the range into the cache key so each range
+// gets its own row instead of overwriting the previous one.
+std::string cacheKey(const Resource& resource) {
```
louwers: This can be unit tested.

louwers (changes requested):
I think this is a good approach because we can use it to incorporate other headers into the cache key as well.

However some parts are very chatty (probably due to being LLM generated). Please have a look over those extensive comments and (re)write them in your own words.

Also, the unit tests are failing...

test/storage/offline_database.test.cpp
```diff
@@ -396,6 +396,70 @@ TEST(OfflineDatabase, PutTile) {
...
 }
 
+// Resources that differ only by their byte range — e.g. PMTiles header vs. directory vs.
+// tile sub-requests against the same .pmtiles URL — must each occupy their own cache row
+// instead of overwriting one another. See https://github.com/maplibre/maplibre-native/issues/3690.
+TEST(OfflineDatabase, PutGetResourceWithDataRange) {
+    FixtureLog log;
+    OfflineDatabase db(":memory:", fixture::tileServerOptions);
+
+    Resource r1{Resource::Kind::Source, "https://example.com/x.pmtiles"};
```
louwers: Maybe a simpler test to add would be to check expected cache keys.

twopastmidnight:
Made the following changes:
- Removed logging
- Cleanup unnecessary comments
- Align the empty data check Copilot mentioned with the rest
- Removed the previous two offline database tests and added a simpler new one (had to expose `cacheKey`)

@louwers can we rerun the CI?

louwers:
We updated the documentation in #3714, I think this limitation is now removed.

louwers:
@scottrhoyt @jj10133 If you're still interested in this feature could you help test this PR?

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
```diff
@@ -229,14 +235,40 @@ class PMTilesFileSource::Impl {
         return fileSource;
     }
 
+    // Cache-first byte-range fetch that skips conditional revalidation on hit, since
+    // PMTiles archives are immutable per upload (validated once by `getHeader()`)
```
louwers: @twopastmidnight Could you explain this comment?

What is "conditional revalidation on hit"?

What does "immutable per upload" mean?

Also why does this code exist here because there does not seem to be anything PMTiles-specific about it?

twopastmidnight: PMTiles files are immutable in the sense that they can not be written into once built, if you want to make changes to the tiles you need to rebuild the entire file. This means we don't need to worry about individual byte-range freshness as long as we know the file itself has not changed (which is checked in `getHeader`). A bit of a sidenote, but when it does change, all the byte offsets will change too, which is why I didn't factor in an automated cache eviction (which would be computationally costly since we'd have to do a wildcard lookup).

What "conditional revalidation on hit" refers to is the cache -> network flow when using `LoadingMethod::All`. I noticed this behavior while testing, where network requests would go out to revalidate cache freshness of the tile even if it's already cached. These are meaningless requests in the context of PMTiles as the archive does not change, and when it does, the old byte ranges are no longer valid anyway.

The reason why this helper is here is because I didn't want to touch too many surfaces with this PR. I think the proper way to handle it would be a new `LoadingMethod`, something like `LoadingMethod::CacheOrNetwork`, but I'd say this would be an out-of-scope change in this PR. If there was a method that did Cache -> Network without firing a cache revalidation request, then this helper would not need to exist in the first place.

twopastmidnight: I just realized I wrote that comment through the lens of having a full planet archive hosted on a CDN, but it doesn't read right if you're embedding a small-area file locally. I'll reword this a bit later along with updating the docs.

twopastmidnight:
@louwers The caching logic in this PR can cause unexpected behavior depending on implementation. I stumbled across https://github.com/maplibre/maplibre-native/issues/3658, which is now defensively guarded as of this PR but can still result in de-synchronisation between cache and the remote file.

When the CDN .pmtiles file is updated while a user has the map open, the new byte ranges will not match because we're now caching the archive header. In the original implementation, every tile request ALSO hits the header for every single tile, which does prevent de-synchronisation from happening but is extremely wasteful.

In the implementation of this PR, updated remote archives will only start working after an app restart. https://github.com/maplibre/maplibre-native/pull/4290/changes/b8daa083dec4e60c25e3f862cdf6a94eb1eeec4d changes `emplace` to `insert_or_assign`, otherwise it would take 2 app restarts before the correct header & metadata go into effect.

I'm not sure how prevalent this edge case is to be honest, but I think it's relatively easy to fix by tracking the ETag on every tile request in `cacheFirstThenNetwork`. If the ETag changes on any tile request, we just clear the `header_cache` & `directory_cache` and the new archive should work as normal. We could perhaps do a release with the current changes, and apply ETag checking in `cacheFirstThenNetwork` if reports come in?

twopastmidnight:
I figured it would be better to fix the issue with CDN archive changes during active user sessions, rather than potentially introducing regressions for library consumers. In https://github.com/maplibre/maplibre-native/pull/4290/changes/150ad27129091db0d06a3fb39786bc3a78045679 we now keep track of the ETag and invalidate caches if a tile request receives one different from what we have stored.

@louwers would it be possible to create an rc release so that others can test this without having to build from source?

platform/default/src/mbgl/storage/pmtiles_file_source.cpp
louwers: This is architecturally unsound. You need to update `bool DatabaseFileSource::canRequest(const Resource& resource)` so that it can store PMTiles files.

Please have a look at `main_resource_loader.cpp`.

twopastmidnight:
Reverted the helper-based implementation in `pmtiles_file_source.cpp` and changed LoadingMethod to All across the board.

twopastmidnight:
Hey @louwers, are we good to merge?

louwers (approved):

louwers:
I think this could use some more testing, but PMTiles sources are still a bit experimental so I am OK with merging this.

Issue #4296, opened 2026-05-15
hillshade-shadow-color and hillshade-highlight-color reject spec-valid interpolate expressions since #3965


dts:
### MapLibre iOS Version

6.24.0 through 6.26.0 (current main, including commit bccfcdd)

### iOS Version

iOS 26.4.2 (also reproduces on 26.4.1 simulator)

### Device

iPad Air 11-inch (M2) — also reproduces on iPhone 17 Pro Simulator (iOS 26.4.1)

### What happened?

Some hillshade layers do not appear as expected on iOS (but work unmodified in web), specifically the one that's part of LINZ's vector style.

### Steps to reproduce

Got a minimal reproduction at [gitlab.com/dts1/maplibre-native-hillshade-repro 
](https://gitlab.com/dts1/maplibre-native-hillshade-repro).

### Relevant log output

```shell
[4]: Expected array<color> but found string instead.
```

### Additional context

So grateful for everybody's work on this library, it's really come a long way in the last few months!! PR that fixes this is here: https://github.com/maplibre/maplibre-native/pull/4297

louwers:
Thanks for creating this issue and suggesting a fix.

Please take a moment to review our AI Policy. https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md

We discourage contributors from using LLMs in communication.

dts:
Read through it - makes sense (and is a good policy).

PR #4297, opened 2026-05-15
fix: accept single-color expressions for colorArray properties

dts:
This is a fix for issue https://github.com/maplibre/maplibre-native/issues/4296. There's an overly restrictive validation step that is preventing animated hillshade layers from rendering (e.g. animating opacity during a zoom between levels).

AI Disclosure: I used AI to create the minimal reproduction & the fix here, I am confident it fixes the underlying issue (I've tested it on my personal device).

Repro of bug: https://gitlab.com/dts1/maplibre-native-hillshade-repro

### DIFF
--- a/src/mbgl/style/conversion/property_value.cpp
+++ b/src/mbgl/style/conversion/property_value.cpp
@@ -3,6 +3,10 @@
 #include <mbgl/style/conversion/rotation.hpp>
 #include <mbgl/style/conversion_impl.hpp>
 #include <mbgl/style/conversion.hpp>
+#include <mbgl/util/color.hpp>
+
+#include <type_traits>
+#include <vector>
 
 namespace mbgl {
 namespace style {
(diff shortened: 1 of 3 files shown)


louwers:
Please make sure to follow our AI Policy https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md

dts:
Shoot, I didn't look at that before, thanks for the heads-up - how would you like me to proceed? Rewrite the PR description and include an AI disclosure? I'm pretty confident in the solution, but it is AI-generated. I validated it as best I could (testing on physical devices, simulators, etc).

louwers:
@dts No worries! It's pretty new and only mentioned in the contributor guidelines.

1. If you can, just write the problem this solves in your own words.
2. Briefly mention your process of fixing it and how you used AI.
3. Take a critical look at the code and especially the comments. Is the solution appropriate? Does the test actually test the problem that was fixed? Is there any needless verbosity or is there a simpler fix?

Then, once you're confident. Mark the PR as ready for review. Then, a maintainer will have a look. And here be prepared to answer questions and/or follow up requests.

Does that sound reasonable?

louwers:
Hey, if this is ready for review, please mark it as such. Thanks!

PR #4302, opened 2026-05-20
core: clear vulkan allocator handle after destruction

sargunv:
`RendererBackend::destroyResources()` destroys the VMA allocator but leaves the member pointing at the released allocator object. This isn't currently a problem as `destroyResources` is only called in the destructor, but if it ever gets called in another path it risks a use-after-free bug without this `= nullptr` patch.

Previously I thought I needed this PR to fix a bug in MLN-FFI, but I was chasing a red herring. Feel free to close or merge this minor defensive cleanup.

### DIFF
--- a/src/mbgl/vulkan/renderer_backend.cpp
+++ b/src/mbgl/vulkan/renderer_backend.cpp
@@ -638,6 +638,7 @@ void RendererBackend::destroyResources() {
     commandPool.reset();
 
     vmaDestroyAllocator(allocator);
+    allocator = nullptr;
 
     usingSharedContext ? void(device.release()) : device.reset();
 


adrian-cojocaru:
`vmaDestroyAllocator` accepts `nullptr` as a valid parameter and handles it as no-op.
```
vmaDestroyAllocator(VmaAllocator VMA_NULLABLE allocator) {
  if(allocator != VK_NULL_HANDLE)
  {
      ...
  }
}
```
Setting `allocator = nullptr` fixes a double free issue, but this is called only in the destructor and shouldn't be possible.

TimSylvester (approved):

sargunv:
> vmaDestroyAllocator accepts nullptr as a valid parameter and handles it as no-op.

I'll take a look! Perhaps I made a mistake in my debugging and chased a red herring

sargunv:
Yup, I was indeed chasing a red herring. 

I've scoped this down to just the `= nullptr` to mitigate the theoretical use-after-free if `destroyResources` ever gets called outside the destructor, but no longer need it for mln-ffi

The PR is now just a minor defensive cleanup. feel free to close or merge

adrian-cojocaru (approved):

PR #4316, opened 2026-05-31
Fix mlt and icu not included in the macOS amalgamation

ciscorn:
`mbgl-core` on Darwin links `mlt-cpp` and `mbgl-vendor-icu`, but neither was merged into the amalgamation on macOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 ### 🐞 Bug fixes
 
 - *...Add new stuff here...*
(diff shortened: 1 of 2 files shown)


louwers:
Thanks! Does this solve your problem?

louwers (approved):

ciscorn:
Yes. maplibre-native-rs uses the amalgamation when `MLN_PRECOMPILE=1` is set, but linking fails on macOS because these symbols are missing.

louwers:
Do you need a new release?

ciscorn:
@louwers No rush on my side. Thanks for asking!

PR #4315, opened 2026-05-31
core: Add HarmonyOS platform

sargunv:
Resolves #3749 

## Summary

This branch adds support for [OpenHarmony](https://en.wikipedia.org/wiki/OpenHarmony) (OHOS) and [HarmonyOS](https://en.wikipedia.org/wiki/HarmonyOS) (HMOS). HMOS is Huawei's commercial OS competing with Android and iOS in China. OHOS is an open source project, making up the core of HMOS and other projects like [Oniro](https://oniroproject.org).

I've only tested this on HarmonyOS, but since I've stuck to OHOS SDK only, it _should_ work similarly on Oniro too.

### Platform

I used existing `platform/default` sources for most components like storage, text, i18n. The OHOS-specific components include:

- `logging_hilog.cpp` maps MapLibre logging to OHOS hilog
- `image.cpp`: decodes platform images through OHOS image APIs
- `http_file_source.cpp`: implements MapLibre HTTP requests through OHOS `libnet_http.so`. 

### Core

I also needed to make a few changes to the core. I carved these out into #4317, #4318, #4319, #4320; see those PRs for details.

## Validation

The branch also includes a sample app, bridging a minimal set of APIs to ArkTS and using them in a live interactive map with the MapLibre demo style, OFM Bright, and OFM Liberty.

The sample app is validated on real hardware, a Huawei MatePad Pro MRDI-W00 running HarmonyOS 6.1.0.117, with both Vulkan and OpenGL. The emulator [doesn't yet support Vulkan](https://developer.huawei.com/consumer/en/doc/harmonyos-guides/ide-emulator-specification#section168125802410)—an upcoming beta release (that I couldn't figure out how to acquire) will but works with OpenGL after the stencil clip-mask workaround mentioned above.

I've also configured CI, using the OHOS SDK from https://github.com/openharmony-rs/ohos-sdk/releases. I'm only building the MapLibre Native library right now in CI, not running any tests, nor building the sample app.

I didn't run iOS/Android/other builds or tests locally, as I assume they're covered in CI.

## Process

I used GPT-5.5 in Codex CLI to put together this PR.

- First, a /goal loop for initial exploration and proof of concept build, commit d8297e949d0a7a7e53ecdc17c34536c86b450639
- Then, real-time pairing with the agent to squash bugs and fill in gaps, roughly up to 95426f7152c48090794c128757343550a1215256
[...]

PR #4317, opened 2026-06-01
core: Tolerate stencil clipping setup failures

sargunv:
To support https://github.com/maplibre/maplibre-native/pull/4315, I needed to work around a limitation in the HarmonyOS emulator's OpenGL implementation.

The HarmonyOS emulator’s OpenGL path breaks under MapLibre’s stencil-based tile clipping.

Here we validate stencil clip-mask setup by checking the required position attribute before enabling stencil clipping. When validation fails, it renders the tile layer group unclipped.


platform/linux/src/gl_functions.cpp
```diff
@@ -3,7 +3,6 @@
 #define GL_GLEXT_PROTOTYPES
 
 #include <GLES3/gl3.h>
-#include <GLES3/gl3ext.h>
```
sargunv, line 6: bonus: openharmony's build environment was missing this header, and it seems unused to me, so I removed it rather than making it conditional

src/mbgl/gl/attribute.hpp
```diff
@@ -67,12 +69,19 @@ class AttributeLocations<TypeList<As...>> final {
 
     AttributeBindingArray toBindingArray(const gfx::AttributeBindings<TypeList<As...>>& bindings) const {
         AttributeBindingArray result;
-        result.resize(sizeof...(As));
+
+        auto reserveLocation = [&](const std::optional<AttributeLocation>& location) {
```
louwers: This seems to be an unrelated change?

src/mbgl/gl/context.hpp
```diff
@@ -79,6 +79,7 @@ class Context final : public gfx::Context {
     void setColorMode(const gfx::ColorMode&);
     void setCullFaceMode(const gfx::CullFaceMode&);
     void setScissorTest(const gfx::ScissorRect&);
+    bool hasStencilBuffer();
```
louwers: Can be marked `const`.

src/mbgl/renderer/paint_parameters.hpp
```diff
@@ -97,7 +97,11 @@ class PaintParameters {
 
     // Stencil handling
 public:
-    void renderTileClippingMasks(const RenderTiles&);
+    /// Whether GL stencil clipping can be used for non-3D drawables in the current render target.
+    bool stencilClippingAvailable = true;
+
+    /// @return False when the clipping mask could not be rendered and stencil clipping should be disabled.
```
louwers: I think a better comment should be just "returns true on success" or similar.

Whether or not it should be disabled is not the responsibility of this class.

src/mbgl/renderer/paint_parameters.hpp
```diff
@@ -97,7 +97,11 @@ class PaintParameters {
 
     // Stencil handling
 public:
-    void renderTileClippingMasks(const RenderTiles&);
+    /// Whether GL stencil clipping can be used for non-3D drawables in the current render target.
+    bool stencilClippingAvailable = true;
```
louwers, line 109: Why is this stored? And why here?

louwers:
See comments, please re-request a review once ready.

src/mbgl/gl/attribute.hpp
sargunv: Ah yeah, this fix snuck in bc it was adjacent to the stencil clipping fallback.

It's still a valid bug fix, as I don't think GL guarantees positions are adjacent and so `result.resize(sizeof...(As));` was valid. Relevant docs: https://docs.gl/gl3/glBindAttribLocation (though it doesn't explicitly say that can be sparse).
 
I don't remember exactly why I ran into this while working on OHOS, and my notes don't justify it; happy to carve this out too if you'd like.

src/mbgl/renderer/paint_parameters.hpp
sargunv: the decision about clipping is made in `DrawableGL::makeStencilMode()` and `PaintParameters` is the context passed into it, so it seemed like a natural place to me, but yeah it's not really a style `paint` parameter and should probably live higher up somewhere (I'm unsure where)

Do you have an alternative suggestion?

louwers: Not entirely sure either...

louwers (approved):

louwers:
Feel free to merge.

PR #4318, opened 2026-06-01
core: Add source location fallback

sargunv:
To support https://github.com/maplibre/maplibre-native/pull/4315, I added a fallback source location helper.

OpenHarmony's build environment lacks the `<source_location>` header. We previously had "A temporary shim for partial C++20 support" that patched the struct into the `std` namespace ([which is undefined behavior)](https://timsong-cpp.github.io/cppwp/n3337/namespace.std), but even that wasn't sufficient as the header itself was included unconditionally.

Here I replace that shim with a custom `mbgl::SourceLocation`, which aliases to `std::source_location` when available and falls back to file/function/line otherwise.

### DIFF
--- a/src/mbgl/util/source_location.hpp
+++ b/src/mbgl/util/source_location.hpp
@@ -11,12 +11,12 @@
 namespace mbgl {
 
 #if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
-using SourceLocation = std::source_location;
-#define MBGL_CURRENT_SOURCE_LOCATION ::mbgl::SourceLocation::current()
+using source_location = std::source_location;
+#define MLN_CURRENT_SOURCE_LOCATION ::mbgl::source_location::current()
 #else
-class SourceLocation final {
+class source_location final {
 public:
-    constexpr SourceLocation(const char* fileName_ = "",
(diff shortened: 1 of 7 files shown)


adrian-cojocaru (approved):

src/mbgl/util/source_location.hpp
```diff
@@ -0,0 +1,42 @@
...
+#if __has_include(<source_location>)
+#include <source_location>
+#endif
+#endif
+
+namespace mbgl {
+
+#if defined(__cpp_lib_source_location) && __cpp_lib_source_location >= 201907L
+using SourceLocation = std::source_location;
+#define MBGL_CURRENT_SOURCE_LOCATION ::mbgl::SourceLocation::current()
```
louwers: Plese use `MLN` prefix.

src/mbgl/util/source_location.hpp
```diff
@@ -0,0 +1,42 @@
...
+#else
+class SourceLocation final {
```
louwers: Let's use `source_location` so it's clear that it follows `std::source_location` exactly.

louwers:
Maybe we can get rid of it completely, let me ask.

louwers (approved):

PR #4319, opened 2026-06-01
core: Update Vulkan headers

sargunv:
To support https://github.com/maplibre/maplibre-native/pull/4315, I updated the Vulkan-Headers submodule to the latest v1.4.x patch to get the official VK_OHOS_surface declarations.

This required dealing with some deprecations, as we were previously on v1.3.x. My understanding is this upgrade doesn't have significant runtime or compatibility implications on our other targets, but I'm not at all an expert here—please check me.

### DIFF
--- a/include/mbgl/vulkan/renderer_backend.hpp
+++ b/include/mbgl/vulkan/renderer_backend.hpp
@@ -31,6 +31,11 @@ class ProgramParameters;
 
 namespace vulkan {
 
+using DynamicLoader = vk::detail::DynamicLoader;
+using DispatchLoaderDynamic = VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE;
+template <typename OwnerType>
+using ObjectDestroy = vk::detail::ObjectDestroy<OwnerType, DispatchLoaderDynamic>;
+
 class RendererBackend : public gfx::RendererBackend {
 public:
     RendererBackend(gfx::ContextMode);
@@ -40,7 +45,7 @@ public:
     void initShaders(gfx::ShaderRegistry&, const ProgramParameters& programParameters) override;
     void init();
 
-    const vk::DispatchLoaderDynamic& getDispatcher() const { return dispatcher; }
+    const DispatchLoaderDynamic& getDispatcher() const { return dispatcher; }
     const vk::UniqueInstance& getInstance() const { return instance; }
     const vk::PhysicalDevice& getPhysicalDevice() const { return physicalDevice; }
     const vk::UniqueDevice& getDevice() const { return device; }
@@ -97,8 +102,8 @@ protected:
     void destroyResources();
 
 protected:
-    vk::DynamicLoader dynamicLoader;
-    vk::DispatchLoaderDynamic dispatcher;
+    DynamicLoader dynamicLoader;
+    DispatchLoaderDynamic dispatcher;
 
     vk::UniqueInstance instance;
     vk::UniqueDebugUtilsMessengerEXT debugUtilsCallback;
--- a/src/mbgl/vulkan/renderer_backend.cpp
+++ b/src/mbgl/vulkan/renderer_backend.cpp
@@ -271,26 +271,26 @@ void RendererBackend::triggerFrameCapture([[maybe_unused]] uint32_t frameCount, 
 
 #ifdef ENABLE_VULKAN_VALIDATION
 
-static VKAPI_ATTR VkBool32 VKAPI_CALL vkDebugUtilsCallback(VkDebugUtilsMessageSeverityFlagBitsEXT messageSeverity,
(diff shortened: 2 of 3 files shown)


include/mbgl/vulkan/renderer_backend.hpp
```diff
@@ -31,6 +31,11 @@ class ProgramParameters;
 
 namespace vulkan {
 
+using DynamicLoader = vk::detail::DynamicLoader;
+using DispatchLoaderDynamic = VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE;
+template <typename OwnerType>
+using ObjectDestroy = vk::detail::ObjectDestroy<OwnerType, DispatchLoaderDynamic>;
+
```
sargunv, line 38: flag for reviewers, I don't really understand what's going on here, why we'd alias `vk::detail` or use a template, but gpt seemed pretty confident this is the pattern we want

adrian-cojocaru: The only benefit that I can see, is that it can resolve `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE` without having to expand it in every usage (similar to how it's used in `vulkan.hpp`).

src/mbgl/vulkan/renderer_backend.cpp
```diff
@@ -601,9 +600,6 @@ void RendererBackend::initDevice() {
                           .setPEnabledExtensionNames(extensions)
                           .setPEnabledFeatures(&physicalDeviceFeatures);
 
-    // this is not needed for newer implementations
-    createInfo.setPEnabledLayerNames(layers);
```
adrian-cojocaru, line 605: Was this flagged by the validation layers/AI? It's mostly a compatibility fix for older implementation and is a no-op on everything else.

adrian-cojocaru (approved):

src/mbgl/vulkan/renderer_backend.cpp
sargunv: Here's the actual source from https://raw.githubusercontent.com/KhronosGroup/Vulkan-Headers/refs/heads/main/include/vulkan/vulkan_structs.hpp

```cpp
#  if !defined( VULKAN_HPP_DISABLE_ENHANCED_MODE )
    VULKAN_HPP_DEPRECATED( "unused" )
    DeviceCreateInfo & setPEnabledLayerNames( ArrayProxyNoTemporaries<char const * const> const & pEnabledLayerNames_ ) VULKAN_HPP_NOEXCEPT
    {
      detail::ignore( pEnabledLayerNames_ );
      return *this;
    }
#  endif /*VULKAN_HPP_DISABLE_ENHANCED_MODE*/
```

and `detail::ignore` is:

```cpp
  template <typename T>
  VULKAN_HPP_CONSTEXPR bool ignore(T const&) VULKAN_HPP_NOEXCEPT {
      return true;
  }
```

So my read is it's literally implemented as a no-op in the header unless `VULKAN_HPP_DISABLE_ENHANCED_MODE` is defined.

(I'm not sure what `VULKAN_HPP_DISABLE_ENHANCED_MODE` is, but I see no evidence of us defining it)

Vulkan deprecation notes: https://docs.vulkan.org/guide/latest/deprecated.html

> **Fallback Strategy**
> 
> Since device layers were deprecated very early, there’s no need for a fallback strategy. All Vulkan implementations should support instance layers instead of device layers.

sargunv: so as far as I can tell:

- vulkan recommends not having a fallback here
- maybe upgrading the header is a breaking change for some "early in vulkan's life" implementation, not included in Vulkan's "all vulkan implementations" claim
- I have no idea how to identify what sort of implementations that'd be (a git blame on this line didn't surface anything interesting) but it seems it'd be sometime in the first few months of Vulkan 1.0 (~mid-2016 vulkan loader)

sargunv: AI tells me, after searching the web:
- Early Android 7.0/7.1 devices (Vulkan was introduced in API 24 / 7.0)
  - This one is notable since MapLibre currently supports down to API 23 / Android 6.0
- Windows systems frozen on very early 2016 Vulkan beta/initial drivers
- Linux desktop distros frozen since 2016

sargunv: AI also tells me this is used only in validation/debug/etc layers, but I'm too out of my depth to validate that claim

adrian-cojocaru: Didn't notice that the C++ headers disabled them by default. I was under the impression that they're still passed to the driver to decide if it wants to ignore them or not. Safe to remove then.

PR #4329, opened 2026-06-05
replace node-pre-gyp-github with gh release

acalcutt:
This PR attempts to fix the publishing issue found when publishing the last node release where multiple releases were made. This removes the custom node-pre-gyp-github, which I have been maintaining, with standard gh release commands

Testing: I have tested this release workflow on my test branch https://github.com/WifiDB/maplibre-native/tree/test-node-gh-release and created the test release https://github.com/WifiDB/maplibre-native/releases/tag/node-v6.5.0-pre.0, which seemed to work as expected

AI Use: I used Claude AI here and asked it to remove node-pre-gyp-github which I have been maintaining. I also asked it to replace the version actions with the version we are pulling in the first job

### DIFF
--- a/.github/workflows/node-release.yml
+++ b/.github/workflows/node-release.yml
@@ -279,7 +279,7 @@ jobs:
 
   publish_npm:
     runs-on: ubuntu-latest
-    needs: publish_binaries
+    needs: [publish_binaries, release-check]
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/node-release.yml
```diff
@@ -34,14 +34,57 @@ jobs:
...
+          if [[ "$version" == *-* ]]; then
+            echo "prerelease=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "prerelease=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Extract changelog for version
+        working-directory: platform/node
+        run: |
+          awk -v ver="## ${{ needs.release-check.outputs.version }}" 'BEGIN{p=0} $0==ver{p=1; next} /^## / && p==1{exit} p==1{print}' CHANGELOG.md > changelog_for_version.md
```
louwers, line 70: Consider factoring this into an external script for easier testing.

louwers (approved):

.github/workflows/node-release.yml
acalcutt: This part was just moved from the publish section to the create release sections, so i would rather not change it right now since I know it works. but i would say maybe in a future PR maybe if it is worth making an external script

Issue #2730, opened 2024-08-16
String expressions operate on bytes instead of characters


1ec5:
String expression operators operate on bytes rather than characters, with unexpected results for any input that isn’t pure ASCII. The [style specification](https://maplibre.org/maplibre-style-spec/expressions/#length) unhelpfully only mentions lengths and indices but doesn’t define them in terms of anything. Even so, a typical developer is very likely to assume it counts the number of characters. A byte count would be the least expected value.

## Example

Create a new iOS project with the following view controller code:

```swift
import UIKit
import MapLibre

// Set this constant to any non-ASCII string.
let name = "ñ"

class ViewController: UIViewController, MLNMapViewDelegate {
	var mapView: MLNMapView!

	override func viewDidLoad() {
		super.viewDidLoad()
		
		mapView = MLNMapView(frame: view.bounds)
		mapView.delegate = self
		mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
		mapView.styleURL = MLNStyle.predefinedStyle("Streets")?.url
		view.addSubview(mapView)
	}
	
	func mapView(_ mapView: MLNMapView, didFinishLoading style: MLNStyle) {
		let feature = MLNPointFeature()
		feature.coordinate = CLLocationCoordinate2D(latitude: 0, longitude: 0)
		feature.attributes = [
			"name": name,
		]
		let source = MLNShapeSource(identifier: "test", shape: feature)
		style.addSource(source)
		
		let lengthLayer = MLNSymbolStyleLayer(identifier: "length", source: source)
		lengthLayer.text = NSExpression(mglJSONObject: ["to-string", ["length", ["get", "name"]]])
		lengthLayer.textFontNames = NSExpression(forConstantValue: ["Open Sans Semibold"])
		lengthLayer.textFontSize = NSExpression(forConstantValue: 30)
		lengthLayer.textOffset = NSExpression(forConstantValue: CGPoint(x: 0, y: 0))
		style.addLayer(lengthLayer)
	}
}
```

## Expected and actual behavior

The label should indicate the length of the `name` constant in characters:

`name` | UTF-8 | Expected label | Actual label
----|----|:--:|:--:
`A` | 41 | 1 | 1
`ñ` | C3 B1 | 1 | 2
`丐` | E4 B8 90 | 1 | 3
`𦨭` | F0 A6 A8 AD | 1 | 4
[...]

PR #4344, opened 2026-06-15
Make string expressions operate on unicode

graysail:
As described in #2730, string expressions currently operate on bytes instead of (unicode) characters. This yields surprising results: E.g. a "slice" expression on a utf8-compliant strings will corrupt the utf8 encoding in case the offsets used for the slice expression don't align with the byte offsets of the character boundaries of the input string; in that case, the "slice" expression will cut the character at the boundary within its multi-byte sequence and thus transform a valid utf8 string into a corrupted string.

Scope:
- Expressions "length", "slice" and "index-of" (for these, the maplibre style spec explicitly mentions "UTF-16 surrogate pair[s]", i.e. unicode code points instead of bytes as relevant unit)

Regarding utf8 validation:
- AFAICS, it is assumed throughout the codebase that strings are utf-8 encoded, so we lean on that assumption here as well.
- Note that the "slice" operator has the capability of modifying a string; if a non-utf8 / garbled string were to enter the system, the slice operator would allow extracting a substring of that which starts with a utf8 leading byte and ends just before an utf8 leading byte (for comparison, the current mainline allows extraction of arbitrary byte sub-sequences from a string)

Further notes:
- As defined in the maplibre style specification (and beyond the scope of #2730 ; though operating on graphemes is mentioned as an idea there), grapheme clusters such as  "f̲" (LATIN SMALL LETTER F (U+0066) +  COMBINING LOW LINE (U+0332)) are still treated as divisible units; hence the "slice" operator can in this example be used to remove the combining diacritic (_) from the base character (f).

AI assistance:
- Github copilot (used for generating the test "TEST(Utf8OpHelpers, unicodeLengthOnValidatedUtf8)")
- Claude:claude-opus-4.6 for code review

The program was tested solely for our own use cases, which might differ from yours.
Stephan Schmid <stephan.s.schmid@mercedes-benz.com> , Mercedes-Benz AG (https://group.mercedes-benz.com/provider/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -733,6 +733,8 @@ list(APPEND SRC_FILES
     ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/parsing_context.cpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/slice.cpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/step.cpp
+    ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/utf8_op_helpers.cpp
+    ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/utf8_op_helpers.hpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/util.cpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/util.hpp
     ${PROJECT_SOURCE_DIR}/src/mbgl/style/expression/value.cpp
--- a/bazel/core.bzl
+++ b/bazel/core.bzl
@@ -391,6 +391,8 @@ MLN_CORE_SOURCE = [
     "src/mbgl/style/expression/parsing_context.cpp",
     "src/mbgl/style/expression/slice.cpp",
     "src/mbgl/style/expression/step.cpp",
+    "src/mbgl/style/expression/utf8_op_helpers.cpp",
+    "src/mbgl/style/expression/utf8_op_helpers.hpp",
     "src/mbgl/style/expression/util.cpp",
     "src/mbgl/style/expression/util.hpp",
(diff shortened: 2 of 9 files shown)


graysail:
Digging into the logs of 'Android Instrumentation Tests (Vulkan)', I can see that the test on 'Google Pixel 7 Pro' succeeded while the test on 'Google Pixel 7a' somehow crashed ("INSTRUMENTATION_RESULT: shortMsg=Process crashed").
Could this be some kind of transient error?

louwers:
@graysail Could you merge main into this branch? Then this issue should be resolved.

graysail:
> @graysail Could you merge main into this branch? Then this issue should be resolved.

@louwers  I just rebased on main

TimSylvester:
We have a dependency on ICU, shouldn't we use that for low-level operations like counting code points?

louwers:
> We have a dependency on ICU, shouldn't we use that for low-level operations like counting code points?

Not on all platforms, but iOS/macOS probably has an implementation for that too.

graysail:
> We have a dependency on ICU, shouldn't we use that for low-level operations like counting code points?

I did have a look into ICU when I implemented this. I rejected it that because I could not really find a nice way to do "get the byte offset for a code point offset" in ICU which is needed for the "slice" operator; The closest I could find was the U8_GET() macro in utf8.h, but this extracts a code point for a given offset instead of returning its byte offset. Besides, ICU is fairly macro-heavy and I struggle to really wrap my head around its failure behavior.

graysail:
CI issue on mainline is fixed AFAICS

-> rebased on main and squashed the 3rd patch (resolving minor ci findings) into the 1st commit.

louwers (approved):
I think this is simple enough that I would rather include this little bit of code rather than use a heavy dependency. ICU is not set up for all platforms, and having different implementations per platform seems overkill.

PR #2955, opened 2024-10-22
Android surface pre-rotation

adrian-cojocaru:
https://developer.android.com/games/optimize/vulkan-prerotation

PR #4348, opened 2026-06-16
feat(android): Vulkan graphics backend support for Android custom layers

Vladyslav-Odobesku-Intellias:
## Summary

- Implements `CustomLayerHost` with Vulkan backend: shader compilation, pipeline creation, and per-frame rendering via command buffer injection.
- Introduces `CustomLayerInitParameters` hierarchy to pass `vk::Device`, `vk::PhysicalDevice`, and `vk::DispatchLoaderDynamic` to `initialize()` in case of Vulkan graphics backend.
- Extracts `CustomLayerHost` into a standalone header (no `layer.hpp` dependency) and ships it via Prefab in the Android AAR so external consumers can derive from it. This one is to simplify shipping custom layer related headers as a prefab, without splitting those files transitive includes would require to ship quite many other maplibre-native headers.

Thing is, I have no idea if people are really utilizing custom layers in the scope of Android. And if some projects do then how exactly do they use them. Team I'm working work was simply copying the CustomLayerHost definition from maplibre-native into their project (that consumes maplibre-android) and then derived from it, implemented OpenGL custom layer, feed it into the Android Custom Layer by giving the pointer to native object and it works for them so far.

The last commit is solely about improving the developer experience for consumers of the library — it eliminates the need to manually copy-paste headers by bundling them directly in the AAR via Prefab.

## How to verify

I've made a sample project with the help of AI: https://github.com/Vladyslav-Odobesku-Intellias/maplibre-vulkan-custom-layer-prefab-consumer

What it does is that it consumes the built MapLibre Android Vulkan AAR library from maven local cache and then builds a simple example map view, adds custom Vulkan layer that doesn't render anything but logs init/render methods.

1. Build and publish the AAR to local Maven:
```bash
cd platform/android
./gradlew :MapLibreAndroid:publishVulkandebugPublicationToMavenLocal \
    -Pmaplibre.abis=arm64-v8a -Pmapbox.stl=c++_shared
```

2. Clone https://github.com/Vladyslav-Odobesku-Intellias/maplibre-vulkan-custom-layer-prefab-consumer
3. Open in Android Studio and run on an arm64 device.


include/mbgl/style/layers/vulkan/custom_layer_render_parameters.hpp
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Vulkan subclass of CustomLayerRenderParameters
+ */
+struct CustomLayerRenderParameters : mbgl::style::CustomLayerRenderParameters {
+    /// Dynamic dispatcher.
+    const vk::DispatchLoaderDynamic& dispatcher;
+    /// Vulkan device
+    vk::Device device;
+    /// Compatible render pass handle — use when creating VkPipeline.
+    vk::RenderPass renderPass;
```
adrian-cojocaru, line 21: Maplibre implements [pre-rotation](https://developer.android.com/games/optimize/vulkan-prerotation) and since this exposes and uses the main `renderPass` it leads to every custom layer having to implement it in their shaders similar to https://github.com/maplibre/maplibre-native/pull/2955.

`CustomLayerRenderParameters` needs to export [SurfaceRenderableResource::getRotation()](https://github.com/maplibre/maplibre-native/blob/main/include/mbgl/vulkan/renderable_resource.hpp#L72) or to disable the feature when custom layers are enabled.

Vladyslav-Odobesku-Intellias: Done, I've propagated the value of `getRotation` into that struct.
Also, added the example code in the docs on how that value should be used.

adrian-cojocaru (approved):

louwers (approved):

PR #4350, opened 2026-06-18
fix(macos): preserve exception RTTI in amalgam archive

ciscorn:
Follow-up to #4316.

While trying to use the macOS amalgam from maplibre-native-rs, I found that some exceptions thrown from the core fail to be caught by downstream code.

### Mechanism

`armerge --keep-symbols 'mln.*'` localizes every non-matching symbol, including the libc++ RTTI symbols (`__ZTI*` / `__ZTS*`) emitted for standard exceptions. On macOS, localized RTTI can prevent exception type matching across the archive boundary, causing the `catch` to be skipped.

### Solution

This PR runs `llvm-objcopy` after `armerge` to restore global visibility for those RTTI symbols.

### AI assistance

GPT-5.6 Sol and Claude Opus 4.7 were used to investigate the symbol export coverage and review the fix.


platform/macos/macos.cmake
```diff
@@ -170,6 +183,14 @@ if(MLN_CREATE_AMALGAMATION)
             $<TARGET_FILE:mbgl-vendor-icu>
             $<TARGET_FILE:mlt-cpp>
             ${STATIC_LIBS}
+        # armerge localizes non-kept symbols, including libc++ RTTI emitted for
+        # standard exceptions. On macOS, localized RTTI can prevent exceptions
+        # from being caught downstream, so restore global symbol visibility.
+        COMMAND ${LLVM_OBJCOPY} --wildcard
+            --globalize-symbol=__ZTISt*
```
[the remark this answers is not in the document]
ciscorn: I could not reproduce this with Homebrew LLVM 22.1.6.

louwers (commented):
Sorry, what is `__ZTISt`? I supposed it's a magnled name but can we deduplicate it and add a comment?

ciscorn:
@louwers  I updated the code comment and combined the three patterns into `__ZT[ISV]St*`.

(It's a Mach-O symbol using Itanium C++ ABI mangling. `ZT[ISV]` identify typeinfo structures, typeinfo names, and vtables, while `St` denotes `std::`)

---

Separately, I noticed that `--keep-symbols 'mbgl.*'` needs to be updated to `'mln.*'` following the namespace rename. I've addressed that in a separate PR.

- #4531

platform/macos/macos.cmake
```diff
@@ -144,12 +144,25 @@ add_test(
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
 
 find_program(ARMERGE NAMES armerge)
+find_program(
+    LLVM_OBJCOPY
+    NAMES llvm-objcopy
+    HINTS
+        /opt/homebrew/opt/llvm/bin
+        /opt/homebrew/opt/llvm@18/bin
```
louwers: I don't want to hardcode LLVM versions because this will get out of date.

ciscorn: Good catch. I've removed the version-specific lines.

louwers (approved):

PR #4352, opened 2026-06-18
fix(android): Add missing benchmark app module

adrian-cojocaru:
Android benchmark runs are failing due to `NoClassDefFoundError`.

```
Process: org.maplibre.android.testapp, PID: 31952
java.lang.NoClassDefFoundError: Failed resolution of: Landroidx/tracing/Trace;
```


adrian-cojocaru:
<del>This fixes the issue, but it's hitting the 150 minutes hard limit from AWS device farm.
We might need to reduce the number of locations or iterations from the benchmark runs. @louwers <del>

If a style fails to load, the benchmark runs won't advance further, leading to a device farm timeout (fixed).
`AWS Open Data Standard Light` is no longer accessible and is returning `User is not authorized to access this resource with an explicit deny in an identity-based policy`.

adrian-cojocaru:
!benchmark android

louwers:
You fixed it!

louwers (approved):

PR #4354, opened 2026-06-18
doc: fix mdbook links

michaelkirk:
Many of them were due to https://github.com/rust-lang/mdBook/issues/984 

Most of the rest seem due to re-organzing the docs into a platforms/ subdir


louwers (approved):

louwers:
Thanks.

PR #4355, opened 2026-06-18
core: Update MLT to latest

michaelkirk:
Presumably this is gated on a tagged release of mlt/cpp or something? I'm not sure if there's more to the mlt cpp release process.

I'm particularly interested in including https://github.com/maplibre/maplibre-tile-spec/commit/f162a222 due to a decode error I'm hitting.

My CPP chops are pretty weak, so do scrutinize.

### DIFF
--- a/src/mbgl/tile/vector_mlt_tile_data.cpp
+++ b/src/mbgl/tile/vector_mlt_tile_data.cpp
@@ -78,7 +78,10 @@ const PropertyMap& VectorMLTTileFeature::getProperties() const {
 }
 
 FeatureIdentifier VectorMLTTileFeature::getID() const {
-    return feature.getID();
+    if (const auto id = feature.getID()) {
+        return *id;
+    }
+    return NullValue{};
 }
 
 namespace {
@@ -191,7 +194,12 @@ std::unique_ptr<GeometryTileLayer> VectorMLTTileData::getLayer(const std::string
     if (data && !tile) {
         try {
             mlt::DataView tileData{data->data(), data->size()};
-            tile = std::make_shared<MapLibreTile>(mlt::Decoder().decode(tileData));
+#if defined(MLT_WITH_FASTPFOR) && MLT_WITH_FASTPFOR
+            constexpr bool supportFastPFOR = true;
+#else
+            constexpr bool supportFastPFOR = false;
+#endif
+            tile = std::make_shared<MapLibreTile>(mlt::Decoder(supportFastPFOR).decode(tileData));
         } catch (const std::exception& ex) {
             Log::Warning(Event::ParseTile, "MLT parse failed: " + std::string(ex.what()));
         }


src/mbgl/tile/vector_mlt_tile_data.cpp
```diff
@@ -78,7 +78,10 @@ const PropertyMap& VectorMLTTileFeature::getProperties() const {
 }
 
 FeatureIdentifier VectorMLTTileFeature::getID() const {
-    return feature.getID();
+    if (const auto id = feature.getID()) {
```
michaelkirk, line 81: Adapts to https://github.com/maplibre/maplibre-tile-spec/commit/8e957fb9 which made `feature.getID()` return an optional.

src/mbgl/tile/vector_mlt_tile_data.cpp
```diff
@@ -191,7 +194,12 @@ std::unique_ptr<GeometryTileLayer> VectorMLTTileData::getLayer(const std::string
     if (data && !tile) {
         try {
             mlt::DataView tileData{data->data(), data->size()};
-            tile = std::make_shared<MapLibreTile>(mlt::Decoder().decode(tileData));
+#if defined(MLT_WITH_FASTPFOR) && MLT_WITH_FASTPFOR
+            constexpr bool supportFastPFOR = true;
+#else
+            constexpr bool supportFastPFOR = false;
+#endif
+            tile = std::make_shared<MapLibreTile>(mlt::Decoder(supportFastPFOR).decode(tileData));
```
michaelkirk, line 202: Adapts to https://github.com/maplibre/maplibre-tile-spec/commit/3370876f213f133b1d9511f6e86628e1d15d7139, which replaced the default constructor with a one parameterized by `supportFastPFOR`

src/mbgl/tile/vector_mlt_tile_data.cpp
CommanderStorm: Would this be more ergonomic?
At least in rust, I would use `.unwrap_or_else(..)`

https://en.cppreference.com/cpp/utility/optional/or_else

Timmylad:
FYI: your changes will likely address only a part of the overall mlt encoding in MLN.

see @TimSylvester pr trying to address the same issue here - https://github.com/maplibre/maplibre-native/pull/4146/files

michaelkirk:
> FYI: your changes will likely address only a part of the overall mlt encoding in MLN.
> 
> see @TimSylvester pr trying to address the same issue here - https://github.com/maplibre/maplibre-native/pull/4146/files

I wasn't intending to fix fastpfor with this PR, I was interested in incorporating https://github.com/maplibre/maplibre-tile-spec/pull/1239

But maybe you're trying to tell me I can't have one without the other? I'm pretty out of the loop with mlt+cpp and MLN.

src/mbgl/tile/vector_mlt_tile_data.cpp
michaelkirk: We don't have `or_else` because we're using ` "-std=c++20"` (`or_else` was introduced in c++23)

Luckily, we have `value_or` (introduced c++17), which is a better fit anyway since we don't need a callable.

...but we can't easily use that.


```
// fails
return feature.getID().value_or(NullValue{});
```

 Because `feature.getID()`, is an `optional<uint64_t>`, `value_or` expects another uint64_t, but NullValue is not a uint64_t.

I think maybe we could do something like:
```
 return feature.getID()
          .transform([](Feature::id_t id) { return FeatureIdentifier(id); })
          .value_or(NullValue{});
```

But that's not really much better IMO, and `transform` is c++23 anyway.

This is as clean as I know how to get it.

Timmylad:
> > FYI: your changes will likely address only a part of the overall mlt encoding in MLN.
> > see @TimSylvester pr trying to address the same issue here - https://github.com/maplibre/maplibre-native/pull/4146/files
> 
> I wasn't intending to fix fastpfor with this PR, I was interested in incorporating [maplibre/maplibre-tile-spec#1239](https://github.com/maplibre/maplibre-tile-spec/pull/1239)
> 
> But maybe you're trying to tell me I can't have one without the other? I'm pretty out of the loop with mlt+cpp and MLN.

Yeah, I believe Tim’s PR is also attempting to update the MLT encoding in MLN. So, it’s possible that you can’t have one without the other.

louwers:
Thanks for the initiative. This has been fixed with https://github.com/maplibre/maplibre-native/pull/4146

PR #4366, opened 2026-06-27
Android: setOverrideSynchronousUpdate for existing sources

systemed:
#4092 replaced `setGeoJsonSync` on Android with a new `.withSynchronousUpdate(true)` option when creating a new GeoJSONSource.

However, this removed the possibility to use synchronous updates for existing GeoJSON sources defined in the stylesheet. Previously we could write:

```kotlin
val turnSource = map.style?.getSource("turn_highlights") as GeoJsonSource? ?: return
turnSource.setGeoJsonSync(FeatureCollection.fromFeatures(lines))
```

which is no longer possible.

This PR reinstates this possibility by giving the user the option to override the synchronous updates setting for a GeoJSON source. With this, we can write:

```kotlin
val turnSource = map.style?.getSource("turn_highlights") as GeoJsonSource? ?: return
turnSource.setOverrideSynchronousUpdate(true)
turnSource.setGeoJson(FeatureCollection.fromFeatures(lines))
```

This fixes #4187 using the approach I described in the comment and subsequently commented on by @louwers. I've chosen to fix it at the code level, rather than the stylesheet level, because this is a matter of client implementation rather than map style - in other words, it's very possible that one client using a given stylesheet might want synchronous updates, and another might not. 

I've tested it locally using a special build of cycle.travel's app and it works well for me.

AI disclosure: I wrote the code and text myself, but did ask Claude to verify if I'd missed anything along the way, because it's better at C++ than I am. ;)


louwers (approved):
LGTM, @alexcristici maybe you can also have a look.

sjg-wdw:
@alexcristici can you take another look?

alexcristici (approved):

PR #3890, opened 2025-10-16
Fixes wrong map view dimension on CarPlay.

StephanPartzsch:
This was caused by enabling the smart zoom setting on iOS26.

This fixes https://github.com/maplibre/maplibre-native/issues/3851

PR #4373, opened 2026-07-03
Fix blurry map view in landscape on iPad (due to an orientation-varying scale factor)

### LINKED ISSUE #4349: Landscape orientation leads to blurry/pixelated rendering
### MapLibre iOS Version

6.25.1

### iOS Version

18.x

### Device

iPad mini 6th gen

### What happened?

I tested the app I'm building on my iPad in landscape orientation and found that the rendering was far from crisp:

Image

Image

It should be crisp like :

iPhone in portrait:
Image

after a temporary fix:
Image

Everything was fine for the same app launched in MacOS, even in "landscape sized" windows.

### Steps to reproduce

1. Run any MapLibre-based iOS app on an iPad in landscape orientation.
2. Observe that map tiles, text labels, and icons appear blurry compared to portrait.

Confirm with the following diagnostic (swizzle layoutSubviews on MLNMapView):
```
screen.nativeBounds.w     : 1488
screen.bounds.w           : 1133
nativeBounds/bounds ratio : 1.313  ← this is what MapLibre uses as scale factor
MTKView.contentScaleFactor: 1.31
screen.nativeScale        : 2.0    ← correct value, never used

```

**Expected behavior**

MapLibre should render at the device's native pixel density (2.0× on iPad mini 6th gen) regardless of orientation. The Metal drawable should be 2266 × 1488 px for a 1133 × 744 pt logical view.

**Actual behavior**

MapLibre renders at 1.313× in landscape. The Metal drawable is 1488 × 977 px. All content appears blurry.

## Root cause

MLNEffectiveScaleFactorForView computes the pixel ratio as:
```objc
// platform/darwin/src/MLNDisplayUtils.mm
CGFloat logicalToPixelRatio = screen.nativeBounds.size.width / screen.bounds.size.width;
```
UIScreen.nativeBounds is always returned in portrait orientation, regardless of the actual device orientation. This is documented Apple behavior. In landscape on iPad mini 6th gen:

|Property	|Value|	Meaning|
| --- | --- | --- |
|screen.nativeBounds.width|	1488|	Portrait short side (physical height)|
|screen.bounds.width|	1133|	Landscape long side (logical width)|
|Computed ratio|	1488 / 1133 = 1.313|	❌ Cross-axis division — meaningless|
|screen.nativeScale|	2.0|	✅ Always correct, orientation-independent|

The wrong scale factor (1.313) is then used in MLNMapViewMetalImpl::layoutChanged() (platform/ios/src/MLNMapView+Metal.mm) to size the Metal drawable:

```objc
void MLNMapViewMetalImpl::layoutChanged() {
    const auto scaleFactor = MLNEffectiveScaleFactorForView(mapView); // = 1.313 (WRONG)
    const auto screenSize = mapView.bounds.size;                       // = {1133, 744}

    resource.mtlView.contentScaleFactor = scaleFactor;                 // = 1.313
    resource.mtlView.drawableSize = CGSizeMake(
        screenSize.width  * scaleFactor,   // 1133 × 1.313 = 1488 px  (should be 2266)
        screenSize.height * scaleFactor    //  744 × 1.313 =  977 px  (should be 1488)
    );

    size = { mtlView.drawableSize.width,   // cached: 1488  (should be 2266)
             mtlView.drawableSize.height }; // cached:  977  (should be 1488)
}
```
The Metal drawable is 1488 × 977 pixels instead of 2266 × 1488. Since framebufferSize() reads mtlView.drawableSize live at render time, and Renderable::size is cached from the same wrong value, both the render viewport and the geometry pipeline operate on wrong dimensions.

Why portrait / iPhone is unaffected: in portrait, both nativeBounds.width and bounds.width refer to the same physical dimension (short side), so the ratio is correct (e.g. 3.0 on iPhone 14 Pro).

Why it is not obviously wrong: the fallback screen.nativeScale in MLNEffectiveScaleFactorForView is never reached because logicalToPixelRatio > 0 is always true, even when the ratio is wrong.

## Fix

In platform/darwin/src/MLNDisplayUtils.mm, replace the orientation-sensitive ratio with screen.nativeScale, which is always correct:
```objc
// BEFORE (buggy):
CGFloat logicalToPixelRatio = screen.nativeBounds.size.width / screen.bounds.size.width;
if (logicalToPixelRatio > 0.0) {
    return logicalToPixelRatio;
}
return [screen respondsToSelector:@selector(nativeScale)] ? screen.nativeScale : screen.scale;

// AFTER (fixed):
CGFloat logicalToPixelRatio = screen.nativeScale;  // orientation-independent, always correct
if (logicalToPixelRatio > 0.0) {
    return logicalToPixelRatio;
}
return screen.scale;
```
Alternatively, to preserve the original intent of computing a ratio from native dimensions while being orientation-safe, use the minimum dimension of each (short side / short side):
```objc
CGFloat nativeShort  = MIN(screen.nativeBounds.size.width, screen.nativeBounds.size.height);
CGFloat logicalShort = MIN(screen.bounds.size.width, screen.bounds.size.height);
CGFloat logicalToPixelRatio = nativeShort / logicalShort; // 1488/744 = 2.0 ✓
```

## Temporary workaround (waiting for a fix and release)

Add the following Objective-C file to your iOS app target. It uses ObjC runtime swizzling to intercept MapLibre's Metal setup and substitute the correct dimensions. No MapLibre source access or fork is required — it works purely at runtime.
```objc
// MLNMapViewScalePatch.m

// Temporary workaround for maplibre-native bug:
// MLNEffectiveScaleFactorForView returns wrong scale in landscape on iPad.
// See: https://github.com/maplibre/maplibre-native/issues/XXXX

#import <Foundation/Foundation.h>
#import <UIKit/UIKit.h>
#import <MetalKit/MetalKit.h>
#import <objc/runtime.h>

@interface NSObject (MLNMapViewScalePatch)
- (void)mlnPatch_layoutSubviews;
@end

@implementation NSObject (MLNMapViewScalePatch)

+ (void)load {
    // (A) Intercept MTKView.setDrawableSize: so that MLNMapViewMetalImpl::layoutChanged()
    //     caches the correct Renderable::size (it reads drawableSize right after writing it).
    Class mtkClass = NSClassFromString(@"MTKView");
    if (mtkClass) {
        SEL   sel = NSSelectorFromString(@"setDrawableSize:");
        Method m  = class_getInstanceMethod(mtkClass, sel);
        if (m) {
            IMP origIMP = method_getImplementation(m);
            IMP patchIMP = imp_implementationWithBlock(^(UIView *self_, CGSize proposed) {
                if (self_.window &&
                    [NSStringFromClass([self_.superview class]) isEqualToString:@"MLNMapView"]) {
                    CGFloat scale   = self_.window.screen.nativeScale;
                    CGSize  b       = self_.superview.bounds.size;
                    CGSize  correct = CGSizeMake(round(b.width  * scale),
                                                 round(b.height * scale));
                    ((void (*)(id, SEL, CGSize))origIMP)(self_, sel, correct);
                } else {
                    ((void (*)(id, SEL, CGSize))origIMP)(self_, sel, proposed);
                }
            });
            method_setImplementation(m, patchIMP);
        }
    }

    // (B) Swizzle MLNMapView.layoutSubviews to fix contentScaleFactor after the original
    //     runs (which sets the wrong value from MLNEffectiveScaleFactorForView).
    Class mlnClass = NSClassFromString(@"MLNMapView");
    if (mlnClass) {
        Method orig  = class_getInstanceMethod(mlnClass, @selector(layoutSubviews));
        Method patch = class_getInstanceMethod([self class], @selector(mlnPatch_layoutSubviews));
        if (orig && patch) method_exchangeImplementations(orig, patch);
    }
}

- (void)mlnPatch_layoutSubviews {
    [self mlnPatch_layoutSubviews]; // call original

    UIView *mapView = (UIView *)self;
    if (!mapView.window) return;

    CGFloat correctScale = mapView.window.screen.nativeScale;
    for (UIView *sub in mapView.subviews) {
        if (![sub isKindOfClass:[MTKView class]]) continue;
        if (sub.contentScaleFactor != correctScale) {
            sub.contentScaleFactor = correctScale;
        }
        break;
    }
}

@end
```

**How to use:**

Add the file to your Xcode project's iOS app target (not the framework target).
No imports or explicit calls needed — +load fires automatically at app startup before MapLibre initializes.
If you use Xcode 16's PBXFileSystemSynchronizedRootGroup, dropping the file into your app folder is sufficient; no .pbxproj editing required.
The patch references MapLibre classes by name (NSClassFromString) and uses no MapLibre headers, so it compiles regardless of how MapLibre is linked (SPM, CocoaPods, XCFramework).
Remove the file once you upgrade to a version of maplibre-native that includes the fix.

albintornqvist:
First of all, thank you for doing a great job with this project. Being able to transition to MapLibre from MapBox v6 has saved us a lot of trouble.

The bug I found is that the map view occurs to get a lower resolution than usual when presented on an iPad in landscape mode. This is easily testable by running the "App" target in the MapLibre Xcode project on an iPad and switching between portrait and landscape. Texts and vector features appears sharper in portrait than in landscape. 
I tested this on an Ipad Pro 11" 2 gen (A2230) building with Xcode 26.3, using the "main" branch with commit SHA 58d76e16174d3766f1a5061af461271e5afd7b7a.

The problem seems to be that the scale factor becomes too low in landscape mode since it effectively becomes a comparison of the short side vs the long side instead of comparing the short sides and the long sides.

Disclaimer: I have definitely used AI (Claude, Opus 4.8) to come to this conclusion and the code should be thoroughly reviewed by someone who knows this code.

Let me know if I should remove the comment on the function (Claude can be a bit chatty) or if there is anything else I can do to improve the code.

According to Claude, this is a regression from #3890 (ios-v6.19.2).

Some screenshots below of running the "App" target in the Xcode project on the previously mentioned iPad in landscape mode. Cropped to make it easier to notice the difference.

Built from the main-branch:
IMG_0347

Built from this branch (fix/landscape-scale-factor):
IMG_0345


maximeg:
Hey! Thank you for the PR. It fixes #4349.

May I suggest to first check if we have `nativeScale` exposed on `screen` as it is the always correct number we seek here.
```
if ([screen respondsToSelector:@selector(nativeScale)]) {
  return screen.nativeScale;
}
```

louwers (approved):

louwers:
Thanks @albintornqvist could you address the comment from @maximeg?

We'll include this fix in the next release.

albintornqvist:
Thanks for the feedback! @louwers That's great to hear.

@maximeg That seems like a very good idea, I'll fix that. And also, great work on that issue! It really explains the problem well.

louwers:
Could you run:

```
pre-commit run --all-files
```

Since this PR is coming from an organization our pre-commit bot cannot push commits to this PR.

louwers:
By the way, @maximeg @albintornqvist we have a monthly meeting tonight to discuss MapLibre Native development. You would be very welcome there. If you want an invite, please send an e-mail to team@maplibre.org

albintornqvist:
> Could you run:
> 
> ```
> pre-commit run --all-files
> ```
> 
> Since this PR is coming from an organization our pre-commit bot cannot push commits to this PR.

Ah, I see. I'll try that!

PR #4377, opened 2026-07-06
core: add CustomVectorSource for binary tile data delivery

Vladyslav-Odobesku-Intellias:
Introduces CustomVectorSource — a new source type that lets applications deliver pre-encoded binary tile data (MVT) directly to the rendering pipeline, bypassing the GeoJSON encode/decode round-trip of CustomGeometrySource.

  - Core C++ API with fetch/cancel callbacks and async setTileData/setTileError/invalidateTile
  - Android/Kotlin platform binding using coroutines for ergonomic async tile fetching
  - Actor-based concurrency with cooperative cancellation
  - Reuses existing well-tested MVT infrastructure instead of introducing yet one more binary tile data format

**Design Proposal**

  Included in this PR: design-proposals/2025-07-05-custom-vector-source.md

**Motivation**

  Existing options for programmatic tile delivery have limitations:
  - URL-based vector sources require HTTP
  - CustomGeometrySource operates on GeoJSON (encode/decode overhead), uses bounding-box requests (unclear tile semantics), and manages its own threadpool (vs. Kotlin suspendible aka coroutines)

  CustomVectorSource fills the gap for applications that generate tiles on-device, decode from proprietary archives, or stream over non-HTTP transports.

  Here is the description of what problem we are solving by introducing such data source. Newer Google Maps SDK provides tile data through GRPC service. That service is launched along the app as a separate process and simply serves tiles through GRPC in its` own format that resembles vector tiles. We cannot request it to service tiles through GRPC + in the vector tiles format. So we need to adapt it. Proposed interface declares a simple interface for the tile data provider to implement.
```
interface CustomVectorTileProvider {
    suspend fun fetchTile(z: Int, x: Int, y: Int): TileData
}
```

  It everything. Only says that this function is suspendable (coroutine) and under the hood it can be anything - some exotic HTTP tile service or as in our case some GRPC service. For us to now adapt such GRPC service to MapLibre is just about implementing CustomVectorTileProvider by doing a suspendable call to GRPC, which will naturally be suspended and then transform data from one binary format to vector tiles.

**Check it yourself**
  This change also adds one entry to the test scenarios of the android demo app. It simply injects trivial CustomGeometrySource that provides cross tile data (2 lines that form a cross) and sets the style of those lines to be red.

**AI Disclosure**

  AI assistance (Claude) was used for drafting the design proposal document and PR description. All code and architectural decisions are human-authored.


design-proposals/2025-07-05-custom-vector-source.md
```diff
@@ -0,0 +1,183 @@
+# Custom Vector Source
+
+## Motivation
+
+MapLibre Native currently supports two ways to provide vector tile data:
+
+1. **URL-based sources** (`vector` source type) — tiles are fetched from a remote server via HTTP.
```
louwers: Not neccesarily via HTTP. You can also load vector sources from local files, i.e. `file://` or even `asset://` URIs.

design-proposals/2025-07-05-custom-vector-source.md
```diff
@@ -0,0 +1,183 @@
...
+2. **CustomGeometrySource** — accepts GeoJSON features programmatically, but internally re-encodes them into vector
+   tiles, adding overhead and limiting control over the tile encoding.
+
+This proposal comes from the need for an efficient way to connect a proprietary data source.
```
louwers: Why does it have to be "proprietary"?

design-proposals/2025-07-05-custom-vector-source.md
```diff
@@ -0,0 +1,183 @@
...
+That data source provides a gRPC service that has a simple API to request tiles by X/Y/Z (similarly to vector tiles) and the
```
louwers: I don't see how this is gRPC specific. It would be good to write the design proposal to be more data source agnostic.

sjg-wdw:
Thoughts @TimSylvester @alexcristici @adrian-cojocaru ?

design-proposals/2025-07-05-custom-vector-source.md
Vladyslav-Odobesku-Intellias: It does not. I just mentioned that this proposal is the result of a need to connect proprietary data source to MapLibre. It doesn't have to be proprietary.

design-proposals/2025-07-05-custom-vector-source.md
Vladyslav-Odobesku-Intellias: That's just some details about the reasoning behind introducing such data source (to connect GRPC based tile service to MapLibre).
If those details are a noise and should not be included in here then I can just strip it out.

Vladyslav-Odobesku-Intellias:
@louwers I've removed mentions of GRPC.
+ I've made design proposal to sound more generic.
As for the reasoning and description of how did we come up needing such custom source I've instead put it into this pull request description so that it does not pollute the design doc.

louwers:
@Vladyslav-Odobesku-Intellias

We will discuss this during the Technical Steering Comittee Meeting tomorrow, 7:30PM CEST.

Do you want to join? If so, please e-mail team@maplibre.org and I will send you an invite.

louwers:
Hi @Vladyslav-Odobesku-Intellias could you merge in main? That should resolve the CI issue.

Also there is a merge conflict.

Vladyslav-Odobesku-Intellias:
> Hi @Vladyslav-Odobesku-Intellias could you merge in main? That should resolve the CI issue.
> 
> Also there is a merge conflict.

done, all green

louwers (approved):

PR #4343, opened 2026-06-12
Add fill extrusion style property that enables rounded corners for extruded buildings (Core feature)

alexcristici:
Added new `fill-extrusion` layout style property:
 `"fill-extrusion-rounded-corner-distance": 10` (unit meters),
that enables extruded buildings to have rounded corners when corner distance is bigger than `0`.

Rounded corners will consume 5x more memory when enabled:

Memory consumption when disabled:
```
fillExtrusionBucketsMem: 2288582 bytes
fillExtrusionBucketsCount: 5
zoom: 15.3155
```

Memory consumption when enabled:
```
fillExtrusionBucketsMem: 12121630 bytes
fillExtrusionBucketsCount: 5
zoom: 15.3155
```

Rounded corner 3D buildinds

PR #4379, opened 2026-07-06
Add fill-extrusion-edge-radius: rounded footprint corners with smooth facade normals

Kotelberg:
## Motivation

Building extrusions in MapLibre render every footprint corner perfectly sharp, which reads fine for most buildings but looks wrong for anything that is actually round or chamfered in real life - rounded plazas, curved facades, chamfered towers. There's no way today to soften those corners short of hand-editing the source geometry, which isn't practical for data pulled from OSM or similar sources at scale.

This PR adds a `fill-extrusion-edge-radius` layout property (in meters, default `0`, fully opt-in) that rounds each footprint corner into a short arc at bucket time, and - where the renderer supports it - blends wall normals across gentle corners so the shaded facade reads as a smooth curve instead of a series of flat facets.

Before (default, `fill-extrusion-edge-radius` unset):

default

*Sharp corners, stock geometry - this is also the byte-identical baseline (see Evidence below).*

After (`fill-extrusion-edge-radius: 2`, a realistic ~2m physical radius):

rounded, radius 2

*Subtle but real corner softening at building scale (visible on the L-shaped notch in the block).*

Sanity check (`fill-extrusion-edge-radius: 8`, exaggerated to make the effect unmistakable):

rounded, radius 8

*Same viewpoint, radius pushed well past realistic building scale to confirm the mechanism is genuinely rounding corners and not a no-op.*

## Design notes

- **Bucket-time, not shader-time.** Rounding happens in `FillExtrusionBucket::addFeature`, before tessellation: each ring is walked and every corner whose radius-adjusted arc fits is replaced with a short quadratic-Bézier arc (control point at the original corner, sampled at `t = 1/3` and `2/3` - four points per rounded corner: start, two midpoints, end). This keeps the change local to geometry generation; the rest of the extrusion pipeline (earcut tessellation, wall-quad emission, index buffers) is unmodified.
- **Plan-view half-radius.** The configured radius is a meters-in-plan-view quantity. It's converted to tile units per canonical zoom (`radiusM * EXTENT / tileMetersAtZoom`, using an equator-circumference approximation, which is adequate for a visual radius) before being applied to the footprint ring. Only the horizontal footprint is rounded - the vertical top edge (where the extrusion meets its flat roof) is unaffected by this property; see "Top bevel is a non-goal" below.
- **Guards against degenerate geometry:**
  - Rings with fewer than 4 points (triangles) are left untouched - rounding a triangle's corners degenerates the shape rather than softening it.
  - Corners whose turn angle is nearly straight (`< ~0.20 rad`, about 11.5°) are skipped - rounding a corner that's already almost a straight line only adds vertices for no visible benefit.
  - The cut length on each side of a corner is clamped to at most half of the shorter adjacent edge (minus a small margin), so two nearby corners on a short edge can never produce overlapping/self-intersecting arcs.
  - If the clamped cut collapses below 1 tile unit, or the requested radius converts to less than ~1.5 tile units at the current zoom, the corner (or the whole ring, respectively) is left sharp rather than emitting a degenerate near-zero-length arc.
- **Crease-aware smooth wall normals, `dot > 0.5` (60°).** Each wall quad is otherwise flat-shaded using its own edge's perpendicular normal, so a facade approximated by several short rounded-corner edges would show visible facet banding instead of a smooth curve. Where the two edges meeting at a footprint vertex have normals whose dot product exceeds `0.5` (i.e. the turn between them is 60° or gentler - exactly the kind of turn a rounded corner produces), the vertex normal is blended between the two edge normals so the fragment shader interpolates a smooth gradient across the curve. Sharper corners (turn greater than 60°, i.e. real building corners that weren't rounded, or a rounding radius small enough that adjacent points still turn sharply) keep their own edge's distinct normal and stay visually crisp. At radius `0` this smoothing is disabled outright (`radiusM > 0.0` gate) so wall normals are exactly the stock faceted values - see Evidence for why this matters.
- **Path-dependent today: this smoothing is not yet universal.** The facade-normal blending described above lives entirely inside `#if !MLN_USE_FILL_EXTRUSION_INSTANCING` in `fill_extrusion_bucket.cpp`. `MLN_USE_FILL_EXTRUSION_INSTANCING` is on for the Metal and Vulkan backends, which means builds using the instanced fill-extrusion vertex path (notably macOS/iOS via Metal) get the rounded *geometry* - the plan-view corner rounding applies unconditionally, on every path - but render it with per-edge faceted wall normals rather than the smooth blend, i.e. rounded corners with visible facets rather than a smooth curve. Non-instanced (GL) builds get both the rounded geometry and the smooth normal blend. Extending the crease-aware normal blend to the instanced path is left as follow-up work; it wasn't done here to keep this PR's diff scoped to a single, reviewable mechanism rather than touching the instancing vertex layout as well.
- **Top bevel is an explicit non-goal.** A rounded *vertical* transition where the extrusion's side walls meet its flat roof (a bevel, not just a rounded footprint) would need paint-time height interaction - the bevel geometry depends on how tall the extrusion actually is, which is a data-driven paint property evaluated later in the pipeline, not something baked once at bucket time alongside the rest of this geometry. That's a materially different (and larger) change and is out of scope for this PR, which only rounds the plan-view footprint.
- **No `maximum` on the property (diverges from Mapbox).** See the [companion style-spec PR ](https://github.com/maplibre/maplibre-style-spec/pull/1748)for the full rationale: Mapbox GL JS v3's equivalent property caps at `maximum: 1`, but real building corners routinely exceed 1 meter, and the guards described above already make arbitrarily large radii degrade gracefully rather than break, so this proposal leaves the property uncapped (`minimum: 0` only).

### Default-off: byte-identical to stock

With `fill-extrusion-edge-radius` unset (the default, `0`), this branch must render pixel-for-pixel identically to upstream - no opt-in, no behavior change. Verified with `mbgl-render`, comparing an unmodified-upstream binary against this branch's binary on the same style and two real viewpoints over a live building tileset:

```
$ cmp base_b.png branch_b.png
(no output - byte-identical, both 21428 bytes)
```

That's the clean result. A second viewpoint (`base_a.png` / `branch_a.png`) showed a 3-byte file-size difference and 2 differing bytes out of 1,049,088 raw scanline bytes (both ±1, classic sub-pixel antialiasing jitter). Before treating that as a regression, we re-ran the *baseline* binary against itself twice at that same viewpoint and got the identical 2-byte diff at the identical offsets and values - i.e. this is pre-existing rendering nondeterminism in the baseline itself, unrelated to this change. This branch's own binary was fully self-consistent across repeated runs at that viewpoint (byte-identical to itself). So: one viewpoint is literally byte-identical, the other is byte-identical modulo baseline-only jitter that reproduces with or without this branch in the comparison. Net verdict: default-off behavior is unchanged.

### Feature-on smoke test

With the property set to a realistic value (`2`, ~2m), both viewpoints render without crashing, and a pixel diff against the default-off render shows roughly 3% of pixels changed, tracing exactly the building silhouette edges/corners - consistent with a geometric rounding effect rather than noise. A zoomed crop of an L-shaped notch corner shows a subtle but real softening at that scale. Pushing the radius to `8` (well past realistic building scale, included above purely as a mechanism check) makes the rounding unambiguous at every corner of the same notch, confirming the subtlety at radius `2` is a matter of scale, not a broken or no-op feature.

### Test coverage

Three new bucket-level unit tests (`test/renderer/fill_extrusion_bucket.test.cpp`):
- A square ring at radius `0` produces byte-identical vertex/index output to a captured stock (pre-this-branch) golden - right-angle corners stay exactly as upstream emits them.
- A regular octagon (45° turns at every corner - well under this PR's rounding-skip threshold and the smoothing crease threshold) at radius `0` also matches a stock golden byte-for-byte, generalizing the identity guarantee beyond a right-angled shape. Note: on this project's buildable local configuration (Metal), this exercises the shared/instancing-agnostic geometry path's identity at radius 0; the smooth-normal crease gate itself lives on the non-instanced path (see "Path-dependent today" above) and wasn't directly exercised by this fixture in that build - it's still a meaningful regression check on the geometry side.
- A square ring at radius `2` produces strictly more vertices than the same ring at radius `0` (17 vs. 5), confirming rounding actually adds the expected arc geometry.

Both golden fixtures were captured from a pure-upstream build (no edge-radius concept at all) before this branch's bucket changes existed, so they assert against real stock output rather than against this branch's own prior behavior.

Full existing test suite run alongside these three: 979 tests run (976 upstream baseline + these 3 new), 977 passed, 2 failed - the same 2 pre-existing `MapSnapshotter` failures present on a clean upstream checkout, unrelated to this change. No new failures, no regressions.

## Companion style-spec PR

[Style-spec PR](https://github.com/maplibre/maplibre-style-spec/pull/1748) - adds the same `fill-extrusion-edge-radius` property definition to `maplibre-style-spec`. Until that lands in a release, this PR vendors the identical JSON block directly into `scripts/style-spec-reference/v8.json` so generated style code (accessors, docs comments, platform bindings) can be produced without waiting on a spec release.

## Commits

1. `Add fill-extrusion-edge-radius layout property to the style spec and generated code` - vendors the property into `scripts/style-spec-reference/v8.json` and regenerates the core layer accessors (`getFillExtrusionEdgeRadius`/`setFillExtrusionEdgeRadius`/`getDefaultFillExtrusionEdgeRadius`, `FillExtrusionEdgeRadius` layout-property type).
2. `Round fill-extrusion footprint corners per fill-extrusion-edge-radius` - the bucket-time corner-rounding implementation (`roundRingCorners`/`roundPolygonCorners`) and its guards, plus the crease-aware smooth wall-normal blend for the resulting curved facades. The normal blend - and the per-ring `edgeNrm` precompute that feeds it - is gated on a positive radius, so at the default radius `0` the bucket takes the stock faceted-`perp` path with no extra allocation and byte-identical output to upstream.
3. `Add fill-extrusion edge-radius bucket tests` - the three unit tests described above (byte-identical-at-radius-0, rounded-square vertex count, octagon coverage caveat).
4. `Regenerate platform style code for fill-extrusion-edge-radius` - Android platform bindings (`PropertyFactory`, `FillExtrusionLayer`, JNI layer, test, CHANGELOG entries for iOS and Android). Darwin (`MLNFillExtrusionStyleLayer` and friends) is intentionally not included: those per-layer files are generated build artifacts that aren't tracked in this repository at all (confirmed against upstream history and the GitHub API) - running the generator locally reproduces them correctly, but there's nothing to commit.

## P.S.
We are [HataHub](https://www.hatahub.com.ua/) - the fastest growing PropTech startup in Ukraine. Map is key to our product, but mapbox license fees are outrageous at our scale. We have decided to go with maplibre and since maintained a fork where we have migrated a lot of mapbox v3 inspired functionality to both native and webgl - 3D layer, shadows, etc. We want to contribute back to community and will be raising PRs with core pieces back to maplibre over this week.

### DIFF
--- a/include/mbgl/style/layers/fill_extrusion_layer.hpp
+++ b/include/mbgl/style/layers/fill_extrusion_layer.hpp
@@ -19,6 +19,12 @@ public:
     FillExtrusionLayer(const std::string& layerID, const std::string& sourceID);
     ~FillExtrusionLayer() override;
 
+    // Layout properties
+
+    static PropertyValue<float> getDefaultFillExtrusionEdgeRadius();
+    const PropertyValue<float>& getFillExtrusionEdgeRadius() const;
+    void setFillExtrusionEdgeRadius(const PropertyValue<float>&);
+
     // Paint properties
 
     static PropertyValue<float> getDefaultFillExtrusionBase();
--- a/platform/android/CHANGELOG.md
+++ b/platform/android/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### ✨ Features and improvements
 
+- Add `fill-extrusion-edge-radius` layout property — bucket-time rounding of extruded footprint corners with smooth facade normals (default 0 = off).
 - Reduce Android runtime symbol resolution ([#4356](https://github.com/maplibre/maplibre-native/pull/4356)).
 
 ### 🐞 Bug fixes
--- a/platform/android/MapLibreAndroid/src/cpp/style/layers/fill_extrusion_layer.cpp
+++ b/platform/android/MapLibreAndroid/src/cpp/style/layers/fill_extrusion_layer.cpp
@@ -39,6 +39,17 @@ FillExtrusionLayer::~FillExtrusionLayer() = default;
 
 // Property getters
 
+jni::Local<jni::Object<>> FillExtrusionLayer::getFillExtrusionEdgeRadius(jni::JNIEnv& env) {
+    using namespace mbgl::android::conversion;
+    auto layer = layerPtr.get();
+    if (!layer) {
+        return std::move(
+            *convert<jni::Local<jni::Object<>>>(env, style::FillExtrusionLayer::getDefaultFillExtrusionEdgeRadius()));
+    }
+    return std::move(
+        *convert<jni::Local<jni::Object<>>>(env, toFillExtrusionLayer(*layer).getFillExtrusionEdgeRadius()));
+}
+
 jni::Local<jni::Object<>> FillExtrusionLayer::getFillExtrusionOpacity(jni::JNIEnv& env) {
     using namespace mbgl::android::conversion;
     auto layer = layerPtr.get();
@@ -288,6 +299,7 @@ void FillExtrusionJavaLayerPeerFactory::registerNative(jni::JNIEnv& env) {
         jni::MakePeer<FillExtrusionLayer, jni::String&, jni::String&>,
         "initialize",
         "finalize",
+        METHOD(&FillExtrusionLayer::getFillExtrusionEdgeRadius, "nativeGetFillExtrusionEdgeRadius"),
         METHOD(&FillExtrusionLayer::getFillExtrusionOpacityTransition, "nativeGetFillExtrusionOpacityTransition"),
         METHOD(&FillExtrusionLayer::setFillExtrusionOpacityTransition, "nativeSetFillExtrusionOpacityTransition"),
         METHOD(&FillExtrusionLayer::getFillExtrusionOpacity, "nativeGetFillExtrusionOpacity"),
--- a/platform/android/MapLibreAndroid/src/cpp/style/layers/fill_extrusion_layer.hpp
+++ b/platform/android/MapLibreAndroid/src/cpp/style/layers/fill_extrusion_layer.hpp
@@ -26,6 +26,8 @@ public:
 
     // Properties
 
+    jni::Local<jni::Object<jni::ObjectTag>> getFillExtrusionEdgeRadius(jni::JNIEnv&);
+
     jni::Local<jni::Object<jni::ObjectTag>> getFillExtrusionOpacity(jni::JNIEnv&);
     void setFillExtrusionOpacityTransition(jni::JNIEnv&, jlong duration, jlong delay);
     jni::Local<jni::Object<TransitionOptions>> getFillExtrusionOpacityTransition(jni::JNIEnv&);
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/layers/FillExtrusionLayer.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/layers/FillExtrusionLayer.java
@@ -144,6 +144,18 @@ public class FillExtrusionLayer extends Layer {
   // Property getters
 
   /**
+   * Get the FillExtrusionEdgeRadius property
+   *
+   * @return property wrapper value around Float
+   */
+  @NonNull
+  @SuppressWarnings("unchecked")
+  public PropertyValue<Float> getFillExtrusionEdgeRadius() {
+    checkThread();
+    return (PropertyValue<Float>) new PropertyValue("fill-extrusion-edge-radius", nativeGetFillExtrusionEdgeRadius());
+  }
+
+  /**
    * Get the FillExtrusionOpacity property
    *
    * @return property wrapper value around Float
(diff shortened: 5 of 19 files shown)


louwers:
>  We have decided to go with maplibre

That's awesome to hear! Would you be able to join the monthly meeting tomorrow?

If you can e-mail me at team@maplibre.org I will send you an invite.

Would be a good place to discuss this PR (since someone else is also working on this).

By the way, maybe take a look at our [AI Policy](https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md). We prefer human written PR descriptions and require disclosure of AI usage.

alexcristici:
Some corners are not rounded correctly because the new added points have coordinate as `short` which lose the floating point precision.

Screenshot 2026-07-08 at 20 38 26

louwers:
Superseded by https://github.com/maplibre/maplibre-native/pull/4343

Please let me know if this implementation works for your use case.

PR #4380, opened 2026-07-07
Add WGPU default platform includes to mbgl-core

gatesn:
Adds platform/default/include to mbgl-core when WGPU is enabled, so core-only source builds can compile the WebGPU headless backend header from the default platform include tree.

### DIFF
--- a/cmake/webgpu.cmake
+++ b/cmake/webgpu.cmake
@@ -9,6 +9,12 @@ target_compile_definitions(
         MLN_RENDER_BACKEND_WEBGPU=1
 )
 
+target_include_directories(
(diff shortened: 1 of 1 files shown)


louwers (approved):

louwers:
Do you need a new release?

Murmele:
> Do you need a new release?

For maplibre-native-rs we use always the hash and build the project directly so we don't need a new release. For prebuilds which can be used also for maplibre-native-rs it would be nice but I think it does not have rush yet

louwers:
OK, if you want you can also make your own releases at some point.

PR #4391, opened 2026-07-08
Fix Darwin style codegen Bazel requirement

gatesn:
Defer Darwin style-code Bazel lookup until configuring the SDK target that consumes generated style sources.

Sorry, I'm relatively new to CMake so I'm not sure if this is the right idea! But I'm trying to remove configure-time dependencies that are never used as part of the build to avoid requiring e.g. bazel when it's never used as part of the mbgl-core build


louwers:
Did you use AI for this PR?

louwers (approved):

PR #3917, opened 2025-10-29
Adds some network delegate methods

AtlasProgramming:
This PR adds the ability to intercept the network requests before and after the actual call.  This will enable the ability to add headers, etc to the request and handle handle specific types of errors/etc in the responses.

PR #4393, opened 2026-07-09
Fix MLNNetworkConfiguration not forwarding 'didReceiveResponse' to its delegate.

albintornqvist:
I found that the MLNNetworkConfiguration isn't calling all of the protocol functions of the MLNNetworkConfigurationDelegate. More specifically the `didReceiveResponse` is never being called, resulting in any implementation of MLNNetworkConfigurationDelegate not being able to listen to network responses or modify the response. 

AI usage: I have used Claude Fable 5 to help me find the problem and come up with a fix for it.

Description written by Claude:

## Root cause

The forwarding method in `MLNNetworkConfiguration.mm` checks `respondsToSelector:` and converts the internal response into an `MLNNetworkResponse` for the delegate — but never actually invokes the delegate. It just converts the unchanged response back and returns it:

```objc
if ([self.delegate respondsToSelector:@selector(didReceiveResponse:)]) {
  MLNNetworkResponse *tempResponse = [MLNNetworkResponse responseWithData:response.data
                                                              urlResponse:response.response
                                                                    error:response.error];
  // the delegate call was missing here
  if (tempResponse) {
    ...
```

The dead `if (tempResponse)` nil-check suggests the delegate call was intended but accidentally left out. This has been the case since the API was introduced in #3917, so it appears the delegate method has never worked.

## Fix

Invoke the delegate and use the response it returns, so delegates can observe or modify responses. One behavioral note: a delegate returning `nil` now suppresses the response (the method returns `nil`), which the pre-existing nil-check clearly anticipated but which could not previously happen — flagging it in case reviewers see that differently.

## Testing

Added two unit tests to `MLNNetworkConfigurationTests`:

- The delegate receives the response exactly once, and modifications the delegate makes (replacing the `NSURLResponse`) are reflected in the returned internal response.
- Without a delegate, the response is returned unchanged.

### DIFF
--- a/platform/darwin/src/MLNNetworkConfiguration.mm
+++ b/platform/darwin/src/MLNNetworkConfiguration.mm
@@ -101,6 +101,7 @@ NSString *const kMLNDownloadPerformanceEvent = @"mobile.performance_trace";
     MLNNetworkResponse *tempResponse = [MLNNetworkResponse responseWithData:response.data
                                                                 urlResponse:response.response
                                                                       error:response.error];
+    tempResponse = [self.delegate didReceiveResponse:tempResponse];
     if (tempResponse) {
       MLNInternalNetworkResponse *internalResponse =
           [MLNInternalNetworkResponse responseWithData:tempResponse.data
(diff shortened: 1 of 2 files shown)


louwers (approved):

louwers:
Thanks for your contribution.

PR #4397, opened 2026-07-11
core: Fix Vulkan snapshot timing for headless too

ciscorn:
Follow-up to #4339, which fixed the surface snapshot timing but not the headless backend.

The headless readback copied the color image without a barrier against rendering, so some (slower?) drivers (e.g. MoltenVK on macOS CI) could read back a stale frame.

### DIFF
--- a/platform/default/src/mbgl/vulkan/headless_backend.cpp
+++ b/platform/default/src/mbgl/vulkan/headless_backend.cpp
@@ -26,8 +26,8 @@ public:
     void bind() override {}
 
     void swap() override {
+        queueSurfaceRead();
         SurfaceRenderableResource::swap();
-        static_cast<Context&>(backend.getContext()).waitFrame();
     }
 };
 
@@ -41,8 +41,6 @@ HeadlessBackend::HeadlessBackend(const Size size_,
 
 HeadlessBackend::~HeadlessBackend() {
     gfx::BackendScope guard{*this, gfx::BackendScope::ScopeType::Implicit};
-
-    texture.reset();
 
     // Explicitly reset the renderable resource
     resource.reset();
@@ -79,21 +77,11 @@ PremultipliedImage HeadlessBackend::readStillImage() {
         resource = std::make_unique<HeadlessRenderableResource>(*this);
     }
 
-    auto& contextImpl = static_cast<Context&>(*context);
     auto& resourceImpl = static_cast<HeadlessRenderableResource&>(*resource);
 
-    if (!texture) {
-        texture = std::make_unique<Texture2D>(contextImpl);
-        texture->setFormat(gfx::TexturePixelType::RGBA, gfx::TextureChannelDataType::UnsignedByte);
-        texture->setUsage(Texture2DUsage::Read);
-    }
-
-    texture->setSize(size);
-
-    contextImpl.waitFrame();
-    texture->copyImage(resourceImpl.getAcquiredImage(), size);
-
-    return std::move(*texture->readImage());
+    auto image = resourceImpl.readImage();
+    assert(image);
+    return std::move(*image);
 }
 
 RendererBackend* HeadlessBackend::getRendererBackend() {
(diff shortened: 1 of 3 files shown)


platform/default/src/mbgl/vulkan/headless_backend.cpp
```diff
@@ -26,8 +26,8 @@ class HeadlessRenderableResource final : public vulkan::SurfaceRenderableResourc
     void bind() override {}
 
     void swap() override {
+        queueSurfaceRead();
```
adrian-cojocaru, line 29: This queues the surface read at the end of the frame and it will skip the first rendered frame (since all render tests use only one frame they will fail).

adrian-cojocaru:
Does this fix the stale frame read back? The timing issue was mostly for a surface read, the headless implementation bypasses almost all surface logic and ends up with a very similar code path as before (using one command buffer instead of two).

ciscorn:
@adrian-cojocaru  A follow-up commit (d201ebd4) adds a barrier before the headless readback copy, the same way the surface path barriers its read.

Headless needs it because the copy reads the frame that was just rendered, so it has to be ordered against the render's color image writes - the same class of timing issue as surface.

platform/default/src/mbgl/vulkan/headless_backend.cpp
ciscorn: For headless, the copy runs in this same `swap()`, so it reads the current frame. My downstream (`maplibre-native-rs`) single-frame render tests still pass.

adrian-cojocaru: I was thinking about [this](https://github.com/maplibre/maplibre-native/blob/62d716aed5457df592bb06e9216d500ed5409a09/src/mbgl/vulkan/renderable_resource.cpp#L108) (that gets applied at the start of the frame), but it's needed only for surfaces. Not an issue for headless.

src/mbgl/vulkan/renderable_resource.cpp
```diff
@@ -536,22 +538,22 @@ void SurfaceRenderableResource::copySurfaceToReadTexture() {
...
-        const auto barrier = vk::ImageMemoryBarrier()
-                                 .setImage(swapchainImage)
-                                 .setOldLayout(vk::ImageLayout::eTransferSrcOptimal)
-                                 .setNewLayout(vk::ImageLayout::ePresentSrcKHR)
-                                 .setSrcAccessMask(vk::AccessFlagBits::eMemoryRead)
-                                 .setDstAccessMask(vk::AccessFlagBits::eTransferWrite)
-                                 .setSrcQueueFamilyIndex(VK_QUEUE_FAMILY_IGNORED)
-                                 .setDstQueueFamilyIndex(VK_QUEUE_FAMILY_IGNORED)
-                                 .setSubresourceRange({vk::ImageAspectFlagBits::eColor, 0, 1, 0, 1});
+        const auto presentBarrier = vk::ImageMemoryBarrier()
```
ciscorn, line 539: renamed for `-Wshadow`

louwers (commented):
@adrian-cojocaru Do you think this is good to merge?

src/mbgl/vulkan/renderable_resource.cpp
```diff
@@ -502,25 +504,25 @@ void SurfaceRenderableResource::copySurfaceToReadTexture() {
...
     auto& commandBuffer = contextImpl.getCommandBuffer();
 
-    if (surface) {
-        const auto barrier = vk::ImageMemoryBarrier()
+    const auto oldLayout = surface ? vk::ImageLayout::ePresentSrcKHR : vk::ImageLayout::eTransferSrcOptimal;
+    const auto srcStage = surface ? vk::PipelineStageFlagBits::eTopOfPipe
+                                  : vk::PipelineStageFlagBits::eColorAttachmentOutput;
+    const vk::AccessFlags srcAccess = surface ? vk::AccessFlags{}
+                                              : vk::AccessFlags{vk::AccessFlagBits::eColorAttachmentWrite};
+    const vk::AccessFlags dstAccess = surface ? vk::AccessFlagBits::eTransferWrite : vk::AccessFlagBits::eTransferRead;
```
adrian-cojocaru:
```suggestion
    const auto srcStage = vk::PipelineStageFlagBits::eColorAttachmentOutput;
    const vk::AccessFlags srcAccess = vk::AccessFlagBits::eColorAttachmentWrite;
    const vk::AccessFlags dstAccess = vk::AccessFlagBits::eTransferRead;
```
Stage and flags can be merged for both scenarios (`dstAccess` should have always been read).

ciscorn: Thanks! I kept the original values to be conservative.

adrian-cojocaru (approved):

PR #4399, opened 2026-07-12
Convert a PMTiles metadata decompression failure into an error response

bonesyjr:
## Summary

A PMTiles archive whose JSON metadata block fails to gunzip currently **aborts the process**. `util::decompress` throws `std::runtime_error`, and the call at `pmtiles_file_source.cpp:453` is the one decompress call site in this file with no handler around it — the exception unwinds out of the response callback on the `PMTilesFileSource` thread, hits `std::terminate`, and the app dies with `SIGABRT`.

The neighbouring paths already do the right thing: the tile path catches and reports `"Error decompressing PMTiles tile:"` (added in #4090), and the directory path catches and reports `"Error parsing PMTiles directory:"`. This makes the metadata path behave the same way — a corrupt or unexpected metadata block becomes a failed request, not a crash.

## How we hit it

We serve ~2GB PMTiles archives over HTTP from object storage. Two things put non-gzip bytes at `json_metadata_offset`, and both end the same way:

- **A rebuilt archive at a stable URL.** The header carries byte offsets into one specific build. A client holding a cached header from the previous build range-fetches the metadata at the old offset, gets whatever now lives there, and aborts. (We've since made our archives immutable, but the crash is reachable by any deployment that overwrites an archive in place.)
- **A CDN that ignores `Range`.** We observed a ranged request for a multi-GB archive answered with `200` and the entire body. `decompress()` then runs over the whole archive and throws.

Neither is MapLibre's fault, but neither should be able to take the process down — from the app's perspective these are just bad bytes on the wire, exactly like a corrupt tile. Crash reports show only `abort` inside `mbgl::util::decompress` with no application frames, which makes this very hard to diagnose downstream.

## Changes

- Wrap the metadata `util::decompress` call in try/catch and surface `Response::Error{Reason::Other, "Error decompressing PMTiles metadata: ..."}`, mirroring the directory handler directly below it.
- Add `PMTilesFileSource.CorruptGzipMetadata`, mirroring the existing `CorruptGzipTile` test.
- Add the `corrupt-gzip-metadata.pmtiles` fixture plus the generator script that produces it, mirroring `corrupt-gzip-tile.py`. The fixture sets `internal_compression = GZIP` with a valid (empty) root directory, so the failure under test is isolated to the metadata block.

Without the source change, the new test aborts the test binary rather than failing.


louwers (approved):

louwers:
@bonesyjr CI is red, could you have a look?

louwers:
Sorry, seems like a GitHub issue. Triggering a new CI run.

PR #4405, opened 2026-07-14
Fix Android feature state instrumentation tests

louwers:
Fixes the Android feature state instrumentation tests. As per the [spec](https://maplibre.org/maplibre-style-spec/expressions/#feature-state):

> Note that ["feature-state"] can only be used with paint properties that support data-driven styling.

And this is ~probably~ true in the MapLibre Native implementation too...

For reference here is what GPT 5.6 spat out.

> ## Summary
> 
> - replace the unsupported `feature-state` layer filter in the Android test fixture with a feature-state-driven `circle-radius` paint expression
> - query away from the feature center so the rendered hit area still proves that state changes are applied
> - assert state update dispatch, direct state retrieval, source-wide key removal, and final state removal
> 
> ## Root cause
> 
> Feature state is supported in data-driven paint properties, but not in layer filter expressions. The previous fixture used it in a filter, which is evaluated while building the circle bucket without feature state. The feature was therefore excluded permanently, and `queryRenderedFeatures` continued to return zero after setting state.
> 
> ## Testing
> 
> - focused `connectedOpenglDebugAndroidTest` run on a local Android 17 emulator: 2/2 passed
> - three additional focused instrumentation runs using the installed test package: 6/6 passed
> 
> The failure reproduced locally before this change, so no additional AWS Device Farm run was submitted.
>

### DIFF
--- a/platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/testapp/style/GeoJsonSourceTests.java
+++ b/platform/android/MapLibreAndroidTestApp/src/androidTest/java/org/maplibre/android/testapp/style/GeoJsonSourceTests.java
@@ -23,6 +23,7 @@ import org.hamcrest.Matcher;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
+import android.graphics.PointF;
 import android.view.View;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


louwers:
@brandonroeder Please have a look if you can.

HarelM (approved):

PR #4408, opened 2026-07-16
fix(android): Add option to move symbol location indicator bearing down the image stack

adrian-cojocaru:
Currently the symbol location indicator uses a background/foreground/bearing image stack, while the specialized implementation and overall documentation use the bearing under foreground and background images. 
This PR adds the `bearingOnTop` option to move the bearing image to make both implementation consistent.

Bearing image is visible only when using `RenderMode.Compass` (all other modes are unaffected) and maplibre default values don't overlap making the draw order irrelevant for this scenario.
<p align="center">
default bearing </br>
<sub>Default bearing drawable</sub>
</p>

<p align="center">Custom bearing drawable:</p>

<table align="center">
    <tr>
        <td align="center">
            
            </br><sub>bearingOnTop(true) - main</sub>
        </td>
        <td align="center">
            bearingOnTop(true) - current implementation
            </br><sub>bearingOnTop(false) -  PR</sub>
        </td>
    </tr>
</table>

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/location/LocationComponentOptions.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/location/LocationComponentOptions.java
@@ -133,6 +133,7 @@ public class LocationComponentOptions implements Parcelable {
   private RectF trackingMultiFingerProtectedMoveArea;
   private String layerAbove;
   private String layerBelow;
+  private Boolean bearingOnTop;
   private float trackingAnimationDurationMultiplier;
   private boolean compassAnimationEnabled;
   private boolean accuracyAnimationEnabled;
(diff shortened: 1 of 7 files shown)


louwers (commented):
Could you share a screenshot of how different it looks?

adrian-cojocaru:
@louwers Samples and default values are unaffected by this change. Added a couple of images for a custom drawable that overlaps with the foreground.

louwers:
@adrian-cojocaru So I guess it's not a breaking change anymore, correct?

adrian-cojocaru:
It's breaking for anybody that uses custom images for the bearing drawable.

louwers:
Would flipping the default value of `bearingOnTop` make it backward compatible?

adrian-cojocaru:
Yes, it makes it backwards compatible (but leaves the two implementations with different results)

louwers (approved):

PR #4411, opened 2026-07-17
feat(android): Create default OkHttpClient only when needed

mjaakko:
Currently, the default OkHttpClient is always created even when a custom HTTP client is provided and the default client is never used.

In my use case, I want all HTTP traffic to be done with a custom OkHttpClient. I'm using `HttpRequestUtil.setOkHttpClient(httpClient)` to set the HTTP client just before initializing the map. This caused the default client to be created even when not needed. Doing this from the UI thread causes a StrictMode violation:
```
StrictMode policy violation; ~duration=18 ms: android.os.strictmode.CustomViolation: newSSLContext
  at android.os.StrictMode$AndroidBlockGuardPolicy.onCustomSlowCall(StrictMode.java:1650)
  at android.os.StrictMode.noteSlowCall(StrictMode.java:2785)
  at okhttp3.internal.platform.Android10Platform.newSSLContext(Android10Platform.kt:63)
  at okhttp3.internal.platform.Platform.newSslSocketFactory(Platform.kt:195)
  at okhttp3.OkHttpClient.<init>(OkHttpClient.kt:293)
  at okhttp3.OkHttpClient$Builder.build(OkHttpClient.kt:1382)
  at org.maplibre.android.module.http.HttpRequestImpl.<clinit>(HttpRequestImpl.java:52)
  at org.maplibre.android.module.http.HttpRequestUtil.setOkHttpClient(HttpRequestUtil.java:51)
```

Existing behavior should be preserved. The default client gets created lazily if no custom client is provided and setting the client to `null` restores the default client.

### DIFF
--- a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/module/http/HttpRequestImpl.java
+++ b/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/module/http/HttpRequestImpl.java
@@ -1,5 +1,7 @@
 package org.maplibre.android.module.http;
 
+import static org.maplibre.android.module.http.HttpRequestUtil.toHumanReadableAscii;
+
 import android.os.Build;
 import android.text.TextUtils;
 import android.util.Log;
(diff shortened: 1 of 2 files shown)


louwers (approved):

louwers:
Thanks for your contribution.

This will be included in the next release.

PR #4412, opened 2026-07-17
build(android): Use cacheable Gradle task for downloading Vulkan validation layers

mjaakko:
Working on #4411 was really painful with a slow, unreliable connection as the build process downloads Vulkan validation layers for every build.

This PR creates a cacheable task for downloading the Vulkan validation layers. The task assumes that the URL content is static. If the input URL has not changed, Gradle can reuse results from previous builds.


louwers (approved):
Thanks for your contribution!

adrian-cojocaru (approved):

PR #4413, opened 2026-07-17
Remove AWS key from iOS/Android test apps

louwers:
This key has been revoked. If we want to use the AWS Style for benchmarking we should bake it in the test app via a GitHub secret.


adrian-cojocaru (approved):

platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/activity/benchmark/BenchmarkActivity.kt
```diff
@@ -125,7 +124,6 @@ class BenchmarkActivity : AppCompatActivity() {
             ),
             styleURLs = listOf(
 //                {"message":"User is not authorized to access this resource with an explicit deny in an identity-based policy"}
```
adrian-cojocaru:
```suggestion
```

PR #4417, opened 2026-07-20
implement rotateBy by specifying an angle in degree

Murmele:
The idea is that you can rotate the map using the scroll wheel of the mouse.

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -506,6 +506,16 @@ void Transform::rotateBy(const ScreenCoordinate& first,
     easeTo(CameraOptions().withBearing(bearing), animation);
 }
 
+void Transform::rotateBy(double angleDegree, const AnimationOptions& animation) {
+    const double bearing = -util::rad2deg(state.getBearing()) - angleDegree;
+    easeTo(CameraOptions().withBearing(bearing), animation);
+}
+
+/*!
+ * \brief Transform::getBearing
+ * Bearing in radiant
+ * \return
+ */
 double Transform::getBearing() const {
     return state.getBearing();
 }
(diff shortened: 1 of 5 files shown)


louwers:
Please provide more context and what you expect this to do and why you need this.

Right now, positive values can sometimes rotate clockwise or counterclockwise. Is this what you intended?

Murmele:
> Please provide more context and what you expect this to do and why you need this.
> 
> Right now, positive values can sometimes rotate clockwise or counterclockwise. Is this what you intended?

I added the reason to the description. It shall be able to rotate the map using the mouse wheel. What do you mean by "can sometimes rotate clockwise or counterclockwise"?

src/mbgl/map/transform.cpp
```diff
@@ -506,6 +506,16 @@ void Transform::rotateBy(const ScreenCoordinate& first,
...
 
+void Transform::rotateBy(double angleDegree, const AnimationOptions& animation) {
+    const double bearing = -util::rad2deg(state.getBearing()) - angleDegree;
+    easeTo(CameraOptions().withBearing(bearing), animation);
+}
+
+/*!
+ * \brief Transform::getBearing
+ * Bearing in radiant
+ * \return
```
louwers:
```suggestion
 * \return bearing in radians
```

louwers:
@Murmele `easeTo` uses the shortest distance, so If you do `rotateBy(270)` it will rotate 90 degrees counterclockwise. Also `rotateBy(360)` will not rotate at all.

But yes, if you jump small amount with a scroll wheel, then it will do what you expect.

If you can add some tests we can merge this I think.

src/mbgl/map/transform.cpp
louwers: radiant -> radians

Murmele:
Thanks. Yes I will add

PR #4418, opened 2026-07-20
Maplibre-Core on android

Murmele:
This is an initial work to get maplibre-core working on android using maplibre-native-rs.

The idea is to have custom interfaces so the Rust Functionality for Http and so on can be used instead of using the android java code. This code is far from finish, just a prototype yet but I wanna hear your opinion about.

@louwers

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -371,6 +371,53 @@
         "MLN_WITH_VULKAN": "ON",
         "MLN_WITH_GLFW": "ON"
       }
+    },
+    {
+      "name": "android",
+      "hidden": true,
+      "generator": "Ninja",
+      "cacheVariables": {
+        "CMAKE_SYSTEM_NAME": "Android",
+        "CMAKE_ANDROID_NDK": "$env{ANDROID_NDK_ROOT}",
+        "CMAKE_TOOLCHAIN_FILE": "$env{ANDROID_NDK_ROOT}/build/cmake/android.toolchain.cmake",
+        "ANDROID_PLATFORM": "23",
+        "ANDROID_ABI": "arm64-v8a",
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "MLN_WITH_OPENGL": "OFF",
+        "MLN_WITH_VULKAN": "OFF"
+      }
+    },
+    {
+      "name": "android-arm64-v8a-opengl",
+      "binaryDir": "${sourceDir}/build-android-arm64-v8a-opengl",
+      "inherits": "android",
+      "displayName": "Android ARM64 OpenGL ES",
+      "description": "Android ARM64 build (OpenGL ES renderer)",
+      "cacheVariables": {
+        "MLN_WITH_OPENGL": "ON"
+      }
+    },
+    {
+      "name": "android-arm64-v8a-vulkan",
+      "binaryDir": "${sourceDir}/build-android-arm64-v8a-vulkan",
+      "inherits": "android",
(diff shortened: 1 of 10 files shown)


louwers (commented):
Have you seen https://maplibre.org/maplibre-native/docs/book/rust.html?

You could probably make a Rust implementation of `HTTPFileSource` on this side.

vendor/harfbuzz.cmake
```diff
@@ -2,9 +2,18 @@ if(TARGET mbgl-harfbuzz)
...
+    if (MLN_WEBGPU_IMPL_FFI)
+		add_library(mbgl-harfbuzz SHARED
+			${CMAKE_CURRENT_LIST_DIR}/harfbuzz/src/harfbuzz.cc
+		)
+        # This target is consumed as a shared dependency by other DSOs.
+        # Keep symbols visible so runtime resolution (e.g. hb_ft_font_create) succeeds on Android.
+        set_target_properties(mbgl-harfbuzz PROPERTIES CXX_VISIBILITY_PRESET default VISIBILITY_INLINES_HIDDEN OFF)
+	else()
+		add_library(mbgl-harfbuzz STATIC
+			${CMAKE_CURRENT_LIST_DIR}/harfbuzz/src/harfbuzz.cc
```
louwers, line 14: Maybe you can look into HarfRust. ;-)

CMakePresets.json
```diff
@@ -371,6 +371,53 @@
...
+    {
+      "name": "android-arm64-v8a-wgpu-ffi",
+      "binaryDir": "${sourceDir}/build-android-arm64-v8a-wgpu-ffi",
+      "inherits": "android",
+      "displayName": "Android ARM64 Vulkan",
+      "description": "Android ARM64 build (Vulkan renderer)",
+      "cacheVariables": {
+        "MLN_WITH_WEBGPU": "ON",
+        "MLN_WEBGPU_IMPL_WGPU": "ON",
+        "MLN_WEBGPU_IMPL_FFI": "ON"
```
louwers, line 419: Just noticed this, but a default should be set with a description in `CMakeLists.txt`.

platform/android/src/jni.cpp
```diff
@@ -7,6 +7,12 @@ namespace android {
 
 JavaVM* theJVM;
 
+#if MLN_WEBGPU_IMPL_FFI
```
louwers, line 10: What's this?

Murmele:
> Have you seen https://maplibre.org/maplibre-native/docs/book/rust.html?
> 
> You could probably make a Rust implementation of `HTTPFileSource` on this side.

The problem then is still that the http rust library is then statically linked once on the maplibre-native side and once on the rust application side?

platform/android/src/jni.cpp
Murmele: I need to set the jvm from externally to be able to use maplibre-native with the ndk

CMakePresets.json
Murmele: You are right I forgott to do it in my previous PR. Thanks I will add

vendor/harfbuzz.cmake
Murmele: But then we have to write the ffi to C++ again :). If maplibre-native is open to implement parts in Rust I am open for it :)

CMakePresets.json
Murmele: I have seen again. They are documented in vendor/webgpu.cmake also with a default value

PR #4420, opened 2026-07-20
feat(ios): add feature state api

MarkusWendorf:
Implements feature state for iOS. feature-state was already implemented in the core.

This is an alternative implementation to #3858, deliberately aligned with the API shape that shipped for Android in #4219.

I also added a vector-tile example to Android, so that we now have a GeoJSON example AND a vector-tile example for feature-state on both iOS and Android.

Tests added:

  - platform/ios/test/MLNFeatureStateTests.swift —
  MLNFeatureStateTests.testFeatureStateRoundTrip (integration test through the Swift source
  API).
  - test/api/query.test.cpp — Query.QuerySourceFeatureStates and
  Query.RemoveSourceFeatureState (set/get round-trip and key-/feature-level removal via the
  renderer).
  - test/map/map.test.cpp — Map.FeatureStateChangesRenderedStyle, an end-to-end test asserting
  that a feature-state change actually alters rendered pixel output through a data-driven
  ["case", ["boolean", ["feature-state", "active"], …]] fill-color expression.
  
  Co-authored with AI: Anthropic Fable/Opus 4.8


louwers:
Thanks Markus. I think an implementation that aligns with Android makes more sense.

Will have a look at this tomorrow.

platform/ios/app-swift/Sources/FeatureStateExample.swift
```diff
@@ -0,0 +1,128 @@
...
+        let features = mapView.visibleFeatures(at: location, styleLayerIdentifiers: ["state-fills"])
+
+        guard let feature = features.first, let featureID = feature.identifier.map({ "\($0)" }) else {
+            return
+        }
+
+        // Toggle the selection state of the tapped feature. Selection is not
+        // exclusive: any number of states can be selected at the same time.
+        let isSelected = statesSource.featureState(featureID: featureID)?["selected"] as? Bool ?? false
+        statesSource.setFeatureState(featureID: featureID, state: ["selected": !isSelected])
```
MarkusWendorf, line 92: @louwers not sure if I jumped the shark here but I assumed your PR is out-dated now that the `source` centric API from Android was merged.

zhengxuehua424-debug (approved):

platform/darwin/src/MLNSource.mm
```diff
@@ -108,6 +113,96 @@ - (BOOL)removeFromStylable:(id<MLNStylable>)mapView
...
+
+  mbgl::Renderer *renderer = self.mgl_renderer;
+  if (!renderer || !featureID || !state) {
+    return NO;
+  }
+
+  mbgl::FeatureState featureState;
+  for (NSString *key in state) {
+    NSExpression *expression = [NSExpression expressionForConstantValue:state[key]];
+    featureState[key.UTF8String] = expression.mgl_constantMBGLValue;
```
brandonroeder: I think nested JSON values crash here because this conversion only handles scalars. Could you use the existing recursive `mgl_propertyMap` conversion and add a nested state test? Android supports this through `JsonObject`

brandonroeder:
Tested this out and it works great. Left one comment but otherwise lgtm 

cc: @louwers

platform/darwin/src/MLNSource.mm
MarkusWendorf: good catch, fixed

louwers (approved):

louwers:
Will be included in the next release.

Thank you!

MarkusWendorf:
@louwers is the next iOS release around the corner? thought this would be released together with 6.29. Thanks

louwers:
Sorry this was intended to be merged earlier.

PR #4424, opened 2026-07-22
core: check if edge length is zero for rounded corners 

zarechnyy:
This PR adds an additional check if nearby points are too close to be rounded.

Current implementation works well for 3D buildings without [buildings:parts](https://wiki.openstreetmap.org/wiki/Key:building:part) enriched geometry, but when the map style has building:part we can observe some artifacts on the map. 
For duplicate vertex - the edge vector length is zero, so `util::normal` computes 0.0 / 0.0 which leads to `NaN`. 

Prerequisites:
Use style with building:part geometry for extruded buildings.
Set any value greater than zero for `fill-extrusion-rounded-corner-distance`

**BEFORE:**
image

image

**AFTER:**
Simulator Screenshot - BAZEL_TEST_iPhone 11 Pro_26 4 - 2026-07-22 at 13 20 25
Simulator Screenshot - BAZEL_TEST_iPhone 11 Pro_26 4 - 2026-07-22 at 13 20 13

### DIFF
--- a/src/mbgl/tile/geometry_tile_data.cpp
+++ b/src/mbgl/tile/geometry_tile_data.cpp
@@ -316,12 +316,20 @@ GeometryCollectionFloat roundPolygonCorners(GeometryCollection& polygon, double 
             auto cornerPoint = convertPoint<double>(ring[i]);
             auto nextPoint = convertPoint<double>(ring[(i + 1) % nVertices]);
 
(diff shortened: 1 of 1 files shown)


louwers:
How hard would it be to add a render test to have a regression test for this behavior?

louwers (approved):

tobi-stephen (approved):

alexcristici (approved):

PR #4428, opened 2026-07-23
core: add Rust-backed MLT tile data

louwers:
The Rust MLT decoder has a safe FastPFor implementation.

This PR is a vibe coded (GPT 5.6) integration with [`mlt-core`](https://crates.io/crates/mlt-core).

### DIFF
--- a/rustutils/src/mlt.rs
+++ b/rustutils/src/mlt.rs
@@ -81,11 +81,7 @@ mod ffi {
             feature_index: usize,
             name: &str,
         ) -> MltValue;
-        fn property_values(
-            &self,
-            layer_index: usize,
-            feature_index: usize,
-        ) -> &[MltValue];
+        fn property_values(&self, layer_index: usize, feature_index: usize) -> &[MltValue];
         fn property_string_value(
             &self,
             layer_index: usize,
@@ -178,7 +174,10 @@ pub fn decode_mlt_tile(data: &[u8]) -> Result<Box<MltTile>, String> {
     let mut decoder = Decoder::default();
     let mut result = Vec::with_capacity(layers.len());
     for layer in layers {
-        match layer.decode_all(&mut decoder).map_err(|err| err.to_string())? {
+        match layer
+            .decode_all(&mut decoder)
+            .map_err(|err| err.to_string())?
+        {
             Layer::Tag01(layer) => {
                 let triangles = feature_triangles(layer.geometry_values());
                 result.push(convert_layer(&layer, triangles)?);
@@ -230,7 +229,8 @@ fn parse_layers(data: &[u8]) -> Result<Vec<Layer<'_>>, String> {
 
 fn feature_triangles(geometry: &mlt_core::GeometryValues) -> Vec<Vec<u32>> {
     let mut result = vec![Vec::new(); geometry.feature_count()];
-    let (Some(triangle_counts), Some(index_buffer)) = (geometry.triangles(), geometry.index_buffer())
+    let (Some(triangle_counts), Some(index_buffer)) =
+        (geometry.triangles(), geometry.index_buffer())
     else {
         return result;
     };
@@ -431,9 +431,19 @@ fn append_geometry_parts(
                 let Some(ring_index) = offset(parts, part_index) else {
                     return;
                 };
-                push_vertex_range(vertices, offset_range(rings, ring_index), false, coordinates);
+                push_vertex_range(
+                    vertices,
+                    offset_range(rings, ring_index),
+                    false,
+                    coordinates,
+                );
             } else {
-                push_vertex_range(vertices, offset_range(parts, part_index), false, coordinates);
+                push_vertex_range(
+                    vertices,
+                    offset_range(parts, part_index),
+                    false,
+                    coordinates,
+                );
             }
             push_part_offset(coordinates, part_offsets);
         }
@@ -446,10 +456,19 @@ fn append_geometry_parts(
             }) else {
                 return;
             };
-            append_polygon_parts(vertices, parts, rings, part_index, coordinates, part_offsets);
+            append_polygon_parts(
+                vertices,
+                parts,
+                rings,
+                part_index,
+                coordinates,
+                part_offsets,
+            );
         }
         GeometryType::MultiPoint => {
-            let Some(geometry_range) = geometry_offsets.and_then(|offsets| offset_range(offsets, feature_index)) else {
+            let Some(geometry_range) =
+                geometry_offsets.and_then(|offsets| offset_range(offsets, feature_index))
+            else {
                 return;
             };
             for index in geometry_range {
@@ -487,7 +506,12 @@ fn append_geometry_parts(
                     let Some(ring_index) = offset(parts, index) else {
                         continue;
                     };
-                    push_vertex_range(vertices, offset_range(rings, ring_index), false, coordinates);
+                    push_vertex_range(
+                        vertices,
+                        offset_range(rings, ring_index),
+                        false,
+                        coordinates,
+                    );
                 } else {
                     push_vertex_range(vertices, offset_range(parts, index), false, coordinates);
                 }
@@ -504,7 +528,14 @@ fn append_geometry_parts(
                 return;
             };
             for part_index in geometry_range {
-                append_polygon_parts(vertices, parts, rings, part_index, coordinates, part_offsets);
+                append_polygon_parts(
+                    vertices,
+                    parts,
+                    rings,
+                    part_index,
+                    coordinates,
+                    part_offsets,
+                );
             }
         }
     }
@@ -538,11 +569,7 @@ fn vertex(vertices: &[i32], index: usize) -> Option<ffi::MltCoordinate> {
     })
 }
 
-fn push_vertex(
-    vertices: &[i32],
-    index: usize,
-    coordinates: &mut Vec<ffi::MltCoordinate>,
-) {
+fn push_vertex(vertices: &[i32], index: usize, coordinates: &mut Vec<ffi::MltCoordinate>) {
     if let Some(coordinate) = vertex(vertices, index) {
         coordinates.push(coordinate);
     }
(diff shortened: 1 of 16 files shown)


rustutils/src/mlt.rs
```diff
@@ -0,0 +1,949 @@
...
+
+    #[derive(Clone, Copy)]
+    struct MltValue {
+        typ: MltValueType,
+        bool_value: bool,
+        i64_value: i64,
+        u64_value: u64,
+        f32_value: f32,
+        f64_value: f64,
+    }
```
CommanderStorm, line 39: A bit sad that there is no simpler way to do this.

rustutils/src/mlt.rs
```diff
@@ -0,0 +1,949 @@
...
+
+    #[derive(Clone, Copy)]
+    struct MltFeatureId {
+        has_id: bool,
+        value: u64,
+    }
```
CommanderStorm, line 45: Interesting that this is not yet implement in cxx

test/tile/vector_tile.test.cpp
```diff
@@ -97,8 +97,8 @@ TEST(VectorTileData, MLTParseResults) {
 
         std::vector<std::string> layerNames = data.layerNames();
 
-        if (!testCase.fastPFOREnabled && testCase.useFastPFOR) {
-            // If fast PFOR is not enabled, the fast PFOR tile should fail to parse
+        if (!testCase.fastPFOREnabled && testCase.useFastPFOR && layerNames.empty()) {
```
CommanderStorm, line 100: If layerNames is empty, how can size be anything but zero? 
Is this still testing something?

src/mbgl/tile/geometry_tile_data.hpp
```diff
@@ -48,15 +45,15 @@ class GeometryCollection : public std::vector<GeometryCoordinates> {
...
+    void setTriangles(std::shared_ptr<const void> owner, std::span<const std::uint32_t> triangles_) {
         triangleOwner = std::move(owner);
         triangles = triangles_;
     }
 
 private:
     GeometryCollection(const GeometryCollection&) = default;
 
-    std::shared_ptr<const mlt::MapLibreTile> triangleOwner;
+    std::shared_ptr<const void> triangleOwner;
```
CommanderStorm, line 56: Is this still save?
What is const void here?
Why do we need to keep it?

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace mbgl {
+namespace {
+
+std::string toStdString(rust::String value) {
+    return {value.data(), value.size()};
+}
```
CommanderStorm, line 25: rust::String implements the explicit converter constructor (not sure what the name is), so you can just std::string(value).
https://cxx.rs/binding/string.html

Also applies to below rust::Str.
Also, Inline afterwards ^^

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+rust::Str toRustStr(const std::string& value) {
+    return {value.data(), value.size()};
+}
+
+struct RustMLTTile {
+    explicit RustMLTTile(rust::Box<rustutils::MltTile> tile_)
+        : tile(std::move(tile_)) {}
+
+    rust::Box<rustutils::MltTile> tile;
```
CommanderStorm, line 35: Do we need to box here? 
What happens if we don't?

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+        case rustutils::MltValueType::I64:
+            return value.i64_value;
+        case rustutils::MltValueType::U64:
+            return value.u64_value;
+        case rustutils::MltValueType::F32:
+            return static_cast<double>(value.f32_value);
+        case rustutils::MltValueType::F64:
+            return value.f64_value;
+        case rustutils::MltValueType::String:
+            return toStdString(getString());
```
CommanderStorm, line 56: Where is getString coming from? It is not imported or defined it seems.

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+    std::optional<Value> getValue(const std::string& key) const override {
+        const auto value = tile->tile->property_value_by_name(layerIndex, featureIndex, toRustStr(key));
+        return convertValue(
+            value, [&] { return tile->tile->property_string_value_by_name(layerIndex, featureIndex, toRustStr(key)); });
+    }
+
+    const PropertyMap& getProperties() const override {
+        if (!properties) {
+            properties.emplace();
```
CommanderStorm, line 126: Curious: why and what exactly does this do here?
I don't see a definition for properties, so this likely is fine, but why are we embracing with nothing if we are already false?

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+        return *properties;
+    }
+
+    FeatureIdentifier getID() const override {
+        const auto id = tile->tile->feature_id(layerIndex, featureIndex);
+        return id.has_id ? mapbox::feature::identifier(id.value) : mapbox::feature::null_value;
+    }
+
+    const GeometryCollection& getGeometries() const override {
+        MLN_TRACE_FUNC();
```
CommanderStorm, line 148: Seems to be the only one being traced. 
Why and when do we trace?

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+class VectorMLTTileData::Impl {
+public:
+    explicit Impl(std::shared_ptr<const std::string> data_)
+        : data(std::move(data_)) {}
+
+    std::unique_ptr<GeometryTileLayer> getLayer(const std::string& name) const {
+        MLN_TRACE_FUNC();
+
+        loadFull();
```
CommanderStorm, line 243: In the rest of the code we are lazy. Here we are eager. Is this intentional?

src/mbgl/tile/vector_mlt_tile_data.rs.cpp
```diff
@@ -0,0 +1,330 @@
...
+    return std::make_unique<VectorMLTTileData>(*this);
+}
+
+std::unique_ptr<GeometryTileLayer> VectorMLTTileData::getLayer(const std::string& name) const {
+    return impl->getLayer(name);
+}
+
+std::vector<std::string> VectorMLTTileData::layerNames() const {
+    return impl->layerNames();
+}
```
CommanderStorm, line 328: Should they not also be in the class? Why are they below the class?

rustutils/src/mlt.rs
```diff
@@ -0,0 +1,949 @@
...
+            result.push(MltLayer {
+                name: layer.name().to_string(),
+                extent: layer.extent().get(),
+                property_names: layer
+                    .iterate_prop_names()
+                    .map(|name| name.to_string())
+                    .collect(),
+                property_values: Vec::new(),
+                ffi_property_values: Vec::new(),
+                part_offsets: vec![0],
```
CommanderStorm, line 211: This will mean we always alloc here, can we make the first index implied 0 in the cpp code or in the decoding instead?

I assume this is reasonably hot

rustutils/src/mlt.rs
```diff
@@ -0,0 +1,949 @@
...
+fn missing_value() -> ffi::MltValue {
+    value(ffi::MltValueType::Missing)
+}
+
+fn null_value() -> ffi::MltValue {
+    value(ffi::MltValueType::Null)
+}
+
+fn value(typ: ffi::MltValueType) -> ffi::MltValue {
+    ffi::MltValue {
```
CommanderStorm, line 626: nit: I'd inline these, derive Default for the last and .. Default::default(), but that is likely fine too.

rustutils/src/mlt.rs
```diff
@@ -0,0 +1,949 @@
...
+            typ: ffi::MltValueType::F32,
+            f32_value: *f32_value,
+            ..value(ffi::MltValueType::F32)
+        },
+        MltPropertyValue::F64(f64_value) => ffi::MltValue {
+            typ: ffi::MltValueType::F64,
+            f64_value: *f64_value,
+            ..value(ffi::MltValueType::F64)
+        },
+        MltPropertyValue::String(_) => value(ffi::MltValueType::String),
```
CommanderStorm, line 671: Is this what we want? Where is the string stored?

rustutils/src/mlt.rs
```diff
@@ -0,0 +1,949 @@
...
+        };
+        let Some(feature) = layer.features.get(feature_index) else {
+            return 0;
+        };
+        let offset_index = feature.part_offset_start + part_index;
+        let Some(start) = layer.part_offsets.get(offset_index) else {
+            return 0;
+        };
+        let Some(end) = layer.part_offsets.get(offset_index + 1) else {
+            return 0;
```
CommanderStorm, line 860: Not 💯 sure if this compiles, but my intuition would be to do this.

```suggestion
        let Some([start,end]) = layer.part_offsets.get(offset_index..offset_index +1) else {
            return 0;
        };
```

CommanderStorm (approved):
Mostly LGTM, a few aspects are a bit weird though. 

How is perf compared to cpp?

louwers:
> How is perf compared to cpp?

About the same.

Regarding your Rust comments, this was 100% vibe coded as a soft of proof of concept. I don't really know Rust.

nyurik:
@louwers any progress on this?  BTW,,,, you may want to consider these two approaches:
* Zngur - https://hkalbasi.github.io/zngur/
* Crubit (Google) - https://github.com/google/crubit

PR #4446, opened 2026-07-31
Changes to the custom Metal layer to allow more flexible rendering

sjg-wdw:
I added our custom weather renderer to MapLibre and it needed to build a whole set of encoders on its own (blit, compute, render) before the render encoder is created for MapLibre.  

The solution I settled on was a preDraw callback, empty by default, which provides the Metal command encoder and runs before the normal render encoder is created.

Also needed access to the paint parameters within the layers to get the command encoder down to the custom layer host.

This did all end up working for the multi-stage rendering we do and shouldn't change how other Metal plugins work.


louwers:
@copilot Fix the compile error in ios-build-cmake.

sjg-wdw:
We seem to be down to things that were already there but are suddenly being flagged by the tests.

louwers (approved):

louwers:
Thanks Steve!

sjg-wdw:
Thank you!  It was an interesting experience to go through the details myself.

Issue #3396, opened 2025-04-15
Additional Hill-shade Options


NathanMOlson:
[MapLibre GL JS issue](https://github.com/maplibre/maplibre-gl-js/issues/5665)

# User Story
As a map maker/owner, I would like to have multiple options when choosing the hill-shade/relief-shade type that MapLibre renders using my terrain-RGB tiles so that I can deliver high-quality cartographic products.

# Rationale
Currently, only a standard hill-shade is available which is a great general hillshade.

By introducing three additional algorithms for rendering hill-shades from terrain RGB tiles, MapLibre will be increasingly useful for advanced cartography. GDAL currently allows 'Combined', 'Multidirectional' and 'Igor', among others. Choosing from these three options when using terrain RGB tiles would allow map makers to select the right theme and feel for their project.

Further explanation of hill-shade types:

[Combined](https://gdal.org/en/stable/programs/gdaldem.html#cmdoption-combined) combined shading, a combination of slope and oblique shading.
[Multidirectional](https://gdal.org/en/stable/programs/gdaldem.html#cmdoption-multidirectional) multidirectional shading, a combination of hillshading illuminated from 225 deg, 270 deg, 315 deg, and 360 deg azimuth. Multidirectional hill-shading applies the formula of http://pubs.usgs.gov/of/1992/of92-422/of92-422.pdf.
[Igor](https://gdal.org/en/stable/programs/gdaldem.html#cmdoption-igor) shading, which tries to minimize effects on other map features beneath. Igor's hill-shading uses a formula from Maperitive http://maperitive.net/docs/Commands/GenerateReliefImageIgor.html.

![Image](https://github.com/user-attachments/assets/973536a9-b795-40d1-a416-274d1dd34fad)

![Image](https://github.com/user-attachments/assets/b419f9d0-5f25-4954-a70c-2b10800b93e2)

![Image](https://github.com/user-attachments/assets/c3af2061-0939-470c-8c3e-d99d513f1153)

GDAL’s alternate hillshade algorithms — from top to bottom: multidirectional, combined, and igor. Images derived from the [USGS National Map](https://apps.nationalmap.gov/downloader/).

acalcutt:
Just an FYI. When I try to use a style with 'hillshade-method' like 
```
    {
      "id": "hillshade",
      "type": "hillshade",
      "layout": {
          "visibility": "visible"
      },
      "paint": {
        "hillshade-method": "standard",
[...]

PR #4454, opened 2026-08-03
fix(darwin): accept a scalar for colorArray/numberArray style properties

### LINKED ISSUE #4453: iOS: setting a scalar on colorArray/numberArray properties (hillshade colors, illumination direction) crashes with -[UIDeviceRGBColor count] since 6.24.0
### Summary

Since ios-v6.24.0, setting `hillshadeShadowColor` / `hillshadeHighlightColor` to a **single** color — or `hillshadeIlluminationDirection` / `hillshadeIlluminationAltitude` to a single number — crashes the app with

```
NSInvalidArgumentException: -[UIDeviceRGBColor count]: unrecognized selector sent to instance 0x11750b700
  CoreFoundation      __exceptionPreprocess
  libobjc.A           objc_exception_throw
  CoreFoundation      -[NSObject(NSObject) doesNotRecognizeSelector:]
  CoreFoundation      ___forwarding___
  MapLibre            <-[MLNHillshadeStyleLayer setHillshadeShadowColor:]>
  ...
```

This is a regression: the same code works on ios-v6.23.0 and it works on Android with the same style values.

### Cause

In 6.24.0 the style spec retyped these properties (multi-directional hillshade, #3396):

| property | ≤6.23 | ≥6.24 |
| --- | --- | --- |
| `hillshade-shadow-color` | `color` | `colorArray` |
| `hillshade-highlight-color` | `color` | `colorArray` |
| `hillshade-illumination-direction` | `number` | `numberArray` |
| `hillshade-illumination-altitude` | – | `numberArray` |

so the generated setters now instantiate `MLNStyleValueTransformer<std::vector<mbgl::Color>, NSArray<MLNColor *> *, mbgl::Color>`. For a constant-value expression, `toPropertyValue` takes the fast path straight into the array overload, which sends `count` to the raw value without a type check — `platform/darwin/src/MLNStyleValue_Private.h`:

```objc
  // Array
  void getMBGLValue(ObjCType rawValue, std::vector<MBGLElement> &mbglValue) {
    mbglValue.reserve(rawValue.count);   // rawValue is a UIColor / NSString here
```

A scalar is legal for these properties everywhere else:

* the spec itself keeps scalar defaults for them (`"type": "colorArray", "default": "#000000"`),
* `mbgl::style::conversion` explicitly accepts a scalar — `src/mbgl/style/conversion/constant.cpp`:

  ```cpp
  std::optional<std::vector<Color>> Converter<std::vector<Color>>::operator()(...) const {
      if (!isArray(value)) {
          // Try single color
          auto color = convert<Color>(value, error);
          if (!color) return std::nullopt;
          return std::vector<Color>{*color};
      }
  ```

  (same for `std::vector<float>`), which is why the JSON/JNI paths on Android and the style-JSON path are unaffected — only the darwin ObjC fast path is,
* MapLibre GL JS accepts a scalar as well.

`NSCAssert`s in that area are compiled out in the released framework, so instead of an assertion the app takes an `NSException` from native code. For Flutter/React Native wrappers that call the setter through FFI this is fatal and uncatchable: it terminates the process.

### Steps to reproduce

```objc
MLNHillshadeStyleLayer *layer = [[MLNHillshadeStyleLayer alloc] initWithIdentifier:@"hillshade"
                                                                           source:demSource];
layer.hillshadeShadowColor = [NSExpression expressionForConstantValue:UIColor.redColor];  // 💥
layer.hillshadeIlluminationDirection = [NSExpression expressionForConstantValue:@335];    // 💥
```

Wrapping the value in an `NSArray` works, so this is purely about scalar tolerance.

### Expected

A scalar should be accepted and treated as a one-element array (as `mbgl::style::conversion` does), or at worst raise a descriptive `NSInvalidArgumentException` instead of `doesNotRecognizeSelector`.

### Impact

Real-world: 83 production crashes across 4 users in our Flutter app (`flutter-maplibre` 0.3.5 → MapLibre iOS 6.25.x) simply for opening a screen whose hillshade layer sets a shadow color. Any app that carried a hillshade color over from ≤6.23 crashes on launch of that map.

### Environment

* MapLibre iOS 6.24.0 – 6.25.x (6.23.0 unaffected)
* iOS 26.6, iPhone17,5 / iPad11,2, release builds
* via `flutter-maplibre` 0.3.5 (`maplibre_ios`), which builds the constant expression as `NSExpression(forConstantValue: UIColor)`

PR with a fix (scalar → one-element vector, mirroring the core converter) plus a regression test follows.

cmaart:
Fixes #4453.

### Problem

Since the style-spec change in 6.24.0 (`hillshade-shadow-color` / `hillshade-highlight-color` → `colorArray`, `hillshade-illumination-direction` / `-altitude` → `numberArray`, #3396), the generated setters instantiate `MLNStyleValueTransformer<std::vector<mbgl::Color>, NSArray<MLNColor *> *, mbgl::Color>`. A constant-value expression takes the fast path in `toPropertyValue` straight into the array overload, which sends `-count` to the raw value with no type check:

```objc
  // Array
  void getMBGLValue(ObjCType rawValue, std::vector<MBGLElement> &mbglValue) {
    mbglValue.reserve(rawValue.count);   // rawValue is a UIColor / NSNumber here
```

So `layer.hillshadeShadowColor = [NSExpression expressionForConstantValue:UIColor.redColor]` — code that worked in 6.23.0 and still works on Android — dies with

```
NSInvalidArgumentException: -[UIDeviceRGBColor count]: unrecognized selector sent to instance
```

The `NSCAssert`s nearby are compiled out in the released framework, so shipping apps get an `NSException` from native code rather than an assertion. Through an FFI wrapper (Flutter, React Native) that is uncatchable and terminates the process.

### Fix

Treat a scalar as a one-element vector, which is what everything else already does:

* the spec keeps scalar defaults for these properties (`"type": "colorArray", "default": "#000000"`),
* `Converter<std::vector<Color>>` / `Converter<std::vector<float>>` in `src/mbgl/style/conversion/constant.cpp` explicitly accept a scalar (that is why the JSON and JNI paths never hit this),
* MapLibre GL JS accepts a scalar too.

Arrays keep the existing code path, so no behaviour changes for callers that already pass an `NSArray`.

### Tests

New `platform/darwin/test/MLNColorArrayTests.mm` (picked up by the existing `test/*.mm` glob in `platform/darwin/BUILD.bazel`):

* `colorArray` property set from a single `MLNColor` → one-element color array
* `colorArray` property set from an `NSArray` of two colors → unchanged behaviour
* `numberArray` property set from a single `NSNumber` → one-element float array

Disclosure: I develop on Windows and cannot build or run the darwin test suite locally, so these two changes have not been compiled here — they need your CI. The diagnosis behind them is verified against 6.19.0/6.23.0/6.24.0/6.25.0 sources and against 83 production crashes in our app.

### DIFF
--- a/platform/darwin/src/MLNStyleValue_Private.h
+++ b/platform/darwin/src/MLNStyleValue_Private.h
@@ -273,6 +273,22 @@ private:  // Private utilities for converting from mgl to mbgl values
 
   // Array
   void getMBGLValue(ObjCType rawValue, std::vector<MBGLElement> &mbglValue) {
+    if (![rawValue isKindOfClass:[NSArray class]]) {
+      // The style spec allows a scalar wherever a numberArray or colorArray is expected (it
+      // even keeps scalar defaults for those properties), and Converter<std::vector<T>> in
+      // mbgl::style::conversion turns a scalar into a one-element vector. Do the same here
+      // instead of sending -count to a color or a string, which raises
+      // doesNotRecognizeSelector and terminates the app.
+      id scalarValue = rawValue;
+      if ([scalarValue isKindOfClass:[NSExpression class]] &&
+          [scalarValue expressionType] == NSConstantValueExpressionType) {
+        scalarValue = [scalarValue constantValue];
+      }
+      MBGLElement mbglElement;
+      getMBGLValue(scalarValue, mbglElement);
+      mbglValue.push_back(mbglElement);
+      return;
+    }
     mbglValue.reserve(rawValue.count);
     for (id obj in rawValue) {
       id constantObject = obj;
--- a/platform/darwin/test/MLNColorArrayTests.mm
+++ b/platform/darwin/test/MLNColorArrayTests.mm
@@ -1,0 +1,64 @@
+#import <Mapbox.h>
+#import <XCTest/XCTest.h>
+
+#import "MLNStyleLayer_Private.h"
+
+#include <mbgl/style/layers/hillshade_layer.hpp>
+
+/// Regression tests for https://github.com/maplibre/maplibre-native/issues/4453.
+///
+/// `colorArray` and `numberArray` properties must accept a scalar, the same way the style spec
+/// (which keeps scalar defaults for them), `mbgl::style::conversion` and MapLibre GL JS do.
+@interface MLNColorArrayTests : XCTestCase
+@end
+
+@implementation MLNColorArrayTests
+
+- (MLNHillshadeStyleLayer *)hillshadeLayer {
+  MLNPointFeature *feature = [[MLNPointFeature alloc] init];
+  MLNShapeSource *source = [[MLNShapeSource alloc] initWithIdentifier:@"sourceID"
+                                                                shape:feature
+                                                              options:nil];
+  return [[MLNHillshadeStyleLayer alloc] initWithIdentifier:@"layerID" source:source];
+}
+
+- (void)testColorArrayPropertyAcceptsSingleColor {
+  MLNHillshadeStyleLayer *layer = [self hillshadeLayer];
+  auto rawLayer = static_cast<mbgl::style::HillshadeLayer *>(layer.rawLayer);
+
+  layer.hillshadeShadowColor = [NSExpression expressionForConstantValue:[MLNColor redColor]];
+
+  mbgl::style::PropertyValue<std::vector<mbgl::Color>> propertyValue =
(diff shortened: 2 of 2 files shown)


cmaart:
Closing this — I opened it prematurely without the repo owner's sign-off on my side. The analysis in #4453 stands; feel free to take the patch from there if it is useful.

PR #4458, opened 2026-08-04
core: Add error reporting for corrupted symbols

adrian-cojocaru:
This PR adds `onSymbolError` to the list of observable events (logged by default in the action journal). This is triggered when a corrupted symbol is detected (see detection logic added in https://github.com/maplibre/maplibre-native/pull/2744).

AI disclosure: used Claude Code + Opus 5 to propagate the event bindings and delegates.


louwers:
Are we still hoping to find the cause of this at some point?

adrian-cojocaru:
@louwers Removing the `MLN_SYMBOL_GUARDS` is a performance increase (both CPU and memory).

louwers (commented):
Should we mark the APIs as unstable if we're going to remove this in the future?

sjg-wdw:
> Should we mark the APIs as unstable if we're going to remove this in the future?

I love the optimism that we'll find this ancient bug and have to remove the callbacks.

louwers (approved):

PR #4383, opened 2026-07-07
Reduce symbol buffers memory by using instancing (Core optimization)

alexcristici:
Before:

symbolBucketsMem: 141_748_976 bytes
symbolBucketsCount: 79
zoom: 16.4259

After:

symbolBucketsMem: 56_696_400 bytes
symbolBucketsCount: 79
zoom: 16.4259

Improvement: 2.5x

(with symbol instances +72 MB)

PR #4460, opened 2026-08-04
feat(core): Implement Rendered Feature Introspection

TimSylvester:
These changes allow a map instance to be configured at startup to track rendered features, allowing programmatic access to which features were rendered by each style layer and in what region of the framebuffer.

The effect is similar to `queryRenderedFeature`, but instead of getting the set of features relevant to a region, returns the region for features, allowing an application to determine which layer(s) rendered a feature and where, how many features were rendered by a specific layer (e.g., a highlight layer), or whether a feature was entirely on-screen in a particular view.

This is currently intended primarily to assist with testing of host applications.  Some examples are included in the Android Espresso tests.

This could be extended to provide "screen reader" type functionality, but that is not in scope here.  Notably that would probably entail some style spec changes, describing how to construct an alt label from the properties of a feature.


TimSylvester:
argh #4511

louwers:
Sorry.

TimSylvester:
Updated to account for #4383, and ready to merge.

TimSylvester:
@louwers Can you review this, or nominate someone else?  I'd like to get it off the list of things to pay attention to.

src/mln/map/map.cpp
```diff
@@ -604,4 +607,57 @@ const std::unique_ptr<util::ActionJournal>& Map::getActionJournal() {
     return impl->actionJournal;
 }
 
+void Map::getRenderedFeatures(const std::optional<std::string>& featureId,
```
louwers, line 610: Wouldn't it be a bit cleaner if this returned an iterator?

TimSylvester: It's a search/filter operation, so it would have to be a custom iterator class.  In another language I might use a continuation-style iterator, but that's impractical here.

louwers (approved):
LGTM but definitely an 'enterprise feature' right now. Quite heavy for something that is only used for testing. I think adding an example of how to use it to the Android and/or iOS documentation would be a good idea, so we're sure that the people that need something like this know it exists. That's something I can pick up when this is merged as well.

The potential for future accesibility use cases is quite interesting I think. Also sounds like something you could apply for some kind of grand for.

PR #4467, opened 2026-08-06
accept libpng with different release/patch number

xabbu42:
I did not find a clear official statement but the definition of PNG_LIBPNG_VER_SHAREDLIB and the used soname at least imply that this versions should be abi compatible. We ran into this because an update of the stable opensuse leap increased the release/patch number and we do not automatically recompile on updates.

There is some relevant prior discussion in https://github.com/maplibre/maplibre-native/issues/3685.


louwers (approved):
Makes sense.

PR #4475, opened 2026-08-11
Add cross-platform C plugin API

louwers:
This PR implements a cross-platform plugin C API for MapLibre Native. More like a proof of concept than anything else.

We might want to expose a C++ API as well, but I don't think a C++ API is a viable option for a thriving plugin ecosystem since this will require rebuilding all plugins for each release. And probably you would need to compile MapLibre Native and the plugins with exactly the same toolchain.

My idea is that we make it as easy as possible for people to start writing/vibe coding and distributing plugins by providing a [template repository](https://github.com/louwers/maplibre-native-plugin-template/) with all the build tooling and release workflows set up for them.

Here you can see an example how a plugin that adds [shadows](https://github.com/louwers/maplibre-native-plugin-template/tree/main/plugins/fill-extrusion-shadows) to fill extrusions is registered.

Releated Discussion: https://github.com/maplibre/maplibre-native/discussions/4321


stefankarschti:
What would be interesting as a proof of concept is to be able to define a proper C interface and evict an entire layer type implementation, then dynamically link against its newly created library. 
This includes:
- the renderer layer code
- render source 
- bucket 
- tweaker
- style layer 
- shader code / multiple backends

We could pick the RasterLayer as a first candidate.

Expanding on this idea, we could imagine a slim renderer core with a C plugin architecture that could abstract away style parsing, tile pyramid, tile loaders, source types and more. That plus a set of standard layer types, default stype parser, MVT and MLT decoders delivered as standard side cars.
Maintenance would be easier and the community could benefit by writing and swapping in they own plugins much easier without recompilation of the core.

platform/android/MapLibrePluginApi/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+            afterEvaluate { from(components["release"]) }
+            pom {
+                name.set("MapLibre Android Plugin API")
+                description.set("Pure-C native plugin ABI and Android registry contract for MapLibre Native")
+                url.set(project.extra["mapLibreArtifactUrl"].toString())
+            }
+        }
+    }
+    repositories {
+        val target = providers.gradleProperty("reposiliteUrl")
```
louwers, line 81: To be cleaned up.

PR #4479, opened 2026-08-13
macOS: report the drawable size to the renderer

bouk:
The renderable size stayed zero, so every frame computed a zero-sized scissor rectangle that could clip the whole map away.

This fixed a flickering issue I had on macOS where many frames would be rendered empty while panning/zooming

Screenshot 2026-08-13 at 15 27 47
Screenshot 2026-08-13 at 15 27 49

Disclosure: this PR was authored using Claude Fable 5


louwers (approved):

louwers:
Glad to see someone playing around with the macOS platform again.

PR #4480, opened 2026-08-13
macOS: render the Metal map on demand

bouk:
The MTKView ran unpaused, re-rendering an unchanged map at the display refresh rate and burning CPU and GPU while the map sat idle. Pause the view and mark it through a display() hook, as iOS does, so an idle map doesn't use any resources.

Drawing now happens inside the display cycle, where an asynchronous present races the Core Animation commit and can flash a cleared, not yet drawn drawable. Commit, wait and present explicitly, as iOS does with presentsWithTransaction.

Disclosure: This PR was authored using Claude Fable 5

### DIFF
--- a/platform/macos/src/MLNMapView+Impl.h
+++ b/platform/macos/src/MLNMapView+Impl.h
@@ -22,6 +22,8 @@ public:
 
   virtual CGLContextObj getCGLContextObj() { return nullptr; }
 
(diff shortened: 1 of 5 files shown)


louwers (approved):

louwers:
Thanks for giving macOS some love!

PR #4490, opened 2026-08-15
fix(core): canonicalize default MapLibre glyph URLs

### LINKED ISSUE #4403: Uncaught std::regex_error in createTokenMap aborts the app during offline pack download when style glyphs are hosted on demotiles.maplibre.org
## Platform / version

- iOS, MapLibre Native `ios-v6.27.0` (via `@maplibre/maplibre-react-native` 11.3.4), observed in production on iOS 26.5.2
- The affected code is shared (`src/mbgl/util/mapbox.cpp`), so Android should reproduce as well

## Steps to reproduce

1. Use any style whose `glyphs` points at the default MapLibre tile server host, e.g.
   `https://demotiles.maplibre.org/font/{fontstack}/{range}.pbf`
   (tiles can come from anywhere — only the glyph URL matters; sources in our style are on a different host entirely)
2. Create an offline pack from that style (`OfflineDownload` / `MLNOfflineStorage addPack…`)
3. As soon as the style resource response is processed, the process aborts with SIGABRT

100% reproducible.

## Root cause

`OfflineDownload::activateDownload()`'s style response callback calls
`mbgl::util::mapbox::canonicalizeGlyphURL(tileServerOptions, url)`. Because the glyph URL matches
the default `TileServerOptions::MapLibreConfiguration()` `baseURL` (`https://demotiles.maplibre.org`),
`isNormalizedSourceURL(...)` returns true and `createTokenMap(tileServerOptions.glyphsTemplate(), url)` runs.

`createTokenMap` (`src/mbgl/util/mapbox.cpp`) builds a `std::regex` from the glyphs template after
substituting **only** the `{domain|path|directory|filename|extension}` tokens:

```cpp
std::regex tokenPattern(R"(\{domain\}|\{path\}|\{directory\}|\{filename\}|\{extension\})");
std::string templatePattern = std::regex_replace(urlTemplate, tokenPattern, "(.+)", std::regex_constants::match_any);
std::regex r2(templatePattern);   // <-- throws
```

The default MapLibre glyphs template is `/font/{fontstack}/{start}-{end}.pbf`, so `{fontstack}`,
`{start}` and `{end}` survive into `templatePattern`. `{f` etc. are invalid ECMAScript brace
quantifiers, so the `std::regex` constructor throws `regex_error`. Nothing on the
`DatabaseFileSource` thread catches it → `std::terminate` → SIGABRT.

(`isNormalizedSourceURL` happens to survive because its token pattern `\{.+\}` greedily swallows
`{fontstack}/{start}-{end}` in one replacement — which is why execution reaches `createTokenMap`
at all.)

## Symbolicated crash (thread `org.maplibre.mbgl.DatabaseFileSource`)

```
libc++abi.dylib  __cxa_throw
MapLibre  std::__1::__throw_regex_error[abi:ue170006]<(std::__1::regex_constants::error_type)8>()  (regex:0)
MapLibre  std::__1::basic_regex<...>::__parse_ERE_dupl_symbol<...>(...)  (regex:0)
MapLibre  std::__1::basic_regex<...>::__parse_ecma_exp<...>(...)  (regex:4283)
MapLibre  std::__1::basic_regex<...>::basic_regex[abi:ue170006]<...>(std::string const&, syntax_option_type)  (regex:2719)
MapLibre  mbgl::util::mapbox::createTokenMap(std::string const&, std::string const&)  (mapbox.cpp:84)
MapLibre  mbgl::util::mapbox::canonicalizeGlyphURL(mbgl::TileServerOptions const&, std::string const&)  (mapbox.cpp:392)
MapLibre  mbgl::OfflineDownload::activateDownload()::$_6 response callback  (function.h:364)
MapLibre  mbgl::OfflineDownload::ensureResource(...) response lambda  (function.h:364)
MapLibre  mbgl::FileSourceRequest::setResponse(mbgl::Response const&)  (file_source_request.cpp:29)
MapLibre  mbgl::Mailbox::receive()  (mailbox.cpp:149)
MapLibre  mbgl::util::RunLoop::process()  (run_loop.hpp:133)
MapLibre  mbgl::util::Thread<mbgl::DatabaseFileSourceThread> thread body  (thread.hpp:70)
```

## Suggested fix

Any (ideally both) of:

1. In `createTokenMap` (and `isNormalizedSourceURL`), regex-escape the non-token parts of the
   template — or at least strip/escape leftover `{...}` tokens — before constructing `std::regex`,
   so arbitrary `glyphsTemplate` values can't produce an invalid pattern.
2. Wrap the canonicalization calls in `OfflineDownload` in `try/catch` and fall back to the raw
   URL, so a canonicalization failure degrades gracefully instead of terminating the host app.

## Related

- #177 reported offline downloads crashing with demotiles in 2022 and already suspected the regex
  validation in `mapbox.cpp`; it was closed as "demotiles doesn't support offline" after the
  maintainer hit a different symptom (304 handling). The uncaught-throw path reported here is
  still present in `ios-v6.27.0`, and it triggers even when only the *glyphs* live on
  demotiles.maplibre.org (a common configuration, since demotiles is MapLibre's public font
  endpoint) while all tile sources are elsewhere.
- #4295 is the same failure class on a different code path (uncaught exception on the offline
  path terminating the host app).

## Workaround for anyone hitting this

Host your glyphs anywhere that is **not** `demotiles.maplibre.org` (we moved to OpenFreeMap's
font endpoint); the canonicalizer then leaves the URL untouched and the download completes.

fallenmi:
## Summary

- change the default MapLibre glyph template to the existing `{path}` grammar used by URL canonicalization
- preserve `{fontstack}/{range}` in the canonical `maplibre://fonts/...` URL
- add a regression for the demotiles glyph URL reported in #4403

## Root cause

`createTokenMap` expands only its generic URL placeholders. The default MapLibre glyph template contained `{fontstack}`, `{start}`, and `{end}`, so those braces reached `std::regex` and could raise `std::regex_error` while an offline pack was being prepared.

This change is intentionally scoped to the default MapLibre/demotiles configuration. It does not broaden the generic template parser for arbitrary custom glyph templates.

Fixes #4403.

## Validation

- confirmed the new `MapLibre.CanonicalURL` regression throws `std::regex_error` before the source change
- passed the focused regression 100 consecutive times after the change
- passed 22 adjacent Mapbox, MapLibre, MapTiler, and `TileServerOptions` URL tests
- passed ClangFormat 20.1.8, targeted pre-commit whitespace/line-ending hooks, and `git diff --check`

The full project suite was not completed locally; remote CI is pending.

## AI assistance disclosure

OpenAI Codex (GPT-5) assisted with investigation, test drafting, implementation, validation, and this draft description. The contributor must personally review every changed line and approve or rewrite this description before marking the PR ready for review.

### DIFF
--- a/src/mbgl/util/tile_server_options.cpp
+++ b/src/mbgl/util/tile_server_options.cpp
@@ -225,7 +225,7 @@ TileServerOptions TileServerOptions::MapLibreConfiguration() {
                                     .withSourceTemplate("/tiles/{domain}.json", "", {})
                                     .withStyleTemplate("{path}.json", "maps", {})
                                     .withSpritesTemplate("/{path}/sprite{scale}.{format}", "", {})
-                                    .withGlyphsTemplate("/font/{fontstack}/{start}-{end}.pbf", "fonts", {})
+                                    .withGlyphsTemplate("/font{path}", "fonts", {})
                                     .withTileTemplate("/{path}", "tiles", {})
                                     .withDefaultStyles(styles)
                                     .withDefaultStyle("Basic")
--- a/test/util/mapbox.test.cpp
+++ b/test/util/mapbox.test.cpp
@@ -313,6 +313,9 @@ TEST(MapLibre, CanonicalURL) {
     EXPECT_EQ("https://demotiles.maplibre.org/font/{fontstack}/{start}-{end}.pbf",
               mln::util::mapbox::normalizeGlyphsURL(
                   mapboxFixture::mapLibreTileServerOptions, "maplibre://fonts/{fontstack}/{start}-{end}.pbf", ""));
+    EXPECT_EQ("maplibre://fonts/{fontstack}/{range}.pbf",
+              mln::util::mapbox::canonicalizeGlyphURL(mapboxFixture::mapLibreTileServerOptions,
+                                                      "https://demotiles.maplibre.org/font/{fontstack}/{range}.pbf"));
 
     EXPECT_EQ("maplibre://tiles/tiles/{z}/{x}/{y}.pbf",
               mln::util::mapbox::canonicalizeTileURL(mapboxFixture::mapLibreTileServerOptions,


miakh:
Confirmed this fixes josxha/flutter-maplibre#535: I reproduced the old `std::regex_error` and verified the merged fix plus 22 adjacent URL tests.

PR #4505, opened 2026-08-19
core: update terrain3d branch to main with namespace changes

acalcutt:
Merge main up to just before the namespace rename (#4487) for a cleaner merge,
then merge in that rename, then apply it to this branch's own files that #4487
never saw.

Also syncs shaders/symbol_*.vertex.glsl with their generated headers - the
terrain occlusion fix had only been applied to the generated gl/symbol_*.hpp,
so re-running generate_shader_code.mjs silently reverted it. The sources were
gating calculate_visibility on u_dem_enabled (elevation) instead of
u_depth_enabled, so labels could show through ridges.

merge rather than squash 

AI Used: Claude AI (Opus)


louwers (approved):

acalcutt:
@louwers it doesn't let me do a regular merge, only a squash and merge. is tht an option somewhere? i wanted to keep the history

image

louwers:
@acalcutt You should be able to push manually to that branch without a PR.

PR #4516, opened 2026-08-24
feat(core): js-parity global-state expression support

### LINKED ISSUE #3302: Support global map state
This is a placeholder for global map state implementation. See: https://github.com/maplibre/maplibre-gl-js/issues/4964

NEKOYASAN:
resolve #3302 

## Summary
- add the [global-state](https://maplibre.org/maplibre-style-spec/expressions/#global-state) expression and the root [state](https://maplibre.org/maplibre-style-spec/root/#state) property, with the runtime API style::Style::setGlobalStateProperty / getGlobalState
- track the state keys referenced by each expression so a change only re-evaluates and relayouts the layers and sources referencing a changed key (parity with gl-js getGlobalStateRefs)
- align failed zoom-only property expression evaluation with GL JS: evaluation failures now use the property's style-spec default instead of value-initializing its type. This applies to all zoom-only expressions, not only `global-state` expressions
- reject global-state updates until the style JSON has loaded, matching GL JS, and expose `style::Style::isLoaded` so platform bindings can distinguish this point from full resource loading
- and platform apis 

## Verification
- `cmake --build build-macos-metal --target mbgl-test-runner` and the style/expression suites, including the `new GlobalStateExpression.*`, `Style.GlobalState*`, `Style.Visibility*`, and `Layer.SymbolLayerOverrides` tests
- Android: `:MapLibreAndroid:testOpenglDebugUnitTest` plus instrumentation tests on an emulator (RuntimeStyleTests, MapSnapshotterTest), arm64 native build
- iOS: `bazel test //platform/ios/test:ios_test --//:renderer=metal` (MLNStyleTests, MLNExpressionTests)
- Node: built the addon locally and ran the tape tests (test/js/global_state.test.js, map.test.js)

## Note
The code in this PR was written with the help of Claude (Fable 5), under my instruction and review.


louwers (commented):
Looked through it, looks quite good!

Let's port the render tests from MapLiibre GL JS. https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/render/tests/global-state

NEKOYASAN:
@louwers thank you! 
I ported the rendering test from GL JS!
I also fixed an issue with how unspecified (default) values were being handled.

image

louwers:
Nice!

NEKOYASAN:
It looks like CI was failing because I had copied v8.json directly from @maplibre/maplibre-gl-style-spec. I’ve now made a few Native-specific adjustments to it.
I’m not entirely sure this is the ideal approach, but I found several other PRs that handled similar style-spec synchronization issues this way, so I followed that precedent for now.

NEKOYASAN:
I'm truly sorry that this turned into such a huge PR...

Based on the comments below, I have added tests and confirmed that everything is working as expected.
https://github.com/maplibre/maplibre-style-spec/issues/886#issuecomment-5409389686

vendor/expected-lite
johncarmack1984, line 1: This moves the submodule off the commit main is on (3583e95 to 95b9cb0), and nothing in the diff uses new expected-lite API, so I think it came along by accident. `git checkout upstream/main -- vendor/expected-lite` puts it back.

src/mln/tile/tile.hpp
```diff
@@ -83,6 +83,11 @@ class Tile : public TileLoaderObserver {
     virtual void setLayers(const std::vector<Immutable<style::LayerProperties>>&) {}
     virtual void setMask(TileMask&&) {}
 
+    // Notifies this tile of the style's current global state, used to
+    // evaluate "global-state" expressions during layout. Must be called
+    // before setLayers() to take effect for the subsequent re-layout.
+    virtual void setGlobalState(const std::shared_ptr<const GlobalStateMap>&) {}
```
johncarmack1984: The worker's `setLayers` already takes the state as an argument. Could `Tile::setLayers` take it the same way? That would drop the two `setGlobalState` calls in TilePyramid and the "must be called before setLayers" rule in the comment.

include/mln/renderer/query.hpp
```diff
@@ -22,6 +24,10 @@ class RenderedQueryOptions {
     std::optional<std::vector<std::string>> layerIDs;
 
     std::optional<style::Filter> filter;
+
+    /// The style's current global state, injected by the renderer so that
+    /// `global-state` expressions in the filter can be evaluated.
+    std::shared_ptr<const GlobalStateMap> globalState;
```
johncarmack1984: RenderOrchestrator copies the options and overwrites this, so a value a caller sets here never gets used. Does it need to be on the public struct, or could the orchestrator hand it to FeatureIndex and the tiles' `querySourceFeatures` next to the options?

src/mln/renderer/data_driven_property_evaluator.hpp
```diff
@@ -24,18 +24,31 @@ class DataDrivenPropertyEvaluator {
...
+            if (needsFeature) {
                 auto returnExpression = expression;
                 returnExpression.setUseIntegerZoom(true);
+                returnExpression.captureGlobalState(parameters.globalState);
                 return ResultType(returnExpression);
             }
-            return ResultType(expression.evaluate(std::floor(parameters.z)));
+            return ResultType(expression.evaluate(style::expression::EvaluationContext(std::floor(parameters.z))
+                                                      .withGlobalState(parameters.globalState.get()),
+                                                  defaultValue));
```
johncarmack1984, line 42: This changes the fallback for every zoom-only expression that fails to evaluate, not only global-state ones. Previously `T()`, now the property default. That's what GL JS does too, so it's correct for the design, it's just not visible from the title or the body. A sentence there would help whoever bisects to it later.

src/mln/style/style_impl.cpp
```diff
@@ -105,6 +107,9 @@ void Style::Impl::parse(const std::string& json_) {
 
     transitionOptions = parser.transition;
 
+    globalStateDefaults = parser.globalStateDefaults;
+    globalState = std::make_shared<const GlobalStateMap>(globalStateDefaults);
```
johncarmack1984, line 111: A `setGlobalStateProperty` call that lands before the style has loaded gets replaced here by the defaults, silently. GL JS throws in that case (`_checkLoaded`), and the Kotlin snapshotter grew an `isStyleLoaded` gate to step around it. Would you rather core warn or reject before load, or keep runtime-set keys across the parse? Keeping them would let the Kotlin gate go. Either way `Style::isLoaded()` is new public API and isn't in the body or changelog yet.

johncarmack1984 (commented):
Saw Bart asking for reviews on this in the newsletter thread, thought I'd pitch in.

Read the core and platform halves against GL JS's implementation (style.ts, the spec's global_state.ts and visibility.ts). The semantics line up as far as I can trace. Null when unset, defaults from `state`, null resets to the default, re-evaluation scoped to the keys that changed. Filters in the query paths, format overrides and the color ramps too.

For anyone looking at the two red checks: those are the Android render-test legs from a `workflow_run` on main at be3f03b8, which failed once and passed on the rerun. Every run on this PR's own head (fc32f1f) is green.

Looked into @HarelM's note about porting the rest of the GL JS global-state tests: 59 of the 69 are in here. The other 10 are sky, projection, and the two `*-layer-opacity` properties, none of which exist in native yet. Should stubs or `// TODO` comments go in to keep parity, or something else?

vendor/expected-lite
NEKOYASAN: Thank you for the review.
I will fix this right away. I will also check the other points raised and address them. Please give me a little time.

NEKOYASAN: thank you! fixed at https://github.com/maplibre/maplibre-native/pull/4516/commits/b8969dd17496acf914832931dd79a61e653d05b4

src/mln/tile/tile.hpp
NEKOYASAN: You're completely right, so I addressed this in https://github.com/maplibre/maplibre-native/pull/4516/commits/86ab6e6eeb66d40fa09594fb9166c88f89cad8fb.
Along with that, I also removed setGlobalState itself since it's not needed.

include/mln/renderer/query.hpp
NEKOYASAN: Sorry, you're right. I removed `globalState` from the public `RenderedQueryOptions` and `SourceQueryOptions`. `RenderOrchestrator` now passes the current state separately through the internal query paths to `FeatureIndex` and the tiles' `querySourceFeatures`.
4.

src/mln/renderer/data_driven_property_evaluator.hpp
NEKOYASAN: Thanks. I added a note to the PR summary clarifying that this fallback change applies to all zoom-only property expressions, not only `global-state` expressions. 
I assume that anything not in a commit message will likely be squashed, so I decided this way would be easier to track.

src/mln/style/style_impl.cpp
NEKOYASAN: Thank you! I chose to match GL JS and reject global-state updates before the style JSON has loaded, rather than queue mutations across style parsing. (https://github.com/maplibre/maplibre-native/pull/4516/commits/b300d1c646919892eb7ba89b5b41409a57f9a484)

Core now throws `MisuseException` before marking the style as mutated, and `MapSnapshotter#setGlobalStateProperty` throws `IllegalStateException` instead of silently dropping the update. I also added Core and Node regression tests.

and I documented `Style::isLoaded()` in the changelog. It reports that the style JSON has been parsed and does not wait for source or sprite resources to finish loading.

NEKOYASAN:
@johncarmack1984 Thank you for your review!

> For anyone looking at the two red checks: those are the Android render-test legs from a workflow_run on main at https://github.com/maplibre/maplibre-native/commit/be3f03b86ec5dffc9ecb09ba1478fac84e9f66d0, which failed once and passed on the rerun. Every run on this PR's own head (https://github.com/maplibre/maplibre-native/commit/fc32f1ff209ae7a292b1c1357ee5296472a6591e) is green.

I rebased upstream/main, and all 50 checks passed.

> The other 10 are sky, projection, and the two *-layer-opacity properties, none of which exist in native yet. Should stubs or // TODO comments go in to keep parity, or something else?

I think it is fine to exclude these for now. I could not find any precedents in the Native codebase where stubs or TODO comments were added for Mapbox GL JS rendering test features that are not yet supported on the Native side. so I believe it would be better to port them together when implementing those features themselves.

platform/darwin/src/MLNStyle.mm
```diff
@@ -533,6 +534,22 @@ - (MLNImage *)imageForName:(NSString *)name {
...
+  for (const auto &property : state) {
+    globalState[@(property.first.c_str())] = MLNJSONObjectFromMBGLValue(property.second);
+  }
+  return [globalState copy];
+}
+
+- (void)setGlobalStateValue:(id)value forProperty:(NSString *)propertyName {
+  MLNLogDebug(@"Setting global state property: %@", propertyName);
+  self.rawStyle->setGlobalStateProperty(propertyName.UTF8String, MLNValueFromJSONObject(value));
+}
```
johncarmack1984, line 555: One iOS side effect of the new throw: `MLNMapSnapshotter.style` hands out an `MLNStyle` before the style has loaded, and this call goes straight into core. A pre-load call used to drop the value; now it terminates the process, since nothing on the ObjC side catches it. The sibling methods in this file wrap the core call in try/catch and raise an `NSException`. Same wrapper here would match the Kotlin `IllegalStateException`. Not blocking.

johncarmack1984 (commented):
Read the four new commits against the threads. Submodule is back on main's commit. `Tile::setLayers` carries the state now, and `setGlobalState` went with it. The query paths hand the state over next to the options instead of on the public structs. The pre-load call throws before the style is marked mutated, and `isLoaded()` is in the changelog. Range-diff of the rebased series against what I read last week is all `=`, so nothing else moved. All 49 checks green on the head. Good from me (with the caveat that I am just a citizen-contributor, not a maintainer!). Hope this was helpful! :)

platform/darwin/src/MLNStyle.mm
NEKOYASAN: Thank you for your review! your review is very helpful to me!

I was satisfied after fixing the Android side... I have now also fixed the iOS side to throw an `NSInternalInconsistencyException`.

I also wanted to review the globe view PR in appreciation of your review, but it turned out to be quite difficult... ;_;

johncarmack1984: haha yeah that one is HUGE, but thanks for taking a crack at it :)

src/mln/renderer/layers/render_color_relief_layer.hpp
```diff
@@ -52,6 +52,7 @@ class RenderColorReliefLayer final : public RenderLayer {
     // Color ramp data
     uint32_t colorRampSize = 256;
     bool colorRampChanged = true;
+    std::shared_ptr<const GlobalStateMap> colorRampGlobalState;
```
louwers, line 56: What are these?

metrics/ignores/platform-macos-webgpu.json
```diff
@@ -9,6 +9,7 @@
     "render-tests/fill-extrusion-pattern/feature-expression": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
     "render-tests/fill-extrusion-pattern/function": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
     "render-tests/fill-extrusion-pattern/function-2": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
+    "render-tests/fill-extrusion-pattern/global-state": "https://github.com/mapbox/mapbox-gl-js/issues/3327",
```
louwers, line 12: This issue is not related to global state.

src/mln/renderer/layers/render_color_relief_layer.hpp
NEKOYASAN: These store the immutable `global-state` snapshot used to generate the cached color ramp for `color-relief`, `heatmap-color`, and `line-gradient`.
Since the ramps are precomputed, they need to be regenerated when a referenced global-state value changes. The stored snapshot is also passed in when evaluating the ramp expression.
~Should I add this as a comment?~
Resolved the conflicts and added a comment.

metrics/ignores/platform-macos-webgpu.json
NEKOYASAN: As with other `fill-extrusion-pattern` tests, this test is known to cause rendering mismatches in the macOS WebGPU environment.
I am logging that issue because the newly added test is also expected to pass correctly once this fill-extrusion-pattern problem is resolved.

src/mln/renderer/layers/render_color_relief_layer.hpp
louwers: Makes sense, thanks.

include/mln/style/expression/format_section_override.hpp
```diff
@@ -27,7 +27,16 @@ class FormatSectionOverride final : public Expression {
         }
 
         return defaultValue.match(
-            [&context](const style::PropertyExpression<T>& e) { return e.getExpression().evaluate(context); },
+            [&context](const style::PropertyExpression<T>& e) {
+                // Evaluating the raw inner expression would bypass the
```
TimSylvester: This could be more clear.  Why is the context sometimes stateless?  It looks like the global state is never provided when properties are evaluated through binders, presumably for simplicity?  Can both the context and the expression have state, and is it correct that the context's state overrides the captured state in that case?

include/mln/style/color_ramp_property_value.hpp
```diff
@@ -14,6 +14,7 @@ namespace style {
 class ColorRampPropertyValue {
 private:
     std::shared_ptr<expression::Expression> value;
+    std::shared_ptr<const std::set<std::string>> globalStateRefs_;
```
TimSylvester: The convention in this code is to use a trailing underscore for arguments which conflict with members, not for members themselves.

include/mln/style/color_ramp_property_value.hpp
```diff
@@ -48,6 +60,8 @@ class ColorRampPropertyValue {
 
     using Dependency = style::expression::Dependency;
     Dependency getDependencies() const noexcept { return value ? value->dependencies : Dependency::None; }
+
+    const std::set<std::string>* getGlobalStateRefs() const noexcept { return globalStateRefs_.get(); }
```
TimSylvester: This could use an explanation of why the color ramp property value needs to track state dependencies.  Something like, if I understand correctly, this is a special-case property which is represented by a dummy `Color` value in the evaluated paint slot and separately evaluated by the layers where it's used (heatmap, relief, line).  Those separate evaluations need to know what keys are relevant in order to avoid unnecessary re-evaluations.

src/mln/style/layers/symbol_layer_impl.hpp
```diff
@@ -17,20 +17,38 @@ struct FormatSectionOverrides;
...
+    template <typename Property, typename T, typename U, typename V>
+    static void setOverride(const T& overrides,
+                            U& overridable,
+                            const V& unevaluated,
+                            const std::shared_ptr<const GlobalStateMap>& globalState) {
         if (hasOverride<Property>(overrides.template get<TextField>())) {
+            auto& value = overridable.template get<Property>();
+            // Wrapping a folded constant would drop the GlobalState dependency
+            // from the wrapper and exempt the bucket from relayout tracking.
+            if (value.isConstant()) {
```
TimSylvester, line 29: Just thinking out loud here, but I wonder if it would be better to prevent expressions containing global state references from being collapsed into constants rather than having to handle it specially when they have been.

TimSylvester (approved):
Looks really good overall, just a couple of nits.  

It would be more idiomatic to use `Immutable<GlobalState>` rather than `shared_ptr`, but I'm not going to insist that you change it.

Thanks!

TimSylvester:
I don't see a coverage run, how's the coverage of the new code from those test cases?

include/mln/style/expression/format_section_override.hpp
NEKOYASAN: In the current code, the `EvaluationContext` created by the binders never has `globalState` set. Pass the global state through all of the binder APIs would have made the change considerably broader, so the style-state snapshot is captured on the `PropertyExpression` for deferred evaluation instead.
`FormatSectionOverride::evaluate()` still accepts a general `EvaluationContext`, so the context and the expression can technically both have state, although this does not currently happen through the binder path. If a caller explicitly provides context state, it takes precedence; the captured state is used only as a fallback when the context has none.

I'm not quite sure how to explain this in a in-source comment clearly, but I'll think about it for a bit.

include/mln/style/color_ramp_property_value.hpp
NEKOYASAN: Thank you! It seems I may not have properly understood the coding guidelines...

I fixed to the file you pointed out and its related files.
Please let me know if I misunderstood anything or if the scope of the changes is too large.

include/mln/style/color_ramp_property_value.hpp
NEKOYASAN: Yes, that understanding is correct. `ColorRampPropertyValue` implements the subset of the `PropertyValue<T>` interface needed by the `color-ramp` properties. Generic paint-property evaluation stores a placeholder `Color`, while the `heatmap`, `color-relief`, and `line-gradient` render layers evaluate the expression separately to build their color-ramp textures. The referenced global-state keys are retained so those textures are rebuilt only when a relevant key changes. 

I added more description to the comment. Is it clear enough?

src/mln/style/layers/symbol_layer_impl.hpp
NEKOYASAN: Thank you for the comment. I took another look at this, but I think the current targeted handling is preferable for the following reasons:

- A paint expression (property?) that depends on global-state but not on a feature can normally be evaluated once by `DataDrivenPropertyEvaluator` and represented as a constant. This avoids evaluating the same value separately for every feature.
- format expression (type?) are a special case because the override wrapper is added after the paint property has been evaluated. If the evaluated constant were wrapped directly, it would no longer carry the original global-state dependency. As a result, Native would not know that the symbol’s per-feature paint data needs to be rebuilt when the referenced state changes.

Keeping all global-state expressions unevaluated would avoid this special case, but it would also move feature-independent paint properties into the per-feature binder path. I therefore restore the original expression only for format-section overrides.

NEKOYASAN:
@TimSylvester thank you for your review!

> It would be more idiomatic to use Immutable<GlobalState> rather than shared_ptr, but I'm not going to insist that you change it.

I understand what you mean, and I completely agree. It appears that my understanding of the language specifications was quite lacking...
However, changing it now would result in a fairly wide-ranging refactoring, so I am thinking of handling it in a separate PR.

> I don't see a coverage run, how's the coverage of the new code from those test cases?

It seems that the upload to Codecov is being skipped. It might be better to fix `upload-coverage.yml` in a separate PR...
Although it only reflects the test coverage for Linux, it looks like 83.3% of the diff is covered.

sargunv:
Nice work! I'm already using this branch's changes in maplibre-compose to provide global state for the accessibility font scale, so expressions can refer to that to scale their labels accordingly.

NEKOYASAN:
I have also started using it a bit for app development on my end, so the basic implementation itself should probably be fine.
The CI failure seems to be due to a download error, so it looks like it needs to be rerun.

Please let me know if there is anything I need to do, such as requesting a re-review or making revisions.

PR #4531, opened 2026-08-28
fix(core): update amalgam symbol filters for mln namespace rename

ciscorn:
Follow-up to #4487.

Updates the amalgam symbol filters to match the namespace rename from `mbgl` to `mln`.


louwers (approved):

louwers:
Missed this. Thank you!

PR #4532, opened 2026-08-28
Convert Android platform files to kotlin

adrian-cojocaru:
Convert Android SDK Java sources to Kotlin.

AI disclosure: Claude Fable/Opus 4.6 assisted with file conversion and formatting.


louwers:
Pre-release can be found here: https://github.com/maplibre/maplibre-native/releases/tag/android-v13.6.1-pre0

louwers (commented):
Could you look into if there's an CLI tool to convert Java to Kotlin?

Maybe we could run an automated tool, and then do an LLM pass for the manual fixes?

Also maybe the formatting can be done in a separate commit?

adrian-cojocaru:
The initial phase used the Android Studio converter, followed by the AI pass to clean up some more code (function calls to properties, class constructors, nullability usage, tests, etc) and formatting. This was squashed into two commits to preserve the file history (since file renaming breaks it).

PR #4535, opened 2026-08-29
fix(ios): reuse annotation container across updates

### LINKED ISSUE #3960: Lags when adding new annotations backed by MLNAnnotationView
### MapLibre iOS Version

6.18.1

### iOS Version

18.3.1

### Device

iPhone 16 pro

### What happened?

In my project I use view annotations and have a logic where new annotations are being added as the camera moves along the map. When there are already quite a few annotations on the map, adding new ones produces significant UI lags. The more annotations are added, the heavier the lags are.

Also, while trying to research the issue, I've noticed, that when I remove out of screen annotations, they are actually not being removed from `annotationContainerView`, but seem to be kept for reusing. Even if I try to remove all the annotations from the map, seems like only the visible annotation are actually being removed from `annotationContainerView`, while non of them are needed any more.

### Steps to reproduce

Preconditions:
1. `mapView(_: MLNMapView, viewFor: any MLNAnnotation) -> MLNAnnotationView?` that returns `MLNAnnotationView` for any added annotation. Same `reuseIdentifier` is user for all annotation views, dequeueReusableAnnotationView is also used prior to new view creation.

Steps to reproduce UI lags
1. Implement `mapViewRegionIsChanging(_: MLNMapView)` method of MLNMapViewDelegate with adding new annotations as the camera is moving (`mapView.addAnnotations()`)
2. move the camera along the map making new annotations to be added
3. the more annotations are added, the more laggy is the movement.

Steps to reproduce annotation view is not actually deleted
1. add 3000 annotations on the map
2. zoom in so that only some of the annotations remain visible
3. delete all the added annotations from the map (`mapView.removeAnnotations(annotations)`)
4. check `annotationContainerView` subviews count - there are still a lot of them

### Relevant log output

```shell

```

### Additional context

That doesn't seem correct, when all the annotations are being removed from the map some of them (in some cases all of them) are actually kept as subviews of `annotationContainerView`.

Also analising MLNMapView.mm code I've notices some points:
1. there is `updateAnnotationContainerViewWithAnnotationViews` method that seems to be called to add new annotation views on the map, and it recreates `annotationContainerView` on each call. I'm not sure why is that needed, but probably this might be a reason of lags when adding new annotations
2. according to `removeAnnotations` method implementation seems like it should delete annotation view from `annotationContainerView` and from reuse queue no matter if it is visible or not, but for some reason that does not happen for out of screen annotations. I've noticed the following code in `removeAnnotations` which seems to be a bug and might be the reason of the issue:
```
if (![annotationViewReuseQueue containsObject:annotationView])
{
    [annotationViewReuseQueue removeObject:annotationView];
}
```

Screencast for deleting all annotations at the moment when only some of them are visible
Image

Screencast for laggy map panning while adding new annotations
Image

(The label on the bottom depicts the number of `annotationContainerView` subviews got as `mapView.subviews.first(where: { $0 is MTKView })?.subviews.first`)

miakh:
## Summary

Closes #3960. Adding view-backed annotations in batches caused `MLNMapView` to replace the annotation container each time and reparent all existing views.

This change keeps the existing container. It also removes deleted annotation views from the reuse queue and clears reusable views once no view-backed annotations remain. Regression tests cover offscreen view removal and container reuse across batches.

## Verification

- `bazel test //platform/ios/test:ios_test --test_filter=MLNAnnotationViewTests --test_output=errors --//:renderer=metal`
- iOS 26.5 Simulator with 3,000 annotation views added in 30 batches
- Annotation-container replacements: 29 before, 0 after

**AI disclosure:** OpenAI Codex (GPT-5) assisted with implementation and tests under human supervision.

### DIFF
--- a/platform/ios/src/MLNAnnotationContainerView.h
+++ b/platform/ios/src/MLNAnnotationContainerView.h
@@ -8,9 +8,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface MLNAnnotationContainerView : UIView
 
-+ (instancetype)annotationContainerViewWithAnnotationContainerView:
-    (MLNAnnotationContainerView *)annotationContainerView;
-
 - (void)addSubviews:(NSArray<MLNAnnotationView *> *)subviews;
 
 @end
--- a/platform/ios/src/MLNAnnotationContainerView.m
+++ b/platform/ios/src/MLNAnnotationContainerView.m
@@ -19,13 +19,6 @@
     return self;
 }
 
-+ (instancetype)annotationContainerViewWithAnnotationContainerView:(nonnull MLNAnnotationContainerView *)annotationContainerView
-{
-    MLNAnnotationContainerView *newAnnotationContainerView = [[MLNAnnotationContainerView alloc] initWithFrame:annotationContainerView.frame];
-    [newAnnotationContainerView addSubviews:annotationContainerView.subviews];
-    return newAnnotationContainerView;
-}
-
 - (void)addSubviews:(NSArray<MLNAnnotationView *> *)subviews
 {
     for (MLNAnnotationView *view in subviews)
--- a/platform/ios/src/MLNMapView.mm
+++ b/platform/ios/src/MLNMapView.mm
@@ -4908,21 +4908,17 @@ static void *windowScreenContext = &windowScreenContext;
     (NSArray<MLNAnnotationView *> *)annotationViews {
   if (annotationViews.count == 0) return;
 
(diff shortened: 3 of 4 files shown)


miakh:
iOS Simulator reproduction with 3,000 annotation views added in 30 batches and then removed.

**Before — the annotation container is replaced 29 times**

https://github.com/user-attachments/assets/7f11e263-755e-4ba3-a487-00c68476dac9

**After — the same container is reused (0 replacements)**

https://github.com/user-attachments/assets/b7fd0726-de3a-4fe5-a00f-dfb96242e994

PR #4537, opened 2026-08-29
iOS: Clear transient edge padding for unchanged camera

### LINKED ISSUE #4449: [iOS] setCamera does not clear transient edgePadding when camera is unchanged
## Description

On iOS, calling `setCamera(... edgePadding: .zero ...)` does not reliably clear a previously applied transient `edgePadding` when the camera itself is unchanged.

This has required MapLibre React Native to keep a 1 pt top/bottom padding workaround since 2021 so that the native camera update is not treated as a no-op.

## Steps to reproduce

```swift
let targetCamera = mapView.camera

mapView.setCamera(
    targetCamera,
    withDuration: 0,
    animationTimingFunction: nil,
    edgePadding: UIEdgeInsets(top: 0, left: 0, bottom: 200, right: 0)
) {
    mapView.setCamera(
        targetCamera,
        withDuration: 0,
        animationTimingFunction: nil,
        edgePadding: .zero,
        completionHandler: nil
    )
}
```

A coordinate located at `targetCamera.centerCoordinate` can be displayed as a point annotation to make the remaining visual offset apparent after the second update.

The same behavior can be encountered when transitioning from a camera update fitted with padding to the same bounds or center with zero transient padding.

## Expected behavior

The second camera update clears the previous transient `edgePadding`. The resulting effective padding should be the map view's persistent `contentInset` only.

## Actual behavior

The second update can return before forwarding new camera options to the core map, leaving the previous transient padding effective.

## Suspected root cause

`MLNMapView.setCamera` first combines the requested `edgePadding` with `contentInset`, then uses this no-op guard:

https://github.com/maplibre/maplibre-native/blob/main/platform/ios/src/MLNMapView.mm#L4295-L4325

```objc
if ([self.camera isEqualToMapCamera:camera] &&
    UIEdgeInsetsEqualToEdgeInsets(_contentInset, edgePadding)) {
    return;
}
```

For a zero transient padding request, the combined `edgePadding` equals `_contentInset`. If the camera is also unchanged, the method returns without considering that the core map may still contain a non-zero transient padding from the previous programmatic camera update.

The no-op decision appears to compare the new request with persistent `contentInset`, rather than with the effective padding currently held by the core map.

## Downstream workaround

MapLibre React Native currently changes zero vertical padding to 1 pt before calling `setCamera`, specifically to bypass this early return:

https://github.com/maplibre/maplibre-react-native/blob/main/package/ios/components/camera/CameraUpdateItem.m#L62-L72

This workaround originated in:

- https://github.com/rnmapbox/maps/pull/1614
- https://github.com/rnmapbox/maps/commit/a2d46c4f272673e248d574f6bcf65407d1196096

## Related work

- https://github.com/maplibre/maplibre-native/issues/171
- https://github.com/maplibre/maplibre-native/pull/188

Those addressed inconsistent padding persistence across gestures, but do not appear to cover clearing transient padding with an otherwise identical programmatic camera update.

## Suggested regression test

1. Set a camera with non-zero transient `edgePadding`.
2. Set the same camera with zero transient `edgePadding`.
3. Assert that the core camera padding equals `contentInset`, with no remaining transient component.

Once fixed and released, MapLibre React Native should be able to remove its 1 pt workaround.

miakh:
Fixes #4449.

`setCamera` now compares the requested effective padding with the padding currently held by the core map before treating an unchanged camera as a no-op. This allows zero `edgePadding` to clear earlier transient padding.

The regression test applies bottom padding, sets the same camera again with zero padding, and verifies that only `contentInset` remains.

Tested with:

- `bazel test //platform/ios/test:ios_test --test_output=errors --//:renderer=metal --test_filter="MLNMapViewGestureRecognizerTests/testSetCameraClearsTransientEdgePaddingWhenCameraIsUnchanged"`

AI usage: OpenAI Codex (GPT-5) assisted with the implementation and tests under human supervision.

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,6 +4,7 @@ MapLibre welcomes participation and contributions from everyone. Please read [`M
 
 ## 6.29.0
 
+- iOS: Clear transient edge padding when setting an unchanged camera.
 - fix(core): accept alpha in hsl colors ([#4435](https://github.com/maplibre/maplibre-native/pull/4435)).
 - fix(core): repaint feature-state-driven symbol paint properties ([#4445](https://github.com/maplibre/maplibre-native/pull/4445)).
 - Make string expressions operate on unicode ([#4344](https://github.com/maplibre/maplibre-native/pull/4344)).
--- a/platform/ios/src/MLNMapView.mm
+++ b/platform/ios/src/MLNMapView.mm
@@ -4317,7 +4317,7 @@ static void *windowScreenContext = &windowScreenContext;
   }
 
   if ([self.camera isEqualToMapCamera:camera] &&
-      UIEdgeInsetsEqualToEdgeInsets(_contentInset, edgePadding)) {
+      UIEdgeInsetsEqualToEdgeInsets(self.cameraEdgeInsets, edgePadding)) {
     if (pendingCompletion) {
       [self animateWithDelay:duration animations:pendingCompletion];
     }
--- a/platform/ios/test/MLNMapViewGestureRecognizerTests.mm
+++ b/platform/ios/test/MLNMapViewGestureRecognizerTests.mm
@@ -67,6 +67,39 @@
   [_styleLoadingExpectation fulfill];
 }
 
+- (void)testSetCameraClearsTransientEdgePaddingWhenCameraIsUnchanged {
+  UIEdgeInsets contentInset = UIEdgeInsetsMake(1, 2, 3, 4);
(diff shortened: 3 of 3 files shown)


miakh:
Reproduction on iPhone 17 Pro Simulator (iOS 26.5), recorded 30 Aug 2026.

Red dot = target coordinate; blue cross = screen center. After clearing padding with the same camera, current main remains 90 pt offset; this branch returns to 0 pt.

Before (current main):

https://github.com/user-attachments/assets/c00a4cdf-2eac-4694-9144-0b8fac2c1520



After (this branch):

https://github.com/user-attachments/assets/861efde8-077b-4663-bee1-d7c23100d846

PR #4538, opened 2026-08-30
fix(core): bundle fastpfor-lib into the Linux/macOS amalgams

ciscorn:
MapLibre Native now enables FastPFor (#4146), so the Linux and macOS amalgams must bundle it alongside `mlt-cpp`. Without it, downstream builds fail.

- #4146


louwers (approved):
Thank you.

PR #4548, opened 2026-09-02
fix(ios): use window scene interface orientation for heading compensation

VandenOstende:
`MLNMapView.updateHeadingForDeviceOrientation` reads the interface orientation with `-[UIApplication statusBarOrientation]`. That API is deprecated and is a no-op as of iOS 27, which has two consequences.

First, the heading is no longer compensated for the interface orientation: the switch always falls through to the portrait default. Second, the method runs on every heading update, so UIKit emits its deprecation notice several times per second. In a navigation app that is hundreds of lines per minute.

This reads the orientation from the view's window scene instead, keeping the old call behind an availability check because the deployment target is still iOS 12.

One behavioural note: when the map view is not yet in a window, `windowScene` is nil and the orientation resolves to `UIInterfaceOrientationUnknown`, which lands on the existing portrait default. Previously the application-wide orientation was used. No heading is rendered in that state, so this should not be observable.


louwers (approved):
Checks out. https://developer.apple.com/documentation/uikit/uiapplication/statusbarorientation

Thanks!

PR #4552, opened 2026-09-03
refactor(storage): write the sqlite3 initializer explicitly

### LINKED ISSUE #907: Remove `MBGL_CONSTRUCTOR` preprocessor macro
https://github.com/maplibre/maplibre-gl-native/blob/8b804a19038c6f5ed01261f0bb4364cc7e386e4a/platform/default/src/mbgl/storage/sqlite3.cpp#L14

Seems unnecessary use of the preprocessor here.

johncarmack1984:
Closes #907.

The `MBGL_CONSTRUCTOR` macro in `sqlite3.cpp` built a static object whose constructor ran the sqlite version check. Same thing, written out: an anonymous-namespace `initialize()` and one static that calls it. The macro, the `NOLINT` it needed, and the last `MBGL_` name in the file go with it.

Behavior is unchanged: a major-version mismatch still throws during static initialization, as before.

Checked: Metal build clean; `Sqlite*`, `OfflineDatabase*` and `*Database*` unit tests, 64 passed; `pre-commit run` clean.

AI Disclosure: I asked Claude Fable 5.1 to scour open issues for something that would be a quick win for @louwers so I could feel slightly less guilty about the monster PR from last week. This was what it found; I wrote the fix according to the discussion in the issue and on #905.

### DIFF
--- a/platform/default/src/mln/storage/sqlite3.cpp
+++ b/platform/default/src/mln/storage/sqlite3.cpp
@@ -11,16 +11,6 @@
 #include <mln/util/traits.hpp>
 #include <mln/util/logging.hpp>
 
-#define MBGL_CONSTRUCTOR(f) \
-    static void f(void);    \
-    struct f##_t_ {         \
-        f##_t_(void) {      \
-            f();            \
-        }                   \
-    };                      \
-    static f##_t_ f##_;     \
-    static void f(void)
-
 namespace mapbox {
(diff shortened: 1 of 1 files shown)


johncarmack1984:
Closing: #4476 does the same thing and already has an approval. Should have checked the linked PRs on the issue before opening.

PR #4553, opened 2026-09-03
fix(core): break a label before a left parenthesis, not after it

### LINKED ISSUE #2448: Allow breaking before left parenthesis, not after
> Some people and some organisations don't put a whitespace before a left parenthesis. Currently Maplibre only breaks after a left parenthesis, not before. 
> This will put the left parenthesis at the end of the line.
> 
> You can notice this problem for example in "Freiburg(Breisgau) Hbf" which is the official station name

https://github.com/maplibre/maplibre-gl-js/pull/4138

I committed the render tests on my fork: https://github.com/louwers/maplibre-native/tree/text-breaking-left-paren

image

johncarmack1984:
Closes #2448. Port of maplibre/maplibre-gl-js#4138.

Station names like "Freiburg(Breisgau) Hbf" have no space before the parenthesis. The shaper only knew how to break *after* a left parenthesis, so the line ended on a dangling "(" and the next one opened with "Breisgau)". A break is now allowed before a left parenthesis and no longer after it, same as GL JS. GL JS's guard comes along: no break before a parenthesis that is the last character, so it never becomes a line of its own.

`util::i18n::allowsWordBreakingBefore` sits next to `allowsWordBreaking`; `determineLineBreaks` checks the next character with it.

| main | this PR |
|---|---|
| main | this PR |

`text-breaking/left-parenthesis` rendered on Metal. Left: main ends the line on "Freiburg(". Right: the break moves before "(Bresigau)".

Tests: `Shaping.BreakBeforeLeftParenthesis` in `test/text/shaping.test.cpp` fails on main (9 and 10 glyphs per line instead of 8 and 11) and passes here. The two render tests are the ones from GL JS #4138, `text-breaking/left-parenthesis` and `text-breaking/whitespace`, unchanged from @louwers' `text-breaking-left-paren` branch: `left-parenthesis` fails on main and passes here, `whitespace` passes on both.

Checked: Metal: `Shaping.*` 3 passed; every `render-tests/text-*` test, 235 passed, 1 passed but ignored, 2 ignored, 0 failed; the full render manifest and the unit suite (860) unchanged. Same tree on Linux GL (llvmpipe, clang-tidy 18 clean), MoltenVK, Windows MSVC (unit 1030 passed) and the iOS simulator: both new tests pass everywhere and every other result matches main. `pre-commit run` clean. Semver: patch.

AI Disclosure: I asked Claude Fable 5.1 to scour open issues for something that would be a quick win for @louwers so I could feel slightly less guilty about the monster PR from last week. This was what it found. I wrote the fix according to the issue body and maplibre/maplibre-gl-js#4138. Testing, verification, and iteration performed locally by Claude Fable.


louwers (approved):

louwers:
Thank you!

johncarmack1984:
The Vulkan instrumentation device run passed (`result=PASSED`, 26 min); the red X is the checks-action step that reports it back, which died in 7 s, so the check is stuck at "in progress". Re-running the failed job should clear it, which is maintainer-only. I didn't want to trigger a whole 2-hour 50 job re-run just for a single job that could be re-run in 30 min from the dropdown 😄

PR #4556, opened 2026-09-05
metal: pass API validation with 3D terrain on and off

AndrewN81:
## What

Two Metal API validation failures with the terrain-3d branch, one with terrain off and one with it on. Xcode's Run scheme enables Metal API Validation by default, so with this branch any app hits one of them on launch (terrain off) or on the first draped frame (terrain on).

1. **Missing sampler bindings with terrain off.** The symbol, circle and fill-extrusion shaders declare the DEM and depth samplers unconditionally, but their tweakers only bind textures when a terrain is present, so validation aborts with `missing sampler binding at index N for demSampler/depthSampler`. `gfx::Context::getPlaceholderTexture2D()` now provides a 1x1 texture that the tweakers bind when `parameters.terrain` is null (the shaders never sample it, `dem_enabled` / `depth_enabled` are 0).
2. **Pipeline state reuse across attachment formats with terrain on.** `mtl::ShaderProgram::getRenderPipelineState` cached states keyed only on the colour mode and vertex layout, so a state created for the BGRA8 screen was reused for the RGBA8 drape and hillshade render targets (`Set Render Pipeline State Validation ... BGRA8Unorm vs RGBA8Unorm`, and a stencil format mismatch). The cache key now folds in the colour, depth and stencil attachment formats.

## How it was verified

- Reproduced both aborts on the iOS simulator with `SIMCTL_CHILD_MTL_DEBUG_LAYER=1 xcrun simctl launch <udid> <bundle>`; both are gone with this change, with terrain on and off.
- Running in a shipped iOS app (Metal, iPhone 16 Pro and simulator) since 2026-09-05.

## Notes

### DIFF
--- a/include/mln/gfx/context.hpp
+++ b/include/mln/gfx/context.hpp
@@ -155,6 +155,12 @@ public:
     /// Create a new vertex attribute array
     virtual gfx::VertexAttributeArrayPtr createVertexAttributeArray() const = 0;
 
+    /// A shared 1x1 zero texture to bind wherever a shader declares a sampler the current
+    /// frame has no real texture for (e.g. the DEM / terrain-depth slots of the symbol, circle
+    /// and fill-extrusion shaders when 3D terrain is off). Metal's API validation aborts on an
+    /// unbound sampler even when the shader never samples it; Vulkan already binds a dummy.
+    const Texture2DPtr& getPlaceholderTexture2D();
+
     /// Create a new drawable builder
     virtual UniqueDrawableBuilder createDrawableBuilder(std::string name) = 0;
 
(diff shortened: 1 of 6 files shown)


pdrhlik:
I've been running this PR merged into feature/terrain-3d. One thing to flag: the shared placeholder texture lives in `gfx::Context`, whose members are destroyed after the backend context's destructor body - where debug builds assert that no rendering resources are left. So every map created and destroyed in a debug build aborts there (I hit it in LocalGlyphRasterizer.* and Annotations.* in mbgl-test-runner), with terrain on or off.

Releasing it with each backend's own resources, before the check, fixes it - one line in each of the three contexts:

    placeholderTexture2D.reset();

in `src/mln/gl/context.cpp` (after `uboAllocator.reset()`), `src/mln/mtl/context.cpp` (after the global uniform buffers are cleared) and `src/mln/vulkan/context.cpp` (after `destroyResources()`).

Commit, if it's easier to take: https://github.com/mapriot/maplibre-native/commit/1089de0c64
Happy to open it as a PR against your branch instead.

PR #4557, opened 2026-09-05
raster-dem: let inline minzoom/maxzoom override the TileJSON

AndrewN81:
## What

A `raster-dem` source's inline `minzoom` / `maxzoom` are ignored on native when the source also has a TileJSON (`url`, or a PMTiles archive header): the TileJSON's zoom range replaces them. maplibre-gl-js lets the inline values win. The practical effect on this branch: a Terrarium PMTiles archive with a z12 header pins the terrain mesh and the drape cover at z12, so every road and building goes soft as soon as the map is tilted at z14+, even though the style asked for `maxzoom: 16` (overzoom).

`RasterDEMSource` now records inline `minzoom` / `maxzoom` from the style (`SourceOptions`) and reapplies them over the loaded tileset's zoom range, matching gl-js precedence.

## How it was verified

- Simulator A/B at z14.3, pitched: with the style's inline `maxzoom: 16` honoured, the drape is rendered from z14 tiles and roads stay crisp; before, identical output for `maxzoom` 14 and 16 (both ignored, cover stuck at the archive's z12).
- Shipped in an iOS app since 2026-09-05.

## Notes

### DIFF
--- a/include/mln/style/sources/raster_dem_source.hpp
+++ b/include/mln/style/sources/raster_dem_source.hpp
@@ -10,6 +10,12 @@ namespace style {
 struct SourceOptions {
     std::optional<Tileset::RasterEncoding> rasterEncoding = std::nullopt;
     std::optional<Tileset::VectorEncoding> vectorEncoding = std::nullopt;
+    // Inline style overrides applied on top of a fetched TileJSON, matching maplibre-gl-js
+    // (`extend(tileJSON, options)` in load_tilejson.ts). A raster-dem source served from an
+    // archive whose TileJSON stops at its deepest stored zoom can declare a higher `maxzoom`
+    // so terrain meshes and drapes at the display zoom while the DEM overzooms.
+    std::optional<uint8_t> minzoom = std::nullopt;
+    std::optional<uint8_t> maxzoom = std::nullopt;
 };
 
 // NOTE: Any derived class must invalidate `weakFactory` in the destructor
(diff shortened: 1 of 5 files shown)


include/mln/style/sources/raster_dem_source.hpp
```diff
@@ -10,6 +10,12 @@ namespace style {
 struct SourceOptions {
     std::optional<Tileset::RasterEncoding> rasterEncoding = std::nullopt;
     std::optional<Tileset::VectorEncoding> vectorEncoding = std::nullopt;
+    // Inline style overrides applied on top of a fetched TileJSON, matching maplibre-gl-js
```
louwers: If you want to reference a source file in another repo, it better be a permalink.

louwers (changes requested):
Needs tests.

include/mln/style/sources/raster_dem_source.hpp
AndrewN81: Done, it now links the gl-js line by commit permalink.

AndrewN81:
Added unit tests in ce4876b:

- `SourceOptions` conversion (test/style/conversion/source_options.test.cpp): `minzoom`/`maxzoom` parsed alone and alongside `encoding`; a non-numeric or out-of-range value is rejected with an `invalid minzoom`/`invalid maxzoom` error.
- `RasterDEMSource` (test/style/source.test.cpp): after `loadDescription` an inline `maxzoom` replaces the TileJSON's while an unset `minzoom` keeps the TileJSON value, and with no inline options the TileJSON range is kept as-is.

Checked that the override test fails against the previous code (max stays at the TileJSON's 12) and passes with the change. All 17 `SourceOptions.*` / `Source.RasterDEM*` tests pass locally on macOS Metal.

PR #4560, opened 2026-09-05
metal: compile shader libraries off the render thread (opt-in); drawables wait instead of blocking

AndrewN81:
## What

Every Metal program is compiled from source on first use with a synchronous `newLibrary`, on the render thread, which on iOS is the main thread. With a cold shader cache that is roughly 0.4 to 0.6 s per program, one after another: a first launch stalls the UI for seconds, and a 3D-terrain style's first draped frame (terrain pass, depth pass, the symbol permutations that appear at a new pitch) stalled it for over a second (Xcode reports "Hang detected" for each).

- `gfx::Context::setAsyncShaderCompilation(bool)`, implemented by the Metal context. When on, `createProgram` returns a `ShaderProgram` whose functions arrive from Metal's asynchronous `newLibrary` completion handler (`ShaderProgram::PendingFunctions`, adopted on the render thread by `isReady()`).
- A drawable whose program is not ready skips its draw for that frame, and `Renderer` keeps requesting frames while `Context::hasPendingShaderCompiles()`, so everything appears a frame or two later instead of blocking.
- The clipping mask program stays synchronous (it is drawn by `Context::renderTileClippingMasks`, not through a drawable).
- Off by default, so single-frame renders (render tests, snapshots) are unchanged. `MLNMapView` turns it on through `Renderer::setAsyncShaderCompilation(true)`. Compile durations are logged at Info on both paths.

Note for reviewers: `getRenderPipelineState` does not add attachment formats to its cache key here; that is #4556. This branch is independent of it apart from that adjacent line.

## How it was verified

- iOS simulator with the app's Metal shader cache deleted: 13 programs compiled concurrently off-thread (400 to 560 ms each), the map renders identically, a draped terrain frame renders identically; no hangs. With a warm cache compiles report 0 to 25 ms.
- Running in a shipped iOS app since 2026-09-05.

## Notes

### DIFF
--- a/include/mln/gfx/context.hpp
+++ b/include/mln/gfx/context.hpp
@@ -172,6 +172,16 @@ public:
     /// Get the generic shader with the specified name
     virtual gfx::ShaderProgramBasePtr getGenericShader(gfx::ShaderRegistry&, const std::string& name) = 0;
 
+    /// Compile shader programs off the render thread where the backend supports it (Metal).
+    /// A program's drawables are skipped for the frames its library is still compiling and the
+    /// renderer keeps requesting frames until every pending compile has landed, so first use of
+    /// a shader (a new symbol permutation, the terrain pass) no longer stalls the render thread
+    /// - on iOS that thread is the main thread, and a terrain style's first draped frame stalled
+    /// it for over a second. Off by default so single-frame renders (render tests, snapshots)
+    /// stay deterministic.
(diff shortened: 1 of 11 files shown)


louwers:
Not sure why this is targetting the `feature/terrain-3d` branch.

AndrewN81:
Fair point, it isn't terrain-specific. Re-opened against `main` as #4563 (same change rebased, with the pre-commit fixes folded in).

PR #4566, opened 2026-09-06
keep digits and short codes upright in CJK line labels

### LINKED ISSUE #4565: Draw digits and short uppercase codes upright in vertical CJK line labels
This issue tracks porting the fix from maplibre/maplibre-gl-js#8205 to MapLibre Native.

With `symbol-placement: line` or `line-center` and `text-keep-upright: true`, CJK characters in vertical line labels are drawn upright, while digits and Latin letters remain rotated along the line. For example, the “21” in `반포대로21길` appears sideways relative to the surrounding characters.

MapLibre Native should match the updated GL JS behavior:

- Draw numbers and short uppercase alphanumeric codes such as `JR`, `A1`, and `A16` upright.
- Use vertical presentation forms for supported punctuation surrounded by upright characters.
- Preserve the existing orientation of lowercase words, longer uppercase words, complex scripts, and runs containing combining marks.

The scope should match the GL JS fix, including its regression tests.

Original bug report: maplibre/maplibre-gl-js#5404
GL JS fix: maplibre/maplibre-gl-js#8205

NEKOYASAN:
Fixes #4565.

## Details

Ports maplibre/maplibre-gl-js#8205 to MapLibre Native.

In vertical CJK line labels, digits such as “21” in `반포대로21길` were rotated relative to the surrounding characters. Numbers of any length and uppercase alphanumeric codes of up to three characters now render upright. Supported punctuation surrounded by upright characters also uses its vertical presentation form.

Lowercase words, longer uppercase words, complex scripts, and runs containing combining marks retain their existing orientation.

Includes all 21 GL JS unit regression cases and both new render fixtures, plus coverage for Native-specific glyph and formatting boundaries.

## Validation

- 45 related unit tests passed.
- 18 `text-writing-mode` render tests passed on macOS Metal.
- Confirmed that the relevant regression tests fail with the pre-fix shaping implementation.

## Note


louwers (changes requested):
Thank you. Some of the existing tests need to be updated. See CI failure.

NEKOYASAN:
@louwers thankyou! fixed test & qt builds

src/mln/util/i18n.cpp
```diff
@@ -685,6 +691,30 @@ bool isWhitespace(char16_t chr) {
     return chr == u' ' || chr == u'\t' || chr == u'\n' || chr == u'\v' || chr == u'\f' || chr == u'\r';
 }
 
+bool isDigit(char16_t chr) {
```
louwers: This looks like it should be in platform specific code, with a platform specific implementation.

Could you explore that option?

NEKOYASAN: Just to clarify, would moving these functions into `platform/default/` (ICU) and `platform/qt/` (QChar), with the build system selecting the implementation instead of using `#ifdef` in the common code, address your concern? Or did you mean using each OS's native character classification APIs?

I used QChar because the Qt builds on Windows and macOS don't directly depend on ICU. The non-Qt platforms already use ICU, so my preference would be to share that implementation to preserve the current behavior and keep the change focused.

louwers: Yes that's what I mean. We try to avoid `#ifdef`s for different platforms and instead use the approach you describe.

NEKOYASAN: Thank you! I should have checked the existing implementation and followed it. I'm sorry.

fixed at https://github.com/maplibre/maplibre-native/pull/4566/commits/097f06db53de05ac8b238e3b4dec402ff49e2037

louwers (approved):

src/mln/util/i18n.cpp
louwers: No worries that's what PR review is for.

louwers:
Thanks for your contribution!

PR #4568, opened 2026-09-08
Core: compress Metal shader sources with generated tooling

louwers:
This picks up the idea from @mwilsnd to compress the shader sources. It seems to save 2.3%.

The Metal shader sources are moved to a directory with a `manifest.json` that a script reads. See here the `shader_source.cpp` that it outputs. https://gist.github.com/louwers/24f39ffc738ef2bfc4789baad1c8e936 The source is decompressed lazily at runtime.

I wanted to try out the GitHub Stack functionality. See #4569 for a PR that builds on this one. 

Generated with GPT-6 Astral.

---

<details><summary>AI generated PR description</summary>
This moves Metal shader sources into `.metal` files and generates a single compressed bundle. The initial iOS arm64 benchmark showed a **180,152-byte (2.30%)** reduction in the stripped library.

This is the first layer of a two-PR stack revisiting the size-related ideas from #1848. The next layer, #4569, shares shader-group code using reflection.

- Add a manifest for 34 built-in shaders and their shared source fragments. Remove leading and trailing blank lines from the extracted fragments.
- Generate the bundle automatically in Bazel and CMake with `scripts/generate_metal_shader_code.ts`. Store each shared fragment once and compress the bundle with zlib. Update the Bazel Node toolchain to 24.11.1 for native TypeScript execution.
- Decompress lazily once, with thread-safe static initialization, and return views into immutable storage. Shader registration does not inflate the bundle. Source assembly still produces a null-terminated string for Metal.
- Run `node --test` from `scripts/` in CI and include the TypeScript files in the existing type check.

The retained source payload is about 198 KiB (about 256 KiB of string capacity with the measured Apple libc++ toolchain). First shader use now includes decompression; device startup performance has not been benchmarked.

Initial size measurement, recorded before the whitespace cleanup and Node toolchain update, against `57d198149785cb6068ca6d62d478670c64a60358`, using Xcode 26.4 (17E192), Bazel 8.6, and Bloaty 1.1 with matching dSYMs:

| iOS arm64 dynamic library | Baseline | Shader tooling | Change |
| --- | ---: | ---: | ---: |
| Stripped file | 7,840,296 B | 7,660,144 B | -180,152 B (-2.30%) |
| gzip-compressed file | 3,052,833 B | 3,054,236 B | +1,403 B |

Bloaty attributes the saving primarily to moving roughly 197 KiB out of `__cstring` into a roughly 30 KiB compressed constant. This improves installed size; the gzip result shows no download-size saving for this layer.

Build command used for both measurements:

```sh
bazel build --compilation_mode=opt --features=dead_strip,thin_lto \
  --objc_enable_binary_stripping --apple_generate_dsym --output_groups=+dsyms \
  --//:renderer=metal //platform/ios:MapLibre.dynamic --jobs=10
```

Validation:

- iOS Release Bazel builds passed for arm64 device and arm64/x86_64 simulator.
- Generator tests passed with Node 24 using `node --test` from `scripts/`: compression round trips, shared ranges, UTF-8 byte offsets, and deterministic output. Type checking and workflow/whitespace hooks also passed.
- All 54 fragments differ from the initial extraction only in boundary blank lines. The TypeScript conversion preserves generated C++ apart from the generator path comment when run with the same Node version. Bazel and CMake generation both pass; their bundles decode to identical shader text. C++ accessor checks previously passed for all 34 shaders on eight concurrent threads.
- The prior combined experiment containing both shader layers passed macOS Metal CMake builds and all six ShaderRegistry tests. Across 97 selected render cases, baseline and experiment had identical statuses and no unexpected failures; 95 images were byte-identical, and the other two differed by one channel value in five pixels and one pixel, respectively.

Reflection changes are in the next PR, so this layer retains the existing per-shader group implementations.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>
</details>


shaders/mtl/raster.prelude.metal
```diff
@@ -0,0 +1,32 @@
+
```
louwers: These two newlines in every shader should be removed.

scripts/generate_metal_shader_code.test.ts
```diff
@@ -0,0 +1,45 @@
+import assert from 'node:assert/strict';
```
louwers, line 1: Do we actually run this test on CI?

.pre-commit-config.yaml
```diff
@@ -10,7 +10,8 @@ repos:
       - id: check-yaml
         args: [ --allow-multiple-documents, --unsafe ]
       - id: end-of-file-fixer
-        exclude: 'vendor/.*'
+        # Metal fragments preserve the exact bytes of the original shader strings.
```
louwers: Remove this.

shaders/mtl/README.md
```diff
@@ -0,0 +1,22 @@
...
+build. It stores each shared fragment once, compresses the concatenation with
+zlib, and emits C++ source accessors. Generated C++ belongs in the build tree.
+To inspect the output manually:
+
+```sh
+node shaders/generate_metal_shader_code.mjs --out /tmp/metal-shader-source.cpp
+node --test shaders/generate_metal_shader_code.test.mjs
+```
+
+Shader text is preserved byte for byte, including comments, whitespace, and
```
louwers: Slop. This file should just describe the tooling, not the result of one run.

PR #4571, opened 2026-09-08
perf(core): avoid sorting features with constant sort keys

### LINKED ISSUE #57: Unnecessary sorting of features if `sort-key` is constant, but not `undefined`
*(This is a reminder about 2 performance bugs discovered and discussed in https://github.com/maplibre/maplibre-gl-js/pull/78)*

---

I checked if the sorting-bug is also present in maplibre-gl-native.

This should already be avoiding bug “1.“, but bug “2.” is still present:

https://github.com/maplibre/maplibre-gl-native/blob/cd8d56467962cfb4c99418b0226d9ea2adb48c30/src/mbgl/layermanager/line_layer_factory.cpp#L30 

\- it can probably be modified to `![...].isConstant()`. 

Other instances exist, too: Just look for `circleSortKey`, `symbolSortKey`, `lineSortKey`, `fillSortKey`.

For circles and symbols it’s also in the renderer:

https://github.com/maplibre/maplibre-gl-native/blob/cd8d56467962cfb4c99418b0226d9ea2adb48c30/src/mbgl/renderer/layers/render_circle_layer.cpp#L128

I’m currently not willing (or able) to spend time to send a PR though

NEKOYASAN:
fixes #57

## Details
As mentioned in #57, an unnecessary sort was executed even when every feature had the same sort key. With an explicit constant sort key, each feature was inserted via `lower_bound` (insertion sort), making layout O(n^2).

With this change, it becomes O(n) when the sort key is constant.

When `lower_bound` was used previously, if a feature with the same `sort-key` existed, it was added before that feature.
To maintain this conventional insertion behavior, the array is reversed once after adding the elements to the end, so that rendering output stays identical to before.

Circle layouts insert directly at the front of the list.

Symbol placement also skips searches for already ordered ranges, and rendering groups segments with constant keys while preserving cross-tile drawing order and `symbol-z-order` behavior.

The table below shows the results of measuring the time it took to generate the layout for a large number of features with the same sort-key specified in a Debug build.

| Layer type | feature count | before | after |
|---|---:|---:|---:|
| fill | 2,000 | 20.63 ms | 0.57 ms |
| fill | 8,000 | 276.14 ms | 2.29 ms |
| line | 2,000 | 18.00 ms | 0.56 ms |
| line | 8,000 | 276.53 ms | 2.23 ms |
| circle | 2,000 | 11.82 ms | 0.35 ms |
| circle | 8,000 | 169.94 ms | 1.36 ms |
| symbol | 2,000 | 189.61 ms | 4.31 ms |
| symbol | 8,000 | 2,906.93 ms | 16.89 ms |

## Validations 
- New parameterized unit tests (`test/layout/sort_key.test.cpp`) cover undefined / constant / camera / data-driven sort keys for fill, line, circle, and symbol layouts, asserting the preserved feature order and the generated sort-key ranges.
- New render test `symbol-sort-key/constant-tile-boundary` verifies cross-tile text/icon ordering, which per-tile serial keys would break.

## Note


louwers (commented):
How did you run the benchmark?

Could you try updating the C++ benchmark?

NEKOYASAN:
@louwers I was comparing before/after by own scripts.
I created benchmark code that performs a similar process and compared the results before and after the change again.

| Layer | Features | Before (ms) | After (ms) |
| --- | ---: | ---: | ---: |
| fill | 2,000 | 18.02 | 0.57 |
| fill | 8,000 | 281.64 | 2.22 |
| line | 2,000 | 18.33 | 0.56 |
| line | 8,000 | 278.54 | 2.24 |
| circle | 2,000 | 10.38 | 0.35 |
| circle | 8,000 | 169.81 | 1.42 |
| symbol | 2,000 | 189.22 | 4.18 |
| symbol | 8,000 | 2952.61 | 16.49

louwers:
By the way, do you use constant sort keys? If so, why?

NEKOYASAN:
To be honest, I just worked on this because it caught my eye while I was browsing old issues.
I think it is commonly used like `["get", "priority"]`.

It is very embarrassing, but I must admit that just once, I mistakenly thought that sort-key worked across layers, and I worked really hard to write constant sort-keys into everything...

Issue #4498, opened 2026-08-17
`file:///android_asset` not understood as pointing to app's asset folder (like `asset://` does)


westnordost:
MapLibre-Native on Android does not understand `file:///android_asset` to point to the app's asset folder.

MapLibre-Native on Android already has the functionality to get resources from the app's asset folder, using the `asset://` uri schema. `file:///android_asset` is mentioned in the documentation [here](https://developer.android.com/reference/android/webkit/WebSettings.html#setAllowFileAccess(boolean)) in the context of WebView. Yeah, that's thin, but `asset://` is not mentioned at all.

**MapLibre Android Version**: any, reproducible in 13.5.0

### Reproduction

1. Place glyphs font resources into the app's `assets/` folder, in the subfolder `glyphs`.

2. Have a style json which refers to these resources. E.g. in the style.json, have
```json
  "glyphs": "file:///android_asset/glyphs/{fontstack}/{range}.pbf",
```

3. Load the map. You will see that MapLibre-Native fails to find these resources.

In comparison, **this** works:
```json
  "glyphs": "asset://glyphs/{fontstack}/{range}.pbf",
```

### Log output

e.g.

```shell
E  {RenderThread 10}[Style]: Failed to load glyph range 64256-64511 for font stack Roboto Regular:( )
```

### Additional context

Compose Multiplatform Resources places its resources into the Android assets folder (on Android).

When one resolves a URI within that folder via [Res.getUri](https://kotlinlang.org/docs/multiplatform/compose-multiplatform-resources-usage.html#compose-multiplatform-resources-as-android-assets), e.g. `Res.getUri("a/path/in/resources")`, it resolves to `file:///android_asset/composeResources/<app package name>.resources/a/path/in/resources`.

louwers:
`file://` uses a local file path, `asset://` uses the asset dir.

`file:///android_asset` seems to be some Android webkit specific thing? I don't know why MapLibre Native should support it.

westnordost:
`asset://` is not mentioned anywhere in the Android documentation. As far as I know, this protocol was basically invented by various (popular) third party libraries, such as ReactNative, Coil and the like. Sure, it is a de-facto standard, I am not arguing to remove it. `file:///android_asset/` on the other hand is mentioned in Android documentation. Yes, only in context of WebView, though.

The strongest reason to support it in my opinion would be that Compose Multiplatform Resources will return such a path when using `Res.getUri`.

sargunv:
[...]

PR #4572, opened 2026-09-08
fix(core): load file:///C:/ URLs and non-ASCII paths on Windows

sargunv:
## Bug

Two problems with `file:` URLs on Windows.

1. `LocalFileSource` strips `file://` and passes the rest to the filesystem. For `file:///C:/dir/style.json`, the form defined in [RFC 8089 appendix E.2](https://www.rfc-editor.org/rfc/rfc8089#appendix-E.2) and produced by Java's [`Path.toUri()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Path.html#toUri()), Rust's [`Url::from_file_path`](https://docs.rs/url/latest/url/struct.Url.html#method.from_file_path), and [browsers](https://url.spec.whatwg.org/#file-state), that leaves `/C:/dir/style.json`, which is not a valid [Windows path](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats). Only `file://C:/dir/style.json` works, and that is what the existing test builds from `getcwd()`.
2. Percent-decoded paths are UTF-8, but `stat()` and `std::ifstream(std::string)` decode narrow strings with the ANSI code page on Windows (the [native narrow encoding](https://en.cppreference.com/w/cpp/filesystem/path) of `std::filesystem::path`). Paths with characters outside that code page fail to open. [`/utf-8`](https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8) only affects source and execution character sets, not runtime path decoding.

## Repro

On Windows, `Map::getStyle().loadURL("file:///C:/data/style.json")` fails with `NotFound`. Same for `file://C:/data/ké/style.json` with percent-encoded UTF-8.

## Fix

- `util::pathFromUTF8()` builds a `std::filesystem::path` through the `char8_t` constructor on Windows and returns the string unchanged elsewhere. [`u8path`](https://en.cppreference.com/w/cpp/filesystem/path/u8path) is deprecated in C++20 and MSVC warns under `/WX`.
- `read_file`, `readFile`, `copyFile` open through it. `write_file` and `deleteFile` are not used by the file source and are unchanged.
- `requestLocalFile` uses `std::filesystem::status` wherever the deployment target supports it. `MLN_HAS_STD_FILESYSTEM` centralizes the availability check in the filesystem helper header; older Apple deployment targets, including the supported iOS 12 minimum, retain `stat`.
- `LocalFileSource` drops the slash before a drive letter. UNC paths are unchanged.

## Tests

`toAbsoluteURL()` now builds `file:///C:/...` URLs on Windows, so the existing tests cover that form. `LocalFileSource.NonASCIIPath` loads a file from a directory named `ké 地図`. The file-source tests also exercise the filesystem status path on supported POSIX platforms. Windows CI covers drive-letter and UTF-8 conversion behavior.

Related: #4498, the `file:///android_asset/` form, which #4501 handles on the MapLibre Android side.

Carried in maplibre-native-ffi since maplibre/maplibre-native-ffi#562. AI disclosure: prepared with Claude Code (Claude Fable 5.1 for analysis, review, and this description; Claude Opus 5 for code edits under its instructions).

### DIFF
--- a/platform/default/src/mln/storage/local_file_request.cpp
+++ b/platform/default/src/mln/storage/local_file_request.cpp
@@ -3,8 +3,7 @@
 #include <mln/util/filesystem.hpp>
 #include <mln/util/io.hpp>
 
-#if defined(_WIN32)
-#include <filesystem>
+#if MLN_HAS_STD_FILESYSTEM
 #include <system_error>
 #else
 #include <sys/types.h>
@@ -17,7 +16,7 @@ void requestLocalFile(const std::string& path,
                       const ActorRef<FileSourceRequest>& req,
                       const std::optional<std::pair<uint64_t, uint64_t>>& dataRange) {
     Response response;
-#if defined(_WIN32)
+#if MLN_HAS_STD_FILESYSTEM
     std::error_code error;
     std::filesystem::file_status status;
     try {
--- a/platform/default/src/mln/storage/local_file_request.cpp
+++ b/platform/default/src/mln/storage/local_file_request.cpp
@@ -1,12 +1,14 @@
 #include <mln/storage/file_source_request.hpp>
 #include <mln/storage/response.hpp>
+#include <mln/util/filesystem.hpp>
 #include <mln/util/io.hpp>
 
+#if defined(_WIN32)
+#include <filesystem>
+#include <system_error>
+#else
 #include <sys/types.h>
 #include <sys/stat.h>
-
-#if defined(_WIN32) && !defined(S_ISDIR)
-#define S_ISDIR(m) (((m) & S_IFMT) == S_IFDIR)
 #endif
 
 namespace mln {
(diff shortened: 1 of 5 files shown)


platform/default/src/mln/storage/local_file_request.cpp
```diff
@@ -15,12 +17,19 @@ void requestLocalFile(const std::string& path,
                       const ActorRef<FileSourceRequest>& req,
                       const std::optional<std::pair<uint64_t, uint64_t>>& dataRange) {
     Response response;
+#if defined(_WIN32)
+    std::error_code error;
+    const auto status = std::filesystem::status(util::pathFromUTF8(path), error);
```
louwers: Can we just use this on all platforms?

sargunv: Requires bumping min iOS version iirc

sargunv: okay I changed it to "if not ios or ios >= 13" rather than "if windows".

though if you're comfortable with a min ios version bump, I think that's better. iOS 12 is pretty old.

src/mln/util/filesystem.hpp
```diff
@@ -2,8 +2,39 @@
 
 #include <string>
 
+// libc++ filesystem requires macOS 10.15, iOS/tvOS 13, or watchOS 6.
```
louwers, line 5: Honestly it's time to bump the minimum iOS version...

sargunv: Sgtm. Separate PR?

I remember there being something weird with the min macOS version too, I'll see if I can dig that up in my mlnffi history

sargunv: removed the compatibility code in efcbc9cb8098684029aa32ed8686fe380f5c83e9

PR #4575, opened 2026-09-08
fix(core): keep in-use on-demand images registered when a request also has missing images

sargunv:
## Bug

[`ImageManager::getImages()`](https://github.com/maplibre/maplibre-native/blob/57d198149785cb6068ca6d62d478670c64a60358/src/mln/renderer/image_manager.cpp#L123) calls `removeRequestor()` first, which drops the requestor from every entry in `requestedImages`. `checkMissingAndNotify()` re-registers it [only in the branch where nothing is missing](https://github.com/maplibre/maplibre-native/blob/57d198149785cb6068ca6d62d478670c64a60358/src/mln/renderer/image_manager.cpp#L280-L285). A request that mixes on-demand images the manager already has with images it still needs leaves the present ones unregistered. [`reduceMemoryUse()`](https://github.com/maplibre/maplibre-native/blob/57d198149785cb6068ca6d62d478670c64a60358/src/mln/renderer/image_manager.cpp#L176) treats them as unused and evicts them through `onRemoveUnusedStyleImages` while the tile still shows them.

## Repro

Host supplies icons through `onStyleImageMissing`. Enough icons to exceed `DEFAULT_ON_DEMAND_IMAGES_CACHE_SIZE`. A tile requests icons a0..a5 (all present) plus b0..b5 (missing). Next cleanup evicts a0..a5; they get re-requested; then b0..b5 are evicted. This alternates indefinitely. maplibre/maplibre-native-ffi#694, maplibre/maplibre-compose#1344.

## Fix

Register the requestor for the on-demand images that are already present in the same loop that finds the missing ones, whether or not anything is missing. Remove the now-redundant loop in the else branch. Images added with `addImage` are unaffected; they have no `requestedImages` entry.

## Test

`ImageManager.MixedDependenciesRetainProvidedImages`. Fails on main at the first mixed request, passes with the fix. Also checks the images are reclaimed once the requestor is gone.

Carried in maplibre-native-ffi since maplibre/maplibre-native-ffi#695. AI disclosure: prepared with Claude Code (Claude Fable 5.1 for analysis, review, and this description; Claude Opus 5 for code edits under its instructions).

### DIFF
--- a/src/mln/renderer/image_manager.cpp
+++ b/src/mln/renderer/image_manager.cpp
@@ -226,10 +226,10 @@ void ImageManager::checkMissingAndNotify(ImageRequestor& requestor, const ImageR
     for (const auto& dependency : pair.first) {
         if (!images.contains(dependency.first)) {
             missingDependencies.emplace(dependency);
-        } else if (auto it = requestedImages.find(dependency.first); it != requestedImages.end()) {
+        } else if (requestedImages.contains(dependency.first)) {
             // The missing-image path below never registers images that are
             // already present.
-            it->second.emplace(&requestor);
+            requestedImages.at(dependency.first).emplace(&requestor);
         }
     }
 
--- a/src/mln/renderer/image_manager.cpp
+++ b/src/mln/renderer/image_manager.cpp
@@ -226,6 +226,10 @@ void ImageManager::checkMissingAndNotify(ImageRequestor& requestor, const ImageR
     for (const auto& dependency : pair.first) {
         if (!images.contains(dependency.first)) {
             missingDependencies.emplace(dependency);
+        } else if (auto it = requestedImages.find(dependency.first); it != requestedImages.end()) {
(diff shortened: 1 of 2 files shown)


src/mln/renderer/image_manager.cpp
```diff
@@ -226,6 +226,10 @@ void ImageManager::checkMissingAndNotify(ImageRequestor& requestor, const ImageR
     for (const auto& dependency : pair.first) {
         if (!images.contains(dependency.first)) {
             missingDependencies.emplace(dependency);
+        } else if (auto it = requestedImages.find(dependency.first); it != requestedImages.end()) {
```
louwers: Use C++20 `contains` (would probably be a good idea to add C++ version in `AGENTS.md`).

louwers (changes requested):
Looks legit, but see comment.

PR #4602, opened 2026-09-09
android: read byte ranges from APK assets

### LINKED ISSUE #4360: `AssetManagerFileSource` does not support byte-range reads
**Describe the bug**

Using a PMTiles archive bundled in the app's `assets/` folder (`pmtiles://asset://filename.pmtiles`) causes a fatal native crash in the `PMTilesFileSour` thread. The crash occurs every time, immediately on map load.

**To Reproduce**
1. Place a valid PMTiles archive (e.g. [watercolor.pmtiles](https://demotiles.maplibre.org/pmtiles/raster/watercolor.pmtiles)) in `src/main/assets/`.
2. Add a `RasterSource` referencing it:

```kotlin
val source = RasterSource(
    "watercolor",
    "pmtiles://asset://watercolor.pmtiles",
    256
)
style.addSource(source)
style.addLayer(RasterLayer("watercolor", "watercolor"))
```

3. Run the app. The process crashes.

**Expected behavior**

The source loads and tiles render, matching the behavior of `pmtiles://https://` and `pmtiles://file://`.

**Observed crash**

Thread: `PMTilesFileSour`

```
#06  mbgl::util::decompress(...)+592
#07  mbgl::PMTilesFileSource::Impl::getMetadata(...)+328
```

Fatal unhandled C++ exception thrown from `decompress()` inside `getMetadata()`.

**Root cause**

`PMTilesFileSource` uses byte-range reads for all internal requests (header, metadata, directory entries, tiles). It sets `resource.dataRange` on each `Resource` before calling `getFileSource()->request(resource, ...)`.

`AssetManagerFileSource::Impl::request` (in `platform/android/MapLibreAndroid/src/cpp/asset_manager_file_source.cpp`) ignores `dataRange` entirely:

```cpp
void request(const std::string& url, ActorRef<FileSourceRequest> req) {
    const std::string path = mbgl::util::percentDecode(url.substr(8));
    Response response;
    if (AAsset* asset = AAssetManager_open(assetManager, path.c_str(), AASSET_MODE_BUFFER)) {
        response.data = std::make_shared<std::string>(
            reinterpret_cast<const char*>(AAsset_getBuffer(asset)),
            AAsset_getLength64(asset));  // returns entire file
        AAsset_close(asset);
    }
    // ...
}
```

It always returns the complete file. When `getMetadata` requests the compressed JSON metadata at a specific byte offset (e.g. bytes 3215–3441 in `watercolor.pmtiles`), it receives the full 18 MB file instead. `decompress()` is called on data beginning with the PMTiles magic bytes `PMTiles\x03…`, which are not gzip-compressed. `zlib::inflate` fails and throws `std::runtime_error`, which propagates as an unhandled exception and terminates the process.

Note: `getHeader` (which requests bytes 0–511) also receives the full file, but happens to succeed because the PMTiles header occupies the first 127 bytes of the file and is read as a fixed binary struct without decompression.

**Workaround**

Use `pmtiles://file://` with an absolute path to a file on device storage:

```kotlin
val path = context.getExternalFilesDir(null)?.absolutePath + "/watercolor.pmtiles"
val source = RasterSource("watercolor", "pmtiles://file://$path", 256)
```

Files on device storage go through `LocalFileSource` → `requestLocalFile()` → `util::readFile()`, which does implement `seekg`-based range reads. Remote archives via `pmtiles://https://` also work correctly (HTTP 206 range responses).

**Suggested fix**

Add range-read support to `AssetManagerFileSource::Impl::request`. The Android NDK provides the necessary API:

```cpp
void request(const std::string& url,
             const std::optional<std::pair<uint64_t, uint64_t>>& dataRange,
             ActorRef<FileSourceRequest> req) {
    const std::string path = mbgl::util::percentDecode(url.substr(8));
    Response response;
    if (AAsset* asset = AAssetManager_open(assetManager, path.c_str(), AASSET_MODE_RANDOM)) {
        if (dataRange) {
            uint64_t offset = dataRange->first;
            uint64_t length = dataRange->second - dataRange->first + 1;
            AAsset_seek64(asset, static_cast<off64_t>(offset), SEEK_SET);
            std::string data(length, '\0');
            AAsset_read(asset, data.data(), length);
            response.data = std::make_shared<std::string>(std::move(data));
        } else {
            response.data = std::make_shared<std::string>(
                reinterpret_cast<const char*>(AAsset_getBuffer(asset)),
                AAsset_getLength64(asset));
        }
        AAsset_close(asset);
    } else {
        response.error = std::make_unique<Response::Error>(
            Response::Error::Reason::NotFound, "Could not read asset");
    }
    req.invoke(&FileSourceRequest::setResponse, response);
}
```

The `AssetManagerFileSource::request` public method would need to pass `resource.dataRange` through to `Impl::request` via the actor invoke call.

**Platform**

- Platform: Android
- MapLibre Native version: current `main` (confirmed with prebuilt `.so` from the test app)
- Device: arm64-v8a emulator/physical device (crash is in native `.so`)

**Additional context**

The same limitation likely affects any other file source that reads assets via `AssetManagerFileSource` but is unrelated to PMTiles-specific behavior—PMTiles just requires range reads to function at all.

**AI Disclosure**

Assisted with debugging and parsing crash logs by Claude Sonnet 4.6

sargunv:
## Bug

MapLibre Android's `AssetManagerFileSource` opens every asset with [`AAssetManager_open`](https://developer.android.com/ndk/reference/group/asset#aassetmanager_open) in `AASSET_MODE_BUFFER` and returns the whole file. It ignores `Resource::dataRange`. PMTiles reads an archive piecewise: the [127-byte header](https://github.com/protomaps/PMTiles/blob/main/spec/v3/spec.md) first, then directories and tiles at the offsets the header and directories give. Each of those is a ranged request, and each one got the entire archive back, so `pmtiles://asset://` sources never loaded.

## Repro

Put any `.pmtiles` file in `src/main/assets/` and add `RasterSource("x", "pmtiles://asset://x.pmtiles")` or the vector equivalent. The source reports a load error and renders nothing, while the same archive works through `pmtiles://file://` from device storage. #4360.

## Fix

Open with `AASSET_MODE_RANDOM`, seek to the requested offset with [`AAsset_seek64`](https://developer.android.com/ndk/reference/group/asset#aasset_seek64), and read only the requested length, clamped to the asset size. Requests without a range still return the whole asset. An inverted or out-of-range request returns an empty body, and a failed seek or short read reports `Response::Error::Reason::Other`; a missing asset still reports `NotFound`. This matches what `util::readFile` does for `file://` paths, so `file://` and `asset://` behave the same behind `pmtiles://`.

The PMTiles page in the MapLibre Android docs no longer says assets are unsupported.

## Test

`StyleLoaderTest.testRasterSourceFromPMTilesAsset` (MapLibre Android instrumentation tests) adds a `RasterSource` for `pmtiles://asset://geography-class-png.pmtiles`, a copy of the existing core test fixture, and asserts the source loads, which only happens once the header and TileJSON reads succeed.

Resolves #4360. Composes with #4501, which adds the `file:///android_asset/` form to the same source.

AI disclosure: prepared with Claude Code (Claude Fable 5.1). Draft until I have reviewed the generated changes.


louwers (approved):
Very nice, thanks!

PR #4604, opened 2026-09-09
Add semiliteral expression to core and mobile SDKs

### LINKED ISSUE #4401: Add `semiliteral` expression for computed arrays and objects

Many style properties accept arrays, but today these arrays must be static when authored in expressions. While `literal` can construct arrays and objects, it intentionally does not evaluate expressions within them. This makes some data-driven styling use cases impossible or requires verbose workarounds.

As the style specification adding support for `semiliteral` in MapLibre Native would be nice for feature parity.

Implement support for the `semiliteral` expression operator in MapLibre Native.

`semiliteral` should evaluate expressions within arrays and objects while preserving the existing behavior of `literal`, enabling dynamically computed array- and object-valued expressions in a backwards-compatible way.

Workarounds such as selecting between precomputed arrays or encoding values in other forms are often verbose, limited, or cannot express all desired styling behavior.

Changing the semantics of `literal` is not a viable alternative because it would break existing styles.

**Additional context**

This is intended to bring MapLibre Native in line with https://github.com/maplibre/maplibre-style-spec/issues/950.

sargunv:
### Description

Adds `semiliteral` so array-valued style properties can calculate their elements with expressions, including `text-offset` and `text-variable-anchor-offset`. Matches the behavior shipped in maplibre/maplibre-style-spec#951: array elements are evaluated, while non-array arguments remain literal.

Fixes #4401, unblocks https://github.com/maplibre/maplibre-compose/issues/151

### Validation

semiliteral fixtures adapted from the style-spec PR with native serialization expectations.

### AI disclosure

Built with GPT-6 Astra in Codex


louwers (approved):

louwers:
Note to self: need to update the `sdk-support` table once this is release in MapLibre iOS and MapLibre Android.

sargunv:
> Note to self: need to update the sdk-support table once this is release in MapLibre iOS and MapLibre Android.

I think it's out of date for JS too. I'll spin up a draft.

PR #4612, opened 2026-09-10
core: add first version of C plugin API with ngon layer as example

louwers:
This pull requests add a C based [plugin API](https://github.com/maplibre/maplibre-native/pull/4612/changes#diff-7ec62784cf94daa19fd58bd6378d856d75dfd997ca48ce8cfdf1265bc1e09827) along with required  C++ changes to support it. All this code is guarded by `MLN_USE_PLUGIN` ifdefs so we can merge this first version of the plugin API without shipping it.

I have also created a proof-of-concept [template repository](https://github.com/louwers/maplibre-native-plugin-template/) where all the build tooling to create and release plugins for MapLibre Android and MapLibre iOS is set up. My hope is that this will kick off a plugin ecosystem for MapLibre Native. However this is out of scope for now so we can focus on the core changes first.

I have added a relatively simple "n-gon layer" which is installed in the GLFW app. Run it with:

```
bazel run --//:plugins=true --//:renderer=metal \
    --linkopt=-L/opt/homebrew/opt/icu4c/lib \
    //platform/glfw:glfw_app -- \
    --style "$PWD/plugins/ngon-layer/examples/interpolation.json" \
    --lon 0 --lat 0 --zoom 10.5
```

image

The idea is to iteratively expand (and modify) the plugin API so it can support various use cases. Until we are happy with its capabilities, then release and stabilize it.

Assisted by GPT-6 Astra


plugins/ngon-layer/tests/ngon.test.cpp
```diff
@@ -0,0 +1,109 @@
+#include "ngon_layer.hpp"
+#include <cmath>
+#include <cstdlib>
+#include <iostream>
+#include <string_view>
+#include <vector>
+
+#define CHECK(condition)                                   \
```
louwers, line 8: Not sure if reinventing a test framework is the best idea.

include/mln/shaders/shader_defines.hpp
```diff
@@ -545,6 +545,9 @@ enum {
 };
 
 static constexpr uint32_t maxAttributeCountPerShader = std::max({
+#if MLN_WITH_PLUGINS
+    16u,
```
louwers: Magic number

zarechnyy (approved):

louwers:
So I've guarded it off completely with `MLN_WITH_PLUGINS`. My goal is to merge it, have us hack on it and extend it until we're happy about the design, implementation, scope and stability, before really releasing it.

After this is merged I would like to enable the plugin API for iOS and Android pre-releases.

TimSylvester (approved):

PR #4614, opened 2026-09-11
darwin: runtime terrain API for iOS and macOS (MLNTerrain, MLNStyle.terrain, MLNMapView.setTerrain)

koekiedaf:
`TERRAIN.md` lists "Runtime styling API for terrain (`setTerrain`) on iOS/macOS (Android has it)" under the work left before merging. Today terrain can only be switched on from the style JSON on Apple platforms, so an app cannot toggle 3D, change the exaggeration from its own UI, or pick a load mode. This adds the missing binding, mirroring the Android API one for one.

### What it adds

**`MLNTerrain`** (`platform/darwin/src/MLNTerrain.{h,mm}` plus a private header) is the darwin counterpart of `Terrain.kt`: a value object holding the raster-dem source identifier and the exaggeration multiplier, with copying, equality, hashing and a description, and the usual conversion to and from `mln::style::Terrain`.

**`MLNStyle.terrain`**, a nullable property, reads and writes `mln::style::Style::setTerrain` / `getTerrain`, the same calls `NativeMapView::setTerrain` / `removeTerrain` / `getTerrainSourceId` / `getTerrainExaggeration` make on Android. Setting it to `nil` removes terrain.

**`MLNMapView`**, on iOS and macOS both:

- `-setTerrainWithSourceIdentifier:exaggeration:` (`setTerrain(sourceIdentifier:exaggeration:)` in Swift), a convenience that goes through the style property and logs a warning instead of crashing when the style has not finished loading;
- `terrainLoadMode`, binding `Map::setTerrainLoadMode` / `getTerrainLoadMode`;
- `terrainSkirtLength`, binding `Map::setTerrainSkirtLength` / `getTerrainSkirtLength`.

**`MLNTerrainLoadMode`** and **`MLNTerrainSkirtLength`** are declared in `MLNTypes.h` with their ordinals pinned to `mln::TerrainLoadMode` and `mln::TerrainSkirtLength`, and a `static_assert` in each `MLNMapView.mm` catches a future reordering of the core enums at compile time rather than as a silent runtime mismatch. Their documentation is the Android enums' documentation, adapted.

### Tests

`platform/ios/test/MLNTerrainTests.m` covers the value object (round trip, default exaggeration, equality and inequality), the style property (set, read back, clear), the convenience method including its `nil` path, both map-level knobs, and the enum ordinals.

### Notes

- No behaviour changes for anything that does not call the new API: a style with a `terrain` root property keeps working exactly as before.
- `TERRAIN.md` gains a Swift snippet in its API Usage section beside the existing C++, Android and style JSON ones, and loses the outstanding-work line this closes.
- Verified by building the iOS and macOS dynamic XCFrameworks with `--//:renderer=metal` and running the API against a real style on the iOS simulator and on an AppKit map view.


koekiedaf:
Reopening from a branch that carries only this change: the previous head also carried three unrelated cherry-picks (#4556, #4557, #4559) from my own build branch.

PR #4626, opened 2026-09-12
Transform::easeTo: honour the caller's easing instead of silently reusing flyTo

AndrewN81:
## Summary

`Transform::easeTo` reuses `flyTo(camera, animation, /*linearZoom=*/true)` for every unbounded, non-gesture, timed ease, "to achieve constant ground speed". At constant zoom the van Wijk path's ground speed is **not** constant, and on that path the caller's `AnimationOptions.easing` is never consulted. So an explicit easing — e.g. the linear `UnitBezier{0,0,1,1}` the Android SDK passes for `easeCamera(update, duration, easingInterpolator = false)` — is silently replaced by a slow-fast-slow profile.

This change takes the `flyTo` shortcut only when no easing was requested, so callers who ask for a specific curve get it. Callers who don't are unchanged.

## Evidence (Android, `easeCamera(update, 5000, false)`, 300 m at constant zoom 15.5, camera centre sampled at 10 Hz)

Velocity as a fraction of the mean, by half-second:

| | 0 | 0.5 | 1.0 | 1.5 | 2.0 | 2.5 | 3.0 | 3.5 | 4.0 | 4.5 s |
|---|---|---|---|---|---|---|---|---|---|---|
| before, Galaxy S10e | 0.64 | 0.86 | 0.98 | 1.17 | 1.18 | 1.27 | 1.17 | 1.04 | 0.86 | 0.74 |
| before, Pixel 6 emulator | 0.65 | 0.82 | 0.97 | 1.15 | 1.24 | 1.25 | 1.14 | 1.05 | 0.90 | 0.71 |
| control: same 300 m via `moveCamera` every vsync | 1.03 | 1.00 | 0.99 | 1.00 | 1.03 | 1.00 | 1.00 | 0.99 | 0.99 | 0.98 |
| **after** | 0.96 | 1.01 | 0.99 | 1.01 | 1.00 | 0.96 | 1.02 | 0.98 | 1.02 | 1.01 |

How it was noticed: a navigation app issuing one linear ease per GPS fix (aimed a little ahead of the rider) showed a visible speed step at every fix — each restart reset the map to the slow phase of the curve. With this change the within-fix velocity is flat.

## Notes

- The solver and the JNI mapping are fine: `UnitBezier{0,0,1,1}` solves to exactly `t`, and `NativeMapView::easeTo` does set it for `easing == false`; the branch above simply never reached it.
- iOS `-[MLNMapView setCamera:withDuration:animationTimingFunction:]` passes an explicit easing too, so it is affected the same way.

### DIFF
--- a/src/mln/map/transform.cpp
+++ b/src/mln/map/transform.cpp
@@ -110,8 +110,13 @@ void Transform::easeTo(const CameraOptions& inputCamera, const AnimationOptions&
     CameraOptions camera = inputCamera;
 
     Duration duration = animation.duration.value_or(Duration::zero());
-    if (state.getLatLngBounds() == LatLngBounds() && !isGestureInProgress() && duration != Duration::zero()) {
-        // reuse flyTo, without exaggerated animation, to achieve constant ground speed.
+    // Reuse flyTo (without the zoom-out) for a plain timed ease — but only when the caller has
+    // not asked for a specific easing curve. flyTo moves along the van Wijk path, whose ground
+    // speed at constant zoom is not constant (slow-fast-slow), so an explicit easing such as a
+    // linear one would otherwise be silently replaced. Measured on Android: a 5 s "linear" ease
+    // ran at 0.64x its mean speed in the first 500 ms and 1.27x mid-way.
+    if (state.getLatLngBounds() == LatLngBounds() && !isGestureInProgress() && duration != Duration::zero() &&
+        !animation.easing) {
         flyTo(camera, animation, true);
         return;
     }


louwers:
Could you maybe share a video of the difference?

AndrewN81:
Here's a side by side (Android 15 emulator, the test app built from this branch with and without the change). The camera does one linear easeCamera(update, 1100, false) per simulated GPS fix at 40 m/s and constant zoom, each ease aimed at the next fix so it's replaced just before it finishes, the way a nav app drives it. The graph is camera ground speed ÷ vehicle speed. On main it swings about 0.7–1.3× every fix; with the change it stays at 1.0×.

https://github.com/user-attachments/assets/56cb6448-d08e-4025-86f0-e6135fcea870

PR #4621, opened 2026-09-11
fix(ios,core): Use @autoreleasepool blocks on worker threads

adrian-cojocaru:
There are a few sources of `@autoreleasepool content` (metal texture, strings, etc) generated on worker threads without any `@autoreleasepool` in scope.

PR #4628, opened 2026-09-13
core: Avoid rendering fully hidden symbol drawables

### LINKED ISSUE #3223: Text rendering performance issues
We just migrated from Mapbox to Maplibre and discovered degraded performance especially on iOS (React Native app for both platforms) when zooming in/out or quickly panning the map. We have 3 symbol layers with small/medium/large variants of POI markers to let collision use the largest possible icon available for a nice map UX with hundreds of POIs in a city. I discovered the rendering performance goes massively down when I add a text in the style and especially when I set text halo (`textHaloWidth`, `textHaloColor`) so I had to turn off text halo effects to keep at least some acceptable performance. The Mapbox component seems to have some more performance optimization around text collision detection or text rendering in general. It's surprising the Android performance is still acceptable (even though worse than Mapbox) and it's sluggish on iPhones.

I'd appreciate any recommendations on how to improve the performance by changing the style - I ended up with this:
```jsx
const useTextHalo = Platform.OS === 'android';

const baseStyle: SymbolLayerStyle = {
  textHaloWidth: 2,
  textHaloBlur: 0,
  textAnchor: 'top',
  textMaxWidth: 10,
  textFont: ['get', 'textFont'],
  iconAllowOverlap: false,
  textAllowOverlap: false,
  iconOptional: false,
  iconAnchor: ['get', 'iconAnchor'],
  symbolSortKey: ['get', 'priority'],
  textOffset: [0, 0.1],
};

export const pinLargeStyle: SymbolLayerStyle = {
  ...baseStyle,
  textField: ['get', 'title'],
  textSize: SIZE_L,
  textColor: ['get', 'textColor'],
  textHaloColor: ['get', 'textHaloColor'],
  iconImage: ['get', 'large'],
  iconSize: ICON_SIZE_MULTIPLIER,
};

export const pinMediumStyle: SymbolLayerStyle = {
  ...baseStyle,
  textField: ['get', 'title'],
  iconImage: ['get', 'medium'],
  textSize: ['case', ['==', ['get', 'type'], 'tour'], SIZE_L, SIZE_M],
  textColor: ['get', 'textColor'],
  iconColor: ['get', 'iconColor'],
  textHaloColor: ['get', 'textHaloColor'],
  iconSize: ICON_SIZE_MULTIPLIER,
};

export const pinSmallStyle: SymbolLayerStyle = {
  iconImage: ['get', 'small'],
  iconColor: ['get', 'iconColor'],
  symbolSortKey: ['get', 'priority'],
  iconAnchor: 'center',
  iconSize: ICON_SIZE_MULTIPLIER,
};

if (!useTextHalo) {
  [pinLargeStyle, pinMediumStyle, pinSmallStyle].forEach((style) => {
    delete style.textHaloWidth;
    delete style.textHaloBlur;
    delete style.textHaloColor;
    delete style.textColor;
  });
}


...

  return (
    <>
      <Images images={images} />
      <ShapeSource id={props.layerId} shape={featuresCollection} onPress={onFeaturePress} hitbox={useMemo(() => ({ width: 0, height: 0 }), [])}>
        <SymbolLayer
          id={`SymbolLayer${props.layerId}Small`}
          minZoomLevel={props.minZoomLevel}
          maxZoomLevel={props.maxZoomLevel}
          style={pinSmallStyle}
        />
        <SymbolLayer
          id={`SymbolLayer${props.layerId}Medium`}
          minZoomLevel={props.minZoomLevel}
          maxZoomLevel={props.maxZoomLevel}
          style={pinMediumStyle}
        />
        <SymbolLayer
          id={`SymbolLayer${props.layerId}Large`}
          minZoomLevel={props.minZoomLevel}
          maxZoomLevel={props.maxZoomLevel}
          style={pinLargeStyle}
        />
      </ShapeSource>
    </>
  );
```

This is the intended result which worked well with Mapbox in terms of performance but is not acceptable with MapLibre on iOS:

image

NEKOYASAN:
close #3223 

## Details

Collision-rejected symbols previously kept their drawables enabled, even when every referenced vertex was fully transparent. These drawables still incurred attribute updates, uniform preparation, and draw calls.

This change disables drawables whose opacity data indicates that all referenced symbols are fully transparent and unplaced. Symbols fading out or starting to fade in remain enabled. Visibility is checked again when existing drawables are updated, allowing them to reappear when collisions are resolved.

Disabled drawables also skip uniform preparation. Consolidated uniform arrays contain only enabled drawables.

The benchmark below places icons at the same coordinates with distinct sort keys. Results are median render times from five repetitions of 50 frames, measured with the Metal backend in a Debug build.

| Feature count | Allow overlap | Before | After | Draw calls before → after |
|---|---|---:|---:|---:|
| 1 | false | 0.48 ms | 0.48 ms | 1 → 1 |
| 256 | false | 4.19 ms | 2.22 ms | 256 → 1 |
| 1,024 | false | 16.60 ms | 8.38 ms | 1,024 → 1 |
| 1,024 | true | 17.32 ms | 17.46 ms | 1,024 → 1,024 |

These results represent a collision-heavy workload, rather than an expected speedup for every map.

## Validations

- Four new tests in `test/renderer/symbol_rendering.test.cpp` cover collision-rejected icons, SDF icons, text, and combined text/icons; visibility changes after zooming; fade-state handling; and reappearance after another layer stops blocking symbols.
- Compared 132 existing render-test images before and after on Metal. 130 were byte-identical. The remaining two differed by at most one color value in 3–4 pixels; repeated runs also exhibited this variation with the same binary.
- A separate fixed-time rendering harness compared 105 frames covering fades, collision changes, and reappearance. All frames were byte-identical.

## Note

~For the Metal benchmark measurements, both comparison binaries used the same per-render autorelease pool. That benchmark-local workaround is not included in this PR. Without pool management, the current command-line Metal runner stalls while acquiring a command buffer.~

### DIFF
--- a/test/renderer/symbol_rendering.test.cpp
+++ b/test/renderer/symbol_rendering.test.cpp
@@ -1,5 +1,6 @@
 #include <mln/test/map_adapter.hpp>
 #include <mln/test/stub_file_source.hpp>
+#include <mln/test/stub_map_observer.hpp>
 #include <mln/test/util.hpp>
 
 #include <mln/gfx/headless_frontend.hpp>
@@ -10,6 +11,10 @@
 #include <mln/style/style.hpp>
 #include <mln/util/io.hpp>
 #include <mln/util/run_loop.hpp>
+
+#if MLN_RENDER_BACKEND_METAL
+#include <Foundation/Foundation.hpp>
+#endif
 
 #include <algorithm>
 #include <sstream>
@@ -47,7 +52,12 @@ std::string collidingSymbolsStyle(std::size_t count, double spacing, SymbolKind 
 
 class SymbolRenderingTest {
 public:
-    explicit SymbolRenderingTest(std::size_t count, double spacing = 0, SymbolKind kind = SymbolKind::Icon) {
+    explicit SymbolRenderingTest(std::size_t count,
+                                 double spacing = 0,
+                                 SymbolKind kind = SymbolKind::Icon,
+                                 MapMode mode = MapMode::Static,
+                                 MapObserver& observer = MapObserver::nullObserver())
+        : map{frontend, observer, fileSource, MapOptions().withMapMode(mode).withSize(frontend.getSize())} {
         fileSource->glyphsResponse = [](const Resource&) {
             Response response;
             response.data = std::make_shared<const std::string>(util::read_file("test/fixtures/resources/glyphs.pbf"));
@@ -63,11 +73,8 @@ public:
 
     util::RunLoop loop;
     std::shared_ptr<StubFileSource> fileSource = std::make_shared<StubFileSource>();
-    HeadlessFrontend frontend{Size{256, 256}, 1.0f};
-    MapAdapter map{frontend,
-                   MapObserver::nullObserver(),
-                   fileSource,
-                   MapOptions().withMapMode(MapMode::Static).withSize(frontend.getSize())};
+    HeadlessFrontend frontend{mln::Size{256, 256}, 1.0f};
+    MapAdapter map;
 };
 
 } // namespace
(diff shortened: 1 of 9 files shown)


test/renderer/symbol_rendering.test.cpp
```diff
@@ -0,0 +1,211 @@
...
+        }
+    };
+
+    // Exercise the real repaint/placement scheduling, but fail rather than hang
+    // if hidden drawables never stop drawing or never become visible again.
+    const auto deadline = Clock::now() + Seconds{10};
+    while (phase != Phase::Complete && Clock::now() < deadline) {
+#if MLN_RENDER_BACKEND_METAL
+        // Headless continuous rendering has no application autorelease pool.
+        const auto pool = NS::TransferPtr(NS::AutoreleasePool::alloc()->init());
```
louwers, line 172: This may not be needed anymore after https://github.com/maplibre/maplibre-native/pull/4621

test/renderer/symbol_rendering.test.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    auto blocker = std::make_unique<style::SymbolLayer>("blocker", "points");
+    blocker->setIconImage({"marker"});
+    auto* blockerLayer = blocker.get();
+    test.map.getStyle().addLayer(std::move(blocker));
+    const auto hidden = test.frontend.render(test.map);
+    EXPECT_EQ(hidden.stats.numDrawCalls, 1);
+    EXPECT_EQ(initial.image, hidden.image);
+
+    blockerLayer->setVisibility(style::VisibilityType::None);
```
louwers, line 207: Looks fishy, you're retrieving a raw pointer from a unique pointer, then moving the unique pointer and then using the raw pointer.

test/renderer/symbol_rendering.test.cpp
```diff
@@ -0,0 +1,211 @@
...
+enum class SymbolKind {
+    Icon,
+    SdfIcon,
+    Text,
+    TextAndIcon
+};
+
+std::string collidingSymbolsStyle(std::size_t count, double spacing, SymbolKind kind) {
+    std::ostringstream json;
+    json << R"({"version":8,"sources":{"points":{"type":"geojson","maxzoom":0,"data":{
```
louwers, line 35: There may be a more readable way to construct this JSON.

louwers (commented):
Sounds promising.

Asked @sjg-wdw and @TimSylvester if they could maybe take a look as well...

sjg-wdw:
One problem we've seen with some of these PRs is tests not being able to cover interpolation.
Can we figure out a way to test halfway between states?

test/renderer/symbol_rendering.test.cpp
NEKOYASAN: It is not resolved in #4621, and it seems better to enable `MLN_CREATE_AUTORELEASEPOOL` instead, but I have not yet been able to examine whether turning this ON will cause any issues.

test/renderer/symbol_rendering.test.cpp
NEKOYASAN: It definitely looked suspicious, so I modified it to call getLayer again.

test/renderer/symbol_rendering.test.cpp
NEKOYASAN: Replaced the string concatenation with a `FeatureCollection`, a `GeoJSONSource`, and `SymbolLayer` setters.

NEKOYASAN:
@sjg-wdw Thank you for your review!

I have made it possible to perform verification for each rendering time by explicitly controlling the rendering time.
In the fade process, I make sure to verify that while the target icons change in opacity, the non-target icons remain fully opaque.

I'm a bit concerned about its high level of complexity, but how does this look?  https://github.com/maplibre/maplibre-native/pull/4628/commits/7ed65938172e400e1b4d936f9c4421d58ee55ae2

PR #4629, opened 2026-09-14
Fix PMTiles directory cache evicting a directory still in use (#4421)

### LINKED ISSUE #4421: PMTilesFileSource: "invalid map<K, T> key" on large archives - directory LRU evicts entries still in use (duplicate keys in directory_cache_control)
**Symptom.** Against a planet-scale PMTiles archive (https://build.protomaps.com/20260716.pmtiles,
~136 GB, 177,061,167 tile entries, v3, gzip/gzip), every vector tile load fails with
`Error parsing PMTiles directory: invalid map<K, T> key`. The archive is valid: the header passes
inspection (magic/version/offsets/compression sane) and an independent PMTiles v3 reader parses
the same archive without errors.

**The message is not a parse failure.** `invalid map<K, T> key` is the MSVC STL message for
`std::map::at()` on a missing key (`std::out_of_range`). The compiled object for
`platform/default/src/mbgl/storage/pmtiles_file_source.cpp` carries BOTH vocabularies â€” the
vendored pmtiles.hpp parse exceptions ("malformed directory exception" etc.) AND the STL messages
â€” and the logged text is the STL one, so `pmtiles::deserialize_directory` succeeded and a
`map::at` threw afterwards. The catch at `getDirectory` (`pmtiles_file_source.cpp:559-562`)
relabels it with the misleading "Error parsing PMTiles directory:" prefix.

**Where.** The `try` at `:549-558` invokes the `getTileAddress` callback (`:589-618`)
synchronously, which does `directory_cache.at(url).at(cacheKey)` (`:596-597`) â€” the throwing
lookup (the `header_cache.at(url)` at `:595` cannot plausibly miss once tiles are being
requested).

**Mechanism (hypothesis, consistent with all observations).** `storeDirectory` (`:473-491`)
maintains an LRU capped at `MAX_DIRECTORY_CACHE_ENTRIES = 100` (`:39`):
- `:485` pushes the key into `directory_cache_control` unconditionally â€” NO dedup on the
  miss-path (dedup exists only on the cache-hit path, `:501-513`);
- `:487-489` evicts `control.front()` when `control.size() > 100`.
Under concurrent tile requests (48 in this embedding), interleaved miss-path stores of SHARED
directories (root + shared leaves) insert duplicate keys into the control list. Once >100
directories are resident â€” trivial for a planet archive under panning â€” every store evicts, and
with duplicates in the list `control.front()` can name a directory that is still needed,
including the one just stored in the same call. The immediate `.at(cacheKey)` in the callback
then throws. Refetch/evict churn repeats this on every tile.

A second, related hazard: the reentrant `tasks[req] = ...` self-overwrite pattern
(`:150/:255/:530`) can destroy an in-flight AsyncRequest whose callback is still executing.

**Suggested fixes.** De-duplicate `directory_cache_control` on the miss path (skip the
`emplace_back` when the key is already resident, or purge duplicates before eviction), and/or
size the cache to the archive's working set rather than a flat 100.

**Repro sketch.** Style with a `pmtiles://` vector source pointing at any planet-scale v3
archive; pan/zoom so more than 100 distinct directories are touched under concurrent tile loads;
observe the error on every subsequent tile. A small single-root regional extract does not
reproduce.

mnordgren:
Fixes #4421.

### Problem

`storeDirectory` appended a directory's key to the per-URL LRU list (`directory_cache_control`) on every store. Concurrent tile requests for one directory each fetch and store it (nothing coalesces in-flight directory fetches), so N concurrent misses left N copies of the key. Once the list passed `MAX_DIRECTORY_CACHE_ENTRIES` (100), evicting a duplicate at the front removed a directory that later list entries still named - including the one just stored - and the `directory_cache.at(url).at(key)` in `getTileAddress`'s callback threw. The `catch` in `getDirectory` relabels that as `Error parsing PMTiles directory: map::at: key not found` (libstdc++) / `invalid map<K, T> key` (MSVC), so it reads as a corrupt archive.

Seen in the wild on Android 13.2.0 with a 220 MB regional archive read by two sources (both read the same file, so every directory was fetched twice): a burst of failed z14 tiles on the first view after load, the tiles staying empty until re-requested.

### Fix

Keep one list entry per key. `storeDirectory` treats an already-resident directory as a cache hit and only moves it to the back of the list; the move-to-back is factored into `touchDirectory`, which the existing cache-hit path in `getDirectory` now uses too. Behaviour is otherwise unchanged.

### Test

`PMTilesFileSource.ConcurrentRequestsShareOneDirectoryEntry` issues 130 concurrent requests for the same tile of the root-only `geography-class-png.pmtiles` fixture. All of them miss until the first store lands, so the old code accumulates more than 100 copies of the root directory's key and fails from the 101st store on; with the fix every request succeeds.

### Disclosure

Per the [AI policy](https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md): the analysis, patch and test were drafted with Claude Code and reviewed by me. I have not built maplibre-native locally, so the new test has only been checked by reading; please let CI be the judge and tell me if it needs adjusting.

https://claude.ai/code/session_01VzpD139tuhP61168FEQPAg


louwers:
The session is not public. It's not required to share that, but if you can make it public, I'd be interested.

Could you indicate what part of the PR description was written by you? You say it was reviewed by you, could you explain what your review process looked like?

Also, please set up a local development environment and test your change locally. If you need help with this or get stuck, feel free to ask on Slack or in a discussion thread.

Issue #2790, opened 2024-08-31
`CameraUpdates` overwrite each other (no parallel camera updates)


westnordost:
**Note**
This is probably rather a missing feature than a bug. However, the API suggests that this should be possible, but it turns out that it isn't.

**Description**
Any new camera update posted to `MapLibreMap::easeCamera` cancels the old camera animation and replaces it with a new camera animation.

**To Reproduce**
Calling e.g. 
```kotlin

val zoomTo = CameraUpdateFactory.newCameraPosition(CameraPosition.Builder().zoom(18).build())
map.easeCamera(zoomTo, 1000)

delay(500)

val moveTo = CameraUpdateFactory.newCameraPosition(CameraPosition.Builder().target(12.0,15.0).build())

map.easeCamera(moveTo, 1000)
```

will cancel the zooming to zoom=18 after 500ms and **instead** (rather than additionally) move to the given target position. I.e. any new camera position update animation will overwrite the last.

**Expected behavior**
The animations of different properties of the camera (zoom, tilt, position, rotation, ...) should run in parallel if they don't conflict each other.

**Video**
In this video, we see that the position is being updated in an animation every second or so to focus on the current location. When the user taps the zoom-in button, the move-to animation is cancelled or the other way round (the zoom-in animation is cancelled), depending on the timing:

https://github.com/user-attachments/assets/4d6529fa-9de5-44bb-aafc-ee21fe746fd1

**Important:** This is not only an issue with **zoom vs position**. Any any time, the user may also press a button to change the map's rotation (the compass button) or tilt (changing from (car-)navigation-mode to just-follow-me-mode) while the position is continuously updated.

**Platform information (please complete the following information):**
 - Platform: Android, any version
 - Version 11.2.0

**Additional context**
Originally reported in https://github.com/streetcomplete/StreetComplete/issues/5860

westnordost:
(Regarding the labels: This is probably rather a missing feature, but an issue for cpp-core, as [animations are implemented in native code](https://github.com/maplibre/maplibre-native/blob/5642b1ffa41e7e13582f77cb15d0157cb8ff6146/include/mbgl/map/map.hpp#L67-L73).)

louwers:
Would just including the options from the previous update in the new one be an option?

westnordost:
You mean the previous animation target? I.e. animation A starts with zoom by +1 (zoom 17 to 18), then while animation A is still running, animation B starts with a new position?

[...]

PR #4637, opened 2026-09-17
feat(core): animate independent camera properties concurrently

sargunv:
## Motivation

Native currently runs one camera animation at a time. This is awkward when different parts of an app control different dimensions of the camera, such as location following with zoom buttons.

For example, the user should be able to start a two-second zoom, then start a longer position-only ease halfway through it. Both should continue, with the zoom reaching its target at the original two-second deadline. Currently, the only way for an app to handle this properly is to own the camera animation and send frame-by-frame camera updates to maplibre native.

Here we support independent property animations in the native core so location following, zoom, rotation, and pitch can run together, as requested in #2790.

## Implementation

- Let partial commands animate independently, so rather than an overarching "camera" animation we maintain a zoom animation independent of a position animation, etc.
- Camera "fly" animations couple position and zoom
- Animations with anchors couple position with their requested property

This changes the core behavior by default. Both [Android](https://github.com/maplibre/maplibre-native/blob/310e50438e8513f0751f72f6dd9e162851521ed2/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/maps/Transform.java#L133-L147) and [iOS](https://github.com/maplibre/maplibre-native/blob/310e50438e8513f0751f72f6dd9e162851521ed2/platform/ios/src/MLNMapView.mm#L4359-L4368) explicitly cancel existing transitions before invoking Native and need separate SDK changes to expose independent animations.

## Tests

New tests in `test/map/transform.test.cpp` cover:
- independent deadlines
- partial replacement
- flight and anchor conflicts
- combined constraints
- callback reentrancy
- still-image completion

Carried in maplibre/maplibre-native-ffi#724. AI disclosure: Codex (GPT-6) assisted with investigation, code, tests, and this description.


src/mln/map/transform.hpp
```diff
@@ -115,6 +115,8 @@ class Transform : private util::noncopyable {
     // Transitions
     bool inTransition() const;
     void updateTransitions(const TimePoint& now);
+    // Timing of the most recently started command, even after it finishes.
+    // Other commands may still be active.
     TimePoint getTransitionStart() const { return transitionStart; }
     Duration getTransitionDuration() const { return transitionDuration; }
```
sargunv: I'm not sure I like the new timing getter semantics, but afaict they're only used in tests so I didn't prioritize coming up with a better api for these. If it's too confusing, maybe we should rename to like `getLatestTransitionStart`

westnordost:
One question mark from the feature request (#2790) was how to deal with animations that animate properties that are not disjunct. To illustrate, two use cases:

- **A:** Camera follows a continuously updating position (animated) on the map. Pressing a zoom-button will zoom into the map (animated). The first only animates position and maybe heading, the second only the zoom. No conflict here, they can animate independently

- **B:** Camera follows a continuously updating position and vehicle heading. Pressing a rotate-button will rotate the camera by 90° (animated), i.e. switching to a "side-view". Both animations want to update the heading

**What would happen in case B here?**

The doc comments sound like it might be the case that in case B, the rotate-button animation stops the other animation, and then in turn the camera-following animation which is called every few frames or so in turn stops the or so in turn stops the rotate-button animation? Docs should be very clear about it. Is there also website-docs / maplibre-native-android/maplibre-native-ios docs that need to be updated?

(My example makes it look like some sort of merging the two animations would be desirable, but this is just the use case I could come up with on the spot. Maybe there are other use cases where the opposite would be desirable?)

---

By the way, I cannot help with review because I am neither proficient with C++ nor with the MapLibre codebase. Just dropping in from a use-case / user point of view.

sargunv:
> What would happen in case B here?

Yeah, on each camera dimension (position, zoom, bearing, etc), the latest camera command wins. So if your camera following animation animates position and bearing, then rotation buttons that also animate bearing would still conflict. In that case, I think it's best for the app to maintain its own "target bearing offset" that the rotation buttons mutate, and is consistently applied in the camera follow logic.

The original semantics of "latest animation wins" is still in place; just applied per-dimension so animations on independent dimensions don't conflict.

> Is there also website-docs / maplibre-native-android/maplibre-native-ios docs that need to be updated?

I haven't updated maplibre native android/ios behavior; the SDK layers still explicitly cancel current animations when new animation commands arrive, so didn't update those docs. These libraries offer some compatibility guarantees so updating behavior on that level requires some consideration of either offering different modes, or deprecating the old behavior for the new.

This PR only decouples the animations in the native core, to make it _possible_ to decouple at the sdk layer if desired.

louwers:
> Both Android and iOS explicitly cancel existing transitions before invoking Native and need separate SDK changes to expose independent animations.

Note to self: need to create an issue when this is merged.

I think we should double check if other platforms also cancel animations, otherwise this would be a breaking change for those platforms.

Edit: clicked wrong button.

src/mln/map/transform.hpp
sargunv: Renamed to `getLatestTransitionStart()` / `getLatestTransitionDuration()` in 498cda82590a; only the transform tests use them.

sargunv:
> I think we should double check if other platforms also cancel animations, otherwise this would be a breaking change for those platforms.

- the Node platform exposes immediate camera commands only, so not affected.
- all iOS public setters cancel.
- https://github.com/maplibre/maplibre-native-qt does not cancel.
- the macOS platform's setCenterCoordinate, setZoomLevel, setDirection do not cancel.
- the Android platform's scrollBy, setFocalBearing, and the focal zoom setter do not cancel.

I'll add a commit to add cancellation to such callers in this repo.

louwers:
Prior art https://github.com/maplibre/maplibre-native/pull/3487

sargunv:
I scanned 3487's thread, but not its implementation in detail. Findings:
- Bugs noted there (flyTo, anchors) are covered here with coupling and tests already
- This version has no toggle; the core is independent by default, the SDKs cancel explicitly for backwards compatibility for now. Sufficient for the FFI and Compose, but the issue will remain open for the SDK users until [that follow up](https://github.com/maplibre/maplibre-native/pull/4637#issuecomment-5727290270)

PR #4639, opened 2026-09-17
fix(core): refresh location indicator colors during transitions

sargunv:
## Bug

The location indicator's uniform updater keeps the paint properties captured at the start of a transition. The accuracy-circle fill and border stay at their previous colors after the transition finishes.

Fixes #4638.

## Repro

Render a blue accuracy circle, then change it to red with a 300 ms transition. It stays blue. Change it to green and it becomes red. Zero-duration updates work.

## Fix

Refresh the uniform updater's evaluated properties each frame, as the other layer renderers do.

## Tests

`Map.LocationIndicatorAccuracyColorsTransition` checks the rendered fill and border after a nonzero transition. Both color checks fail without the fix and pass on macOS ARM64 / Metal. Other backends have not been tested locally.

AI disclosure: Codex (GPT-6) assisted with investigation, code, tests, and this description.

Issue #4638, opened 2026-09-17
Location indicator accuracy-circle colors remain one update behind with transitions


sargunv:
**Describe the bug**

Changing `accuracy-radius-color` on an existing `location-indicator` layer with a nonzero paint transition leaves the circle displaying the previous color, even after the transition has finished. A subsequent color change displays the preceding target.

**To Reproduce**

1. Load an empty style and place the camera at latitude/longitude 0, zoom 16, in a 512 × 512 map.
2. Set the style transition duration to 300 ms.
3. Add a source-free `location-indicator` layer with `location: [0, 0, 0]`, `accuracy-radius: 100`, and opaque blue `accuracy-radius-color`. No images are needed.
4. Render until the center pixel is opaque blue.
5. Change only `accuracy-radius-color` to opaque red. Render for one second.
6. Change only that color to opaque green. Render for another second.

**Expected behavior**

The circle reaches red after step 5 and green after step 6, once each transition finishes.

**Platform information:**

- Operating System: macOS 26.6.2 (25G83), ARM64.
- Platform (e.g. Node.js, Qt): Kotlin/JVM via MapLibre Native FFI, Metal renderer.
- Version: `org.maplibre.nativeffi` runtime artifacts `0.202609.2`; MapLibre Compose at [`16d6a01fd`](https://github.com/maplibre/maplibre-compose/tree/16d6a01fd) ([PR #1426](https://github.com/maplibre/maplibre-compose/pull/1426)). Native source inspected: `abd68adf0d0f3e661ae40e7515e0e0cff90d83f8`, the submodule revision at the FFI `bindings/kotlin/v0.202609.2` tag.

AI assistance disclosure: GPT-6 helped investigate and draft this report.

PR #4644, opened 2026-09-17
feat(core): add bearing accuracy sector to location indicator

sargunv:
**I am not experienced with graphics programming and may have committed some sins in the implementation**

## Motivation

The core location indicator can show position accuracy as a circle, but it cannot show uncertainty in its bearing. Apps need a sector that widens as bearing accuracy decreases, like a direction beam behind the location dot.

I originally implemented this as an image on a SymbolLayer, but if I want animated transitions this becomes impractical on normal style layers. So here I add a bearing accuracy sector to `LocationIndicatorLayer` for use in MapLibre Compose (through MapLibre Native FFI).

If this isn't something you'd like to merge; I'll continue vendoring it as a patch in the ffi until the plugin api is ready, then switch to that to replace the built-in location indicator. But given we have a location indicator layer, I think it's a good idea to include this feature on it.

## Implementation

The new paint properties follow the existing position-accuracy API:

- `bearing-accuracy`: angular half-width in degrees. A value of 15 draws a 30° sector centered on `bearing`; 180 draws a full circle.
- `bearing-accuracy-radius`: visual radius in logical pixels, independent of position accuracy in metres. Defaults to 0, hiding the sector.
- `bearing-accuracy-color`: color and opacity, fading smoothly to transparent at the outer edge.

The sector follows the puck's bearing and perspective compensation and draws above the accuracy circle, below the puck images. All three properties support zoom expressions and paint transitions. Includes the drawable backends and legacy OpenGL path, but not the SDK bindings.

This is also affected by #4638 and so will benefit from #4639.

Bearing accuracy sector

## Tests

Three render fixtures cover north-crossing bearings, live updates on a pitched map at 2× pixel density, full/wide sectors, and hidden limits.

AI disclosure: Codex (GPT-6) assisted with investigation, code, tests, and this description.


louwers (commented):
This looks pretty sleek, but probably the location indicator makes a lot more sense as a plugin?

sargunv:
> This looks pretty sleek, but probably the location indicator makes a lot more sense as a plugin?

I think so, yeah. I just added this bearing accuracy indicator on the existing layer because it's currently not a plugin, and I'm now using this layer in maplibre compose.

Once the plugin api merges, are you thinking of carving out the location indicator as a first party plugin for the android/ios sdks? On my end, I'm happy to replace it with a plugin shipped in maplibre-compose. I've [already done that](https://github.com/maplibre/maplibre-compose/blob/ac9a88967adc44638fb30325fbd8718a013bf4b9/lib/maplibre-compose/src/jsMain/kotlin/org/maplibre/compose/layers/GlJsLocationIndicator.kt) with GL JS

louwers:
@sargunv I think this would be a great test driver for the plugin API. I'm not sure what that would look like but I think it's a good idea, since a lot of users are already hacking their own version together.

Edit: keep hitting the wrong button.

sargunv:
I think, ideally:
- carve out this location indicator layer as a showcase first-party plugin, get it out of core
- bundle it with the android/ios SDKs so that LocationComponent api continues to work
- and I bundle it in maplibre-compose for my LocationIndicatorLayer api

sargunv:
I vibe-drafted a fully functioning plugin here: https://github.com/maplibre/maplibre-native-ffi/pull/731

It required a few extensions to the plugin api:
- animation (#4654)
- source-free plugin layers
- premultiply default colors (bug fix, #4663)
- native rotation properties
- some other stuff my agent describes as "inverse screen projection, and source-free feature snapshots" which I haven't yet read or understood why it needed that

will close this out, and maintain this as a patch in the ffi builds for maplibre-compose use until the plugin api can support it

PR #4650, opened 2026-09-18
fix(core): pick raster-dem tiles at the rounded zoom like raster sources

johncarmack1984:
A hillshade at zoom 4.6 loads z4 tiles while a raster layer on the same tileset loads z5. `coveringZoomLevel` rounds the zoom for raster and video sources and floors it for everything else, raster-dem included, so from x.5 up to the next zoom hillshade and color-relief draw one zoom level coarser than a raster layer would. GL JS rounds for both: `RasterDEMTileSource` extends `RasterTileSource`, which sets `roundZoom`. One line: raster-dem joins raster and video in `coveringZoomLevel`.

| main | this PR |
|---|---|
| main | this PR |

The demotiles terrain (256 px tiles) over Innsbruck at zoom 9.6, `mbgl-render` on Metal. Left: main loads z10. Right: z11.

Tests: `TileCover.CoveringZoomLevelRoundsForRasterSources` (the covering zoom per source type) and `Source.RasterDEMTileRoundsCoveringZoom` (a raster-dem source at zoom 4.6 requests z5) fail on main and pass here. Render test `hillshade-zoom/rounded` sits at zoom 10.6 over the z12 terrain tiles already in the repo, which the floored zoom never asks for: fails on main, passes here.

Checked: Metal: `Source.*` and `TileCover.*` 60 passed; every raster, hillshade and color-relief render test, 62 passed, 1 ignored, 0 failed; the full render manifest 1311 passed, 23 passed but ignored, 76 ignored, 0 failed, so no existing expectation moves. Linux GL (llvmpipe in Docker, `-Werror`, clang-tidy 18 clean): the same unit tests 60 passed, the new render test passes; `color-relief/rainbow` fails there with and without the change (it sits at an integer zoom). `pre-commit run` clean. Semver: patch. The cost is the one raster sources already pay: from x.5 up, a raster-dem source asks for the next zoom's tiles, up to four times as many.

AI Disclosure: Claude Fable 5.1 was used to find any globe-prep changes that would reduce the review load on #4553; this was discovered as a pre-existing bug on main and patched accordingly.


louwers (approved):
Thanks, maybe @acalcutt wants to have a look as well but LGTM.

PR #4653, opened 2026-09-18
android: enable native plugins in pre-releases and add demo

louwers:
This PR:

- Packages the existing ngon-layer for Android `plugins/ngon-layer/android`. I did not add any MapLibre Android SDKs so it looks up `mln_plugin_register_v1` in the MapLibre Android `.so`. Bit weird but works.
- Enable the plugin API for Android pre-releases. `pre-x.y.z.`
- Add another Android app in `plugins/android/app` that uses the Android library from `plugins/ngon-layer/android`. It also uses the SDK from `platform/android` so this app can also be used for plugin API development as well as working on the example plugins.


plugins/ngon-layer/android/src/main/cpp/plugin.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+} // namespace
+
+extern "C" JNIEXPORT void JNICALL Java_org_maplibre_plugins_android_NativePlugin_registerNgon(JNIEnv* env,
+                                                                                              jclass,
+                                                                                              jstring libraryName) {
+    const char* name = env->GetStringUTFChars(libraryName, nullptr);
+    if (!name) return;
+    // Resolve from the renderer MapLibre already loaded, without loading a second registry.
+    void* host = dlopen(name, RTLD_NOW | RTLD_NOLOAD);
```
louwers, line 18: This looks fishy.

zarechnyy (approved):

PR #4533, opened 2026-08-28
feat(core): render the map as a globe when the style sets projection

johncarmack1984:
Closes #3161.

`"projection": {"type": "globe"}` renders the map as a globe on Metal, Vulkan, WebGPU and OpenGL, with GL JS's math and GL JS's render tests as the oracle. This PR is the twelve phases of #4525 as 32 commits, one per phase (9 and 10 are one commit per family/backend; 10d is the review's shader-group signature unification), and four more out of the review here: 12i shares the tile zoom function so terrain (#4190) can call it, 12j fixes the flicker in the Android video below, 12k puts the rendered feature capture (#4460) through the globe, so a feature is reported where the sphere draws it and not at all behind the horizon, 12l keeps the location indicator's puck its size away from the globe's center (it grew along the horizon, 13 by 58 pixels for a 16 pixel image at 60 degrees of longitude). Two more, 12m and 12n, came with the rebase over the plugin API, below. Every commit builds and passes its own tests alone.

Each phase is a branch on my fork, chained by base, so each phase can be read as one diff: `phased/1` to `phased/12h`, for example https://github.com/johncarmack1984/maplibre-native/compare/phased/3...phased/4. Rebase-merge would keep the phase commits on main. If you'd rather take phases one at a time, I can open them here in order.

Checked at the opening head: on every commit, the Metal unit tests and the globe render group, plus a Linux OpenGL (Docker) run of the whole render suite, a Bazel build and an MSVC (Windows ARM64) build; the fork's CI on every phase ref, Windows judged against main's own baseline. At that tip: Metal 1362 passed / 79 ignored / 0 failed on the full suite, unit tests 915/915; Linux OpenGL in Docker 1364 passed with the same three environment failures main shows there; Dawn and wgpu 1373 / 86 / 0 on the branch before the instancing merge, and the globe group 59 / 4 / 0.

[...]

PR #4671, opened 2026-09-22
ci(ios): free runner disk space before the size test build and the cache save

johncarmack1984:
`ios-build` is running out of disk on `macos-26`.

On main, eleven of the twelve red ios-ci runs since 09-16 finished every build step and then lost the runner in `Post Set up Bazel`, where setup-bazel tars the disk cache; the twelfth died inside the size-test build with the same disk errors. The one green run that got through that step (ef11fddf) hit `zstd: error 70 … No space left on device` and saved nothing. The disk cache archive is 11.1 GB now, up from 9.5 GB a week ago, and the tar needs that much room on top of everything the job has built.

Pull requests don't save, so it shows one step earlier. A cold size-test build dies in the darwin sandbox with EINVAL around the 50-minute mark (#4533). When the build does finish, the `cp` of the dSYM right after fails with EIO (#4625, #4667, #4668). Same build, same image; #4666 got through a few hours earlier. We're at the edge.

Measured on this PR's runs: 17 GiB free of 320 GiB right before the size-test build. The first expunge gives back 4 GiB (17 s). The warm size-test build then takes 8 GiB, leaving 13 GiB, about where main sits when the tar starts. The second expunge gives back 9 GiB (10 s), so the save starts with 21 GiB free for an 11 GB archive. Enough, not roomy; the cap below is what buys margin.

Two steps. After the CppUnitTests upload, expunge rules_xcodeproj's output base (a second full debug build we don't touch again). After the last Bazel command, expunge the main one so the save has room. `bazel clean --expunge` because Bazel leaves its tree read-only; `rm -rf` couldn't delete it (first run of this PR). Both print `df -h` before and after. `~/.cache` (disk cache, repo cache) stays.

The save only runs on main, so this PR's run proves the steps and the freed space; the first main run after merge proves the save.

Separate follow-up: the disk cache grows forever (setup-bazel has no GC, bazel-contrib/setup-bazel#71). `--experimental_disk_cache_gc_max_size` once a run shows the on-disk size.

actionlint + pre-commit clean.


louwers (approved):

louwers:
Thanks!

